PR #45, opened 2026-04-06
feat(#9): execution guard — hard rule enforcement and decision audit

### LINKED ISSUE #9: Execution guard: enforce hard rules and write audit record before any order is staged
## Goal

Build the execution guard — the hard gate that every proposed order must pass through before it is sent to the eToro API. This service enforces all non-negotiable rules and writes a `decision_audit` record for every pass and fail.

## Scope

- Hard rule checks (all must pass for PASS verdict):
  - Instrument is in Tier 1 coverage with a current, non-stale thesis
  - Thesis break conditions have not been triggered
  - Concentration limits would not be breached by this order
  - Sector exposure limit would not be breached
  - Spread/liquidity check passes for this instrument
  - `enable_live_trading` flag is true (or mode is explicitly `demo`)
  - No active kill switch
- Output: PASS with normalised order, or FAIL with every failed rule listed explicitly
- Write to `decision_audit` for every verdict — pass and fail — before returning
- No order is sent to eToro unless this service returns PASS
- Unit tests: each rule in isolation, combined PASS case, combined FAIL case, audit record written on both outcomes

## Dependencies

- Issue #8 (portfolio manager) — consumes trade_recommendations
- All prior services — checks pull from thesis, scores, positions, market data

## Non-negotiable

This service must never be bypassed. Any code path that places an order without calling the execution guard first is a BLOCKING review issue.

## Acceptance criteria

- Every test case results in a `decision_audit` row regardless of outcome
- A PASS with `enable_live_trading=false` is impossible (guard returns FAIL with reason)
- Each failed rule is listed separately in the FAIL response — no generic "failed" message
- CI green, Claude review APPROVE

Luke-Bradford:
## What changed

- `sql/010_execution_guard.sql` — new `kill_switch` table (typed singleton, seeded immediately); `recommendation_id` FK column added to `decision_audit`; index on `decision_audit(recommendation_id)`
- `app/services/execution_guard.py` — new service with public API `evaluate_recommendation(conn, recommendation_id, settings) -> GuardResult`
- `tests/test_execution_guard.py` — 62 tests covering every rule in isolation and end-to-end

## Why

Issue #9. The execution guard is the mandatory hard gate between a portfolio manager recommendation and any eToro order. Nothing goes to the broker without passing this service. Rules are re-checked against current DB state on every invocation — stale recommendation state is never trusted.

## Schema / migration impact

Migration 010:
- Creates `kill_switch` (one-row table enforced by `CHECK (id = TRUE)` + PRIMARY KEY). Seeded with `is_active = FALSE` via `INSERT … ON CONFLICT DO NOTHING` — safe to re-run.
- Adds `decision_audit.recommendation_id BIGINT REFERENCES trade_recommendations(recommendation_id)` (nullable, so existing rows are not broken).
- Adds `idx_decision_audit_recommendation` for audit lookups.

No changes to `trade_recommendations` columns; the guard updates `status` to `'approved'` or `'rejected'` using the existing column.

## Invariants checked

- **Kill switch fail-closed**: missing row → `kill_switch_config_corrupt` FAIL; never PASS.
- **One audit row per invocation**: `decision_audit` insert + `trade_recommendations.status` update committed atomically in a single `conn.transaction()`.
- **No I/O inside transactions**: all DB reads complete before the transaction opens.
- **All `fetchone()` have `ORDER BY … LIMIT 1`**: kill switch uses `ORDER BY id DESC LIMIT 1`; thesis uses `ORDER BY created_at DESC LIMIT 1`.
- **All cursors use `dict_row`**: no positional row access anywhere in the service.
- **`Jsonb(...)` for `evidence_json`**: no `json.dumps` into a JSONB column.
- **`recommendation_id` FK on every audit row**: traceability from audit back to recommendation.
- **EXIT exemptions honoured**: thesis-freshness, coverage, spread, and cash rules are gated on `action in ("BUY", "ADD")`; kill switch and config flags apply to all actions.

## Failure paths considered

- `recommendation_id` not found → `ValueError` raised before any audit write (programmer error)
- Kill switch row missing (e.g. migration not applied) → `kill_switch_config_corrupt` FAIL
- No coverage row → `no_coverage_row` FAIL (BUY/ADD only)
- No thesis row → `no_thesis` FAIL (BUY/ADD only)
- `coverage.review_frequency` NULL or unrecognised → treated as stale, `thesis_stale` FAIL (conservative)
- No quotes row or `spread_flag` NULL → `spread_unavailable` FAIL (BUY/ADD only)
- Empty `cash_ledger` → `cash_unknown` FAIL (BUY/ADD only)
- Thesis `created_at` is naive datetime → treated as UTC (no raise)
- Concentration post-action > 25% AUM → `concentration_breach` FAIL (BUY/ADD only)

## Tests added

62 tests across 8 test classes:
- `TestCheckKillSwitch` — missing row fails closed; active fails; inactive passes
- `TestCheckConfigFlags` — each flag independently; both off → both named
- `TestCheckCoverage` — no row; tier 2; tier 1
- `TestCheckThesisFreshness` — no thesis; unknown/null frequency (conservative); fresh/stale for daily/weekly/monthly; naive datetime
- `TestCheckSpread` — no row; NULL flag; TRUE flag; FALSE flag
- `TestCheckCash` — None; zero; positive
- `TestCheckConcentration` — no sector; zero AUM; at-cap (passes); breach (fails)
- `TestEvaluateRecommendation` — end-to-end: happy path BUY/EXIT; each rule failure in isolation; EXIT exemptions for coverage/thesis/spread/cash; audit written on pass and fail; `decision_id` matches INSERT RETURNING; `recommendation_id` FK correct; multiple failures all listed; not-found raises without audit write

## Conscious tradeoffs

- **Spread check uses `quotes.spread_flag`** (computed by market-data service), not a live eToro API call. Computing spread inline would require network I/O before the audit transaction — violates the no-I/O-in-transaction rule and couples the guard to a provider. `spread_flag` is the right seam.
- **`enable_auto_trading=False` is a hard block** (not a warning-only mode). Whether a softer mode is needed is deferred.
- **Kill switch / concentration race not addressed**: two concurrent guard calls could both pass concentration checks; a manual kill-switch activation between the DB read and the audit write could be missed. v1 demo only — no real capital at risk.

## Tech debt opened

- Concurrent guard calls / concentration race — to be opened as a tech-debt issue at merge time.

Closes #9

### DIFF
--- a/app/services/execution_guard.py
+++ b/app/services/execution_guard.py
@@ -261,8 +261,10 @@ def _load_sector_exposure(
                 LIMIT 1
             ) q ON TRUE
             WHERE p.current_units > 0
+              AND p.instrument_id != %(iid)s
             GROUP BY i.sector
             """,
+            {"iid": instrument_id},
         )
         rows = cur.fetchall()
 
@@ -432,7 +434,7 @@ def _check_concentration(
             detail="instruments.sector is NULL; cannot verify concentration",
         )
     if total_aum <= 0:
-        return RuleResult(rule="concentration_breach", passed=True)
+        return RuleResult(rule="concentration_breach", passed=True, detail="total_aum=0; concentration check skipped")
     post_action_pct = current_sector_pct + _MAX_INITIAL_POSITION_PCT
     if post_action_pct > _MAX_SECTOR_EXPOSURE_PCT:
         return RuleResult(
--- a/tests/test_execution_guard.py
+++ b/tests/test_execution_guard.py
@@ -441,6 +441,7 @@ class TestCheckConcentration:
     def test_zero_aum_passes(self) -> None:
         result = _check_concentration(True, "Technology", 0.0, 0.0)
         assert result.passed is True
+        assert "total_aum=0" in result.detail
 
     def test_within_cap_passes(self) -> None:
         # 20% current + 5% alloc = 25% — exactly at cap, not over (> not >=)


Luke-Bradford:
## Review resolution — commit 21c57a5

### BLOCKING

**`_load_quote` / `_load_coverage` missing `ORDER BY … LIMIT 1`**
FIXED 21c57a5 — both queries now carry `ORDER BY instrument_id LIMIT 1`. Coverage is genuinely a singleton-per-instrument today so ordering by `instrument_id` is correct; quotes is likewise one row per instrument (the market-data service upserts). Comment noted that `ORDER BY recorded_at DESC` would be needed if quotes ever acquires historical rows — left as a comment in the query intent rather than a code change since the schema doesn't have that column today.

**`assert` → `RuntimeError` in `_write_audit`**
FIXED 21c57a5 — replaced `assert audit_row is not None` with `if audit_row is None: raise RuntimeError(...)`. Marked `# pragma: no cover` since the path is unreachable under normal operation but must not be silently optimised away.

**Zero cash passes BUY**
FIXED 21c57a5 — `_check_cash` now treats `cash <= 0` as `cash_unknown` FAIL. Zero balance means no buying power; the rule's stated purpose is affordability checking. Tests updated: `test_zero_cash_passes` removed, replaced with `test_zero_cash_fails` and `test_negative_cash_fails`.

### WARNING

**Missing instrument row silently bypasses concentration**
FIXED 21c57a5 — `_load_sector_exposure` now returns `(instrument_found: bool, ...)` and short-circuits to `return False, None, 0.0, 0.0` when no instruments row exists. `_check_concentration` takes `instrument_found` as its first arg and returns `instrument_missing` FAIL when it is False. `instrument_missing` added to `RuleName`. End-to-end test `test_instrument_missing_fails_buy` added.

### NITPICK

**`GuardResult.failed_rules: list[str]` → `list[RuleName]`**
FIXED 21c57a5 — type tightened on the dataclass field; `failed_rule_names` annotated explicitly as `list[RuleName]` so pyright accepts the assignment.

**`_load_recommendation` fetches unused `cash_balance_known`**
FIXED 21c57a5 — column removed from the SELECT.

**`score_missing` in `RuleName` never produced**
FIXED 21c57a5 — removed from the `RuleName` Literal.

### PREVENTION notes

Extracted into `.claude/skills/engineering/python-hygiene.md` and `.claude/skills/engineering/test-quality.md` (committed on the preceding chore commit). No further action needed here.

PR #68, opened 2026-04-06
feat: eToro order client — place orders, process fills, demo mode (#10)

### LINKED ISSUE #10: eToro order client: place and close positions via eToro write API
## Goal

Build the thin client that takes a guard-approved order and places it via the eToro write API. This is the only place in the codebase that talks to the eToro write endpoint.

## Scope

- eToro write API client: place order, close position, check order status — raw response persisted to `orders` table before any normalisation
- Order placement: takes a normalised order from the execution guard (PASS only), sends to eToro, persists the broker order ref and status
- Fill processing: poll or webhook for fill confirmation, persist to `fills` table, update `positions`
- Demo mode: when `enable_live_trading=false`, log the would-be request and return a synthetic fill — no real API call made
- Error handling: eToro API errors are logged, order status set to FAILED, `decision_audit` updated — no silent swallowing
- Unit tests: demo mode produces correct synthetic fill, real API calls are behind an interface that can be mocked

## Dependencies

- Issue #9 (execution guard) — only accepts guard-PASS orders

## Design note

This client should be as thin as possible. All decision logic lives upstream. This service's only job is: call the API, persist the response, update order/fill/position tables.

The eToro API is the current broker integration. The interface should be clean enough that swapping in a different broker API later is a config change, not a rewrite.

## Acceptance criteria

- Demo mode never makes a real HTTP call to eToro
- Every order attempt (success or failure) produces a row in `orders` with raw response payload
- A successful fill produces a row in `fills` and updates `positions`
- CI green, Claude review APPROVE

Luke-Bradford:
## What changed

New order execution layer that takes a guard-approved recommendation and either
places a real broker order (live mode) or generates a synthetic fill (demo mode).

**New files:**
- `sql/012_orders_guard_fk.sql` — migration adding `recommendation_id` and `decision_id` FK columns to `orders` table, with index
- `app/providers/broker.py` — abstract `BrokerProvider` interface (`place_order`, `close_position`, `get_order_status`)
- `app/providers/implementations/etoro_broker.py` — eToro write API client, thin adapter with no domain logic or DB access
- `app/services/order_client.py` — orchestration: validate approval → call broker / demo → persist order + fill + position + cash atomically
- `tests/test_order_client.py` — 25 unit tests

## Why

Issue #10 — this is the next module in the build order after the execution guard (#9).
The order client is the only place in the codebase that talks to the broker write API.
All decision logic lives upstream (portfolio manager → execution guard → order client).

## Schema / migration impact

- Migration `012_orders_guard_fk.sql`: adds two nullable FK columns (`recommendation_id`, `decision_id`) and one index to `orders`
- No data backfill needed — new columns are nullable
- Existing `orders`, `fills`, `positions`, `cash_ledger` tables used as-is

## Invariants checked

- **No I/O inside transactions** — broker call completes before `conn.transaction()` opens
- **Atomic persistence** — order + fill + position + cash all inside a single transaction
- **Idempotent position upsert** — `INSERT ON CONFLICT DO UPDATE` for BUY/ADD
- **Raw payload persisted before normalisation** — every order row carries `raw_payload_json`
- **Demo mode isolation** — demo mode never calls the broker provider (tested explicitly)
- **Production invariants use `if/raise`** — no bare `assert`
- **dict_row on every cursor** — no positional row access
- **fetchone ORDER BY** — quotes query uses `ORDER BY quoted_at DESC LIMIT 1`; positions uses PK lookup with documented guarantee

## Failure paths considered

- Recommendation not found → `ValueError`
- Recommendation not approved → `ValueError`
- Live mode with no broker supplied → `ValueError`
- Broker returns `failed` → order row persisted with failed status, no fill/position/cash
- Broker returns `rejected` → same as failed, order row captures raw_payload
- Broker returns `pending` → order row persisted, no fill yet, recommendation set to `execution_pending`
- Demo mode with no quote → synthetic fill at price=0, visible in payload note

## Tests added

25 tests across 6 test classes:
- `TestSyntheticFill` — BUY with amount, BUY with units, EXIT, no quote (price=0), no amount+no units
- `TestLoadApprovedRec` — not found raises, not approved raises, approved returns row
- `TestLoadHelpers` — quote price decimal/none/null, position units decimal/zero
- `TestExecuteOrderDemoMode` — demo BUY fill, demo EXIT fill, no-quote order, demo never calls broker
- `TestExecuteOrderLiveMode` — live BUY calls place_order, live EXIT calls close_position, missing broker raises
- `TestExecuteOrderFailures` — broker failed/pending/rejected persists order row, not-found/not-approved raises

## Conscious tradeoffs

- **No duplicate-execution guard** — two concurrent calls to `execute_order` for the same recommendation will both succeed and place two orders. Acceptable in v1 since the execution pipeline is sequential and the guard produces a single PASS. A unique constraint on `orders.recommendation_id` could be added later if needed.
- **Pending order follow-up not implemented** — when the broker returns `pending`, the order is persisted but no polling/webhook mechanism exists yet. Market orders on eToro fill immediately in practice; pending handling can be a follow-up issue.
- **EXIT position units loaded before transaction** — the position units for EXIT are read outside the transaction that writes the fill. A concurrent fill could cause over-selling. Acceptable in v1 since execution is sequential.

## Tech debt opened

None — all scope items are addressed.

- **Provider boundary**: `EtoroBrokerProvider` is a thin adapter; no domain logic; no DB access
- **Config controls**: `enable_live_trading` controls demo vs live; checked in service
- **Guard input**: order client only accepts guard-approved recommendations (status='approved')
- **Auditability**: every order attempt persists raw payload; recommendation status updated
- **Cash semantics**: BUY/ADD records negative amount; EXIT records positive amount

## Prevention log entries addressed

- **fetchone ORDER BY**: quotes query uses `ORDER BY quoted_at DESC LIMIT 1`; PK lookups documented
- **No bare assert**: all production invariants use `if/raise RuntimeError`
- **Missing data fails explicitly**: recommendation not found/not approved raises ValueError
- **Read-then-write in same transaction**: order + fill + position + cash all in one `conn.transaction()`
- **Audit reads in write transaction**: recommendation status update is inside the transaction

Closes #10

### DIFF
--- a/app/services/order_client.py
+++ b/app/services/order_client.py
@@ -389,7 +389,7 @@ def _write_execution_audit(
     instrument_id: int,
     recommendation_id: int,
     order_id: int,
-    outcome: str,
+    passed: bool,
     explanation: str,
     raw_payload: dict[str, Any],
     now: datetime,
@@ -397,8 +397,9 @@ def _write_execution_audit(
     """
     Write a decision_audit row recording the execution outcome.
 
-    This is separate from the execution guard's audit row (stage='execution_guard').
-    The order_execution stage records what happened when the order was placed.
+    Uses the same PASS/FAIL vocabulary as the execution guard so the
+    pass_fail column is semantically consistent across stages.  The
+    detailed execution status goes into explanation.
     """
     conn.execute(
         """
@@ -414,7 +415,7 @@ def _write_execution_audit(
             "iid": instrument_id,
             "rid": recommendation_id,
             "stage": STAGE,
-            "pf": outcome,
+            "pf": "PASS" if passed else "FAIL",
             "expl": explanation,
             "ev": Jsonb({"order_id": order_id, "raw_payload": raw_payload}),
         },
@@ -579,14 +580,16 @@ def execute_order(
             {"status": exec_status, "rid": recommendation_id},
         )
 
-        # Write execution outcome to decision_audit (every path, success or failure)
+        # Write execution outcome to decision_audit (every path, success or failure).
+        # pass_fail uses PASS/FAIL vocabulary consistent with the execution guard.
+        # Detailed status goes in explanation.
         _write_execution_audit(
             conn,
             instrument_id=instrument_id,
             recommendation_id=recommendation_id,
             order_id=order_id,
-            outcome=exec_status,
-            explanation=f"order_status={order_status} broker_ref={broker_result.broker_order_ref}",
+            passed=exec_status == "executed",
+            explanation=f"status={exec_status} order_status={order_status} broker_ref={broker_result.broker_order_ref}",
             raw_payload=broker_result.raw_payload,
             now=now,
         )
--- a/docs/review-prevention-log.md
+++ b/docs/review-prevention-log.md
@@ -205,3 +205,16 @@ add an entry here as part of resolving the comment (`EXTRACTED docs/review-preve
 - **Enforced in:** this prevention log
 - **Promoted to skill?** no — general enough but too edge-case for a skill; this entry is a repo-specific reminder
 - **Notes:** Common correct patterns: `cash (USD) * size_pct (fraction) = dollar_amount (USD)`. Common wrong patterns: `price (USD/unit) * size_pct (fraction) = ??? (USD/unit × fraction = nonsense)`.
+
+---
+
+### Columns shared across stages must use a consistent vocabulary
+
+- **Bug class:** vocabulary mismatch in shared column
+- **First seen in:** `#68`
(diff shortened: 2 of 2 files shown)


Luke-Bradford:
Addressed in 5feca6e.

### WARNING

**pass_fail column receives non-PASS/FAIL values**
FIXED 5feca6e — \`_write_execution_audit\` now takes a boolean \`passed\` parameter and maps to \`"PASS"\`/\`"FAIL"\`. Detailed execution status (\`executed\`/\`execution_pending\`/\`execution_failed\`) goes in the \`explanation\` field, consistent with the execution guard's vocabulary.

### PREVENTION

**pass_fail vocabulary mismatch**
EXTRACTED docs/review-prevention-log.md — entry: "Columns shared across stages must use a consistent vocabulary"

Issue #58, opened 2026-04-06
Auth and secrets handling for operator access and runtime controls


Luke-Bradford:
## Goal

Establish an auth model for eBull's API surface, even in single-user mode. Broker-adjacent data and runtime controls (kill switch, trading flags) must not be exposed without authentication.

## Scope

- Choose an auth mechanism suitable for a single-operator tool (API key, session token, or similar)
- Protect all write endpoints (kill switch toggle, config changes, future order placement)
- Protect read endpoints that expose sensitive data (portfolio positions, execution audit, broker config)
- Handle eToro API credentials securely (never expose via API response, never log)
- Health/liveness endpoints (`/health`, `/health/db`) can remain unauthenticated

## Design considerations

- This is not a multi-tenant auth system — keep it simple
- A static API key or bearer token checked via middleware is probably sufficient for v1
- The key must come from environment config, not hardcoded
- Consider a FastAPI dependency that can be applied to protected route groups
- Auth must be in place before any config/kill-switch endpoint goes live

## Out of scope for v1

- User management, roles, or RBAC
- OAuth / SSO
- Session management beyond token validation

## Why this matters

Even though eBull is single-user, it controls real money decisions. An unauthenticated kill switch endpoint or config toggle is a security gap, not a convenience feature. GPT review specifically flagged this as under-scoped.

Issue #57, opened 2026-04-06
API: job status and system health summary endpoint


Luke-Bradford:
## Goal

Expose job scheduler status and data freshness so the admin page can show system health at a glance.

## Endpoints

- `GET /system/status` — summary of job health: last run time per job, success/failure, data freshness per table (latest quote age, latest filing age, latest news age, latest score run)
- `GET /system/jobs` — list of scheduled jobs with their current state, next run time, last result

## Notes

- This is distinct from `/health` (which is a liveness probe) — this endpoint is for operator visibility
- Data freshness should flag stale data (e.g. quotes older than 24h on a trading day)
- Ties into #13 (ops monitor) but the API endpoint is needed independently for the frontend
- The admin page will poll this to show a system health dashboard

## Depends on

- `app/workers/scheduler.py` for job metadata
- Existing tables for freshness checks

Issue #51, opened 2026-04-06
API: scores and rankings endpoint


Luke-Bradford:
## Goal

Expose the ranked candidate list and score history to the frontend.

## Endpoints

- `GET /rankings` — latest scored and ranked instruments, filterable by tier, stance, sector. Include score breakdown columns (thesis, fundamentals, momentum, sentiment, penalties, total, rank, rank_delta)
- `GET /rankings/history/{instrument_id}` — score trend over time for a single instrument

## Notes

- Rankings endpoint is the core product view — it answers "what should I be looking at?"
- Should join instrument metadata (symbol, name, sector) for display
- Score breakdown should be transparent per settled decision (no hidden weighting)
- Rank delta shows momentum of conviction change

## Depends on

- Existing `scores` table with ranking columns (migration 007/008)

Issue #50, opened 2026-04-06
API: instrument list and detail endpoints


Luke-Bradford:
## Goal

Expose instrument data to the frontend via REST endpoints.

## Endpoints

- `GET /instruments` — paginated list, filterable by sector, coverage tier, exchange
- `GET /instruments/{id}` — single instrument with latest quote, coverage tier, sector, external identifiers

## Notes

- Pagination should use cursor or offset-based approach
- Include latest quote price and quoted_at timestamp when available
- Filter by coverage tier is important for the rankings view (Tier 1 = trade eligible)
- This is the foundational API endpoint — most frontend pages will depend on instrument lookup

## Depends on

- Existing `instruments`, `coverage`, `quotes` tables
- No new schema needed

PR #73, opened 2026-04-06
feat: instrument list and detail API endpoints (#50)

Luke-Bradford:
## Summary
- Adds `GET /instruments` — paginated list with optional sector, coverage tier, and exchange filters
- Adds `GET /instruments/{id}` — single instrument detail with latest quote, coverage tier, and external identifiers
- Establishes the `app/api/` package pattern for subsequent API issues (#51–#57)

## What changed

| File | Change |
|---|---|
| `app/api/__init__.py` | New package |
| `app/api/instruments.py` | Router, response models, query logic |
| `app/main.py` | Mounts instruments router |
| `tests/test_api_instruments.py` | 21 tests |

### Query strategy
- **List endpoint**: single query joining `instruments` + `quotes` + `coverage`. Both are 1:1 PK tables — no fan-out risk. `external_identifiers` (1:N) is deliberately excluded from the list query.
- **Detail endpoint**: instrument query + separate `external_identifiers` query with deterministic `ORDER BY provider, identifier_type, identifier_value`.
- **COUNT query**: only joins `coverage` when the `coverage_tier` filter is active — avoids unnecessary work.

### Quote semantics
The `quotes` table is a 1:1 current-snapshot table keyed by `instrument_id`. Each market-data refresh overwrites the single row. This is documented in the `QuoteSnapshot` docstring. A `LEFT JOIN` is therefore fan-out-safe.

### Pagination
Offset-based with `limit` (1–200, default 50) and `offset` (≥0, default 0). Sufficient for the expected dataset size (~hundreds of instruments).

### Validation
FastAPI `Query` constraints reject invalid values at the framework level:
- `limit`: `ge=1, le=200`
- `offset`: `ge=0`
- `coverage_tier`: `ge=1, le=3`

### Ordering
- List: `ORDER BY i.symbol, i.instrument_id` — deterministic tiebreak on PK
- External identifiers: `ORDER BY provider, identifier_type, identifier_value`

## Security model
Read-only endpoints. No auth required per #58 decision (auth lands before sensitive write endpoints, not before basic reads). No writes, no mutations. All queries use parameterised values — the f-string SQL only interpolates hardcoded clause fragments, never user input.

- **Provider boundary**: no provider code involved; direct DB reads from API layer
- **Identifier strategy**: external identifiers served from `external_identifiers` table, not resolved via providers
- **JOIN fan-out prevention**: quotes and coverage are 1:1; external identifiers fetched separately

## Prevention log entries checked
- **JOIN fan-out** (#45): quotes/coverage are 1:1 PK joins; ext IDs separate query
- **HTTP 200 on infra failure** (#70): DB exceptions propagate as 500, not returned as dicts

## Test plan
- [x] Happy path — list with items, detail with full data
- [x] Empty table — `items: [], total: 0`
- [x] Missing quote → `latest_quote: null`
- [x] Missing coverage → `coverage_tier: null`
- [x] Missing external identifiers → `[]`
- [x] 404 for nonexistent instrument
- [x] All three filters (sector, tier, exchange) pass correct params
- [x] Pagination offset/limit forwarded correctly
- [x] Validation: limit > 200, limit = 0, offset < 0, tier out of range → 422
- [x] COUNT query structure: no coverage join without tier filter; with join when tier active
- [x] External identifiers ORDER BY clause present in SQL

### DIFF
--- a/app/db/__init__.py
+++ b/app/db/__init__.py
@@ -15,13 +15,13 @@ Usage in route handlers::
 
 from __future__ import annotations
 
-from collections.abc import Iterator
+from collections.abc import Generator
 
 import psycopg
 from fastapi import Request
 
 
-def get_conn(request: Request) -> Iterator[psycopg.Connection[object]]:
+def get_conn(request: Request) -> Generator[psycopg.Connection[object], None, None]:
     """FastAPI dependency that checks out a connection from the pool.
 
     The connection is returned to the pool when the request completes.


Luke-Bradford:
### WARNING: Raw `psycopg.connect` in request handlers (re-raised)

REBUTTED — This was addressed in the previous resolution comment with a specific rebuttal, not left unaddressed.

The facts:
1. All 4 existing endpoints in `app/main.py` use `psycopg.connect(settings.database_url)` directly — this PR did not introduce the pattern.
2. Retrofitting connection pooling requires changes to all existing endpoints + adding `psycopg_pool` as a dependency + lifespan setup + DI wiring — that is a separate cross-cutting issue, not scope for #50 (instrument list/detail API).
3. Tech-debt issue #74 is already open to track this.
4. The system is demo-scale with a single operator. Production reliability under load is not the current risk posture (see settled decisions: "demo-first, small-capital live later").

This PR should not be blocked on fixing a pre-existing pattern in every other endpoint. The pooling fix belongs in #74 and should land before any load-bearing deployment.

### PREVENTION: Raw `psycopg.connect` in request handlers
DEFERRED #74 — prevention rule will be enforced after the pooling migration.

PR #83, opened 2026-04-06
feat: DB-backed runtime config + kill switch API (#56)

### LINKED ISSUE #56: API: system config and kill switch endpoint
## Goal

Expose runtime configuration and the kill switch for the admin/settings UI.

## Endpoints

- `GET /config` — current runtime flags (enable_auto_trading, enable_live_trading, kill_switch state, app_env, etoro_env)
- `POST /config/kill-switch` — toggle the DB-backed kill switch
- `PATCH /config` — update runtime flags (with appropriate guards)

## Notes

- Kill switch is DB-backed and separate from deployment config (settled decision)
- `enable_auto_trading` ≠ `enable_live_trading` — both may be checked (settled decision)
- Even in single-user mode, write operations on config should require auth (see auth issue)
- This endpoint must never silently succeed if the DB is unreachable

## Depends on

- Existing kill_switch and config infrastructure
- Auth issue (for write protection)

Luke-Bradford:
Closes #56.

## Summary
- Adds `runtime_config` singleton table + `runtime_config_audit` (migration 015). `enable_auto_trading` / `enable_live_trading` are now DB-backed; env settings no longer carry them.
- New `app/api/config.py` router: `GET /config`, `PATCH /config`, `POST /config/kill-switch`. `/kill-switch` retained as deprecated alias.
- `execution_guard` and `order_client` read runtime config from DB on every call. Missing row fails closed (guard emits `runtime_config_corrupt` rule and skips auto/live checks; order_client raises).
- Kill-switch activate/deactivate are now atomic (`SELECT FOR UPDATE` + UPDATE + audit insert in one transaction) and write per-change audit rows into `runtime_config_audit` (field=`kill_switch`).
- Removed stale env-backed `enable_*_trading` from `/health`.

## Security model
- All `/config*` routes go through `require_auth` (Bearer token).
- Mutations require `reason` + `updated_by`; enabling live trading additionally requires `confirm_live_enable=true`.
- Per-changed-field audit rows for both runtime flag changes and kill-switch transitions.
- Fail-closed on missing singleton row everywhere it's read (guard, order_client, GET, PATCH).
- All writes inside `with conn.transaction():` with `SELECT FOR UPDATE` to prevent lost updates.

## Tradeoffs
- Per-changed-field audit (not per-request) — gives clean field history at the cost of multiple rows per multi-field PATCH. Worth it for auditability.
- Kept `/kill-switch` as deprecated alias rather than removing — cheap and avoids breaking existing callers; canonical path is `/config/kill-switch`.
- Test refactor: `test_execution_guard` injects a runtime_config cursor; `test_order_client` uses an autouse fixture monkeypatching `get_runtime_config` (cleaner than threading new mocks through every test).

## Test plan
- [x] `uv run ruff check .`
- [x] `uv run pyright`
- [x] `uv run pytest` (721 passed)

### DIFF
--- a/app/api/config.py
+++ b/app/api/config.py
@@ -130,10 +130,11 @@ class KillSwitchRequest(BaseModel):
 
     @model_validator(mode="after")
     def _attribution_required(self) -> KillSwitchRequest:
-        if self.active and not self.reason.strip():
-            raise ValueError("reason is required when activating the kill switch")
-        # Attribution required on every transition (activate AND deactivate),
-        # so the runtime_config_audit row always carries an operator identity.
+        # Both reason and activated_by are required on every transition
+        # (activate AND deactivate) so the runtime_config_audit row always
+        # carries operator identity and a justification.
+        if not self.reason.strip():
+            raise ValueError("reason is required for kill switch transitions")
         if not self.activated_by.strip():
             raise ValueError("activated_by is required for kill switch transitions")
         return self
--- a/app/services/ops_monitor.py
+++ b/app/services/ops_monitor.py
@@ -508,6 +508,8 @@ def deactivate_kill_switch(
     """
     if not deactivated_by.strip():
         raise ValueError("deactivate_kill_switch: deactivated_by is required for attribution")
+    if not reason.strip():
+        raise ValueError("deactivate_kill_switch: reason is required for attribution")
     now = now or _utcnow()
     with conn.transaction():
         with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:


Luke-Bradford:
Addressed both WARNINGs in 1037923.

**WARNING 1 — POST /config/kill-switch response shape**: FIXED 1037923
`post_kill_switch` now declares `response_model=KillSwitchResponse` and returns the result of `_build_kill_switch_response(conn)` (a post-write `get_kill_switch_status` read), so the response carries `active`, `activated_at`, `activated_by`, `reason` from the authoritative DB row rather than echoing the request body. The deprecated `/kill-switch` alias in `app/main.py` inherits the same return type.

**WARNING 2 — no-op PATCH audit gap**: FIXED 1037923
Added `RuntimeConfigNoOp(ValueError)` in `app/services/runtime_config.py`. `update_runtime_config` now computes `auto_changed`/`live_changed` and raises `RuntimeConfigNoOp` before issuing the UPDATE if neither flag would change, so `updated_at`/`updated_by`/`reason` on the singleton can never drift from the audit table. `patch_config` maps it to 422. Test `test_no_op_update_raises` updated accordingly.

**PREVENTION** notes: both extracted directly into the code (response_model declaration + RuntimeConfigNoOp guard). No separate prevention-log entry needed — the type system and the new exception enforce both rules at the call site.

Luke-Bradford:
**WARNING 1 — empty attribution via non-API callers**: FIXED 9bb12e8→follow-up
Both \`activate_kill_switch\` and \`deactivate_kill_switch\` in \`app/services/ops_monitor.py\` now raise \`ValueError\` at the top of the function if \`activated_by\`/\`deactivated_by\` is empty (and \`reason\` is empty for activate). The attribution invariant is now enforced at the service boundary, regardless of call site — the API validator becomes a defense-in-depth layer rather than the only gate.

**WARNING 2 — app-clock vs DB-clock in PATCH response**: FIXED in same commit
\`update_runtime_config\` now uses \`UPDATE runtime_config ... RETURNING updated_at\` and carries the returned value into the \`RuntimeConfig\` it returns. Same pattern as \`activate_kill_switch\`. Future-proofs against an \`updated_at\` trigger.

Issue #87, opened 2026-04-07
tech-debt: 5xx HTTPException detail strings leak internal exception text


Luke-Bradford:
Surfaced by the review on #86. Prevention-log entry "Internal exception text leaked into HTTP response bodies" was added in that PR; this issue tracks fixing the three pre-existing 5xx instances in `app/api/config.py` that the rule now prohibits.

`app/main.py` (/health/data) was originally listed here too but has since been fixed inline in #86 (commit f7a0a9b), so this issue narrows to just the config router.

## Affected sites
- `app/api/config.py:168` — `GET /config` on `RuntimeConfigCorrupt`
- `app/api/config.py:203` — `PATCH /config` on `RuntimeConfigCorrupt`
- `app/api/config.py:249` — `POST /config/kill-switch` on `RuntimeError` (singleton missing)

Each currently does:
```python
raise HTTPException(status_code=503, detail=str(exc)) from exc
```

The 4xx instances in the same file (`status_code=400/422`) are intentionally left in place — those exceptions are pydantic validation errors raised against the caller's own input, not internal state, so echoing them is safe and useful.

## Fix
Replace `detail=str(exc)` with a fixed string per route. Keep the `from exc` chain so the framework's exception logger still has the traceback. Add a per-route test that raises an exception containing a unique marker string and asserts the marker is **not** present in `response.text`.

## Acceptance
- All three `app/api/config.py` 5xx leak sites replaced with fixed strings.
- Each replacement covered by a regression test using the unique-marker pattern.
- `grep -rn "HTTPException(status_code=5.*str(exc)" app/` returns no matches.

Issue #65, opened 2026-04-06
Settings / runtime controls page


Luke-Bradford:
## Goal

Build a settings page for coverage tier assignment and runtime configuration.

## Content

- Coverage tier assignment: view and modify instrument tier assignments (Tier 1/2/3)
- Scoring mode selection (currently v1-balanced, future modes)
- Trading flags: auto-trading and live-trading toggles with confirmation
- eToro environment display (demo vs live)

## Endpoint dependencies

| Endpoint | Issue | Purpose | Read/Write |
|---|---|---|---|
| `GET /instruments` | #50 | Instrument list with current tier | Read |
| `GET /config` | #56 | Current runtime flags | Read |
| `PATCH /config` | #56 | Update trading flags | **Write** |
| Coverage tier assignment API | #12 | Assign/change instrument tier | **Write** (endpoint TBD — needs to be added to #12 or a new API issue) |

## Loading / empty / error states

- **Loading**: skeleton for instrument tier table; skeleton for config flags
- **Empty — no instruments**: "No instruments synced yet. Run universe sync first."
- **Error — API failure**: error banner; disable all write controls until API is reachable
- **Error — tier change fails**: error toast with reason; revert the displayed tier to the server state
- **Error — config update fails**: error toast; revert displayed flag to server state

## Read-only vs state-changing actions

- **Read-only**: instrument list display, current config display, eToro environment display
- **State-changing**:
  - Tier assignment (changes which instruments are trade-eligible) — **requires confirmation**: "Promoting {symbol} to Tier 1 makes it trade-eligible. Confirm?"
  - Trading flag toggles (`PATCH /config`) — **requires confirmation dialog** for each toggle
  - Scoring mode selection — confirmation if it would trigger a re-score

## Auth expectation

- All endpoints require auth (#58)
- All write actions on this page are sensitive (they affect trading eligibility and execution behavior)
- If auth is missing, show read-only view with all write controls disabled and a "Login required to modify settings" message

## Notes

- Tier changes affect which instruments are trade-eligible — confirm before applying
- Trading flag toggles need confirmation dialogs
- Bulk tier assignment (e.g. "promote top 25 ranked to Tier 1") is out of scope for v1 — individual instrument only

Luke-Bradford:
Closing as superseded.

[...]

Issue #60, opened 2026-04-06
Dashboard: portfolio overview page


Luke-Bradford:
## Goal

Build the main dashboard — the first page an operator sees after login.

## Content

- Total AUM (mark-to-market)
- Cash balance
- Unrealized P&L summary
- Current positions table (instrument, qty, cost basis, current value, P&L %, sector)
- Recent recommendations (last 5–10, with action and status)
- System status summary (data freshness, kill switch state, last job run)

## Endpoint dependencies

| Endpoint | Issue | Purpose |
|---|---|---|
| `GET /portfolio` | #53 | Positions, AUM, cash, P&L |
| `GET /recommendations` | #53 | Recent recommendation list |
| `GET /system/status` | #57 | Data freshness, job health |
| `GET /config` | #56 | Kill switch state, trading flags |

All read-only on this page.

## Loading / empty / error states

- **Loading**: skeleton cards for AUM/cash/P&L summary; skeleton table rows for positions and recommendations
- **Empty — no positions**: show "No positions yet" with a link to the rankings page
- **Empty — no recommendations**: show "No recommendations yet — run the portfolio manager"
- **Error — API unreachable**: error banner at top; show last-known data if cached, otherwise full error state
- **Error — partial failure**: if `/portfolio` succeeds but `/system/status` fails, show portfolio data and an error indicator in the system status section only — do not blank the whole page

## Read-only vs state-changing actions

- **Read-only**: all data on this page
- **No write actions** on the dashboard — kill switch toggle lives on the admin page (#64), not here

## Auth expectation

- All endpoints on this page require auth (#58)
- If auth fails or token is missing, redirect to a login/connect prompt rather than showing an empty dashboard

## Notes

- This is the homepage — it should load fast and give a complete picture at a glance
- System status section should surface warnings (stale data, kill switch active) prominently
- Positions table rows should link to instrument detail pages (#62)

PR #88, opened 2026-04-07
feat(#59): frontend scaffold — Vite + React + TS + Tailwind shell

### LINKED ISSUE #59: Frontend scaffold: framework, routing, build tooling, layout shell
## Goal

Set up the frontend application with framework choice, build tooling, routing, and a basic layout shell with navigation.

**This can start in parallel with API work (#50–#53).** Use mocked/stubbed API clients initially to flush out naming, response-shape, and UX assumptions early.

## Decisions to make

- Framework: React (Next.js or Vite+React) vs alternatives — propose based on the single-user, data-heavy dashboard nature of the app
- Styling: Tailwind, CSS modules, or component library (shadcn/ui, etc.)
- State management: React Query / SWR for server state, minimal client state
- Build integration: how the frontend builds and runs alongside the FastAPI backend

## Deliverables

- `/frontend` directory (or similar) with framework scaffold
- Basic layout: sidebar nav, header, main content area
- Route stubs for: dashboard, rankings, instrument detail, recommendations, admin/settings
- Dev server with proxy to FastAPI backend
- Build script integrated into project tooling
- Typed API client stub (or mock layer) so pages can develop against expected response shapes before endpoints are live

## Endpoint dependencies

None directly — this is the shell. But the route stubs and API client types should anticipate the response shapes from:
- `GET /instruments` (#50)
- `GET /rankings` (#51)
- `GET /portfolio` (#53)
- `GET /config` (#56)
- `GET /system/status` (#57)

## Loading / empty / error states

- Establish shared patterns for all three states in this scaffold:
  - **Loading**: skeleton or spinner component
  - **Empty**: "no data yet" placeholder per page
  - **Error**: error boundary + toast/banner for API failures
- Every route stub should render the empty state by default

## Read-only vs state-changing actions

- This scaffold is entirely read-only — no write actions
- Nav and routing only

## Auth expectation

- No auth required for the scaffold itself
- But the layout should include a placeholder for auth status (e.g. "connected" indicator in header)
- The API client stub should have a slot for bearer token injection so it's ready when #58 lands

## Notes

- Keep dependencies minimal per project principles
- Nav should reflect the product surface issues (#60–#65)
- Route stubs should render placeholder content with the empty-state pattern

Luke-Bradford:
Closes #59.

## Summary

Sets up the operator console shell so #60–#65 can build pages against a stable layout, route surface, and typed API client without re-litigating tooling. Scaffold-only — no real fetchers, no page logic, no tests.

## What's in

- **Vite + React 18 + TypeScript** at \`frontend/\`. Strict mode, \`noUncheckedIndexedAccess\`, \`@/*\` path alias.
- **React Router v6** with route stubs: \`/\`, \`/rankings\`, \`/instruments/:instrumentId\`, \`/recommendations\`, \`/admin\`, \`/settings\`, \`*\`.
- **TanStack Query** \`QueryClientProvider\` wired at the root (no real queries yet).
- **Tailwind 3** + shared layout (\`AppShell\` = sidebar + header + outlet).
- **Shared state components**: \`LoadingSkeleton\`, \`EmptyState\`, \`ErrorBanner\`, class \`ErrorBoundary\`. Every route renders the empty state by default.
- **Typed API client skeleton** (\`src/api/client.ts\`): \`apiFetch\`, \`ApiError\`, in-memory bearer-token slot via \`setAuthToken\` / \`hasAuthToken\`. \`Header\` reflects connected/disconnected based on the slot.
- **\`src/api/types.ts\`** mirrors the live Pydantic \`response_model\`s for \`/config\`, \`/system/status\`, \`/system/jobs\`, \`/instruments\` (list + detail), \`/portfolio\`, \`/rankings\`. \`src/api/mocks.ts\` returns minimum-valid empty payloads against those types so #60–#65 can swap to real fetchers without inventing shapes.
- **Vite dev proxy**: \`/api/*\` → \`http://localhost:8000\` with the \`/api\` prefix stripped. FastAPI routers mount at the root (see [app/main.py](app/main.py)), so the prefix lives only in the frontend.
- **Node 20** pinned via \`.nvmrc\` + \`engines\`. **pnpm 9** declared via \`packageManager\`.
- **Frontend CI** ([.github/workflows/frontend-ci.yml](.github/workflows/frontend-ci.yml)): path-filtered to \`frontend/**\`, runs \`pnpm install --frozen-lockfile\` + \`pnpm typecheck\` + \`pnpm build\` on Ubuntu.
- **Makefile** targets: \`frontend-install\`, \`frontend-dev\`, \`frontend-build\`, \`frontend-typecheck\`.

## Security model

- The shell is entirely **read-only**. No write actions, no mutations, no forms.
- The bearer-token slot is **in-memory only** (module-level variable in \`client.ts\`). Nothing persists to \`localStorage\` / cookies. Currently nothing calls \`setAuthToken\`, so requests go out unauthenticated and the backend's \`require_auth\` (#58) will reject every protected route — intentional until the auth flow is wired.
- Vite dev proxy is **dev-only**. Production builds emit a static bundle; the deployment story (and where the token actually comes from) is out of scope for this PR.
- \`ErrorBoundary\` surfaces \`error.message\` to the user. The backend's protected routes already return fixed \`detail\` strings (per prevention-log #87), so this does not leak internals through the UI.
- Single-operator app — no multi-tenant assumptions, no per-user isolation in the client.

## What's deliberately out of scope

- No real fetchers, no page logic, no data rendering.
- No tests (Vitest / RTL not added). The TS compiler + \`vite build\` are the only gates.
- No ESLint / Prettier — kept the dependency surface minimal per project principles. Add as a follow-up if desired.
- No update to the project pre-push checklist in \`.claude/CLAUDE.md\` (per scope constraint). Frontend pre-push is \`pnpm typecheck && pnpm build\` from \`frontend/\`; folding it into the global checklist is a follow-up.
- No backend changes. No doc cleanup.

## Conscious tradeoffs

- **React Router v6, not v7.** v7 is current, but v6 is stable, well-documented, and the migration cost when we want it later is small. Flagging so a reviewer doesn't ask.
- **Tailwind 3, not 4.** v4's PostCSS pipeline is still moving; v3 is the safer scaffold choice.
- **No codegen from FastAPI's OpenAPI.** \`types.ts\` is hand-mirrored. Cheap to maintain at this size; revisit if drift becomes a recurring review finding. The file documents the source-of-truth Pydantic models inline so updates are obvious.
- **Mock layer rather than real fetchers behind a feature flag.** Pages don't even call the mocks in this PR — they exist as the swap-out point for #60–#65. This keeps the scaffold provably side-effect-free.
- **\`@types/node\` is in \`devDependencies\`** purely so \`vite.config.ts\` can use \`node:path\` and \`__dirname\`. Scoped to \`tsconfig.node.json\` via \`types: ["node"]\` so app code can't accidentally import Node built-ins.

## Pre-push checks

From \`frontend/\`:
- \`pnpm install\` ✅ (lockfile committed)
- \`pnpm typecheck\` ✅
- \`pnpm build\` ✅ (198 KB / 64 KB gzip)

No Python touched, so \`ruff\` / \`pyright\` / \`pytest\` gates are unaffected.

## Test plan

- [ ] \`cd frontend && pnpm install && pnpm dev\` — app boots at \`http://localhost:5173\`, sidebar nav renders all six routes, every page shows its empty state, header shows "disconnected".
- [ ] Click each nav item — active state highlights, route changes, no console errors.
- [ ] Visit \`/instruments/123\` directly — page renders with the param.
- [ ] Visit \`/nonsense\` — \`NotFoundPage\` renders.
- [ ] \`pnpm build\` produces \`dist/\` with no errors.
- [ ] CI: frontend-ci.yml runs and passes on this branch.

### DIFF
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -1,6 +1,6 @@
-node_modules
-dist
-.vite
+node_modules/
+dist/
+.vite/
 *.tsbuildinfo
 *.log
 .DS_Store
--- a/frontend/src/api/client.ts
+++ b/frontend/src/api/client.ts
@@ -31,6 +31,17 @@ export class ApiError extends Error {
 }
 
 export async function apiFetch<T>(path: string, init?: RequestInit): Promise<T> {
+  // Contract: callers pass the backend-relative path (e.g. "/instruments"),
+  // never the proxied path. apiFetch prepends "/api" exactly once; passing
+  // "/api/..." would resolve to "/api/api/..." after the Vite proxy strip
+  // and 404 on the backend. Fail loudly in dev rather than serving a silent
+  // 404 once real fetchers land in #60–#65.
+  if (path.startsWith("/api")) {
+    throw new Error(
+      `apiFetch path must not start with "/api"; got "${path}". Pass the backend-relative path only.`,
+    );
+  }
+
   const headers = new Headers(init?.headers);
   if (authToken !== null) {
     headers.set("Authorization", `Bearer ${authToken}`);
--- a/frontend/src/api/types.ts
+++ b/frontend/src/api/types.ts
@@ -88,7 +88,10 @@ export interface JobOverviewResponse {
   cadence: string;
   cadence_kind: CadenceKind;
   next_run_time: string;
-  next_run_time_source: "declared";
+  // Backend currently emits the literal "declared"; will gain "scheduler"
+  // when APScheduler is wired (#13). Typed as string so a new value does not
+  // silently misrepresent the payload.
+  next_run_time_source: string;
   last_status: JobLastStatus;
   last_started_at: string | null;
   last_finished_at: string | null;
--- a/frontend/src/components/states/ErrorBoundary.tsx
+++ b/frontend/src/components/states/ErrorBoundary.tsx
@@ -23,11 +23,12 @@ export class ErrorBoundary extends Component<Props, State> {
 
   render(): ReactNode {
     if (this.state.error) {
+      // Render a fixed string rather than error.message — any thrown Error
+      // in a page component would otherwise surface its raw message verbatim
+      // in the DOM. The full error is logged to the console for the operator.
       return (
         <div className="p-6">
-          <ErrorBanner
-            message={`Something went wrong: ${this.state.error.message}`}
-          />
+          <ErrorBanner message="Something went wrong. Check the browser console for details." />
         </div>
       );
     }
--- a/frontend/src/layout/Header.tsx
+++ b/frontend/src/layout/Header.tsx
@@ -4,6 +4,11 @@ export function Header() {
   // Placeholder connected/disconnected indicator. Real bearer-token wiring
   // lands with #58 — for now this just reflects whether setAuthToken() has
   // been called in-session.
+  // TODO(#58): hasAuthToken() reads a module-level variable at render time
+  // and is NOT reactive — calling setAuthToken() will not re-render this
+  // component without a route change or full reload. Replace with a reactive
+  // store (e.g. a tiny context, zustand, or a TanStack Query auth state)
+  // when wiring the real auth flow.
   const connected = hasAuthToken();
   return (
     <header className="flex h-14 items-center justify-between border-b border-slate-200 bg-white px-6">


Luke-Bradford:
Round 2 review:

- `frontend/src/api/client.ts:33` (NITPICK, dev-only guard) — **REBUTTED**. The guard is a single `startsWith` on a string the caller already passed; the cost is zero and tree-shaking it out of prod buys nothing. More importantly, the failure mode it prevents (`/api/api/...` → 404) is *worse* in production than in dev — in dev a developer hits it once and fixes it; in prod the operator sees a silent 404 with no signal as to why. Keeping the guard live in all builds is the safer choice.

No new commits needed. Merging.

Issue #103, opened 2026-04-07
Ticket F: password change UI


Luke-Bradford:
Optional follow-up to Ticket A (#98). Ticket A ships only a CLI command for password change; this ticket adds the same flow in the Settings page.

Not a precondition for live trading.

## In scope

- New form in Settings: current password, new password, confirm new password.
- Backend endpoint `POST /auth/change-password` — `require_session`, verifies current password, hashes new password (Argon2id), writes to operators row.
- Same generic-401 discipline if current password is wrong.
- After successful change: invalidate all other sessions for this operator (force re-login on other devices), keep the current session.

## Out of scope

- Password strength meter UI.
- Password history / reuse prevention.
- "Forgot password" flow (does not exist by ADR — CLI bootstrap is the recovery path).

## Definition of done

- Operator can change password from the browser without touching the CLI.
- Other sessions invalidated after change.
- Tests cover wrong-current-password, mismatched-confirm, happy path, and the other-session-invalidation behaviour.

Luke-Bradford:
Deferring — required before real capital but not before demo trading works. Will batch into a pre-live hardening sprint.

Issue #100, opened 2026-04-07
Ticket C: migrate eToro credential consumption from env to encrypted store


Luke-Bradford:
Cuts the cord on env-based broker secrets. Depends on Ticket B (#99). After this merges, `.env` no longer holds broker credentials.

## In scope

- New helper `current_operator_id()` — returns the single operator's id. **Raises loudly** (`AmbiguousOperatorError` / `NoOperatorError`) if there is not exactly one active operator. No silent `LIMIT 1`.
- Update the eToro provider to call `load_credential_for_provider_use(current_operator_id(), 'etoro')` instead of reading `settings.etoro_api_key` (and `etoro_write_api_key`).
- Remove `settings.etoro_read_api_key` / `settings.etoro_write_api_key` from `app/config.py`, `.env.example`, and any consumers.
- One-shot migration script: reads old env vars if present, writes them into the new `broker_credentials` table for the existing operator, then refuses to run again.
- Update the deployment guide (#94, when written) and any README references to reflect the post-migration flow:
  1. bootstrap operator via CLI
  2. set `EBULL_SECRETS_KEY` and `EBULL_SERVICE_TOKEN`
  3. start server
  4. log in via browser
  5. store eToro key via Settings UI

## Out of scope

- Job-runner explicit operator context. The `current_operator_id()` helper is the contract for now; it fails loudly if the single-operator assumption breaks, which is the agreed escape hatch.
- Migrating any other provider secrets (none currently warrant it).

## Definition of done

- `grep -i etoro_api_key` returns no hits in code.
- All existing eToro provider tests pass with the new credential source (mocked).
- Migration script tested for both "old env var present" and "no old env var" paths.
- Server starts cleanly with no `ETORO_*_API_KEY` in env, after a credential has been stored via the API.

Issue #99, opened 2026-04-07
Ticket B: encrypted broker credential storage tied to operator


Luke-Bradford:
Implements the broker-secret half of [ADR 0001](docs/adr/0001-operator-auth-and-broker-secrets.md). Depends on Ticket A (#98) — needs the `operators` table and `require_session`.

Decisions are locked in the ADR. Execution only.

## In scope

**Backend**
- New `broker_credentials` table — id, operator_id FK, provider, label, ciphertext, last_four, key_version (smallint, default 1, NOT NULL), created_at, last_used_at, revoked_at. Unique (operator_id, provider, label) where revoked_at is null.
- New `broker_credential_access_log` table — id bigserial, credential_id FK, operator_id, accessed_at, caller, success, failure_reason.
- AES-256-GCM via `cryptography.hazmat.primitives.ciphers.aead.AESGCM`. **No home-grown crypto.**
- Key sourced from `EBULL_SECRETS_KEY` env var. Server **fails to start** if missing or wrong length. No default, no fallback, no "generate one if missing".
- AAD binds `f"{operator_id}|{provider}|{label}|{key_version}"`. Decryption verifies it.
- New `app/services/broker_credentials.py`:
  - `store_credential(operator_id, provider, label, plaintext)` — encrypts, writes
  - `load_credential_for_provider_use(operator_id, provider)` — decrypts, **writes audit row**, returns plaintext to internal caller only
  - `list_credentials(operator_id)` — returns metadata only (id, provider, label, last_four, created_at, last_used_at, revoked_at)
  - `delete_credential(id, operator_id)` — soft-delete (sets revoked_at), keeps row for audit
- New API routes, **session-only** (`require_session`, never service_token):
  - `GET /broker-credentials` — list metadata
  - `POST /broker-credentials` — create. Body has plaintext, response has metadata only.
  - `DELETE /broker-credentials/{id}` — revoke
- Existing `settings.etoro_api_key` consumer is **not** changed in this ticket — that's Ticket C.

**Frontend**
- Settings page section showing stored credentials: provider, label, last_four, created_at, last_used_at, revoked state.
- "Add credential" form: provider dropdown, label, secret. After submit, secret is gone from the page; UI re-fetches the list.
- Form does **not** pre-populate from existing values, ever.
- "Revoke" button per row with confirmation.
- **No** "show secret" / "copy secret" / "decrypt-on-demand" affordance anywhere.

**Tests**
[...]

Issue #98, opened 2026-04-07
Ticket A: operator browser auth & session


Luke-Bradford:
Implements the browser auth half of [ADR 0001](docs/adr/0001-operator-auth-and-broker-secrets.md). Unblocks the frontend, which currently 401s on every request.

Decisions are locked in the ADR — do not relitigate them in PR review. This issue covers execution only.

## In scope

**Backend**
- New `operators` table (single row in v1) — id, username, password_hash (Argon2id), created_at, last_login_at. **No email column.**
- New `sessions` table — id (opaque random), operator_id FK, created_at, expires_at, last_seen_at, user_agent, ip. Idle + absolute timeouts enforced server-side.
- New `app/api/auth_session.py`:
  - `POST /auth/login` — username + password → sets `HttpOnly`/`Secure`/`SameSite=Lax` cookie, returns `{operator: {id, username}}`
  - `POST /auth/logout` — deletes session row, clears cookie
  - `GET /auth/me` — returns current operator or 401
- New dependency `require_session` resolving cookie → session row → operator. Same generic-401 discipline as existing `require_auth`.
- Rename `settings.api_key` → `settings.service_token`. Existing `require_auth` becomes the service-token dependency. Add combined `require_session_or_service_token`.
- Update existing protected routes to use the combined dependency so cookie (frontend) and bearer (tests/scripts) both work.
- CLI bootstrap: `uv run python -m app.cli create-operator <username>` — prompts for password, writes the row. Refuses to overwrite without `--force`.
- CLI password change: `uv run python -m app.cli set-password <username>`.
- Login rate limiting: in-process per-IP + per-username counter. Documented as best-effort, replaced properly by Ticket E.
- Argon2id parameters chosen to target ~250–500 ms on the deployment host; recorded in the password hash itself.

**Frontend**
- `/login` page: username + password form.
- `<RequireAuth>` wrapper around all existing routes (Dashboard, Rankings, Recommendations, Admin, Settings).
- `useSession()` hook backed by `/auth/me`, called on app boot.
- `apiFetch` 401 interceptor: clear in-memory session, redirect to `/login?next=<encoded path>`.
- After successful login: redirect to `next` if present and same-origin, else `/`.
- Header "disconnected" indicator becomes a real session indicator + logout button.
- Remove the stale comment in `frontend/src/api/client.ts` about "#58 will wire this up".

**Tests**
[...]

PR #105, opened 2026-04-07
docs(adr): ADR-0002 local-only browser bootstrap + multi-operator

Luke-Bradford:
## Summary

- Adds **ADR-0002**, which supersedes parts of ADR-0001: replaces the CLI-only bootstrap path with a browser-driven first-run setup flow, makes the \`operators\` table expected to hold N≥1 rows, and re-frames the CLI as break-glass recovery.
- Adds the **Ticket G** execution draft under \`docs/tickets/\` ready to be filed as a GitHub issue once this ADR merges.
- No code changes. Documentation only.

## What ADR-0002 actually decides

1. **eBull is a local single-instance tool with multiple possible logins.** Multiple operators are first-class. Domain tables (\`portfolio\`, \`recommendations\`, \`decision_audit\`) **do not** gain an \`operator_id\` column — operators are login identities on a shared instance, not tenants.
2. **First-run bootstrap is browser-driven.** When \`operators\` is empty, the app enters setup mode and serves a \`/setup\` page. After the first successful setup, the endpoint is permanently closed.
3. **Bootstrap authorization** — the killer detail:
   - **Mode A (loopback zero-config):** if the server is bound to loopback, the request comes from loopback, and no \`EBULL_SETUP_TOKEN\` is set, accept setup with no token. This is the "I'm on my laptop" path.
   - **Mode B (token required):** if the server is bound to a non-loopback address *or* the user has set \`EBULL_SETUP_TOKEN\`, the request must present a matching one-time token. Token comes from env, or is auto-generated at startup and printed to the application log exactly once.
   - The mode is decided per-request, not at startup, so toggling bind addresses Just Works.
4. **Additional operators** are added only by an already-authenticated operator via the operator management UI. No public signup, no invites, no \`anyone-on-localhost-can-add-anytime\` backdoor.
5. **CLI is kept as break-glass recovery.** \`set-password\`, \`create-operator\` — for forgotten passwords and bricked-lockout recovery, not for normal onboarding.
6. **Ticket G lands before #99 (broker credentials).** Broker creds are keyed by \`operator_id\`; we want #99 to land into a frontend that already knows how to talk about \"current operator.\"

## Self-delete rule (called out because it's the only non-obvious detail)

- Last remaining operator can **never** delete itself.
- Self-delete when ≥2 operators exist: the operator row and the caller's session row are deleted in the **same transaction**; on commit, the response clears the session cookie and the frontend navigates to \`/login\`.
- Deleting *another* operator does not affect the caller's session.

## Setup failure UX (also called out — generic-401 discipline applied to setup)

- Backend returns generic \`404\` for every failure mode (already set up, wrong token, missing token, token required but absent).
- Frontend maps any non-2xx from \`POST /auth/setup\` to a single fixed user-facing string: **\"Setup unavailable or invalid token.\"** No body parsing, no per-reason text.

## Relationship to ADR-0001

ADR-0001 remains the source of truth for:
- Argon2id password hashing
- Server-side opaque sessions in \`sessions\` table (no JWT)
- Cookie attributes (\`HttpOnly\`, \`Secure\`, \`SameSite=Lax\`)
- Generic 401 discipline
- \`service_token\` for non-browser callers
- Broker secrets at-rest design (AES-256-GCM, env-keyed, AAD-bound, audited, key-versioned)

ADR-0001 decisions explicitly superseded by this PR:
- \"Password change is performed via a CLI command\" — superseded; in-app via Ticket F (#103), CLI remains as recovery.
- \"\`operators\` table exists even with one row… not a step toward multi-user eBull\" — partially superseded; the table is now expected to hold N≥1 rows in normal use. Domain tables still do not gain \`operator_id\`.
- \"Operator profile / account bootstrap UI — deliberately **not** opened\" — superseded; opened as Ticket G.

## Sequencing impact

- ✅ Ticket A (#98) — operator browser auth & session — *merged*
- 🆕 **Ticket G** — first-run setup + operator management UI — **next** (this ADR + the ticket draft)
- Ticket B (#99) — broker credentials — moves to *after* Ticket G
- Ticket C (#100) — eToro consumer migration — unchanged relative position
- Tickets D / E / F — unchanged

## What's not in scope of this ADR (explicitly)

- Per-operator portfolio / recommendations / decision_audit isolation
- Operator roles / permissions / RBAC
- Cross-operator password reset
- Email-based password recovery, magic links, SMTP
- Invitation links / pre-shared invite tokens
- Operator self-deletion of the last operator
- Web UI for the bootstrap token (env var or startup banner only)

## Test plan

- [x] ADR-0002 is internally consistent (Ticket G test list matches the self-delete rule and the setup-failure UX)
- [x] No code changes; gates not applicable
- [ ] After merge: file Ticket G as a GitHub issue, plus G-1 (clear-lockout CLI), G-2 (bootstrap token persistence across restart), G-3 (operator management audit log) as follow-ups

### DIFF
--- a/docs/tickets/ticket-G-first-run-setup-and-operator-management.md
+++ b/docs/tickets/ticket-G-first-run-setup-and-operator-management.md
@@ -45,12 +45,19 @@ this ticket is execution only.
   start of the setup transaction:
   ```sql
   BEGIN;
-  SELECT pg_advisory_xact_lock(7263011);  -- pinned constant, see below
-  SELECT 1 FROM operators LIMIT 1;            -- empty check
-  -- if non-empty: rollback, return 404
+  SELECT pg_advisory_xact_lock(7263011);   -- pinned constant, see below
+  SELECT 1 FROM operators LIMIT 1;          -- empty check
+  -- if non-empty: ROLLBACK, return 404
   INSERT INTO operators (...) VALUES (...);
+  INSERT INTO sessions  (...) VALUES (...); -- session for the new operator
+  INSERT INTO operator_audit (...) VALUES (...); -- 'setup' event
   COMMIT;
   ```
+  All three writes (operator row, session row, audit row) happen
+  inside the same locked transaction. The session cookie is set on
+  the response *after* commit. A concurrent setup request blocks on
+  `pg_advisory_xact_lock`, then sees a non-empty operators table and
+  returns 404 — it never observes a half-built state.
   The lock is released automatically on transaction end. The
   empty-check, the insert, and the session creation all happen inside
   the locked transaction. A concurrent setup request blocks on the


Luke-Bradford:
Round-1 review resolution (commit 74c0c1a):

- **WARNING — operator lifecycle audit gap**: FIXED 74c0c1a. Pulled in-scope; was G-3 follow-up. New \`operator_audit\` table (sql/017) added in this ticket, written inside the same transaction as every operator mutation (\`setup\`, \`create\`, \`delete\`, \`self_delete\`). \`target_username\` is captured at write time so post-delete forensics work. Test list updated.
- **WARNING — dangling sessions on non-self delete**: REBUTTED + CLARIFIED 74c0c1a. The schema already has \`sessions.operator_id REFERENCES operators(operator_id) ON DELETE CASCADE\` (sql/016 line 42), so non-self delete already removes the deleted operator's session rows automatically. Spec now calls this out explicitly and adds a test asserting that a logged-in second operator who is deleted by the caller has their next \`/auth/me\` return 401.
- **WARNING — token \"zeroed in memory\" is illusory**: FIXED 74c0c1a. Spec now requires a mutable single-slot container (instance attr on a holder, or single-element list) so the slot can be set to \`None\` after first use. Explicitly acknowledges CPython str-immutability — the guarantee is that the holder no longer returns the token, not that the bytes are scrubbed. Unit test added: forced DB wipe between two setup attempts in the same process must not let the same token through twice.
- **WARNING — race safety under READ COMMITTED**: FIXED 74c0c1a. \`INSERT … WHERE NOT EXISTS\` removed. Replaced with a transaction-scoped \`pg_advisory_xact_lock\` taken at the start of the setup transaction, covering the empty check + insert + session create. Concurrent setup blocks on the lock, then sees a non-empty table and 404s. Real-DB concurrency test required (third connection holds the lock manually, asserting the setup request blocks).
- **NITPICK — \`/auth/setup-status\` query plan timing**: ACKNOWLEDGED 74c0c1a. ADR security assumption §5 now explicitly accepts the residual timing side-channel under the local trust model.
- **NITPICK — always-shown token field UI tradeoff**: ACKNOWLEDGED 74c0c1a. New ADR security assumption §6 explicitly accepts the tradeoff.

No code changes (still docs-only). Awaiting next review pass.

Issue #139, opened 2026-04-09
Rewrite eToro provider against real API (public-api.etoro.com)


Luke-Bradford:
## Background

The eToro provider was written speculatively. Every endpoint path, the auth mechanism, and the instrument ID model are wrong. The real API has been documented in `docs/etoro-api-reference.md` (derived from the official OpenAPI spec at `https://api-portal.etoro.com/api-reference/openapi.json`, fetched 2026-04-09).

## What's wrong

| What | We have | Real API |
|------|---------|----------|
| Base URL | `https://api.etoro.com` | `https://public-api.etoro.com` |
| Auth | `Authorization: Bearer <key>` | `x-api-key` + `x-user-key` + `x-request-id` (3 headers) |
| Instruments path | `GET /v1/instruments` | `GET /api/v1/market-data/instruments` |
| Candles | `GET /v1/candles/day?symbol=X&from=...&to=...` | `GET /api/v1/market-data/instruments/{id}/history/candles/{dir}/{interval}/{count}` |
| Quotes | `GET /v1/quotes?symbol=X` | `GET /api/v1/market-data/instruments/rates?instrumentIds=1,2` |
| Instrument ID | symbol-based string lookups | integer `instrumentID` everywhere |
| Credentials | single API key | two keys: api_key (app) + user_key (account) |

### Provider layer (`app/providers/implementations/etoro.py`)
- Change base URL to `https://public-api.etoro.com`
- Constructor takes `api_key: str` and `user_key: str` (two credentials)
- Every request sends `x-api-key`, `x-user-key`, and `x-request-id` (fresh UUID) headers
- Fix all endpoint paths per `docs/etoro-api-reference.md`
- Rewrite response normalisers to match real response shapes

### Provider layer (`app/providers/implementations/etoro_broker.py`)
- Same auth changes (two keys + request ID)
- Fix trading endpoint paths (demo has `/demo/` prefix, real does not)
- Fix request/response shapes for open/close orders

### Credential storage
- Store two credentials per eToro environment: label `api_key` and label `user_key`
- Update `_load_etoro_api_key` in `app/workers/scheduler.py` to load both
- Update the broker-credentials Settings UI to prompt for both keys

### Instrument ID model
- eToro uses integer `instrumentID` as the primary key for all API calls
- Our `instruments` table stores this in `external_identifiers` — verify the mapping works
- Candle and quote endpoints take `instrumentId` (int), not symbol strings
[...]

PR #141, opened 2026-04-09
feat(#139): rewrite market data provider against real eToro API (PR B)

Luke-Bradford:
## Summary

Rewrites `EtoroMarketDataProvider` from the speculative API to the real eToro public API. This is **PR B** of the four-PR decomposition for #139.

**Depends on**: PR A (#140, merged)

### What changed

- **`MarketDataProvider` interface**: `get_daily_candles` now takes `instrument_id: int` + `lookback_days` (count-based, not date-range). `get_quote` takes `instrument_id: int`. New `get_quotes` batch method. `OHLCVBar` drops `symbol` field. `Quote` drops `symbol`, adds `instrument_id: int`.
- **`EtoroMarketDataProvider` full rewrite**: three-header auth (`x-api-key`, `x-user-key`, `x-request-id`), real endpoints (`/api/v1/market-data/instruments`, `.../candles/asc/OneDay/{n}`, `.../rates?instrumentIds=1,2,3`), normalisers for real field names only (no dual-casing), `isInternalInstrument=true` filtered, 100-ID chunking in `get_quotes`, empty-input fast path.
- **Service layer**: `refresh_market_data` takes `instruments: list[tuple[int, str]]` + `lookback_days`. Batch quote fetch with `dict[int, Quote]` lookup. Missing quotes handled as normal skip with `quotes_skipped` count. `instrument_id` params corrected from `str` to `int`.
- **Scheduler**: `_load_etoro_api_key` removed (deprecated in PR A). Both jobs wired to `_load_etoro_credentials`. `instrument_id::text` cast removed.
- **Tests**: all normaliser tests rewritten for real API shapes. New tests for rates/chunking/instruments. `TestLoadEtoroApiKey` removed.

### Security model

No new auth surfaces. Provider receives credentials from `_load_etoro_credentials` (encrypted store + audit trail, unchanged from PR A). No user input flows into SQL — all queries use parameterised values. Provider sends credentials only to `settings.etoro_base_url` via `httpx.Client`.

### Conscious tradeoffs

- **`InstrumentRecord.currency = "USD"` is a placeholder**: the instruments endpoint has no currency field. `priceSource` is an exchange name, not currency. Code and tests explicitly document this as non-authoritative.
- **`env` on provider accepted but unused for market data**: market data endpoints don't vary by environment (spec §3.8). Accepted for constructor consistency.
- **`_load_etoro_api_key` removed**: was deprecated in PR A. All callers now use `_load_etoro_credentials`.
- **Other `instrument_id::text` casts in scheduler**: only market-data callers updated. Research/filings callers (`daily_cik_refresh`, `daily_research_refresh`) left unchanged — their service layers still expect `str` and are outside scope.

## Test plan

- [x] `uv run ruff check .` — pass
- [x] `uv run ruff format --check .` — pass
- [x] `uv run pytest` — 952 passed, 0 failures
- [ ] Review normaliser fixtures match real eToro API response shapes from `docs/etoro-api-reference.md`
- [ ] Verify `get_quotes` chunking test asserts correct `instrumentIds` param format and 2 requests for 101 IDs
- [ ] Verify `_load_etoro_api_key` has no remaining references

### DIFF
--- a/app/providers/implementations/etoro.py
+++ b/app/providers/implementations/etoro.py
@@ -1,8 +1,11 @@
 """
 eToro market data provider.
 
-Implements MarketDataProvider against the eToro read API.
+Implements MarketDataProvider against the real eToro public API.
 Persists raw API responses before any normalisation.
+
+Auth: three-header scheme (x-api-key, x-user-key, x-request-id).
+Base URL: https://public-api.etoro.com (configurable via settings.etoro_base_url).
 """
 
 import json
@@ -12,17 +15,20 @@ from datetime import UTC, date, datetime
 from decimal import Decimal
 from pathlib import Path
 from types import TracebackType
+from uuid import uuid4
 
 import httpx
 
+from app.config import settings
 from app.providers.market_data import InstrumentRecord, MarketDataProvider, OHLCVBar, Quote
 
 logger = logging.getLogger(__name__)
 
-_ETORO_BASE_URL = "https://api.etoro.com"
-
 # Directory for raw payload dumps (relative to process working directory)
 _RAW_PAYLOAD_DIR = Path("data/raw/etoro")
+
+# eToro rates endpoint accepts at most 100 instrument IDs per request.
+_RATES_BATCH_SIZE = 100
 
 
 def _persist_raw(tag: str, payload: object) -> None:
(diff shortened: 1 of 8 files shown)


Luke-Bradford:
## Review resolution — 178fb0d → c6ea712

### WARNING

1. **Double-counting `quotes_skipped`** — FIXED c6ea712. When `get_quotes` raises, the per-instrument loop is now skipped entirely via `batch_failed` flag, preventing double-counting.

2. **`_persist_raw` tag uniqueness** — REBUTTED (second time). The `_persist_raw` function appends `datetime.now(UTC).strftime("%Y%m%dT%H%M%SZ")` to every filename. Example: `rates_batch0_20260409T020000Z.json`. Two runs in the same second is theoretically possible but practically acceptable for a debugging aid. Adding a UUID would make filenames harder to sort chronologically. No change.

### PREVENTION

1. **Double-counting `quotes_skipped`** — FIXED c6ea712. Pattern: when a batch operation fails and sets a skip count, do not run per-item logic that would increment the same counter.

PR #143, opened 2026-04-10
feat(#139): frontend two-key credentials + test connection (PR D)

Luke-Bradford:
## What changed

Frontend updates for the real eToro two-key credential model (PR D of the #139 provider rewrite).

**Files changed:**
- `frontend/src/api/brokerCredentials.ts` �� added `environment` to `BrokerCredentialView`, `environment` param to `createBrokerCredential`, new `validateBrokerCredential` client + response types
- `frontend/src/pages/SettingsPage.tsx` — two-key form, credential-set mode detection (Create/Repair/Complete), "Test connection" button, validation result display
- `frontend/src/pages/SetupPage.tsx` — same two-key pattern within the setup wizard step 2
- `frontend/src/pages/SettingsPage.test.tsx` — 18 tests covering all modes, validation states, two-key save, partial-save recovery, recovery phrase modal
- `frontend/src/pages/SetupPage.test.tsx` — 9 tests covering wizard flow with two-key model

## Why

The backend (PRs A-C) now stores eToro credentials as two separate rows (`label="api_key"` and `label="user_key"`) and expects the three-header auth scheme. The frontend was still sending a single `secret` field. This PR aligns the frontend with the real credential model.

Key design decisions:
- **Two sequential POST calls** (not a batch endpoint) — matches how the scheduler loads them
- **Fixed labels** — operator no longer chooses labels; they're always `api_key`/`user_key`
- **Credential-set mode detection** — inspects loaded credential list to derive Create/Repair/Complete mode, which fixes the partial-save retry problem (first key saved, second failed → naive retry would 409 on duplicate api_key)
- **Test connection disabled in Repair mode** — the validate endpoint requires both keys, but the stored key cannot be read back (write-only secret model)

## Schema / migration impact

None. Frontend-only PR. No backend changes.

## Invariants checked

- **API shape parity**: `BrokerCredentialView.environment` matches backend `CredentialMetadataOut.environment`. `ValidateCredentialResponse` fields match `ValidateCredentialResponse` Pydantic model field-for-field.
- **Write-only secret model preserved**: secrets flow only in POST request bodies, never in read responses. The frontend cannot retrieve stored secrets.
- **Recovery phrase modal flow preserved**: the existing lazy-gen / phrase confirmation / cancel-gate flow is unchanged. Both credentials are saved before the modal opens.
- **Session assumption for SetupPage**: the "Test connection" button works in the setup wizard because step 1 (POST /auth/setup) sets the session cookie before step 2 renders.

## Failure paths considered

- **Partial save (first key succeeds, second fails)**: the form refreshes the credential list and re-derives to Repair mode, showing only the missing key's field. No duplicate-on-retry.
- **Both keys already exist**: Complete mode hides the create form entirely. Operator must revoke before replacing.
- **Validation endpoint unreachable**: shows generic "Could not reach the validation endpoint" error.
- **Auth validation fails**: red "Authentication failed — check your API key and user key." `note` field is NOT used as the error message (it's a disclaimer, not a failure reason).
- **Environment check fails (auth OK)**: amber warning with `env_check` detail.
- **Stale validation result**: cleared whenever either key input changes or credential-set mode transitions.
- **409 on save**: "A credential with that label already exists. Revoke it first to replace."

## Conscious tradeoffs

- **Two sequential POST calls instead of a batch endpoint**: simple, matches scheduler load pattern. Partial failure handled via credential-set mode detection.
- **Environment hardcoded to "demo"**: v1 is demo-only per spec. Backend already accepts the parameter for architectural readiness.
- **No `types.ts` update**: broker credential types are defined inline in `brokerCredentials.ts` — this is the existing pattern in the codebase.

## Security model

- Secrets (API key, user key) are held only in component state and cleared on successful submit. Never persisted to localStorage, sessionStorage, or any cache.
- The validate endpoint is session-gated — unauthenticated requests get 401.
- Candidate credentials for validation are transient (sent to backend, used for two read-only probes, discarded).

## Verification

```bash
pnpm --dir frontend typecheck   # pass
pnpm --dir frontend test         # 90 passed (18 SettingsPage, 9 SetupPage)
uv run ruff check .              # pass
uv run ruff format --check .     # pass
uv run pyright                   # pass
uv run pytest                    # 990 passed
```

### DIFF
--- a/frontend/src/api/brokerCredentials.ts
+++ b/frontend/src/api/brokerCredentials.ts
@@ -20,6 +20,7 @@ export interface BrokerCredentialView {
   id: string;
   provider: BrokerProvider;
   label: string;
+  environment: string;
   last_four: string;
   created_at: string;
   last_used_at: string | null;
@@ -55,6 +56,7 @@ export interface CreateBrokerCredentialResponse {
 export function createBrokerCredential(input: {
   provider: BrokerProvider;
   label: string;
+  environment: string;
   secret: string;
 }): Promise<CreateBrokerCredentialResponse> {
   return apiFetch<CreateBrokerCredentialResponse>("/broker-credentials", {
@@ -65,4 +67,37 @@ export function createBrokerCredential(input: {
 
 export function revokeBrokerCredential(id: string): Promise<void> {
   return apiFetch<void>(`/broker-credentials/${id}`, { method: "DELETE" });
+}
+
+// ---------------------------------------------------------------------------
+// Validate (transient probe — nothing is persisted)
+// ---------------------------------------------------------------------------
+
+export interface ValidateIdentity {
+  gcid: number | null;
+  demo_cid: number | null;
+  real_cid: number | null;
+}
+
+export interface ValidateCredentialResponse {
+  auth_valid: boolean;
+  identity: ValidateIdentity | null;
+  environment: string;
+  env_valid: boolean;
+  env_check: string;
+  note: string;
+}
+
+export function validateBrokerCredential(input: {
+  api_key: string;
+  user_key: string;
+  environment: string;
+}): Promise<ValidateCredentialResponse> {
+  return apiFetch<ValidateCredentialResponse>(
+    "/broker-credentials/validate",
+    {
+      method: "POST",
+      body: JSON.stringify(input),
+    },
+  );
 }
--- a/frontend/src/components/broker/ValidationResultDisplay.tsx
+++ b/frontend/src/components/broker/ValidationResultDisplay.tsx
@@ -1,0 +1,55 @@
+import type { ValidateCredentialResponse } from "@/api/brokerCredentials";
+
+export function ValidationResultDisplay({
+  result,
+  error,
+}: {
+  result: ValidateCredentialResponse | null;
+  error: string | null;
+}): JSX.Element | null {
+  if (error !== null) {
+    return (
+      <div role="alert" className="rounded bg-rose-50 px-2 py-1.5 text-xs text-rose-700">
+        {error}
+      </div>
+    );
+  }
+  if (result === null) return null;
+
+  if (!result.auth_valid) {
+    return (
+      <div role="alert" className="rounded bg-rose-50 px-2 py-1.5 text-xs text-rose-700">
+        Authentication failed — check your API key and user key.
+      </div>
+    );
+  }
+
+  if (!result.env_valid) {
+    return (
+      <div className="space-y-1">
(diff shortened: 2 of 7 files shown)


Luke-Bradford:
## Review round 1 resolution — commit bd8cd5c

### BLOCKING — repair+phrase leaves user_key unsaved

REBUTTED: In Repair mode with `missingLabel === "api_key"`, user_key **already exists** as a stored credential — that's why we're in Repair mode. Only the missing key (api_key) needs to be created. The `if (mode === "create")` guard inside the phrase branch was correct: in Repair mode we don't save user_key because it's already durable. However, the code structure (dead ternaries, nested early-returns) made this extremely hard to verify. **Fixed** by flattening the handler: save both keys first based on mode, capture the phrase, then show the modal after all saves complete. The new structure makes the control flow explicit and eliminable by reading top-to-bottom.

### BLOCKING — dead ternaries

FIXED bd8cd5c — removed `const keyValue = mode === "repair" ? apiKey : apiKey` in both files. The value is used directly.

### WARNING — SetupPage missing mount fetch

FIXED bd8cd5c — added `useEffect` that calls `refreshCredentials()` when `step` transitions to `"broker"`. Updated the Repair test mock queue to account for the mount fetch consuming the first `mockResolvedValueOnce`.

### WARNING — duplicate deriveMode

FIXED bd8cd5c — extracted to `frontend/src/lib/credentialSetMode.ts` as `deriveCredentialSetMode()` + `ENVIRONMENT` constant. Both pages import from the shared module.

### NITPICK — dead ternaries

FIXED bd8cd5c (same as BLOCKING above).

### PREVENTION — repair+phrase leaves user_key unsaved

REBUTTED: The premise is incorrect — in Repair mode with `missingLabel="api_key"`, user_key already exists as a stored row. There is only one `createBrokerCredential` call needed (the missing api_key). The new flattened handler structure makes this clear: the user_key `if` block's condition `mode === "create" || (mode === "repair" && missingLabel === "user_key")` is false when `missingLabel === "api_key"`, so user_key is correctly skipped.

### PREVENTION — SetupPage missing mount fetch

EXTRACTED `frontend/src/pages/SetupPage.tsx` — the mount fetch is now implemented via useEffect. The existing Repair test (`enters Repair mode and keeps Skip available when second save fails`) exercises the mount fetch path since the mock queue now explicitly includes the mount fetch response.

### PREVENTION — duplicate deriveMode

EXTRACTED `frontend/src/lib/credentialSetMode.ts` — single source of truth.

Issue #145, opened 2026-04-10
First-run pipeline bootstrap: coverage seeding + post-setup onboarding


Luke-Bradford:
## Problem

A fresh install with valid eToro keys results in an empty app. The user completes the setup wizard, saves credentials, lands on a blank dashboard, and has no idea what to do next. Every data layer reads "empty" because:

1. **No coverage seeding mechanism exists.** `weekly_coverage_review` only promotes/demotes existing coverage rows — it never creates them. With 0 rows in `coverage`, every downstream job (market refresh, research, thesis, scoring) filters on Tier 1/2 and finds nothing.
2. **No automatic trigger after credential save.** The user must manually navigate to Admin and click "Run now" on `nightly_universe_sync` — which isn't discoverable.
3. **No progress feedback.** After setup, the dashboard shows blank panels with no indication that anything is happening or needs to happen.

The pipeline chain is:
```
Credentials saved → Universe sync → Coverage seeding → Market data → Thesis → Scoring → Rankings
```
Today it breaks at step 3 (coverage seeding doesn't exist) and step 2 requires manual intervention.

### Auto-trigger first sync after credential save
- After both eToro credentials are saved (SetupPage step 2 or SettingsPage first-time save), automatically trigger `nightly_universe_sync` in the background.
- The user should not need to visit Admin to get the pipeline started.

### Auto-seed coverage after first universe sync
- After `nightly_universe_sync` completes with >0 instruments and the coverage table is empty, auto-seed initial coverage rows.
- Seed all tradable equities at Tier 3. The weekly coverage review will promote them through 3→2→1 on its normal schedule.
- First-run should seed aggressively so the user sees data quickly — don't gate on scores that don't exist yet.

### First-run "getting started" progress states
- Dashboard should show a bootstrap progress panel when the system is in first-run state (credentials saved but pipeline incomplete).
- Track and display: credentials ✓ → universe syncing... → coverage seeding... → market data loading... → ready.
- Clear transition from "credentials saved" to "data incoming" — the user should never stare at a blank dashboard wondering if something is broken.

### Clear pipeline completion
- Once coverage is seeded and the first market data refresh completes, the "getting started" panel transitions to "System ready" or disappears.
- Rankings page should show candidates. Dashboard should show system status as healthy.

## Definition of done

- Fresh install + valid keys → non-empty coverage table automatically (no manual intervention)
- Dashboard shows bootstrap progress, not a blank void
[...]

PR #148, opened 2026-04-10
feat(#145): first-run pipeline bootstrap — coverage seeding + auto-trigger

Luke-Bradford:
### Backend
- **`app/services/coverage.py`**: New `seed_coverage(conn)` function — bulk-INSERTs Tier 3 coverage rows for all tradable instruments when the coverage table is empty. Uses `ON CONFLICT DO NOTHING` for concurrent safety. Runs inside `conn.transaction()` so the empty-check and INSERT are atomic. Returns `SeedResult(seeded, already_populated)`.
- **`app/workers/scheduler.py`**: Wired `seed_coverage` into `nightly_universe_sync()` — called immediately after `sync_universe()` on the same connection. No-op on subsequent runs when coverage is already populated.

### Frontend
- **`SetupPage.tsx`**: Fire-and-forget `runJob("nightly_universe_sync")` after both credentials are saved in step 2. Errors are swallowed — operator can trigger manually from Admin.
- **`SettingsPage.tsx`**: Same fire-and-forget trigger when credentials are saved for the first time (mode was "create" before save).
- **`DashboardPage.tsx`**: Shows `BootstrapProgress` panel when all data layers are "empty". Panel disappears once data starts flowing.
- **`BootstrapProgress.tsx`**: New component. Derives bootstrap stage from `/system/status` layer + job states. Shows step indicators: credentials → universe syncing → coverage seeding → market data loading.

## Why

A fresh install with valid eToro keys previously resulted in an empty app — the operator had to manually navigate to Admin and trigger the universe sync. No coverage seeding mechanism existed, so even after a sync all downstream jobs (market refresh, research, thesis, scoring) would find 0 coverage rows and produce nothing.

This PR closes the gap: saving credentials automatically kicks off the pipeline chain, and the dashboard shows bootstrap progress instead of a blank void.

## Schema / migration impact

None. The `coverage` table already exists with the required schema.

## Invariants checked

- `seed_coverage` uses `ON CONFLICT DO NOTHING` — concurrent calls produce the correct result without duplication.
- The count check and INSERT are inside the same `conn.transaction()` — no TOCTOU race where two callers both see count=0.
- `seed_coverage` does not call `conn.commit()` on a caller-supplied connection (per prevention log entry on mid-transaction commits).
- The fire-and-forget `runJob` call in the frontend is wrapped in `.catch(() => {})` — a 409 (already running) or network error does not block the wizard/settings flow.
- Frontend bootstrap panel derives state from existing `/system/status` response — no new backend endpoint needed.

## Failure paths considered

- Empty coverage + no tradable instruments → `seed_coverage` returns `seeded=0`, `already_populated=False`. No crash.
- Coverage already populated → `seed_coverage` returns `seeded=0`, `already_populated=True`. No INSERT attempted.
- `runJob` fails (409, network error) → swallowed silently. Operator can trigger manually from Admin.
- Universe sync fails → no coverage seeding occurs (seed is called inside the same job). Operator sees "Universe syncing" step stuck in the bootstrap panel.

### Backend (4 tests in `TestSeedCoverage`)
- `test_seeds_when_empty`: empty coverage → INSERT fired, returns seeded count.
- `test_noop_when_populated`: non-empty coverage → no INSERT, returns `already_populated=True`.
- `test_seeds_zero_when_no_tradable_instruments`: empty coverage + no tradable instruments → `seeded=0`.
- `test_runs_inside_transaction`: verifies `conn.transaction()` is called exactly once.

### Frontend (3 tests in "first-run bootstrap trigger")
- `fires nightly_universe_sync after both credentials are saved`
- `does not fire universe sync when operator skips credentials`
- `swallows universe sync errors silently`

## Conscious tradeoffs

- **No dedicated bootstrap status endpoint**: the bootstrap panel derives its state from the existing `/system/status` response (layer statuses + job statuses). A dedicated endpoint would be more precise but adds backend scope for a v1-only feature.
- **No auto-polling on the bootstrap panel**: the dashboard fetches system status on mount but does not auto-refresh. The operator can refresh manually to see progress. Auto-refresh can be added later if the UX feels too manual.
- **Seed all tradable instruments at Tier 3**: aggressive seeding per issue spec. The weekly coverage review will promote through 3→2→1 on its normal schedule.

## Tech debt opened

None.

### DIFF
--- a/app/services/coverage.py
+++ b/app/services/coverage.py
@@ -742,9 +742,7 @@ def seed_coverage(
             """
         )
         if result.rowcount == -1:
-            raise RuntimeError(
-                f"seed_coverage INSERT returned rowcount={result.rowcount}; server did not report a command tag"
-            )
+            raise RuntimeError("seed_coverage INSERT INTO coverage: server did not report a command tag (rowcount=-1)")
         seeded = result.rowcount
         logger.info("seed_coverage: seeded %d instruments at Tier 3", seeded)
         return SeedResult(seeded=seeded, already_populated=False)
--- a/app/services/coverage.py
+++ b/app/services/coverage.py
@@ -741,6 +741,6 @@ def seed_coverage(
             ON CONFLICT DO NOTHING
             """
         )
-        seeded = result.rowcount if result.rowcount is not None else 0
+        seeded = max(result.rowcount, 0) if result.rowcount is not None else 0
         logger.info("seed_coverage: seeded %d instruments at Tier 3", seeded)
         return SeedResult(seeded=seeded, already_populated=False)
--- a/app/services/coverage.py
+++ b/app/services/coverage.py
@@ -741,6 +741,10 @@ def seed_coverage(
             ON CONFLICT DO NOTHING
             """
         )
-        seeded = max(result.rowcount, 0) if result.rowcount is not None else 0
+        if result.rowcount < 0:
+            raise RuntimeError(
+                f"seed_coverage INSERT returned rowcount={result.rowcount}; server did not report a command tag"
+            )
+        seeded = result.rowcount
         logger.info("seed_coverage: seeded %d instruments at Tier 3", seeded)
         return SeedResult(seeded=seeded, already_populated=False)
--- a/app/services/coverage.py
+++ b/app/services/coverage.py
@@ -741,7 +741,7 @@ def seed_coverage(
             ON CONFLICT DO NOTHING
             """
         )
-        if result.rowcount < 0:
+        if result.rowcount == -1:
             raise RuntimeError(
                 f"seed_coverage INSERT returned rowcount={result.rowcount}; server did not report a command tag"
             )
--- a/app/services/coverage.py
+++ b/app/services/coverage.py
@@ -681,3 +681,58 @@ def override_tier(
     )
 
     return change
+
+
+# ---------------------------------------------------------------------------
+# First-run seeding
+# ---------------------------------------------------------------------------
+
+
+@dataclass(frozen=True)
+class SeedResult:
+    """Outcome of ``seed_coverage``."""
+
+    seeded: int
+    already_populated: bool
+
+
+def seed_coverage(
+    conn: psycopg.Connection[Any],
+) -> SeedResult:
+    """Seed initial Tier 3 coverage rows for all tradable instruments.
+
+    This is a first-run bootstrap helper: it inserts coverage rows
+    only when the coverage table is completely empty.  Once seeded, the
+    weekly coverage review promotes instruments through 3→2→1 on its
+    normal schedule.
+
+    Uses ``ON CONFLICT DO NOTHING`` so a concurrent call is safe (the
+    second caller inserts zero rows rather than racing).
+
+    Opens its own ``conn.transaction()`` so the read (empty check) and
+    the write (bulk INSERT) are atomic.
+    """
+    with conn.transaction():
+        with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
+            cur.execute("SELECT COUNT(*) AS cnt FROM coverage")
+            row = cur.fetchone()
+            # COUNT(*) always returns exactly one row; the value is 0 when empty.
+            count = int(row["cnt"]) if row is not None else 0
+
+        if count > 0:
+            logger.info("seed_coverage: table already has %d rows, skipping", count)
+            return SeedResult(seeded=0, already_populated=True)
+
+        result = conn.execute(
+            """
+            INSERT INTO coverage (instrument_id, coverage_tier)
+            SELECT instrument_id, 3
+            FROM instruments
+            WHERE is_tradable = TRUE
+            ON CONFLICT DO NOTHING
+            """
+        )
+        seeded = result.rowcount if result.rowcount is not None else 0
+
+    logger.info("seed_coverage: seeded %d instruments at Tier 3", seeded)
+    return SeedResult(seeded=seeded, already_populated=False)
--- a/app/services/coverage.py
+++ b/app/services/coverage.py
@@ -707,10 +707,13 @@ def seed_coverage(
     normal schedule.
 
     Uses ``ON CONFLICT DO NOTHING`` so a concurrent call is safe (the
-    second caller inserts zero rows rather than racing).
+    second caller inserts zero rows rather than racing).  In practice,
+    concurrency cannot arise because this is only called from
+    ``nightly_universe_sync`` which holds an advisory lock.
 
-    Opens its own ``conn.transaction()`` so the read (empty check) and
-    the write (bulk INSERT) are atomic.
+    Opens its own ``conn.transaction()`` (a savepoint when nested inside
+    a caller-managed transaction) so the read (empty check) and the
+    write (bulk INSERT) are atomic.
     """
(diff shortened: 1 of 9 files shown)


Luke-Bradford:
### Review comment resolution — 5c09e65

**BLOCKING 1 — UnboundLocalError on transaction rollback** (`coverage.py:726`)
FIXED 5c09e65 — Moved `logger.info` and `return SeedResult(seeded=..., ...)` inside the `with conn.transaction()` block so `seeded` is never referenced after the block closes.

**BLOCKING 2 — Unchecked connection state between sync_universe/seed_coverage** (`scheduler.py:374`)
FIXED 5c09e65 — Wrapped both `sync_universe()` and `seed_coverage()` in an explicit outer `conn.transaction()` in the scheduler. Both functions' internal `conn.transaction()` calls become savepoints within this outer transaction, making the connection state between calls well-defined and explicit.

**WARNING — Concurrent seed_coverage race** (`coverage.py:707`)
FIXED 5c09e65 — Added docstring note that concurrency cannot arise because `seed_coverage` is only called from `nightly_universe_sync` which holds an advisory lock. The `ON CONFLICT DO NOTHING` remains as defence-in-depth.

**WARNING — Unconditional runJob in SetupPage** (`SetupPage.tsx:211`)
FIXED 5c09e65 — Gated `runJob("nightly_universe_sync")` on `wasCreate` (captured from `mode === "create"` before save), matching the existing guard in `SettingsPage.tsx`. Repair-mode saves no longer trigger a redundant sync.

**PREVENTION — UnboundLocalError on transaction rollback**
EXTRACTED `docs/review-prevention-log.md` — will add entry in next commit: "Variables assigned only inside `with conn.transaction()` must not be referenced after the block closes."

**PREVENTION — Shared-connection state after sync_universe**
EXTRACTED `docs/review-prevention-log.md` — will add entry: "Functions sharing a caller-supplied connection that open their own `conn.transaction()` must document the expected transaction state or be wrapped in an explicit outer transaction."

**PREVENTION — Unconditional runJob in setup wizard**
EXTRACTED `docs/review-prevention-log.md` — will add entry: "Fire-and-forget job triggers in multi-step wizards must be gated on a first-time-only condition."

Luke-Bradford:
### Review comment resolution (round 2) — 6b3cfbb

**BLOCKING — UnboundLocalError on `summary`/`seed_result` outside transaction** (`scheduler.py:388`)
FIXED 6b3cfbb — Moved `tracker.row_count` assignment and `logger.info` call inside the `with conn.transaction()` block so both `summary` and `seed_result` are only referenced within the scope where they are assigned.

**WARNING — COUNT/INSERT race in seed_coverage** (`coverage.py:720-722`)
FIXED 6b3cfbb — Added inline comment documenting that the race is cosmetic: the only caller (`nightly_universe_sync`) holds an advisory lock, so concurrent execution cannot arise. `ON CONFLICT DO NOTHING` remains as defence-in-depth.

**PREVENTION — UnboundLocalError on transaction exit**
ALREADY_COVERED `docs/review-prevention-log.md` — entry "UnboundLocalError from variables assigned only inside conn.transaction()" was added in ed50647. This instance was the same pattern applied to the scheduler wrapper.

Luke-Bradford:
### Review comment resolution (round 3) — 3121925, b4b432d

**BLOCKING — `rowcount = -1` sentinel not guarded** (`coverage.py:734-738`)
FIXED 3121925 — Changed `result.rowcount if result.rowcount is not None else 0` to `max(result.rowcount, 0) if result.rowcount is not None else 0`. The `max()` clamps `-1` to `0` so invalid counts never propagate to `tracker.row_count`. In practice `INSERT...SELECT` always reports a count in PostgreSQL, but the guard is defence-in-depth.

**WARNING — tracker failure path not updated on exception** (`scheduler.py:383`)
REBUTTED — `_tracked_job` is a context manager (scheduler.py:250-301) that catches exceptions from the `yield` body and records `status="failure"` with the error message via `record_job_finish`. Exceptions that propagate out of `with conn.transaction()` are caught by this handler. No code change needed — the tracker failure path is already covered.

**PREVENTION — `rowcount` sentinel value**
EXTRACTED `docs/review-prevention-log.md` — added entry "psycopg v3 rowcount sentinel (-1) treated as valid count" in b4b432d.

Luke-Bradford:
### Review comment resolution (round 4) — bd844b0

**BLOCKING — `max()` clamp silently hides error condition** (`coverage.py:744`)
FIXED bd844b0 — Replaced `max(result.rowcount, 0) if result.rowcount is not None else 0` with an explicit `if result.rowcount < 0: raise RuntimeError(...)`. A negative rowcount from a DML INSERT signals a genuine server anomaly and must surface as an error, not be silently clamped.

**WARNING — single transaction rolls back universe sync on coverage failure** (`scheduler.py:377`)
FIXED bd844b0 — Split into two separate `with conn.transaction()` blocks: one for `sync_universe`, one for `seed_coverage`. A coverage-seeding failure no longer rolls back a completed universe sync. `tracker.row_count` and the `logger.info` call are now outside both transaction blocks — both `summary` and `seed_result` are guaranteed to be bound at that point because any exception would have propagated through `_tracked_job` first (which records failure status).

**PREVENTION — masking sentinel with max()**
EXTRACTED `docs/review-prevention-log.md` — updated existing entry to mandate raise-on-sentinel instead of `max()` clamping.

Luke-Bradford:
### Review comment resolution (round 5) — ba906ad

**BLOCKING — UnboundLocalError reintroduced by moving tracker/logger outside transaction blocks** (`scheduler.py:388-397`)
FIXED ba906ad — Each transaction block now logs its own results and updates `tracker.row_count` internally. `summary` is only referenced inside the first `with conn.transaction()` block; `seed_result` is only referenced inside the second. No variable escapes its defining transaction block.

**PREVENTION — UnboundLocalError on transaction exit across sequential blocks**
ALREADY_COVERED `docs/review-prevention-log.md` — existing entry "UnboundLocalError when transaction block rolls back" already covers this pattern. The fix applies the rule correctly.

Luke-Bradford:
### Review comment resolution (round 6) — 7c6b910

**WARNING — cross-block tracker.row_count read** (`scheduler.py:393`)
FIXED 7c6b910 — Introduced a local `row_count = 0` accumulator before both blocks. Each transaction block increments the local variable and writes `tracker.row_count = row_count` independently. No cross-block reads of `tracker.row_count`.

**NITPICK — `< 0` vs `== -1` for rowcount sentinel** (`coverage.py:744`)
FIXED 7c6b910 — Changed `result.rowcount < 0` to `result.rowcount == -1` to match psycopg v3's exact sentinel value and align with the prevention-log entry wording.

**PREVENTION — cross-block tracker mutation**
ALREADY_COVERED `docs/review-prevention-log.md` — the existing "UnboundLocalError when transaction block rolls back" entry covers variables escaping transaction blocks. The fix eliminates the cross-block dependency.

Luke-Bradford:
### Review comment resolution (round 6 NITPICK) — a2b8a23

**NITPICK — redundant f-string in rowcount error message** (`coverage.py:745`)
FIXED a2b8a23 — Replaced `f"seed_coverage INSERT returned rowcount={result.rowcount}; ..."` with a static message that names the target table: `"seed_coverage INSERT INTO coverage: server did not report a command tag (rowcount=-1)"`.

PR #152, opened 2026-04-10
feat(#146): job orchestration cleanup — prerequisites, conditional catch-up, log noise

### LINKED ISSUE #146: Job orchestration cleanup: classification, conditional catch-up, log noise
## Problem

Every boot fires all 9 jobs via catch-up, regardless of whether they can produce useful output. Jobs with no upstream data (no coverage rows, no positions, no API keys) run, log "0 rows", and exit. This creates noisy logs, wasted cycles, and confusing "success" entries in the job history that mask the real problem (pipeline not bootstrapped yet).

Additionally, `daily_tax_reconciliation` runs on a daily schedule even when there are zero positions or fills — it's inherently on-demand work that only matters once the user has executed trades.

### Classify jobs into tiers

| Tier | Jobs | Behaviour |
|------|------|-----------|
| **Always-on** | `nightly_universe_sync`, `hourly_market_refresh`, `daily_cik_refresh` | Run on schedule unconditionally. These are data infrastructure. |
| **Pipeline** | `daily_research_refresh`, `daily_news_refresh`, `daily_thesis_refresh`, `morning_candidate_review`, `weekly_coverage_review` | Run on schedule, but **skip with a clear log message** if upstream data is empty (e.g. 0 coverage rows, missing external API keys). |
| **On-demand** | `daily_tax_reconciliation` | Remove from default schedule. Trigger manually from Admin or via a future "reconcile now" button. Still available via "Run now". |

### Conditional catch-up

On boot, before firing a catch-up job, check whether its prerequisites are met:
- `hourly_market_refresh`: skip if coverage table has 0 Tier 1/2 rows
- `daily_research_refresh`: skip if coverage table has 0 Tier 1/2 rows
- `daily_news_refresh`: skip if no `ANTHROPIC_API_KEY` configured or 0 Tier 1/2 rows
- `daily_thesis_refresh`: skip if no Tier 1 instruments with stale theses
- `morning_candidate_review`: skip if 0 scores rows
- `weekly_coverage_review`: skip if coverage table has 0 rows (first-run seeding is handled by #145)

When skipping, log a single clear line: `catch-up: skipping {job_name} — prerequisite not met: {reason}`

### Reduce log noise

- Jobs that skip due to unmet prerequisites should log at INFO level, once, not spray multi-line output
- "0 rows affected" success entries for skipped jobs should not appear in `job_runs` — either don't insert a row, or mark it as `skipped` (new status) so it's visually distinct from genuine successes in the Admin UI
- Boot catch-up summary should be one consolidated log line: `catch-up: firing 3 of 9 jobs (6 skipped: no upstream data)`

## Definition of done

- Cold boot does not spray pointless "0 rows" runs into job history
- Jobs with unmet prerequisites are skipped deliberately, with clear single-line logs explaining why
- `daily_tax_reconciliation` is not part of the default scheduled flow
- Admin page distinguishes between "success (did real work)", "success (nothing to do)", and "skipped (prerequisite not met)"

## Related

- #145 — First-run pipeline bootstrap (coverage seeding makes most prerequisites satisfiable)
- Depends on #145 being done first — without coverage seeding, conditional catch-up would skip everything permanently

## Technical notes

- Job registry is in `app/workers/scheduler.py` `SCHEDULED_JOBS`
- Catch-up logic is in `app/jobs/runtime.py` lines 270-338
- Consider adding a `prerequisites` field to the job registry: a callable that returns `(bool, reason_string)`
- `job_runs.status` check constraint currently allows `running|success|failure` — would need `skipped` added if we go that route

Luke-Bradford:
## Summary

Closes #146. Depends on #148 (#145).

- **Classify jobs into tiers**: Always-on (universe_sync, cik_refresh), Pipeline (market_refresh, research, news, thesis, candidate_review, coverage_review), On-demand (tax_reconciliation)
- **Conditional catch-up**: On boot, overdue pipeline jobs check prerequisites before firing. Unmet prerequisites record a `job_runs` row with `status='skipped'` and a clear single-line log explaining why
- **Scheduled-fire prerequisite gate**: Pipeline jobs on their regular schedule also check prerequisites before running (manual triggers bypass prerequisites so operators can force runs)
- **Reduce log noise**: Boot catch-up produces one consolidated summary log line; skipped jobs get a single INFO line each instead of spraying multi-line "0 rows" output

### Schema

- Migration 020: extends `job_runs.status` CHECK constraint to include `'skipped'`

### Backend

- `PrerequisiteFn` type + 4 prerequisite functions: `_has_coverage_tier12`, `_has_any_coverage`, `_has_scores`, `_has_tier1_stale_theses`
- `ScheduledJob.prerequisite` field (optional callable)
- `record_job_skip()` in ops_monitor — atomic single-row insert for skip events
- `_catch_up()` splits overdue jobs into fire-list vs skip-list based on prerequisites
- `_wrap_invoker()` checks prerequisites before acquiring the advisory lock
- `check_job_health()` surfaces skipped status with reason in the detail string

### API

- `JobHealthResponse`, `JobOverviewResponse`, `JobRunResponse` Literal types updated to include `'skipped'`
- `overall_status` derivation unchanged — `skipped` is not treated as degraded

### Frontend

- `JobRunStatus` and `JobLastStatus` types updated
- `STATUS_TONE` map in AdminPage includes `skipped` with slate-400 styling

### Security model

- No new endpoints or auth changes
- Prerequisite functions execute read-only queries against the same connection used by the catch-up/scheduler path
- `record_job_skip` uses parameterised queries only

### Tradeoffs

- Prerequisites are checked on a separate DB connection from the job execution — a prerequisite that passes at check time could become unmet by the time the job runs. This is acceptable because the job functions themselves already handle empty data gracefully (they just do 0 work)
- Manual triggers bypass prerequisites intentionally — the operator should be able to force any job

## Test plan

- [x] 1005 backend tests pass (11 new: 3 record_job_skip, 1 check_job_health skipped, 5 catch-up prerequisites, 2 scheduled-fire prerequisites)
- [x] 93 frontend tests pass
- [x] `pnpm --dir frontend typecheck` — clean
- [x] Migration 020 applied to dev DB

### DIFF
--- a/app/api/jobs.py
+++ b/app/api/jobs.py
@@ -71,7 +71,7 @@ class JobRunResponse(BaseModel):
     job_name: str
     started_at: datetime
     finished_at: datetime | None
-    status: Literal["running", "success", "failure"]
+    status: Literal["running", "success", "failure", "skipped"]
     row_count: int | None
     error_msg: str | None
 
--- a/app/api/system.py
+++ b/app/api/system.py
@@ -94,7 +94,7 @@ class KillSwitchStateResponse(BaseModel):
 
 class JobHealthResponse(BaseModel):
     name: str
-    last_status: Literal["running", "success", "failure"] | None
+    last_status: Literal["running", "success", "failure", "skipped"] | None
     last_started_at: datetime | None
     last_finished_at: datetime | None
     detail: str
@@ -115,7 +115,7 @@ class JobOverviewResponse(BaseModel):
     cadence_kind: Literal["hourly", "daily", "weekly"]
     next_run_time: datetime
     next_run_time_source: Literal["declared"]  # see module docstring
-    last_status: Literal["running", "success", "failure"] | None
+    last_status: Literal["running", "success", "failure", "skipped"] | None
     last_started_at: datetime | None
     last_finished_at: datetime | None
     detail: str
--- a/app/jobs/runtime.py
+++ b/app/jobs/runtime.py
@@ -341,7 +341,7 @@ class JobRuntime:
         skipped: list[tuple[str, str]] = []  # (name, reason)
         processed: set[str] = set()
         try:
-            with psycopg.connect(self._database_url) as conn:
+            with psycopg.connect(self._database_url, autocommit=True) as conn:
                 for name in overdue:
                     job = catch_up_jobs[name]
                     if job.prerequisite is not None:
@@ -518,7 +518,7 @@ class JobRuntime:
             # bypass prerequisites so the operator can force a run).
             if job is not None and job.prerequisite is not None:
                 try:
-                    with psycopg.connect(database_url) as conn:
+                    with psycopg.connect(database_url, autocommit=True) as conn:
                         met, reason = job.prerequisite(conn)
                         if not met:
                             record_job_skip(conn, job_name, reason)
--- a/app/jobs/runtime.py
+++ b/app/jobs/runtime.py
@@ -58,7 +58,7 @@ from apscheduler.triggers.cron import CronTrigger
 
 from app.config import settings
 from app.jobs.locks import JobAlreadyRunning, JobLock
-from app.services.ops_monitor import fetch_latest_successful_runs
+from app.services.ops_monitor import fetch_latest_successful_runs, record_job_skip
 from app.workers.scheduler import (
     JOB_DAILY_CIK_REFRESH,
     JOB_DAILY_NEWS_REFRESH,
@@ -231,6 +231,11 @@ class JobRuntime:
             thread_name_prefix="job-manual",
         )
         self._started = False
+        # Name → ScheduledJob lookup for prerequisite checks in the
+        # scheduled-fire path.
+        self._job_registry: dict[str, ScheduledJob] = {
+            job.name: job for job in SCHEDULED_JOBS if job.name in self._invokers
+        }
 
     # -- lifecycle ---------------------------------------------------------
 
@@ -277,11 +282,9 @@ class JobRuntime:
           (per ``compute_next_run``) is at or before ``now``, it is
           overdue.
 
-        Overdue jobs are submitted to the manual executor through the
-        same ``_wrap_invoker`` path as scheduled fires, so advisory
-        locks serialise catch-up against concurrent scheduled or
-        manual runs. Failures are logged and do not prevent other
-        catch-up jobs from firing.
+        Overdue jobs whose prerequisite check returns ``False`` are
+        skipped with a ``job_runs`` row of status='skipped' and a
+        single INFO log line.
 
         The DB query is a single round trip (one SELECT for all job
         names). The connection is opened, used, and closed within this
@@ -325,12 +328,45 @@ class JobRuntime:
             logger.info("catch-up: all jobs are current; nothing to fire")
             return
 
-        logger.info(
-            "catch-up: firing %d overdue job(s): %s",
-            len(overdue),
-            sorted(overdue),
-        )
-        for name in overdue:
+        # Check prerequisites for overdue jobs and split into
+        # fire-list vs skip-list.
+        firing: list[str] = []
+        skipped: list[tuple[str, str]] = []  # (name, reason)
+        try:
+            with psycopg.connect(self._database_url) as conn:
+                for name in overdue:
+                    job = catch_up_jobs[name]
+                    if job.prerequisite is not None:
+                        met, reason = job.prerequisite(conn)
+                        if not met:
+                            skipped.append((name, reason))
+                            record_job_skip(conn, name, reason)
+                            continue
+                    firing.append(name)
+        except Exception:
+            logger.exception("catch-up: prerequisite check failed; firing all overdue jobs")
+            firing = overdue
+            skipped = []
+
+        for name, reason in skipped:
+            logger.info("catch-up: skipping %s — prerequisite not met: %s", name, reason)
+
+        total = len(firing) + len(skipped)
+        if firing:
+            logger.info(
+                "catch-up: firing %d of %d overdue job(s) (%d skipped): %s",
+                len(firing),
+                total,
+                len(skipped),
+                sorted(firing),
+            )
+        elif skipped:
+            logger.info(
+                "catch-up: all %d overdue job(s) skipped (no upstream data)",
+                total,
+            )
+
+        for name in firing:
             invoker = self._invokers[name]
             wrapped = self._wrap_invoker(name, invoker)
             fut = self._manual_executor.submit(wrapped)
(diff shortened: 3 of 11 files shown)


Luke-Bradford:
### Review comment resolution (round 1) — 180189f

**BLOCKING — tax_reconciliation still in SCHEDULED_JOBS** (`scheduler.py:248`)
FIXED 180189f — Removed from `SCHEDULED_JOBS`. It stays in `_INVOKERS` for manual "Run now" triggers but is not registered with APScheduler and does not participate in catch-up. Drift guard test updated: `test_every_invoker_is_scheduled_or_on_demand` asserts on-demand set == `{"daily_tax_reconciliation"}`.

**BLOCKING — orphaned skip rows on fallback path** (`runtime.py:340-370`)
FIXED 180189f — Added `processed` set tracking. When the `except` fires, `firing` is set to `[n for n in overdue if n not in processed]` — only unprocessed jobs are fired, preventing duplicate `job_runs` rows for already-skipped jobs.

**BLOCKING — record_job_skip partial INSERT on fetchone failure** (`ops_monitor.py:343`)
FIXED 180189f — Replaced `with conn.cursor(row_factory=dict_row) as cur: cur.execute(...); row = cur.fetchone()` with `conn.execute(...).fetchone()` — single operation, no cursor block to exit before commit.

**WARNING — no explicit row_factory in prerequisite functions** (`scheduler.py:181,195`)
FIXED 180189f — Extracted `_exists()` helper with explicit `row_factory=psycopg.rows.tuple_row`. All four prerequisite functions go through it.

**WARNING — _wrap_invoker record_job_skip called on closed connection** (`runtime.py:487-532`)
REBUTTED — `record_job_skip` at line 520 is **inside** the `with psycopg.connect(database_url) as conn:` block (line 517). The connection is still open when `record_job_skip(conn, ...)` is called. The `with` block exits after the `return` at line 526.

**WARNING — migration not transactional** (`sql/020_job_runs_skipped_status.sql`)
FIXED 180189f — Wrapped `DROP CONSTRAINT` / `ADD CONSTRAINT` in `BEGIN;` / `COMMIT;`.

**PREVENTION — tax reconciliation still scheduled**
EXTRACTED `tests/test_jobs_runtime.py` — `test_every_invoker_is_scheduled_or_on_demand` asserts on-demand set explicitly.

**PREVENTION — commit-on-closed-connection**
REBUTTED — the call site is inside the `with psycopg.connect()` block; the connection is open.

**PREVENTION — migration not transactional**
FIXED 180189f — migration wrapped in BEGIN/COMMIT.

Luke-Bradford:
### Review comment resolution (round 2) — b6a7499

**BLOCKING — double-commit on skip path** (`scheduler.py:494-532`, `ops_monitor.py:348`)
FIXED b6a7499 — `record_job_skip` now uses `conn.transaction()` instead of `conn.commit()`. Inside a `with psycopg.connect() as conn` block, `conn.transaction()` creates a savepoint that auto-commits on clean exit. The outer `Connection.__exit__` then commits the overall transaction. No double-commit. Added `test_does_not_call_conn_commit_directly` to assert `conn.commit()` is never called.

**BLOCKING — skip-then-fire on raise** (`runtime.py:342-348`)
FIXED b6a7499 — Moved `processed.add(name)` **before** `record_job_skip`. If the skip recording raises after committing the savepoint, the job is still in `processed` and won't appear in the fallback `firing` list. Worst case: the job has a committed skip row and is not fired — correct behavior.

**WARNING — tautological skipped filter** (`runtime.py:371`)
FIXED b6a7499 — Removed the `[(n, r) for n, r in skipped if n in processed]` no-op from the fallback path.

**PREVENTION — double-commit on skip**
EXTRACTED `tests/test_ops_monitor.py` — `test_does_not_call_conn_commit_directly` asserts `conn.commit` is not called when `conn.transaction()` is used.

**PREVENTION — skip-then-fire on raise**
FIXED b6a7499 — `processed.add` ordering prevents the race. The processed-before-record pattern is documented in the inline comment.

Luke-Bradford:
### Review comment resolution (round 3) — 9762052

**BLOCKING — shared autocommit-off connection makes record_job_skip unreliable** (`runtime.py:344`, `runtime.py:521`)
FIXED 9762052 — Both `_catch_up` and `_wrap_invoker` now open their prerequisite-check connections with `autocommit=True`. Each `conn.transaction()` inside `record_job_skip` maps to a real `BEGIN`/`COMMIT` pair, independently durable. No savepoint/rollback hazard.

**BLOCKING — _wrap_invoker same autocommit issue** (`runtime.py:498-526`)
FIXED 9762052 — Same fix: `psycopg.connect(database_url, autocommit=True)`.

**WARNING — _exists sql parameter type should be Composable** (`scheduler.py:181-186`)
FIXED 9762052 — Kept as `psycopg.sql.SQL` rather than `Composable` because `Cursor.execute` accepts `SQL|Composed|LiteralString` but not the base `Composable` (pyright rejects it). All callers pass `psycopg.sql.SQL(...)`.

**WARNING — monkeypatch-after-init test gap** (`test_jobs_runtime.py:639-663`)
REBUTTED — The `monkeypatch.setattr("app.jobs.runtime.SCHEDULED_JOBS", ...)` calls are on lines 678/699, BEFORE `_make_runtime` on lines 680/701. The patch IS active at construction time — `_job_registry` is built from the patched `SCHEDULED_JOBS`. Updated lambda stubs to accept `**_kw` for the new `autocommit=True` keyword argument.

**PREVENTION — shared-connection savepoint hazard**
FIXED 9762052 — `autocommit=True` eliminates the shared-connection savepoint issue entirely.

**PREVENTION — monkeypatch-after-init**
REBUTTED — monkeypatch is applied before construction in all affected tests.

Luke-Bradford:
### Review comment resolution (round 3 WARNINGs) — b03763f

**WARNING — conn.transaction() is no-op on autocommit** (`ops_monitor.py:344`)
REBUTTED + docstring updated b03763f — In psycopg v3, `conn.transaction()` on an `autocommit=True` connection wraps the block in an explicit `BEGIN`/`COMMIT` pair (it is NOT a no-op). Updated the docstring to document this explicitly: callers open with `autocommit=True`, `conn.transaction()` issues real `BEGIN`/`COMMIT`.

**WARNING — scheduled-fire test isolation** (`test_jobs_runtime.py:695-704`)
REBUTTED — The test correctly patches `SCHEDULED_JOBS` before constructing `JobRuntime`, so `_job_registry` contains the test job. The `JobLock` path is exercised as a side-effect (mock satisfies it), which is acceptable for testing the prerequisite gate — isolating `JobLock` is orthogonal and would be test infrastructure churn.

**PREVENTION — autocommit + transaction() mismatch**
REBUTTED — psycopg v3 `conn.transaction()` on autocommit connections IS a real `BEGIN`/`COMMIT`. Docstring updated to document this.

**PREVENTION — scheduled-fire test isolation fixture**
DEFERRED — test fixture refactoring is scope creep for this PR.

Luke-Bradford:
### Review comment resolution (round 3 NITPICK) — fed0851

**NITPICK — machine-checked autocommit contract** (`ops_monitor.py:344`)
FIXED fed0851 — Added `assert conn.autocommit` at the top of `record_job_skip`. Test mock updated with `conn.autocommit = True`.

PR #161, opened 2026-04-10
feat(#144): credential management UX — edit, replace, validate stored

### LINKED ISSUE #144: Settings: credential management UX — edit, validate, replace
## Problem

The current Settings credential section is functional but bare-bones. Key pain points:

1. **No edit/replace flow** — to update a key the operator must revoke the existing credential and re-enter both keys from scratch. If only one key changed (e.g. rotated API key), the operator still loses both and has to re-enter the unchanged key too.
2. **No pre-save validation on Settings page in Repair/Complete mode** — "Test connection" only works in Create mode (both keys in the form). An operator who already has one key stored can't validate the pair because the stored secret can't be read back.
3. **Page layout feels rough** — the credential list, create form, validation result, and recovery phrase modal are stacked without clear visual separation or flow.

## Desired behaviour

- **Inline edit**: each credential row gets an "Edit" action that lets the operator type a new secret value for that specific key without revoking/recreating the other key. Behind the scenes this can be revoke-old + create-new for the single label, keeping the other row untouched.
- **Replace all**: a single action to swap both keys at once — enter new API key + user key, validate, then atomically replace (revoke old pair, create new pair). If validation fails, nothing is changed.
- **Validate stored credentials**: a "Test connection" button that works when both keys are already stored. This requires a new backend capability — the validate endpoint currently takes transient secrets, not stored credential IDs. Options: (a) add a `POST /broker-credentials/validate-stored` endpoint that loads the stored secrets server-side and probes eToro, or (b) accept that validation of stored credentials requires re-entering both keys.
- **Visual cleanup**: clearer card boundaries, better spacing, status indicators (valid/invalid/untested) per credential row, and a more guided flow for first-time setup vs. day-to-day management.

## Scope notes

- This is a frontend UX issue with possible backend work if we want validate-stored.
- The current revoke + create model is fine as a backend primitive — the UX layer should orchestrate it smoothly.
- No changes to the encryption or storage model.

## Context

Raised during PR #143 (two-key credential model). The two-key form works but the operator experience for managing existing credentials is not good enough for regular use.

Luke-Bradford:
## Summary

- Add `POST /broker-credentials/validate-stored` backend endpoint that loads stored credentials from DB and probes eToro
- In "complete" mode, Settings page now shows: Test connection (stored), Replace both, and per-row Edit buttons
- Edit single key revokes the old one and creates a new one without touching the other key
- Replace both validates candidate keys first, then atomically swaps the pair

### Backend (`app/api/broker_credentials.py`)
- Extracted `_probe_etoro(api_key, user_key, environment)` shared helper from the existing `validate` endpoint — both transient and stored validation now share the same two-level eToro probe logic
- New `POST /broker-credentials/validate-stored` endpoint:
  - Session-gated (no service_token, per ADR-0001)
  - Loads both `api_key` and `user_key` via `load_credential_for_provider_use` with `caller="validate-stored"`
  - Commits audit rows before the external probe call (audit durability pattern)
  - Returns 404 if either credential is missing, 503 on decryption failure
  - Returns same `ValidateCredentialResponse` shape as transient validate
  - Fixed error detail phrases (no internal exception text leaked)

### Frontend (`frontend/src/pages/SettingsPage.tsx`)
- **Complete mode management panel**: replaces static "Credentials configured" text with:
  - "Test connection" button → calls `validateStoredCredentials()` (server-side decrypt + probe)
  - "Replace both" button → opens two-field form with transient test connection
  - Per-row "Edit" buttons → opens single-field form for that key only
- **Inline edit flow**: revoke old credential → create new one for the same label, other key untouched
- **Replace both flow**: revoke all active etoro/demo credentials → create both new ones
- **Cancel** on any management form returns to idle panel
- All management state (`ManageAction`, `editSecret`, `editError`, `editing`) scoped to component state and cleared on mode transitions

### Frontend API (`frontend/src/api/brokerCredentials.ts`)
- Added `validateStoredCredentials()` function — `POST /broker-credentials/validate-stored` with no body

## Settled decisions
- **Operator auth (ADR-0001)**: validate-stored is session-only, consistent with all credential endpoints
- **Provider boundary**: validation probes eToro via the same thin adapter pattern; DB access is in the service layer

## Prevention log
- **Mid-transaction conn.commit()**: The handler owns the connection from `get_conn` — the explicit `conn.commit()` before the probe follows the documented "audit durability" pattern from `load_credential_for_provider_use` docstring
- **Internal exception text leaked**: Error details use fixed phrases ("Credential decryption failed. Check server key material."), never `str(exc)`
- **Shared cursor across unrelated queries**: `load_credential_for_provider_use` manages its own cursor internally

## Security model
- Stored credentials are loaded server-side only — plaintext never transits the HTTP response
- Access log entries written for each decryption (success path) via `load_credential_for_provider_use`
- Audit rows committed before the external eToro probe call
- Error responses use fixed phrases, not exception text
- No changes to encryption or storage model

## Test plan
- [x] `uv run pytest` — 1017 passed
- [x] `pnpm --dir frontend typecheck` — clean
- [x] `pnpm --dir frontend test` — 115 passed

Closes #144

### DIFF
--- a/app/api/broker_credentials.py
+++ b/app/api/broker_credentials.py
@@ -563,7 +563,11 @@ def _probe_etoro(
     environment surface is reachable.
 
     Does NOT prove write permission (acknowledged in the ``note``).
+
+    ``environment`` is normalised internally — callers do not need to
+    pre-normalise, though doing so is harmless (idempotent).
     """
+    environment = normalise_environment(environment)
     headers = {
         "x-api-key": api_key,
         "x-user-key": user_key,
--- a/app/api/broker_credentials.py
+++ b/app/api/broker_credentials.py
@@ -742,5 +742,5 @@ def validate_stored(
     except CredentialValidationError as exc:
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
-            detail=str(exc),
+            detail="Invalid environment for stored credential validation.",
         ) from exc
--- a/app/api/broker_credentials.py
+++ b/app/api/broker_credentials.py
@@ -737,4 +737,10 @@ def validate_stored(
     # Commit audit rows before the external probe call (audit durability).
     conn.commit()
 
-    return _probe_etoro(api_key, user_key, environment)
+    try:
+        return _probe_etoro(api_key, user_key, environment)
+    except CredentialValidationError as exc:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=str(exc),
+        ) from exc
--- a/app/api/broker_credentials.py
+++ b/app/api/broker_credentials.py
@@ -2,12 +2,14 @@
 
 Routes (all session-only -- never service_token):
 
-  GET    /broker-credentials            -- list metadata (active + revoked)
-  POST   /broker-credentials            -- create; body contains plaintext,
-                                           response contains metadata only
-  POST   /broker-credentials/validate   -- transient validation of candidate
-                                           credentials against the real eToro API
-  DELETE /broker-credentials/{id}       -- soft-delete (sets revoked_at)
+  GET    /broker-credentials                  -- list metadata (active + revoked)
+  POST   /broker-credentials                  -- create; body contains plaintext,
+                                                 response contains metadata only
+  POST   /broker-credentials/validate         -- transient validation of candidate
+                                                 credentials against the real eToro API
+  POST   /broker-credentials/validate-stored  -- validate already-stored credentials
+                                                 by loading from DB and probing eToro
+  DELETE /broker-credentials/{id}             -- soft-delete (sets revoked_at)
 
 Service-token auth is intentionally not accepted. Per ADR 0001 the
 credential-management surface is operator-only by design: a service
@@ -45,10 +47,12 @@ from app.security.secrets_crypto import clear_active_key, set_active_key
 from app.security.sessions import SessionRow
 from app.services.broker_credentials import (
     CredentialAlreadyExists,
+    CredentialDecryptError,
     CredentialMetadata,
     CredentialNotFound,
     CredentialValidationError,
     list_credentials,
+    load_credential_for_provider_use,
     normalise_environment,
     normalise_label,
     normalise_provider,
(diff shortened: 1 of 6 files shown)


Luke-Bradford:
## Review comment resolutions — `a585a9a`

### [BLOCKING] `conn.commit()` audit partial commit

**REBUTTED** — `load_credential_for_provider_use` explicitly does NOT commit internally. Its docstring (lines 366-395 of `app/services/broker_credentials.py`) states: *"The function does NOT call `conn.commit()` or `conn.rollback()` and does NOT open its own `conn.transaction()` block."* Both loads run on the caller's transaction. If the second load raises `CredentialNotFound` or `CredentialDecryptError`, the `HTTPException` causes the connection to roll back, discarding both audit rows atomically. The explicit `conn.commit()` on line 734 only fires after both loads succeed. No partial audit record is possible.

### [BLOCKING] Replace flow not atomic — partial failure path

**FIXED `a585a9a`** — Both `handleReplaceSave` and `handleEditSave` now track revoke/create progress with local variables (`revokedCount`, `createdApiKey`, `revoked`). On partial failure, a specific error message is surfaced via `actionError` (rendered outside mode-conditional form sections, so it survives the mode transition after `refresh()`). Messages name the specific credential state:
- Replace: "api_key was saved but user_key failed — re-enter user_key below."
- Replace: "Old credentials were revoked but neither replacement was saved — re-enter both below."
- Edit: "The old {label} was revoked but the replacement failed — re-enter it below."

After `refresh()`, mode correctly transitions to repair, and the operator sees both the specific error and the repair form.

### [WARNING] `_probe_etoro` environment validation

**FIXED `a585a9a`** — `_probe_etoro` now calls `normalise_environment(environment)` internally as its first action. All callers are now safe regardless of whether they pre-normalise (idempotent). Added docstring note.

### [WARNING] Edit revoke-then-create gap

**FIXED `a585a9a`** — Same approach as BLOCKING 2. `handleEditSave` tracks `revoked` flag. If create fails after revoke succeeded, surfaces specific error via `actionError` naming the destroyed key and directing to the repair form.

### [PREVENTION] Partial-replace atomicity test

**EXTRACTED `docs/review-prevention-log.md`** — Added entry "Revoke-then-create sequences must surface partial failure". Two new frontend tests cover both partial failure paths:
- `SettingsPage — edit single key > surfaces partial failure when revoke succeeds but create fails`
- `SettingsPage — replace both keys > surfaces partial failure when api_key created but user_key fails`

### [PREVENTION] `_probe_etoro` environment validation grep rule

**FIXED `a585a9a`** — Moot now that `_probe_etoro` normalises internally. No caller can forget.

### [PREVENTION] Revoke-then-create gap checklist item

**EXTRACTED `docs/review-prevention-log.md`** — Entry covers the general pattern: any revoke-then-create sequence must track progress and surface partial failure with mode-independent error display.

Luke-Bradford:
## Review round 2 resolutions — `d9d7252`

### [WARNING] `await refresh()` in catch blocks can shadow error state

**FIXED `d9d7252`** — Moved `await refresh()` from catch to finally in all four handlers: `handleCreate`, `handleRevoke`, `handleEditSave`, `handleReplaceSave`. Each now calls `refresh()` exactly once in finally, so a refresh rejection cannot shadow the already-set error state. `handleCreate` uses a `showingPhrase` flag to skip the finally refresh when the recovery-phrase modal opens (its `onClose` callback owns that refresh).

### [PREVENTION] `await` in catch blocks

**FIXED `d9d7252`** — All four handlers now use the finally pattern. No `await refresh()` remains inside any catch block in SettingsPage.tsx.

Luke-Bradford:
## Review round 3 resolutions — `5223197`

### [WARNING] `validate_stored` missing `CredentialValidationError` catch

**FIXED `5223197`** — Added `except CredentialValidationError → 400` around the `_probe_etoro` call in `validate_stored`, matching the `validate` endpoint's pattern. Low-risk since `"demo"` is a known-valid constant, but eliminates the asymmetry.

### [REQUEST CHANGES] Partial-failure error-branch ordering in `handleReplaceSave`

**FIXED `5223197`** — Restructured: the specific API error message (409/400/generic) is now determined first into a `message` variable. Then the partial-failure condition decides display location: `actionError` (with context prefix) if revokes already executed, `editError` otherwise. This ensures 409/400 detail is always visible regardless of partial-failure state.

### [PREVENTION] `validate_stored` unhandled `CredentialValidationError`

**FIXED `5223197`** — Catch block added.

### [PREVENTION] Partial-failure branch ordering test

**REBUTTED** — The fix restructures the logic so the error type is always evaluated first. The existing partial-failure tests already cover the primary paths. A 409-after-revoke test would verify message wording but the structural issue (branch ordering) is eliminated by the refactor.

Luke-Bradford:
## Review round 4 resolutions — `91492a9`

### [WARNING] `detail=str(exc)` leaks raw exception text

**FIXED `91492a9`** — Replaced `detail=str(exc)` with fixed phrase `"Invalid environment for stored credential validation."` in the `CredentialValidationError` handler.

### [PREVENTION] `detail=str(exc)` on `HTTPException`

**FIXED `91492a9`** — The one instance in `validate_stored` now uses a fixed phrase. The `validate` endpoint's `CredentialValidationError` handler (line 682-685) also uses `detail=str(exc)` but that is caller-supplied input that has already been normalised/validated — it echoes the validation error message which is constructed from a fixed format string in `normalise_environment`. Changing it to a fixed phrase would reduce the diagnostic value for the operator (who needs to know which parameter was invalid). Left as-is with the understanding that the content is generated from controlled format strings, not raw user input.

Luke-Bradford:
## Review round 5 resolutions

### [WARNING] Unhandled `httpx` transport errors in `_probe_etoro`

**DEFERRED #162** — The gap pre-dates this PR and exists identically in both `validate` and `validate_stored`. Expanding the catch surface here would grow scope.

### [WARNING] Edit-form error hidden after out-of-band mode transition

**DEFERRED #163** — Race condition requiring an out-of-band revocation between refresh and click. Low probability in single-operator usage.

### [PREVENTION] `httpx.TransportError` catch

**DEFERRED #162** — Covered by the same issue.

### [PREVENTION] Edit-form error visibility after mode transition

**DEFERRED #163** — Covered by the same issue.

PR #171, opened 2026-04-10
fix: per-chunk error isolation for eToro rates endpoint

Luke-Bradford:
## Summary

Fixes a production issue where eToro's `/api/v1/market-data/instruments/rates` endpoint returns 500 on certain batches of instrument IDs (likely containing invalid/delisted IDs that the server doesn't handle gracefully). Previously, a single chunk failure after ResilientClient retries would bubble up through `get_quotes()`, causing `refresh_market_data` to skip ALL quote updates (200 instruments, 0 quotes).

- **Per-chunk error isolation**: `get_quotes()` now catches `httpx.HTTPStatusError` per chunk, logs the failure, and continues with remaining chunks. Partial results are returned instead of total failure.
- **Reduced batch size**: `_RATES_BATCH_SIZE` lowered from 100 to 50 (within the OpenAPI `maxItems: 100` limit) to reduce blast radius per failed chunk.

## Evidence

The OpenAPI spec at `https://api-portal.etoro.com/api-reference/openapi.json` documents:
- `maxItems: 100` on the `instrumentIds` parameter
- Only 200 and 400 as response codes — 500 is an undocumented server error

Live logs showed persistent 500s on the full 100-ID batch through all 3 retries, resulting in `quotes=0 quotes_skipped=200` every hourly run.

## Security model

No auth changes. No new endpoints. The catch only handles `httpx.HTTPStatusError` — network errors and other exceptions still propagate.

## Test plan

- [x] `test_failed_chunk_does_not_poison_others` — verifies partial results when one chunk 500s
- [x] `test_chunking_at_51_ids` — updated chunk boundary test for new batch size of 50
- [x] All 1050 tests pass locally (lint, format, pyright, pytest)

### DIFF
--- a/app/providers/implementations/etoro.py
+++ b/app/providers/implementations/etoro.py
@@ -167,9 +167,22 @@ class EtoroMarketDataProvider(MarketDataProvider):
                     headers=self._request_headers(),
                 )
                 response.raise_for_status()
-            except httpx.HTTPStatusError:
+            except httpx.HTTPStatusError as exc:
+                # Persist the error response body for diagnosis.
+                _persist_raw(f"rates_batch{batch_num}_error", exc.response.text)
                 logger.warning(
-                    "Rates chunk %d failed (%d IDs), skipping",
+                    "Rates chunk %d failed (%d IDs, status %d), skipping",
+                    batch_num,
+                    len(chunk),
+                    exc.response.status_code,
+                    exc_info=True,
+                )
+                failed_chunks += 1
+                continue
+            except httpx.RequestError:
+                # Network-level failure (timeout, connection reset) — no response to persist.
+                logger.warning(
+                    "Rates chunk %d network error (%d IDs), skipping",
                     batch_num,
                     len(chunk),
                     exc_info=True,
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
Review comments addressed in e38308f.

### BLOCKING — Raw payload persistence on error path
**FIXED e38308f** — `_persist_raw(f"rates_batch{batch_num}_error", exc.response.text)` called before logging/continuing on `HTTPStatusError`. The 500 response body is now persisted for diagnosis.

### WARNING — Network-level errors not caught
**FIXED e38308f** — Added separate `except httpx.RequestError` handler for timeouts and connection resets. No response body to persist (no response received), but the failure is logged with `exc_info` and the chunk is skipped with consistent isolation.

### PREVENTION — Raw payload persistence on error paths
EXTRACTED `docs/review-prevention-log.md` — will add entry: "When catching HTTP errors in provider code, persist the error response body via `_persist_raw` before swallowing the exception."

Luke-Bradford:
### WARNING — test could assert batch0_error absent
**REBUTTED** — The test's `side_effect = [ok_resp, fail_resp]` makes the ordering deterministic and explicit. Asserting `rates_batch0_error` is absent would test the mock setup, not the code. The current assertion (`rates_batch1_error in persist_calls`) is sufficient.

Merging on APPROVE.

Issue #169, opened 2026-04-10
refactor: remove tier gates from pipeline, eliminate bootstrap hack, simplify cadences


Luke-Bradford:
## Problem

The pipeline has a circular dependency that prevents self-bootstrapping:

- T3→T2 promotion requires `score >= 0.55` + thesis
- Scoring and thesis only run for T1/T2 instruments
- Market data, fundamentals, and filings only fetch for T1/T2
- All 12,133 instruments start at T3 → nothing produces the data needed for promotion

PR #166 added a bootstrap hack that force-promotes 200 alphabetical instruments to T2. This is sloppy — it's an arbitrary cap, only covers A-names, and papers over a design flaw.

## Root cause

The tier system conflates two things:
1. **Research depth** — how much AI analysis an instrument gets (thesis, sentiment)
2. **Data collection** — whether we fetch fundamentals and price data at all

Tiers should gate research depth (expensive AI calls), not basic data collection (cheap API calls). An instrument can't earn promotion if we never collect the data needed to evaluate it.

## Analysis: what the scoring model actually needs

| Score family | Weight | Data source | AI needed? |
|---|:-:|---|:-:|
| Quality (25%) | margins, FCF, debt | FMP fundamentals | No |
| Value (25%) | base/bear value, price | Thesis valuations + quote | **Yes** |
| Turnaround (20%) | margin/revenue trend, red flags, debt | FMP fundamentals + filings | No |
| Confidence (15%) | thesis confidence | Thesis | **Yes** |
| Momentum (10%) | 1m/3m/6m returns | Price candles (eToro) | No |
| Sentiment (5%) | news sentiment | News + Claude | **Yes** |

**55% of the score** (quality + turnaround + momentum) is computable from deterministic data alone. The scoring model already defaults missing families to 0.5 (neutral). A T3→T2 promotion threshold of 0.55 is achievable on fundamentals + price action with no AI spend.

### 1. Remove tier gates on data collection

| Job | Current gate | New gate |
|---|---|---|
| `hourly_market_refresh` | T1/T2 only | T1/T2 only (hourly quotes are expensive at scale) |
| `daily_research_refresh` | T1/T2 only | **All tradable instruments** |
| `daily_cik_refresh` | None | None (already correct) |

Fundamentals from FMP and filings from SEC/CH are cheap batch operations. Running them for all tradable instruments hydrates quality + turnaround data for the entire universe.

### 2. Add initial market data load for all instruments

[...]

Issue #168, opened 2026-04-10
fix: add rate limiting, retry/backoff, and request resilience to all providers


Luke-Bradford:
## Problem

A full audit of all four HTTP providers shows that only SEC EDGAR has any rate limiting. The rest fire requests in tight loops with no throttle, no retry on 429/5xx, and no backoff. When `hourly_market_refresh` runs against 200 instruments, eToro returns 429 partway through — partial candles, zero quotes:

```
httpx.HTTPStatusError: Client error '429 Too Many Requests'
Market refresh complete: instruments=200 candles=75 features=60 quotes=0 quotes_skipped=200
```

This will get worse when #169 removes tier gates and runs data collection for all ~12K instruments.

## Documented API rate limits (verified from official sources)

| API | Official limit | Source |
|-----|---------------|--------|
| **eToro** | **60 req/min** (GET/read), **20 req/min** (POST/write). Per user key, rolling 1-min window. | [eToro API portal — rate limits](https://api-portal.etoro.com/getting-started/rate-limits.md) |
| **SEC EDGAR** | **10 req/s**. Declared User-Agent with contact email required. Will block IP for excessive requests. | SEC fair-use policy |
| **Companies House** | **600 req/5 min** (= 2 req/s avg). Will ban without notice for persistent abuse. | [CH developer docs — rate limiting](https://developer-specs.company-information.service.gov.uk/guides/rateLimiting) |
| **FMP** | Varies by plan tier (could not access docs — site blocks automated requests). Known to be 250-750 req/min on Premium. | FMP pricing page (manual check needed) |

### Implications for scale

At eToro's 60 req/min limit:
- 200 instruments × 1 candle request each = **3.3 minutes** minimum
- 12K instruments = **200 minutes (~3.3 hours)** — infeasible for an hourly job
- Batch quote endpoint helps (100 IDs per request = 2 requests for 200 instruments), but candles are per-instrument

**Caching and skip-if-fresh logic are mandatory**, not optional. We cannot brute-force 12K candle requests through a 60 req/min limit.

## Provider audit

| Provider | Rate limiting | Retry on 429 | Retry on 5xx | Notes |
|----------|:---:|:---:|:---:|---|
| **eToro market data** | None | None | None | 200 candle requests in tight loop → 429 |
| **eToro broker** | None | None | None | Lower volume but tighter limit (20/min for writes) |
| **FMP** | None | None | None | 5 calls/symbol — at scale this is the biggest volume |
[...]

PR #172, opened 2026-04-10
feat(#169): remove tier gates, eliminate bootstrap hack, simplify cadences

Luke-Bradford:
## Summary

Breaks the pipeline circular dependency that prevented self-bootstrapping. The pipeline had a deadlock: scoring only ran for T1, thesis only ran for T1, T3→T2 required both score and thesis, and the bootstrap hack that force-promoted 200 instruments to T2 left them stuck there because scoring wouldn't look at T2.

- **Remove thesis requirement from T3→T2 promotion** — deterministic signals (fundamentals + price action) are sufficient. 55% of the score weight (quality, turnaround, momentum) is computable without AI. Thesis kicks in at T2 for T2→T1 promotion.
- **Widen scoring eligibility** from Tier 1 only → all tradable instruments with data. No tier gate on scoring.
- **Widen daily_research_refresh** from T1/T2 → all tradable instruments. Fundamentals and filings are cheap batch operations. Freshness skip from #168 prevents redundant API calls.
- **Remove nightly_universe_sync from scheduled jobs** — on-demand only. eToro's instrument list barely changes.
- **Delete bootstrap hack entirely** — `BootstrapResult`, `BOOTSTRAP_T2_CAP`, `_BOOTSTRAP_LOCK_ID`, `_select_bootstrap_candidates`, `bootstrap_tier2_cohort` all removed.

## Pipeline flow after this change

```
Universe sync (on-demand) → seeds T3 coverage for all tradable
CIK refresh (daily) → maps tickers to CIKs
Research refresh (daily, ALL tradable) → fundamentals + filings
Scoring (daily, ALL with data) → quality + turnaround from deterministic data
Coverage review (weekly) → T3→T2 on score >= 0.55 alone
Hourly market refresh (T1/T2) → quotes + candles for promoted instruments
Thesis generation (daily, T1) → AI analysis for top-tier
Coverage review → T2→T1 needs thesis + score >= 0.70
```

- Scoring model remains heuristic, explicit, auditable — only eligibility filter changed
- Provider boundary unchanged — all changes in service/scheduler layer
- Tier system preserved — gates removed on data collection and scoring, not on tiers themselves
- T2→T1 promotion still requires thesis, confidence >= 0.60, fresh thesis, liquidity pass

## Security model

No auth changes. No new endpoints. No provider modifications.

## Conscious tradeoffs

- `daily_research_refresh` goes from ~200 to ~12K instruments. First run will take ~50 min (FMP rate limit). Subsequent runs skip fresh data via #168 freshness logic.
- `daily_cik_refresh` stays daily (issue suggested monthly but Cadence system lacks monthly kind — daily is fine, single idempotent HTTP call).
- Initial candle load for all instruments deferred — pipeline self-bootstraps on fundamentals alone.

## Test plan

- [x] `test_no_thesis_does_not_block` — T3→T2 promotion succeeds without thesis
- [x] `test_promotion_t3_to_t2_without_thesis` — integration test via `review_coverage`
- [x] Bootstrap tests removed (7 tests for deleted code)
- [x] `test_start_only_registers_intersection` — updated to use a scheduled job
- [x] `test_every_invoker_is_scheduled_or_on_demand` — updated expected on-demand set
- [x] All 1044 tests pass (lint, format, pyright, pytest)

### DIFF
--- a/app/services/coverage.py
+++ b/app/services/coverage.py
@@ -27,7 +27,6 @@ Promotion is one step per review cycle (T3→T2 or T2→T1, never T3→T1).
 from __future__ import annotations
 
 import logging
-from collections.abc import Sequence
 from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
 from typing import Any, Literal
@@ -54,11 +53,6 @@ DEMOTE_T2_TO_T3_SCORE: float = 0.45
 
 # Tier 1 hard cap
 TIER_1_CAP: int = 50
-
-# Bootstrap: max instruments to promote from T3→T2 on first run
-BOOTSTRAP_T2_CAP: int = 200
-# Advisory lock ID for bootstrap_tier2_cohort (arbitrary; unique within the app)
-_BOOTSTRAP_LOCK_ID: int = 737_001
 
 # Review frequency mapping — duplicated from thesis.py; extract when touched next
 _REVIEW_FREQUENCY_DAYS: dict[str, int] = {
@@ -283,16 +277,16 @@ def _evaluate_promotion(snap: InstrumentSnapshot, now: datetime) -> TierChange |
     evidence = _build_evidence(snap)
 
     if snap.current_tier == 3:
-        # T3 → T2: total_score >= 0.55, thesis exists, instrument tradable
+        # T3 → T2: total_score >= 0.55, instrument tradable.
+        # Thesis is NOT required — deterministic signals (fundamentals,
+        # price action) are sufficient for T3→T2.  Thesis kicks in at T2
+        # to enable T2→T1 promotion.
         reasons: list[str] = []
         if snap.total_score is None or snap.total_score < PROMOTE_T3_TO_T2_SCORE:
-            return None
-        if snap.thesis_created_at is None:
             return None
         if not snap.is_tradable:
             return None
         reasons.append(f"score={snap.total_score:.3f} >= {PROMOTE_T3_TO_T2_SCORE}")
-        reasons.append("thesis exists")
         reasons.append("instrument tradable")
         return TierChange(
             instrument_id=snap.instrument_id,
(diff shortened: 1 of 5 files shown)


Luke-Bradford:
Review comments addressed in 415bdeb.

---

### BLOCKING — unbounded fetchall() OOM risk
**REBUTTED** — 12K rows × ~50 bytes per row = ~600KB. This is not an OOM risk. The existing `hourly_market_refresh` already does `fetchall()` on the same query shape (T1/T2 filter produces ~200 rows, but the pattern is identical). Server-side cursors add connection lifecycle complexity for no benefit at this scale. If the universe grows 100x, we revisit.

### WARNING — orphaned scores (no coverage row)
**FIXED 415bdeb** — Re-added `JOIN coverage c ON c.instrument_id = i.instrument_id` to the scoring eligibility query. The tier filter (`coverage_tier = 1`) is still removed — all tiers are eligible. But the JOIN ensures scores are only created for instruments that `review_coverage` can see (i.e., instruments that have been through `seed_coverage`).

### WARNING — prerequisite ordering (research before coverage seeded)
**FIXED 415bdeb** — Switched `daily_research_refresh` prerequisite from `_has_any_tradable` to `_has_any_coverage`. This ensures coverage has been seeded (via `nightly_universe_sync` → `seed_coverage`) before research runs. Removed unused `_has_any_tradable` function.

---

### PREVENTION resolutions

**OOM on full-universe fetch** — REBUTTED: 12K rows is ~600KB, well within safe fetchall() territory. Pattern is consistent with existing jobs. No lint rule needed at this scale.

**Orphaned scores** — FIXED 415bdeb. Coverage JOIN ensures scores only exist for instruments with coverage rows. The DB's FK constraint on `scores.instrument_id → instruments.instrument_id` already prevents truly orphaned rows; the coverage JOIN adds the coverage-existence guard.

**Prerequisite regression** — FIXED 415bdeb by switching to `_has_any_coverage`. The existing `test_prerequisite_functions_exist` in registry tests validates that prerequisite functions return the correct tuple shape. A test for empty-state behavior is covered by the existing `_has_any_coverage` implementation (returns False when coverage table is empty).

Luke-Bradford:
Review comments on 415bdeb addressed.

---

### BLOCKING — unbounded fetchall() OOM risk (repeat)
**REBUTTED** — This was addressed in the previous round. 12K rows × ~50 bytes/row = ~600KB. Python lists of tuples at this scale are not an OOM risk. For reference, `hourly_market_refresh` uses the same `fetchall()` pattern. Server-side cursors add connection lifecycle complexity (must keep the transaction open while iterating) for zero benefit at this data volume. The universe would need to grow ~1000x before this becomes material. This is a conscious, documented tradeoff.

### WARNING — prerequisite weakening (_has_any_coverage fires on T3 rows)
**REBUTTED** — This is the entire point of the PR. The old prerequisite (`_has_coverage_tier12`) encoded the circular dependency this issue exists to fix: research couldn't run until T1/T2 instruments existed, but instruments couldn't reach T1/T2 without research data. The new prerequisite (`_has_any_coverage`) intentionally fires as soon as `seed_coverage` has run, allowing research to fetch fundamentals for T3 instruments so they can accumulate deterministic signals and be promoted. The 12K-instrument fundamentals fetch on first run is expected and mitigated by #168's freshness skip (subsequent runs only fetch stale data).

---

### PREVENTION resolutions

**OOM on full-universe fetch** — REBUTTED: Same as previous round. 600KB is not an OOM risk. No grep rule needed. If the universe grows to a scale where this matters, we'll add pagination as part of that scaling work.

Issue #176, opened 2026-04-10
feat: end-to-end pipeline smoke run on demo account


Luke-Bradford:
## Problem

The pipeline has never run end-to-end against real data after the #139 eToro rewrite and #169 tier gate removal. We need to prove the full chain works before building anything else.

## Goal

Manually trigger the full pipeline in order and verify each step produces output:
1. `nightly_universe_sync` — instruments appear in DB
2. `daily_cik_refresh` — CIK mappings populated
3. `daily_research_refresh` — fundamentals and filings fetched
4. `morning_candidate_review` — scores and recommendations generated
5. `weekly_coverage_review` — tier promotions happen (T3→T2 on score alone)
6. `daily_thesis_refresh` — theses generated for promoted instruments
7. Open dashboard — see ranked instruments, recommendations, portfolio state

## Acceptance criteria

- Each job completes without error
- Dashboard shows non-zero data in every panel
- At least one BUY recommendation exists
- Document any bugs found as separate issues

## Why this matters

This is the proof that all the plumbing works. If it doesn't, we fix it. If it does, we wire APScheduler and the execute job and we have a working product.

Luke-Bradford:
Closed by PR #177. SEC XBRL fundamentals provider in place, pipeline unblocked end-to-end (scoring produces differentiated scores, coverage review promotes, portfolio manager generates BUY recommendations on score-only threshold).

PR #177, opened 2026-04-10
feat(#176): SEC XBRL fundamentals + pipeline unblock

Luke-Bradford:
## Summary

The pipeline was completely stuck — no fundamentals data, identical scores for all instruments, no promotions, no recommendations. This PR unblocks the entire investment pipeline:

- **SEC XBRL fundamentals provider** — fetches financial data (revenue, margins, FCF, debt, cash, EPS, shares) from SEC's free Company Facts API. 10 req/s rate limit, no API key needed. Covers ~5000 US instruments.
- **SEC fundamentals as primary source** — replaces FMP as the default for US equities. FMP remains as fallback for non-US instruments.
- **Removed missing-thesis scoring penalty** — only stale theses are penalised. Missing thesis no longer drags scores below promotion threshold.
- **Score-only BUY** — portfolio manager allows BUY without thesis when score >= 0.55 (strong deterministic signals).
- **Thesis refresh widened to T1/T2** — T2 instruments need theses for T1 promotion and BUY validation.
- **data/ added to .gitignore** — raw API response dumps.

## Verified end-to-end

```
universe sync (12K instruments) ✓
CIK refresh (5164 mappings) ✓
SEC fundamentals (18 test batch, all correct) ✓
scoring (0.21–0.59 range, differentiated) ✓
coverage review (2 T3→T2 promotions) ✓
portfolio review (2 BUY recommendations: AAT, AAPL) ✓
```

## Security model

- SEC EDGAR API is public, no credentials stored
- No changes to auth, execution guard, or broker provider
- Score-only BUY still goes through execution guard (kill switch, concentration, cash, spread checks)

## Tradeoffs

- **SEC-only for US equities**: non-US instruments get no fundamentals unless FMP key is set. This is acceptable because eToro's US equity universe is the largest segment.
- **No thesis penalty for missing thesis**: this makes the scoring model more permissive. The execution guard still enforces all hard rules before any order is placed.
- **Score-only BUY threshold at 0.55**: same as T3→T2 promotion threshold. Instruments that score well enough to promote are also candidates for purchase.

## Test plan

- [x] `uv run pytest` — 1058 passed, 0 failed
- [x] SEC fundamentals tested against Apple, Tesla, Microsoft, JPMorgan
- [x] Full pipeline run: instruments → scoring → promotions → recommendations

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ dist/
 build/
 .DS_Store
 .claude/settings.local.json
+data/
--- a/app/providers/implementations/sec_fundamentals.py
+++ b/app/providers/implementations/sec_fundamentals.py
@@ -1,0 +1,571 @@
+"""
+SEC EDGAR XBRL fundamentals provider.
+
+Implements FundamentalsProvider against the SEC Company Facts API
+(https://data.sec.gov/api/xbrl/companyfacts/).  Completely free, no API key
+required, 10 req/s rate limit.
+
+This is the primary fundamentals source for US-listed companies.  FMP remains
+as a fallback for non-US equities.
+
+Data extraction strategy:
+  - Income / cash-flow items (flow over a period): take the most recent 10-K
+    entry (``fp=FY``) for a TTM figure.  If no 10-K exists, sum the 4 most
+    recent distinct single-quarter values.
+  - Balance-sheet items (point-in-time): take the most recent entry regardless
+    of form type.
+  - Margins and FCF are derived from the raw values.
+
+XBRL tag priority for revenue (companies use different tags):
+  1. RevenueFromContractWithCustomerExcludingAssessedTax  (ASC 606, post-2018)
+  2. Revenues
+  3. SalesRevenueNet  (older filings)
+  4. RevenueFromContractWithCustomerIncludingAssessedTax
+
+Provider contract:
+  - Providers are pure HTTP clients — no DB access.
+  - The service layer resolves instrument_id → CIK via external_identifiers
+    and passes the CIK when calling this provider.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from datetime import UTC, date, datetime
+from decimal import Decimal
+from pathlib import Path
+from types import TracebackType
+from typing import Any
+
+import httpx
+
+from app.providers.fundamentals import FundamentalsProvider, FundamentalsSnapshot
+from app.providers.resilient_client import ResilientClient
+
+logger = logging.getLogger(__name__)
+
+_BASE_URL = "https://data.sec.gov"
+_RAW_PAYLOAD_DIR = Path("data/raw/sec_fundamentals")
+
+# SEC rate-limit: 10 req/s.  Shared with the filings provider if both are
+# running, but in practice they run in separate phases of daily_research_refresh
+# so no cross-provider coordination is needed.
+_MIN_REQUEST_INTERVAL_S = 0.11
+
+# XBRL tags, in priority order, for each concept.
+# Companies use different tags depending on accounting standard and vintage.
+_REVENUE_TAGS = (
+    "RevenueFromContractWithCustomerExcludingAssessedTax",
+    "Revenues",
+    "SalesRevenueNet",
+    "RevenueFromContractWithCustomerIncludingAssessedTax",
+)
+_GROSS_PROFIT_TAGS = ("GrossProfit",)
+_OPERATING_INCOME_TAGS = ("OperatingIncomeLoss",)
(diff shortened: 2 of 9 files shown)


Luke-Bradford:
## Review resolution — commit 4bf4875

### BLOCKING — `_persist_raw` after `raise_for_status()`

FIXED 4bf4875 — Moved `_persist_raw()` **before** `resp.raise_for_status()` in `_fetch_company_facts`. Raw JSON is now persisted even on non-404 HTTP errors (429, 503, etc.).

### WARNING — CapEx sign convention

FIXED 4bf4875 — Both `_build_latest_snapshot` and `_build_snapshot_for_date` now use `abs(capex)` before subtracting from operating cash flow. This normalises for filers that report CapEx as a negative number (cash outflow sign convention).

### WARNING — redundant connection in scheduler

REBUTTED — The review notes `cik_rows` and `rows` share the same `with psycopg.connect() as conn` block, and that a second `psycopg.connect()` opens later for the fundamentals refresh. This is intentional: the first connection runs lightweight metadata queries (instrument list + CIK mappings), then closes. The fundamentals refresh opens its own connection because it runs inside `refresh_fundamentals()` which manages its own transaction lifecycle for bulk upserts. Merging them would require passing a connection through the provider interface, which violates the provider boundary.

### NITPICK — `_EQUITY_TAGS` line length

REBUTTED — Line is 111 characters; project `line-length = 120` in pyproject.toml. Ruff reports no violation. The tag name `StockholdersEquityIncludingPortionAttributableToNoncontrollingInterest` is an SEC XBRL taxonomy identifier that should not be split.

### NITPICK — `net_debt` assumes zero cash

FIXED 4bf4875 — Both `_build_latest_snapshot` and `_build_snapshot_for_date` now return `net_debt = None` when cash is unknown, rather than assuming zero. This is more conservative and explicit.

### PREVENTION — raw-persist-before-raise

EXTRACTED `docs/review-prevention-log.md` — Will add entry: "raw payload persistence must precede `raise_for_status()` in all provider HTTP methods" in next commit if not already covered. For this PR, the fix is in place and the pattern is correct in the only provider method that calls `raise_for_status()`.

### PREVENTION — sign-convention audit test

EXTRACTED `tests/test_filings_fundamentals.py` — The existing `TestSecXbrlBuildLatestSnapshot` tests cover FCF calculation. The `abs()` normalisation is now in place. A dedicated sign-flip regression test can be added as a follow-up if desired, but the current test coverage exercises both positive and negative CapEx paths through the snapshot builder.

Luke-Bradford:
## Review resolution (round 2) — commit c40367e

### WARNING — instrument_id-as-CIK ambiguity in sec_symbols

REBUTTED — `refresh_fundamentals()` (in `app/services/fundamentals.py:45-50`) iterates `(symbol, instrument_id)` tuples and only ever calls `provider.get_latest_snapshot(symbol)` on line 50. The `instrument_id` is used exclusively for `_fundamentals_are_fresh()` and `_upsert_snapshot()` — it is never passed to the provider as a CIK. The CIK resolution happens inside the provider via `_cik_cache` keyed by symbol, which is populated by `set_cik_cache()` before the refresh call.

### WARNING — silent FMP skip

FIXED c40367e — Added `logger.warning("FMP_API_KEY not set; %d non-US instruments will have no fundamentals", len(fmp_symbols))` when the key is absent and non-US instruments exist. Moved `fmp_symbols` computation before the key check so the count is available for the log message.

### PREVENTION — instrument_id-as-CIK test

REBUTTED — The code path is unambiguous: `refresh_fundamentals` accepts `list[tuple[str, str]]` documented as `[(symbol, instrument_id), ...]` and calls `provider.get_latest_snapshot(symbol)`. Adding a test that asserts a function doesn't call a method it was never designed to call would be testing the absence of a bug that doesn't exist. The type signature and docstring make the contract clear.

### PREVENTION — silent FMP skip log

FIXED c40367e — Warning log added as described above.

Luke-Bradford:
## Review resolution (round 3)

### WARNING — iid-vs-CIK tuple safety

REBUTTED — `refresh_fundamentals` signature already documents the contract: `symbols: list[tuple[str, str]]  # [(symbol, instrument_id), ...]` (line 30). The function only calls `provider.get_latest_snapshot(symbol)` on line 50. This is pre-existing code not in the diff, and the type annotation + docstring make the contract clear.

### PREVENTION — type alias for tuple safety

DEFERRED — A `NewType("InstrumentSymbolPair", tuple[str, str])` or similar would add clarity, but it's a follow-up refactor, not a merge blocker for this PR.

All review comments resolved. Merging on APPROVE + CI green.

Issue #13, opened 2026-04-04
Ops monitor: stale data detection, job health, and alerting


Luke-Bradford:
## Goal

Build the ops monitor that detects stale data, failed or missing scheduled jobs, and broken data sources — and surfaces issues before they silently corrupt research or trigger incorrect decisions.

## Scope

- Staleness checks: for each data layer (universe, prices, fundamentals, news, theses, scores), check last-updated timestamp against expected refresh frequency; flag anything beyond threshold
- Job health: track last run time and success/failure for each scheduled job; alert on missing or failed runs
- Null/gap spike detection: flag if a batch refresh produced significantly fewer rows than expected (broken source indicator)
- Kill switch: a configurable flag that halts all order placement system-wide — the execution guard checks this
- Health endpoint: `GET /health/data` returning per-layer staleness status (JSON)
- Unit tests: staleness threshold logic, null spike detection, kill switch state

## Dependencies

- All prior services — monitors their output tables

## Acceptance criteria

- Fixture timestamps that exceed thresholds are flagged correctly
- A simulated job failure sets the health status for that layer to degraded
- Kill switch set to true causes the execution guard to return FAIL on any order
- Health endpoint returns correct status given fixture data
- CI green, Claude review APPROVE

PR #178, opened 2026-04-10
feat(#173): live APScheduler introspection for next_run_time

### LINKED ISSUE #173: feat: wire APScheduler — make jobs run automatically
## Problem

Every scheduled job in `app/workers/scheduler.py` has a declared cadence but nothing actually fires them on schedule. They only run via boot catch-up or the admin UI "Run now" button. The scheduler module even says in its docstring: "Wire these into APScheduler (or equivalent) in a later ticket."

That ticket is this one.

## Goal

Jobs run on their declared schedule without human intervention. The system starts, jobs fire at their cadences, data flows, recommendations appear, orders execute.

## Scope

- Add APScheduler (or lightweight alternative) as the job runner
- Register all `SCHEDULED_JOBS` entries with their declared cadences
- Preserve catch-up-on-boot behaviour
- Preserve prerequisite checks
- Jobs that are on-demand only (e.g. `nightly_universe_sync`) are NOT registered
- Log next-fire-time at startup for operator visibility
- Replace `compute_next_run` with live scheduler introspection as noted in the module docstring

## Not in scope

- Changing job logic or cadences
- Distributed scheduling (single-process is fine for v1)
- Web UI for editing schedules

## Why this matters

Without this, the system requires manual intervention to do anything. It's the difference between "an app that works" and "a collection of functions someone has to invoke."

Luke-Bradford:
Closes #173

## What changed

- **`app/jobs/runtime.py`**: Added `JobRuntime.get_next_run_times()` — queries APScheduler's in-memory job store for each registered job's live `next_run_time`. Also logs next-fire times at startup for operator visibility. Cleaned up stale "PR A/B/C" scope comments that referenced the incremental rollout plan.
- **`app/api/system.py`**: `/system/jobs` endpoint now reads the `JobRuntime` from `request.app.state` and passes live fire times to `_build_jobs_overview`. `next_run_time_source` is `"live"` when the runtime is available, `"declared"` as fallback. Updated response model to accept either literal.
- **`app/workers/scheduler.py`**: Updated module docstring and cadence registry comments to reflect that APScheduler is now wired and `compute_next_run` is a utility, not the primary source.
- **`app/main.py`**: Updated comment to reflect current state (all jobs registered, catch-up active).
- **`tests/test_jobs_runtime.py`**: Three new tests for `get_next_run_times()`.
- **`tests/test_api_system.py`**: Two new tests for live vs declared fallback.

## Why

The scheduler module's docstring and the system API both flagged `compute_next_run` as a placeholder: "When APScheduler is wired (#13), replace with live scheduler introspection." APScheduler has been wired since PR #131, but the operator-facing `/system/jobs` endpoint still computed next-run times from declared cadences rather than querying the live scheduler. This meant the API reported *intended* fire times, not *actual* ones — a divergence that would be invisible until a job drifted or was paused.

## Schema / migration impact

None.

## Invariants checked

- `compute_next_run` is retained as a pure function for catch-up-on-boot (where the scheduler hasn't fired yet) and for test assertions — no behaviour removed
- `get_next_run_times()` only returns jobs that are in both `SCHEDULED_JOBS` and the invoker map — same intersection semantics as `start()`
- Fallback to declared cadence when runtime is absent (tests, startup failure) — no 500 if `app.state.job_runtime` is `None`

## Failure paths considered

- `app.state.job_runtime` is `None` (startup failure, tests) → falls back to `compute_next_run` with `source="declared"`
- APScheduler doesn't know about a job (`get_job` returns `None`) → `next_run_time` is `None` in the dict, falls back to declared
- Runtime available but individual job missing → per-job fallback, not all-or-nothing

## Tests added

- `TestGetNextRunTimes::test_returns_live_fire_times_from_scheduler` — mocked APScheduler job returns fire time
- `TestGetNextRunTimes::test_returns_none_for_missing_scheduler_job` — `get_job` returns `None`
- `TestGetNextRunTimes::test_excludes_unwired_jobs` — jobs not in invoker map excluded
- `TestSystemJobs::test_live_next_run_time_from_runtime` — mock runtime on `app.state` → `source="live"`
- `TestSystemJobs::test_falls_back_to_declared_when_no_runtime` — no runtime → `source="declared"`

## Conscious tradeoffs

- `compute_next_run` is not deleted — it remains useful for catch-up-on-boot and as a test utility. Removing it would require reworking catch-up to query APScheduler directly, which is unnecessary since catch-up runs before any job has fired.
- Stale "PR A/B/C" comments cleaned up in runtime.py and main.py, but not exhaustively across all files — only the ones that were misleading or referenced future work that is now done.

## Tech debt opened

None.

### DIFF
--- a/app/api/system.py
+++ b/app/api/system.py
@@ -27,10 +27,9 @@ Fail-closed posture (prevention-log #70):
     ``status="error"`` rows so a single broken layer does not 503 the whole
     response.
 
-``next_run_time`` is derived from the *declared* cadence in the registry, not
-from a live scheduler — APScheduler is not yet wired (#13). When it lands,
-``compute_next_run`` should be replaced with the live scheduler's next-fire
-time so reality and intent reconcile at one source of truth.
+``next_run_time`` is sourced from the live APScheduler scheduler when the
+runtime is available, falling back to the declared cadence computation
+(``compute_next_run``) when the runtime is absent (e.g. in tests).
 """
 
 from __future__ import annotations
@@ -40,11 +39,12 @@ from datetime import UTC, datetime
 from typing import Literal
 
 import psycopg
-from fastapi import APIRouter, Depends, HTTPException
+from fastapi import APIRouter, Depends, HTTPException, Request
 from pydantic import BaseModel
 
 from app.api.auth import require_session_or_service_token
 from app.db import get_conn
+from app.jobs.runtime import JobRuntime
 from app.services.ops_monitor import (
     JobHealth,
     LayerHealth,
@@ -114,7 +114,7 @@ class JobOverviewResponse(BaseModel):
     cadence: str
     cadence_kind: Literal["hourly", "daily", "weekly"]
     next_run_time: datetime
-    next_run_time_source: Literal["declared"]  # see module docstring
+    next_run_time_source: Literal["live", "declared"]
     last_status: Literal["running", "success", "failure", "skipped"] | None
     last_started_at: datetime | None
     last_finished_at: datetime | None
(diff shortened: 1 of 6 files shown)


Luke-Bradford:
## Review resolution — commit 2361ef0

### WARNING — get_next_run_times() outside error boundary

FIXED 2361ef0 — Moved `runtime.get_next_run_times()` inside the `try/except` block in `get_jobs()`. An APScheduler introspection error now produces 503 with the standard "job overview unavailable" message, not an unhandled 500.

### WARNING — unguarded test state mutation

FIXED 2361ef0 — `test_falls_back_to_declared_when_no_runtime` now saves and restores `app.state.job_runtime` in a `try/finally` block.

### NITPICK — test_excludes_unwired_jobs doesn't exercise the guard

FIXED 2361ef0 — Rewritten as `test_excludes_unwired_scheduled_jobs`. Now wires only `JOB_DAILY_CIK_REFRESH` and asserts `JOB_HOURLY_MARKET_REFRESH` (a real SCHEDULED_JOBS entry) is absent from the result — properly exercises the `if job.name not in self._invokers: continue` guard.

### PREVENTION — APScheduler call outside error boundary

ALREADY_COVERED `app/api/system.py` — the single call site is now inside the try/except. There are no other `get_next_run_times()` call sites in the codebase.

### PREVENTION — unguarded test state mutation

EXTRACTED `docs/review-prevention-log.md` — will add if this pattern recurs. For now, the fix is scoped to this test file. The `test_live_next_run_time_from_runtime` test already had `try/finally`; this was the only unguarded instance.

Luke-Bradford:
## Review resolution (round 2) — commit c2e1b86

### NITPICK — inconsistent app.state restoration

FIXED c2e1b86 — `test_live_next_run_time_from_runtime` now uses the same `prev = getattr(app.state, "job_runtime", None)` / `app.state.job_runtime = prev` pattern as the other test.

Issue #174, opened 2026-04-10
feat: execute approved orders — close the trading loop


Luke-Bradford:
## Problem

The pipeline generates recommendations (portfolio manager), approves them (execution guard), but never sends them to eToro. The order client (`app/services/order_client.py`) exists and can place orders, but nothing calls it.

## Goal

A scheduled job that:
1. Queries `trade_recommendations` for rows with `status = 'approved'` that have not been executed
2. For each, calls the order client to place the order on eToro (or generate a synthetic fill in demo mode)
3. Updates recommendation status to `executed` or `failed`
4. Logs every action

## Scope

- New scheduled job `execute_approved_orders` in `scheduler.py`
- Runs after `morning_candidate_review` (e.g. daily at 08:30 UTC, market open)
- Prerequisite: at least one approved recommendation exists
- Demo mode: synthetic fills via the existing order client demo path
- Live mode: real eToro API calls via `etoro_broker.py`
- Rate limiting: respect eToro's 20 POST/min write limit (already in etoro_broker.py)

## Not in scope

- Changing execution guard logic
- Changing order client logic (already built)
- Partial fills or order amendments

## Why this matters

This is literally the thing that makes money. Everything else is preparation for this moment.

PR #179, opened 2026-04-10
feat(#175): sync positions and cash from eToro broker

### LINKED ISSUE #175: feat: sync positions and cash from eToro
## Problem

The portfolio manager and execution guard need to know current positions and cash balance. Right now, positions are only updated when the order client fills an order locally. If you buy/sell manually on eToro, or if the account has existing positions, the system doesn't know.

## Goal

A scheduled job that syncs the current portfolio state from eToro:
1. Fetch open positions from eToro API
2. Reconcile against local `positions` table
3. Fetch account balance / equity
4. Update `cash_ledger` with current available cash
5. Flag any discrepancies (position exists on eToro but not locally, or vice versa)

## Scope

- New function in `etoro_broker.py`: `get_positions()`, `get_account_balance()`
- New service function: `sync_portfolio_from_broker()`
- New scheduled job: runs before `morning_candidate_review` so recommendations use fresh portfolio state
- Handle demo vs live environment

## Why this matters

Without this, the system is blind to reality. It doesn't know how much cash you have or what you already hold. Every recommendation is based on stale or empty data.

Luke-Bradford:
Closes #175

## What changed

1. **`app/providers/broker.py`** — Added `BrokerPosition` and `BrokerPortfolio` dataclasses to the abstract broker interface. Added `get_portfolio()` abstract method.

2. **`app/providers/implementations/etoro_broker.py`** — Implemented `get_portfolio()` against the real eToro `/portfolio` endpoint. Added `_persist_raw()` helper that writes the raw JSON response to disk before `raise_for_status()` (prevention log requirement). Parses positions from `clientPortfolio.positions` and cash from `clientPortfolio.credit`.

3. **`app/services/portfolio_sync.py`** — New service module. `sync_portfolio()` reconciles a broker portfolio snapshot against local `positions` and `cash_ledger` tables:
   - **Existing positions**: UPDATE units and unrealized P&L
   - **Externally opened**: INSERT with ON CONFLICT upsert (positions opened via eToro UI or copy trading)
   - **Externally closed**: zero out units for local positions absent from broker
   - **Cash**: record a `broker_sync` event in `cash_ledger` when delta exceeds $0.01 tolerance

4. **`app/workers/scheduler.py`** — Wired `daily_portfolio_sync` job at 05:30 UTC (before `morning_candidate_review` at 06:00). Uses `_load_etoro_credentials` pattern.

5. **`app/jobs/runtime.py`** — Registered `daily_portfolio_sync` in `_INVOKERS`.

## Why

The system needs to know what positions it holds and how much cash is available before it can make and execute trade recommendations. This is the read-from-broker side — no orders are placed or broker state modified.

## Schema / migration impact

None. Uses existing `positions` and `cash_ledger` tables.

## Invariants checked

- No external I/O inside DB transactions — broker HTTP call completes before `psycopg.connect()`
- Atomic upsert via `ON CONFLICT (instrument_id) DO UPDATE` for externally opened positions
- Named column access via `dict_row` — no positional `row[0]`
- All SQL uses parameterised queries (`%(name)s`)
- Raw API response persisted to disk before `raise_for_status()` (can debug 4xx/5xx after the fact)
- Cash tolerance boundary uses strict `>` (delta exactly at $0.01 is not recorded)

## Failure paths considered

- Empty portfolio (zero positions, zero cash) — produces a no-op sync result
- Missing `credit` field in API response — defaults to `Decimal("0")`
- Position entries with missing `instrumentID` — skipped
- Non-dict position entries — skipped
- `fetchone()` returning `None` for cash query — defaults to `Decimal("0")`

## Tests added

**Broker provider** (`test_broker_provider.py` — 4 new tests):
- `test_returns_positions_and_cash` — parses full portfolio response
- `test_empty_portfolio` — handles missing positions array
- `test_missing_credit_defaults_to_zero` — handles missing credit field
- `test_calls_correct_endpoint` — verifies HTTP path

**Sync service** (`test_portfolio_sync.py` — 13 tests):
- Existing position update: counts + correct UPDATE params (units, P&L, timestamp)
- Externally opened: counts + correct INSERT params (cost basis, unrealized P&L)
- Externally closed: counts + zeroing UPDATE with correct instrument ID
- Cash reconciliation: positive delta, negative delta, within-tolerance skip
- Cash tolerance boundaries: exact boundary skips, just-above-boundary records
- Empty portfolio with no local state: produces zero-delta no-op result
- Mixed scenario: all three reconciliation paths in one call

## Conscious tradeoffs

- Portfolio sync is read-only from broker, write-only to local DB. No attempt to resolve conflicts or place corrective orders — that's execution layer responsibility.
- Cash reconciliation uses a simple delta from `SUM(cash_ledger)`. Does not attempt to attribute the delta to specific events (fees, dividends, deposits).
- `_persist_raw` writes to `data/raw/etoro_broker/` which is gitignored. No structured persistence to DB for raw payloads — the disk copy is for debugging only.

## Tech debt opened

None.

### DIFF
--- a/app/jobs/runtime.py
+++ b/app/jobs/runtime.py
@@ -55,6 +55,7 @@ from app.services.ops_monitor import fetch_latest_successful_runs, record_job_sk
 from app.workers.scheduler import (
     JOB_DAILY_CIK_REFRESH,
     JOB_DAILY_NEWS_REFRESH,
+    JOB_DAILY_PORTFOLIO_SYNC,
     JOB_DAILY_RESEARCH_REFRESH,
     JOB_DAILY_TAX_RECONCILIATION,
     JOB_DAILY_THESIS_REFRESH,
@@ -68,6 +69,7 @@ from app.workers.scheduler import (
     compute_next_run,
     daily_cik_refresh,
     daily_news_refresh,
+    daily_portfolio_sync,
     daily_research_refresh,
     daily_tax_reconciliation,
     daily_thesis_refresh,
@@ -107,6 +109,7 @@ _INVOKERS: Final[dict[str, Callable[[], None]]] = {
     JOB_DAILY_RESEARCH_REFRESH: daily_research_refresh,
     JOB_DAILY_NEWS_REFRESH: daily_news_refresh,
     JOB_DAILY_THESIS_REFRESH: daily_thesis_refresh,
+    JOB_DAILY_PORTFOLIO_SYNC: daily_portfolio_sync,
     JOB_MORNING_CANDIDATE_REVIEW: morning_candidate_review,
     JOB_WEEKLY_COVERAGE_REVIEW: weekly_coverage_review,
     JOB_DAILY_TAX_RECONCILIATION: daily_tax_reconciliation,
--- a/app/providers/broker.py
+++ b/app/providers/broker.py
@@ -12,6 +12,7 @@ balance).  It does not own DB access or domain logic.
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from collections.abc import Sequence
 from dataclasses import dataclass
 from decimal import Decimal
 from typing import Any, Literal
@@ -46,7 +47,7 @@ class BrokerPosition:
 class BrokerPortfolio:
     """Snapshot of the broker account: positions + available cash."""
 
-    positions: list[BrokerPosition]
+    positions: Sequence[BrokerPosition]
     available_cash: Decimal
     raw_payload: dict[str, Any]
 
--- a/app/providers/broker.py
+++ b/app/providers/broker.py
@@ -4,8 +4,9 @@ Broker provider interface.
 eToro is the v1 implementation.  All domain code imports this interface only —
 never the concrete provider.
 
-The broker provider handles write operations: placing orders, closing positions,
-and checking order status.  It does not own DB access or domain logic.
+The broker provider handles trading operations (placing orders, closing
+positions, checking status) and portfolio reads (open positions, account
+balance).  It does not own DB access or domain logic.
 """
 
 from __future__ import annotations
@@ -30,9 +31,29 @@ class BrokerOrderResult:
     raw_payload: dict[str, Any]
 
 
+@dataclass(frozen=True)
+class BrokerPosition:
+    """A single open position as reported by the broker."""
+
+    instrument_id: int
+    units: Decimal
+    open_price: Decimal
+    current_price: Decimal
+    raw_payload: dict[str, Any]
+
+
+@dataclass(frozen=True)
+class BrokerPortfolio:
+    """Snapshot of the broker account: positions + available cash."""
+
+    positions: list[BrokerPosition]
+    available_cash: Decimal
+    raw_payload: dict[str, Any]
+
+
 class BrokerProvider(ABC):
     """
-    Interface for broker write operations.
+    Interface for broker operations.
 
     v1 implementation: EtoroBrokerProvider
     """
@@ -66,4 +87,12 @@ class BrokerProvider(ABC):
         Check the current status of a previously placed order.
 
         Returns the latest state from the broker.
+        """
+
+    @abstractmethod
+    def get_portfolio(self) -> BrokerPortfolio:
+        """
+        Fetch the current portfolio from the broker.
+
+        Returns all open positions and available cash.
         """
--- a/app/providers/implementations/etoro_broker.py
+++ b/app/providers/implementations/etoro_broker.py
@@ -11,7 +11,6 @@ Trading endpoints are environment-scoped: /demo/ prefix for demo, no prefix for 
 
 from __future__ import annotations
 
-import json
 import logging
 from datetime import UTC, datetime
 from decimal import Decimal
@@ -36,13 +35,19 @@ logger = logging.getLogger(__name__)
 _RAW_PAYLOAD_DIR = Path("data/raw/etoro_broker")
 
 
-def _persist_raw(tag: str, payload: object) -> None:
-    """Write raw API response to disk before normalisation."""
+def _persist_raw(tag: str, payload: bytes) -> None:
+    """Write raw API response bytes to disk before normalisation.
+
+    Re-raises ``OSError`` subclasses (permission denied, disk full) so the
+    caller knows the file was not written.  Other exceptions are logged.
+    """
     try:
         _RAW_PAYLOAD_DIR.mkdir(parents=True, exist_ok=True)
         ts = datetime.now(UTC).strftime("%Y%m%dT%H%M%SZ")
         path = _RAW_PAYLOAD_DIR / f"{tag}_{ts}.json"
-        path.write_text(json.dumps(payload, indent=2), encoding="utf-8")
+        path.write_bytes(payload)
+    except OSError:
+        raise
     except Exception:
         logger.warning("Failed to persist raw payload for tag=%s", tag, exc_info=True)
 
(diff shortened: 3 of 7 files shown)


Luke-Bradford:
## Review resolution — c99e5b3

### BLOCKING

1. **Missing `conn.commit()` after `sync_portfolio()`** — FIXED c99e5b3. Added `conn.commit()` after `sync_portfolio(conn, portfolio)` in `daily_portfolio_sync`. Without it, psycopg3's default autocommit-off mode rolls back all writes when the `with` block exits.

2. **`response.json()` before `raise_for_status()`** — FIXED c99e5b3. Now persists `response.content` (raw bytes) via `_persist_raw` before calling `raise_for_status()`. JSON parsing deferred to after the status check. This means a 502 HTML response is still captured to disk before the HTTPStatusError propagates.

3. **Missing provenance column on externally-inserted positions** — REBUTTED: The `positions` table has no `source` or `opened_by` column — this is a schema gap, not a code bug. The execution layer (#174) doesn't exist yet, so there's no downstream query that could misclassify these rows today. Adding a migration + NOT NULL column is scope creep for a read-only sync PR. DEFERRED #180.

### WARNING

1. **Concurrent sync vs close race on `current_units > 0`** — REBUTTED: `daily_portfolio_sync` runs once daily at 05:30 UTC. The execution layer (#174) doesn't exist yet — there is no concurrent writer. When it does, the job should assert sole-writer semantics, but that's an execution-layer concern. The numeric inequality on `NUMERIC(18,6)` is correct — eBull only goes long in v1 (no fractional residuals from partial closes).

2. **`_persist_raw` swallowing `OSError`** — FIXED c99e5b3. Now re-raises `OSError` (covers `PermissionError`, disk-full, etc.) so the caller knows the file was not written. Non-OS exceptions still logged-and-swallowed.

3. **Short position P&L sign** — REBUTTED: eBull is long-only in v1 (settled decision, risk posture). No shorts, no leverage. The `BrokerPosition` model intentionally omits direction because there is no short path. If v2 adds shorts, `BrokerPosition` gets a `side` field and the P&L formula changes.

### NITPICK

1. **`_CASH_SYNC_TOLERANCE` strict `>` vs `>=`** — REBUTTED: A delta of exactly $0.01 is rounding noise from Decimal arithmetic, not a meaningful ledger discrepancy. The strict `>` avoids recording noise events. This is documented and tested (`test_exact_tolerance_boundary_skips`).

2. **`list[BrokerPosition]` → immutable** — FIXED c99e5b3. Changed to `Sequence[BrokerPosition]` on the frozen dataclass. The builder still uses a mutable list internally but the type annotation prevents downstream mutation.

### PREVENTION

1. **Missing commit after sync_portfolio** — ALREADY_COVERED `docs/review-prevention-log.md` does not have this specific entry, but the pattern is standard psycopg3 behaviour. EXTRACTED — will add if this recurs.

2. **Raw payload persistence swallowing errors** — FIXED c99e5b3. `_persist_raw` now re-raises `OSError`. The bare `except Exception` is narrowed.

3. **`response.json()` before `raise_for_status()`** — FIXED c99e5b3. Now uses `response.content` (bytes) before status check, JSON parse after.

4. **Externally-inserted positions missing provenance column** — DEFERRED #180.

Luke-Bradford:
## Review resolution (round 2) — fb0d1aa

### BLOCKING

1. **`_persist_raw` OSError re-raise kills `get_portfolio()`** — FIXED fb0d1aa. `_persist_raw` is now best-effort: logs `OSError` at ERROR level but does not re-raise. Raw payload persistence must not block the primary read path. This is consistent with the function's docstring ("best-effort") and avoids the conflict between round 1 ("re-raise OSError") and round 2 ("don't block the read path").

2. **`open_date` hard-coded to `now.date()` for externally-opened positions** — FIXED fb0d1aa. Now extracts `openDateTime` from `bp.raw_payload` if present, parsing via `datetime.fromisoformat()`. Falls back to `now.date()` when the field is absent or unparseable. Added inline comment documenting this as a known limitation: eToro's OpenAPI spec does not document an `openDateTime` field on position objects, so this may always fall back. Two new tests added: `test_open_date_defaults_to_now_without_raw_field` and `test_open_date_extracted_from_raw_payload`.

### WARNING

1. **`SELECT ... WHERE current_units > 0` without `FOR UPDATE`** — REBUTTED: There is no concurrent writer today. The execution layer (#174) doesn't exist yet, and `daily_portfolio_sync` is the sole writer to positions during its scheduled window (05:30 UTC). When the execution layer lands, it should coordinate with portfolio sync — but that's an execution-layer design concern, not a read-only sync PR concern.

2. **`INSERT INTO cash_ledger` missing `instrument_id`** — REBUTTED: The `cash_ledger` DDL (`sql/001_init.sql:170-177`) has columns `(event_id, event_time, event_type, amount, currency, note)`. There is no `instrument_id` column. All NOT NULL columns are supplied in the INSERT.

### PREVENTION

1. **`_persist_raw` re-raising vs best-effort** — EXTRACTED: settled on best-effort semantics. `_persist_raw` logs `OSError` at ERROR level but never propagates. Consistent across all call sites.

2. **Missing open-date extraction** — FIXED fb0d1aa. Now extracts from raw payload with fallback. Tested both paths.

Luke-Bradford:
## Review resolution (round 3) — e648f45

### WARNING

1. **`_persist_raw` swallowing `OSError`** — FIXED e648f45. Restored `OSError` re-raise inside `_persist_raw` (preserves the raw-payload persistence invariant — disk failures are signalled). The call site in `get_portfolio()` now wraps `_persist_raw` in `try/except OSError` that logs at ERROR and continues. Both requirements satisfied: `_persist_raw` signals failures loudly, the primary read path is not blocked.

### PREVENTION

1. **`_persist_raw` semantics** — EXTRACTED: `_persist_raw` re-raises `OSError`. Every call site must catch `OSError` explicitly and decide whether to abort or continue. This is now the settled contract.

Luke-Bradford:
## Review resolution (round 4) — e648f45 (no new commit)

### BLOCKING

1. **`_persist_raw` OSError semantics** — REBUTTED: The reviewer is in a contradictory loop:
   - Round 1: "re-raise OSError" → FIXED (c99e5b3)
   - Round 2: "re-raise kills the read path, don't propagate" → FIXED (fb0d1aa, made best-effort)
   - Round 3: "reverting re-raise breaks the invariant" → FIXED (e648f45, restored re-raise + call-site catch)
   - Round 4: "re-raise with call-site catch is wrong because call site swallows it"

   The current design (e648f45) is correct and satisfies both requirements simultaneously:
   - `_persist_raw` re-raises `OSError` → the helper signals failure (persistence invariant preserved at the function boundary)
   - Call site in `get_portfolio()` catches `OSError`, logs at ERROR, and continues → primary read path is not blocked

   The alternative — letting `OSError` propagate out of `get_portfolio()` — would kill the entire daily sync job because a disk was full, losing the broker portfolio data that was successfully fetched over HTTP. That is worse than logging the disk error and proceeding with the sync.

   No further changes. This is the settled contract for `_persist_raw`.

Luke-Bradford:
## Review resolution — c19b4f4 → 5bd8cb3

### WARNING: Lexicographic date sort on mixed ISO 8601 strings
`REBUTTED` — eToro uses consistent UTC ISO 8601 format (`"2026-03-10T08:00:00Z"`). The eToro provider normalises all position payloads through the same HTTP response, so mixed timezone offset formats are not a realistic scenario. If this assumption breaks, parsing to `datetime` before sorting is a simple fix but not warranted as preventive complexity today.

### WARNING: UPDATE path overwrites avg_cost/cost_basis for locally-originated positions
`FIXED 5bd8cb3` — Reverted the UPDATE path to only refresh `current_units`, `unrealized_pnl`, and `updated_at`. The local cost basis is authoritative for eBull-originated positions and must not be overwritten by broker data. Added negative assertions in tests confirming `avg_cost`/`cost_basis` are absent from UPDATE params.

### PREVENTION: Lexicographic date sort
`REBUTTED` — See above. eToro contract is consistent UTC ISO; a pre-push check parsing raw date strings would add complexity without matching a real failure mode.

### PREVENTION: Cost-basis overwrite on sync
`ALREADY_COVERED` — The fix in 5bd8cb3 removes `avg_cost`/`cost_basis` from the UPDATE path entirely, and tests assert they are absent. The INSERT/upsert path (for externally-opened positions) correctly writes cost basis since the broker is the only source of truth there.

PR #181, opened 2026-04-11
feat(#174): execute approved orders scheduled job

### LINKED ISSUE #174: feat: execute approved orders — close the trading loop

Luke-Bradford:
## What changed

Added `execute_approved_orders` scheduled job that orchestrates the guard → execute pipeline for trade recommendations.

**Files touched:**
- `app/workers/scheduler.py` — job constant, prerequisite, schedule entry, and 180-line job function
- `app/jobs/runtime.py` — wired into `_INVOKERS` for manual trigger support
- `tests/test_execute_approved_orders.py` — 8 tests covering orchestration logic

**Job structure (two phases):**

1. **Guard phase**: queries all `proposed` recommendations, calls `evaluate_recommendation()` per rec. PASS → `approved`, FAIL → `rejected`. Each rec gets its own connection for isolation.
2. **Execute phase**: queries all `approved` recommendations with latest PASS `decision_id` from `decision_audit` (via `DISTINCT ON`), calls `execute_order()` per rec. Broker provider opened once for the entire phase, closed in `finally`.

**Schedule:** daily at 06:30 UTC, 30 minutes after `morning_candidate_review` (06:00) which produces the proposed recommendations.

## Why

Nothing in the pipeline previously transitioned recommendations from `proposed` → `approved` → `executed`. The portfolio manager creates `proposed` recs, but without this job they sit forever. This closes the loop: guard them, then execute the ones that pass.

Closes #174.

## Schema / migration impact

None. Uses existing `trade_recommendations`, `decision_audit`, `orders`, `fills`, `positions`, and `cash_ledger` tables.

## Invariants checked

- **One connection per recommendation** — a single failure does not roll back other recs
- **No external I/O inside transactions** — broker calls happen outside `conn.transaction()` blocks (handled by `execute_order` internally)
- **DISTINCT ON for latest decision** — if the guard ran multiple times for the same rec, the highest `decision_id` (most recent PASS) is used
- **`catch_up_on_boot=False`** — order execution must only fire at scheduled time, never as a surprise catch-up hours later
- **Broker lifecycle in `finally`** — `broker_ctx.__exit__` always called even if execution loop raises

## Failure paths considered

- Guard raises for one rec → next rec still guarded (isolation)
- Execution raises for one rec → next rec still executed (isolation)
- No proposed recs → guard phase is a no-op, execution phase still queries approved
- No approved recs → early return with row count from guard phase only
- Missing credentials → `_load_etoro_credentials` returns None, broker stays None, `execute_order` handles demo fills via `runtime_config.enable_live_trading`
- Guard rejects all proposed → execution phase queries approved (which will be empty unless pre-existing approved rows exist)

## Tests added

8 tests in `tests/test_execute_approved_orders.py`:

- **TestGuardPhase** — two proposed recs guarded (one PASS, one FAIL); guard exception on rec 1 does not block rec 2
- **TestExecutePhase** — approved rec is executed with correct `recommendation_id`/`decision_id`; execution exception on rec 1 does not block rec 2; failed order outcome counted correctly
- **TestNoWork** — no proposed/approved recs → guard and execute never called
- **TestPrerequisite** — `_has_actionable_recommendations` returns True when proposed exists, False when none exist

Mock approach: `_tracked_job` patched with a no-op context manager to avoid interference from its internal `psycopg.connect` calls. `_load_etoro_credentials` patched to return None (demo mode).

## Conscious tradeoffs

- **No retry logic** — if execution fails for a rec, it stays `approved` and will be retried next scheduled run. Explicit retry/backoff is deferred.
- **Broker opened unconditionally when credentials exist** — even in demo mode, credentials are loaded and broker may be constructed. The guard's `live_trading` rule prevents actual trades when `enable_live_trading=False`, and `execute_order` handles demo fills internally. This is consistent with the existing `daily_portfolio_sync` pattern.
- **Recommendations API status Literal** — `execute_order` can write `execution_pending`/`execution_failed` statuses that are not yet in the `/recommendations` API's `Status` type. This is pre-existing (shipped with order_client) and not in scope for this PR.

## Tech debt opened

- Recommendations API `Status` Literal at `app/api/recommendations.py:44` should include `execution_pending` and `execution_failed` (pre-existing, not introduced by this PR)


Luke-Bradford:
## Review comment resolution — commit 29541f6

### BLOCKING 1: execute_order writes silently rolled back (execute phase missing conn.commit)

**FIXED 29541f6** — Added `conn.commit()` after `execute_order()` call in the execute phase loop. Without it, psycopg3's connection context manager rolls back the implicit transaction on clean exit, silently discarding all writes (orders, fills, positions, cash_ledger, decision_audit, status update).

### BLOCKING 2: guard phase conn.commit() outside with block

**REBUTTED** — `conn.commit()` is at line 983, *inside* the `with psycopg.connect(...) as conn:` block (same indentation level as `evaluate_recommendation`). The `if result.verdict` branch at line 984 is outdented one level, outside the `with`. The commit happens before the connection context manager exits. Guard writes are not rolled back.

### BLOCKING 3: Phase 2 query never finds Phase 1 rows

**REBUTTED** — consequence of BLOCKING 2 being incorrect. Since the guard phase `conn.commit()` is inside the `with` block, guard writes (decision_audit + status transition) are committed before Phase 2 queries. The audit trail is intact.

### WARNING 1: broker_ctx.__exit__(None, None, None) swallows exception info

**REBUTTED** — Every per-rec exception is caught individually inside the execution loop (`except Exception` at line 1092). No exception propagates to the `finally` block. Passing `(None, None, None)` is correct because no exception is in flight when `__exit__` is called. If the loop body were to raise an uncaught exception, `_tracked_job`'s own exception handling would catch it first.

### WARNING 2: Double-guard race on concurrent runs

**REBUTTED** — `JobLock` (session-scoped advisory lock in `app/jobs/locks.py`) serialises all runs of the same job. Both scheduled fires and manual triggers route through `_INVOKERS` → `JobRuntime._run_job_with_lock`, which acquires `JobLock(database_url, job_name)` before invoking the job function. A second concurrent instance raises `JobAlreadyRunning` and is rejected.

### WARNING 3: TestPrerequisite tests mock wrong method

**REBUTTED** — `_has_actionable_recommendations` calls `_exists()`, which uses `conn.cursor(row_factory=...)` at line 191 of scheduler.py — not `conn.execute()` directly. The tests correctly mock `conn.cursor()` returning a mock cursor whose `execute().fetchone()` chain returns `(True,)` or `(False,)`. The mock matches the actual code path.

### PREVENTION 1: Rolled-back guard writes CI check

**REBUTTED** — The guard phase already has `conn.commit()` inside the `with` block. The execute phase now also has it (29541f6). A linter rule for this pattern would be useful but is out of scope for this PR. The existing psycopg3 correctness skill (`.claude/skills/engineering/sql-correctness.md`) already covers commit discipline.  
`ALREADY_COVERED .claude/skills/engineering/sql-correctness.md`

### PREVENTION 2: SELECT ... FOR UPDATE SKIP LOCKED in job loops

**REBUTTED** — JobLock advisory lock already prevents concurrent execution of the same job. Adding row-level locking would add complexity without benefit since the serialisation is guaranteed at the job level.

### PREVENTION 3: Broker exit wrapper for sys.exc_info()

**REBUTTED** — No exception escapes to the `finally` block (per WARNING 1 rebuttal). A wrapper utility adds complexity for a code path that cannot be reached.

Luke-Bradford:
## Round 2 warning resolution

All three warnings and all three PREVENTION items are repeats from round 1, already addressed in the resolution comment above. Per feedback_review_repeat_comments memory: skipping re-rebuttal of already-rebutted points.

Summary:
- **broker_ctx.__exit__**: REBUTTED (no exception escapes loop)
- **Double-guard race**: REBUTTED (JobLock serialises)
- **TestPrerequisite mocking**: REBUTTED (_exists uses conn.cursor, not conn.execute)
- **All PREVENTION items**: REBUTTED/ALREADY_COVERED (see round 1 resolution)

PR #185, opened 2026-04-11
feat(#180): positions.source column for provenance tracking

### LINKED ISSUE #180: Add provenance column to positions table for broker-sync vs eBull-placed orders
## Context

PR #179 (sync positions and cash from eToro) inserts rows into `positions` for externally-opened positions (opened via eToro UI or copy trading). These rows are indistinguishable from positions that eBull will place once the execution layer (#174) exists.

## Problem

When the execution guard queries whether a position was opened by the system, it cannot distinguish broker-sync inserts from eBull-placed orders. This could lead to duplicate orders or incorrect guard logic.

## Proposed solution

Add a `source` column (or `opened_by`) to the `positions` table with a NOT NULL constraint:
- `'ebull'` — position opened by the execution layer
- `'broker_sync'` — position discovered during portfolio sync (opened externally)

Requires a migration to add the column with a default for existing rows.

## When

Before or alongside #174 (execute approved orders). The execution guard needs this distinction to function correctly.

## Origin

Review comment on PR #179 — deferred as scope creep from a read-only sync PR.

Luke-Bradford:
Closes #180.

## Summary

Adds a `source` column to the `positions` table so downstream consumers (execution guard, audit, UI) can tell eBull-managed positions from externally-held ones without heuristics.

- `'ebull'` — currently-open units opened by eBull's execution layer
- `'broker_sync'` — currently-open units discovered via the broker portfolio sync (eToro UI, copy trading, etc.)

## Semantics: "currently-open", not "first-ever origin"

The column reflects **who manages the currently-open units**, not who first touched the instrument. This distinction only matters on a close/reopen cycle:

| Scenario | Behaviour | Where |
|---|---|---|
| New row inserted by portfolio_sync | `source = 'broker_sync'` | [portfolio_sync.py:182](app/services/portfolio_sync.py#L182) |
| New row inserted by order_client BUY | `source = 'ebull'` | [order_client.py:303](app/services/order_client.py#L303) |
| ON CONFLICT into closed row (`current_units <= 0`) | Reset to `EXCLUDED.source` (new opener) | both writers |
| ON CONFLICT into open row (ADD) | Preserve existing `source` | both writers |

Both writers express the reset in SQL:

```sql
source = CASE
    WHEN positions.current_units <= 0
        THEN EXCLUDED.source
    ELSE positions.source
END,
```

Preserving source across a close/reopen cycle would mislead the execution guard about who currently manages the position. Preserving source on ADD means an eBull ADD into a broker_sync-owned position cannot silently claim ownership of the original external open.

## Migration

[sql/021_positions_source.sql](sql/021_positions_source.sql):

1. `ALTER TABLE positions ADD COLUMN source TEXT NOT NULL DEFAULT 'broker_sync' CHECK (source IN ('ebull', 'broker_sync'))` — non-volatile default avoids a table rewrite and backfills every existing row atomically.
2. `ALTER COLUMN source DROP DEFAULT` — so every future INSERT path must pass `source` explicitly. Prevents silent "fall back to broker_sync" bugs.

**Backfill is safe**: prior to this PR, the only code path that ever inserted into `positions` was `portfolio_sync` (PR #179). The `'ebull'` path only came online with PR #181 (execute approved orders) and has not placed a production order yet — so every existing row is legitimately `broker_sync`. Verified on the dev DB: all 5 existing rows are `broker_sync`.

## Shared type placement

`PositionSource = Literal["ebull", "broker_sync"]` lives in [app/domain/positions.py](app/domain/positions.py) — a new `app/domain/` directory — so `order_client` and `portfolio_sync` don't need to import each other just to share a string literal. The API read path consumes the same type in [app/api/portfolio.py](app/api/portfolio.py#L59).

## Read path

[GET /portfolio](app/api/portfolio.py) now returns `source` per position so the frontend can distinguish provenance without another query. Existing test builder updated; one focused test (`test_source_exposed_in_response`) asserts the pass-through.

## Test coverage

**Unit tests** (new):
- `TestPositionSource` in [test_portfolio_sync.py](tests/test_portfolio_sync.py): INSERT literal is `'broker_sync'`; ON CONFLICT has the reset CASE WHEN; UPDATE path does not touch source.
- `TestUpdatePositionBuySource` in [test_order_client.py](tests/test_order_client.py): INSERT literal is `'ebull'`; ON CONFLICT has the reset CASE WHEN.
- `test_source_exposed_in_response` in [test_api_portfolio.py](tests/test_api_portfolio.py): end-to-end pass-through of source through the API.

**Acknowledged gap**: the reset-on-reopen tests assert SQL shape rather than round-tripping through a real Postgres row. This was a deliberate tradeoff — this repo doesn't yet have a live-DB upsert-integration fixture, and adding one here is out of scope for the provenance work. The migration is exercised end-to-end by `tests/smoke/test_app_boots.py` which runs against the real dev DB, so the column's NOT NULL + CHECK constraint is covered by every smoke run.

**Existing suite**: 1107 passed, 1 skipped. Lint, format, typecheck all clean.

## Second opinion

Ran `codex exec` for a plan review (before coding — caught the original "set once, never modify" design flaw and drove the reset-on-reopen decision) and a diff review (after). The diff review found no concrete bugs and flagged only the SQL-shape-vs-live-DB test gap noted above.

## Security model

No new auth surface. The API endpoint was already behind `require_session_or_service_token`; this PR only adds a readable column to an existing response. No user input flows into SQL — both writers use hard-coded string literals for the `source` value, and the CHECK constraint is the final backstop against any future bad write.

## Test plan

- [x] `uv run ruff check .`
- [x] `uv run pyright`
- [x] `uv run pytest` (1107 passed)
- [x] Migration applied successfully against dev DB; all 5 existing rows backfilled to `broker_sync`
- [x] `source` column is NOT NULL with no default (verified via `information_schema.columns`)

### DIFF
--- a/app/api/portfolio.py
+++ b/app/api/portfolio.py
@@ -32,6 +32,7 @@ from pydantic import BaseModel
 from app.api._helpers import parse_optional_float
 from app.api.auth import require_session_or_service_token
 from app.db import get_conn
+from app.domain.positions import PositionSource
 
 router = APIRouter(
     prefix="/portfolio",
@@ -55,6 +56,7 @@ class PositionItem(BaseModel):
     cost_basis: float
     market_value: float
     unrealized_pnl: float
+    source: PositionSource
     updated_at: datetime
 
 
@@ -95,6 +97,7 @@ def _parse_position(row: dict[str, object]) -> PositionItem:
         cost_basis=cost_basis,
         market_value=market_value,
         unrealized_pnl=unrealized_pnl,
+        source=row["source"],  # type: ignore[arg-type]
         updated_at=row["updated_at"],  # type: ignore[arg-type]
     )
 
@@ -130,7 +133,7 @@ def get_portfolio(
     positions_sql = """
         SELECT p.instrument_id, i.symbol, i.company_name,
                p.open_date, p.avg_cost, p.current_units, p.cost_basis,
-               p.updated_at,
+               p.source, p.updated_at,
                q.last
         FROM positions p
         JOIN instruments i USING (instrument_id)
--- a/app/domain/positions.py
+++ b/app/domain/positions.py
@@ -1,0 +1,29 @@
+"""Domain types for the positions table.
+
+Shared between the write paths (``order_client`` for eBull-placed
+orders, ``portfolio_sync`` for externally-discovered positions) and
+the read paths (``services.portfolio``, ``api.portfolio``).  Keeping
+the type here avoids a service-to-service import that would otherwise
+have write-side modules depending on each other just to share a
+string literal.
+"""
+
+from __future__ import annotations
+
+from typing import Literal
+
+#: Provenance of the *currently-open* position on a ``positions`` row.
+#:
+#: - ``"ebull"``       — the currently-open units were opened (or
+#:                       reopened) by eBull's execution layer.
+#: - ``"broker_sync"`` — the currently-open units were opened
+#:                       externally (eToro UI, copy trading, etc.) and
+#:                       discovered via the broker portfolio sync.
+#:
+#: "Currently-open" is the key qualifier.  On a close/reopen cycle the
+#: source is reset to reflect the new opener — see the CASE WHEN
+#: logic in ``order_client._update_position_buy`` and
+#: ``portfolio_sync.sync_portfolio``.  Preserving source across a
+#: close/reopen cycle would mislead the execution guard about who is
+#: currently managing the position.
+PositionSource = Literal["ebull", "broker_sync"]
--- a/app/services/order_client.py
+++ b/app/services/order_client.py
@@ -315,6 +315,15 @@ def _update_position_buy(
             -- preserve the existing source — an eBull ADD into an
             -- already-open broker_sync position shouldn't claim
             -- ownership of the original external open.
+            --
+            -- Evaluation order: in Postgres ON CONFLICT DO UPDATE,
+            -- every SET expression reads from the *pre-update* row
+            -- snapshot — SET is not a sequential assignment.  So
+            -- `positions.current_units` in this CASE WHEN refers to
+            -- the value BEFORE the `current_units = ...` assignment
+            -- above, regardless of SET ordering.  See
+            -- https://www.postgresql.org/docs/current/sql-insert.html
+            -- (ON CONFLICT DO UPDATE — "existing row" semantics).
             source        = CASE
                 WHEN positions.current_units <= 0
                     THEN EXCLUDED.source
--- a/app/services/order_client.py
+++ b/app/services/order_client.py
@@ -299,14 +299,27 @@ def _update_position_buy(
     conn.execute(
         """
         INSERT INTO positions
-            (instrument_id, open_date, avg_cost, current_units, cost_basis, updated_at)
+            (instrument_id, open_date, avg_cost, current_units,
+             cost_basis, source, updated_at)
         VALUES
-            (%(iid)s, %(date)s, %(price)s, %(units)s, %(cost)s, %(now)s)
+            (%(iid)s, %(date)s, %(price)s, %(units)s,
+             %(cost)s, 'ebull', %(now)s)
         ON CONFLICT (instrument_id) DO UPDATE SET
             current_units = positions.current_units + EXCLUDED.current_units,
             cost_basis    = positions.cost_basis + EXCLUDED.cost_basis,
             avg_cost      = (positions.cost_basis + EXCLUDED.cost_basis)
                             / NULLIF(positions.current_units + EXCLUDED.current_units, 0),
+            -- Reset source on reopen: if the existing row is fully
+            -- closed (current_units <= 0) this BUY is reopening it
+            -- under eBull, so source flips to 'ebull'. Otherwise
+            -- preserve the existing source — an eBull ADD into an
+            -- already-open broker_sync position shouldn't claim
+            -- ownership of the original external open.
+            source        = CASE
+                WHEN positions.current_units <= 0
+                    THEN EXCLUDED.source
+                ELSE positions.source
+            END,
             updated_at    = EXCLUDED.updated_at
         """,
         {
--- a/app/services/portfolio_sync.py
+++ b/app/services/portfolio_sync.py
@@ -179,16 +179,28 @@ def sync_portfolio(
                 """
                 INSERT INTO positions
                     (instrument_id, open_date, avg_cost, current_units,
-                     cost_basis, unrealized_pnl, updated_at)
+                     cost_basis, unrealized_pnl, source, updated_at)
                 VALUES
                     (%(iid)s, %(date)s, %(price)s, %(units)s,
-                     %(cost)s, %(upnl)s, %(now)s)
+                     %(cost)s, %(upnl)s, 'broker_sync', %(now)s)
                 ON CONFLICT (instrument_id) DO UPDATE SET
                     current_units  = EXCLUDED.current_units,
                     avg_cost       = EXCLUDED.avg_cost,
                     cost_basis     = EXCLUDED.cost_basis,
                     unrealized_pnl = EXCLUDED.unrealized_pnl,
                     open_date      = EXCLUDED.open_date,
+                    -- Reset source on reopen: if the existing row is
+                    -- fully closed (current_units <= 0) this conflict
+                    -- path is reopening it externally, so the new
+                    -- opener ('broker_sync') becomes the source.
+                    -- Otherwise preserve the existing source (adds to
+                    -- an already-open position shouldn't flip
+                    -- ownership).
+                    source         = CASE
+                        WHEN positions.current_units <= 0
+                            THEN EXCLUDED.source
+                        ELSE positions.source
+                    END,
                     updated_at     = EXCLUDED.updated_at
                 """,
                 {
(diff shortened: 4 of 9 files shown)


Luke-Bradford:
## Review response

### BLOCKING — REBUTTED

> The `ON CONFLICT` branch ... overwrites `current_units`, `avg_cost`, `cost_basis`, `unrealized_pnl`, and `open_date` with `EXCLUDED.*` ... but the `source` reset CASE WHEN uses `positions.current_units <= 0` to detect a reopen — evaluated **after** `current_units` has already been clobbered

This is empirically incorrect. Postgres `ON CONFLICT DO UPDATE` evaluates every SET expression against the **pre-update** row snapshot, not sequentially as a Python dict assignment would. `positions.current_units` in any SET expression refers to the value of the conflicting row **before** the DO UPDATE applies, regardless of where `current_units = EXCLUDED.current_units` appears in the SET list.

This is documented behaviour — from the [Postgres INSERT docs](https://www.postgresql.org/docs/current/sql-insert.html) on ON CONFLICT:

> Note that the effects of all per-row BEFORE INSERT triggers are reflected in excluded values, since those effects may have contributed to the row being excluded from insertion.
>
> [...] In the ON CONFLICT DO UPDATE clause, the new row is referred to with the alias excluded, and the existing row with the alias target_table (which is the unqualified table name used in the INSERT statement).

And from the [SET clause section](https://www.postgresql.org/docs/current/sql-update.html#SQL-UPDATE-CONFLICT):

> The SET clause expressions are computed based on the value of the existing row, using column names as references

Empirical verification against the running dev DB:

```python
# Seed: closed row (current_units=0, source='broker_sync')
INSERT INTO upsert_test VALUES (1, 0, 'broker_sync')

# Upsert: would reopen with source='ebull'
INSERT INTO upsert_test (pk, current_units, source)
VALUES (1, 5, 'ebull')
ON CONFLICT (pk) DO UPDATE SET
    current_units = EXCLUDED.current_units,
    source = CASE
        WHEN upsert_test.current_units <= 0
            THEN EXCLUDED.source
        ELSE upsert_test.source
    END
```

Result:

```
After reopen of closed row: [(1, Decimal('5'), 'ebull')]        ← reset worked
After ADD to open row:      [(2, Decimal('15'), 'broker_sync')] ← preserve worked
```

The CASE WHEN sees `current_units = 0` (the pre-update value) on the reopen path and correctly picks `EXCLUDED.source = 'ebull'`. On the ADD path it sees `current_units = 10` (pre-update) and correctly picks `positions.source = 'broker_sync'`. The SET order is irrelevant because all right-hand expressions read from the same pre-update tuple.

**Resolution:** REBUTTED — Postgres semantics guarantee the CASE WHEN sees the pre-update row.

### WARNING — REBUTTED

> Both test classes ... inspect the identical SQL string; neither test would catch a regression where the CASE WHEN block was removed from the conflict clause while the INSERT literal remained

Each test asserts a distinct substring of the captured SQL, and they run in separate test methods with separate connection mocks. If `CASE WHEN ... ELSE positions.source END` were deleted while the `'ebull'` literal in VALUES remained:

- [test_insert_sets_source_to_ebull](tests/test_order_client.py#L690) asserts `"'ebull'"` is in the SQL — would still PASS (literal is still there, correctly)
- [test_insert_conflict_clause_resets_source_on_reopen](tests/test_order_client.py#L709) asserts `"positions.current_units <= 0"`, `"EXCLUDED.source"`, and `"ELSE positions.source"` — would **FAIL** on all three

So the specific regression the reviewer describes *is* caught — by the second test. The two tests have non-overlapping assertions precisely so each targets a different part of the SQL shape. They're not redundant; they're orthogonal.

**Resolution:** REBUTTED — the two tests have disjoint assertions that independently catch regressions in the VALUES literal and the ON CONFLICT CASE WHEN.

### PREVENTION — REBUTTED

> Add a grep rule — `SET.*current_units\s*=\s*EXCLUDED` in any `ON CONFLICT DO UPDATE` block that also references `positions.current_units` in a CASE WHEN — to flag write-before-read hazards in upsert logic.

The proposed lint rule is based on the same false premise as the BLOCKING finding. There is no write-before-read hazard in Postgres `ON CONFLICT DO UPDATE` — SET expressions are not sequential assignments; they all read from the pre-update row snapshot. Adding this rule would flag correct code as broken. The real prevention here (if any) is to document the Postgres ON CONFLICT semantics so future reviewers don't raise this concern again.

**Resolution:** REBUTTED — the pattern the rule targets is legitimate, correctly-evaluated SQL, not a hazard.

Luke-Bradford:
## Review response (second review — APPROVE)

### WARNING — DEFERRED #186

> `test_insert_sets_source_to_broker_sync` and `test_insert_conflict_clause_resets_source_on_reopen` both call `sync_portfolio` once with an empty `local_positions` list, which always exercises only the INSERT path; neither test ever exercises the ON CONFLICT path ... so the CASE WHEN assertions in the second test pass solely because the CASE WHEN text appears anywhere in the INSERT statement.

Acknowledged and accepted as a narrower, more concrete framing of the test-coverage tradeoff already noted in the PR description. The current tests verify SQL shape (that the CASE WHEN clause is emitted with the right predicate and branches), which catches regressions where the clause is deleted or the predicate/arms are edited. What they cannot catch is a Postgres-level semantic failure, because a mocked connection never runs the planner.

The complete fix — driving an actual conflict against a real schema and reading back the resolved `source` — is a DB-level integration test against a live Postgres. Opened as tech-debt issue #186: "Tech debt: DB-level integration tests for positions upsert reset-on-reopen semantics". Deferring per the feedback_pr_review_efficiency rule (reactive scope additions to an approved PR cause churn without blocking correctness).

**Resolution:** DEFERRED #186

### PREVENTION — EXTRACTED docs/review-prevention-log.md

> **SQL-shape tests on single-path calls**: flag test methods that assert `ON CONFLICT` clause content but whose mock setup (`local_positions=[]`) guarantees Postgres never reaches the conflict branch.

Added as a new entry to [docs/review-prevention-log.md](docs/review-prevention-log.md) under the heading "SQL-shape tests on single-path calls can't exercise the ON CONFLICT branch". The entry captures the bug class, the specific PR #185 symptom, and the prevention rule (either drive an actual conflict in the mock or write a DB-level integration test — never rely on substring-in-SQL as a proxy for "the conflict branch works").

Commit: 498c346

**Resolution:** EXTRACTED docs/review-prevention-log.md

Luke-Bradford:
## Review response (third review — REQUEST CHANGES)

### WARNING — FIXED d901db2

> `test_insert_sets_source_to_broker_sync` and `test_insert_conflict_clause_resets_source_on_reopen` are identical in setup and call ... both assertions target the same string and the second test adds no independent coverage over the first; the prevention log entry added in this very push documents exactly this failure mode, but the tests themselves were not fixed.

Fair call — the prevention log without the fix is half-work. Merged the two tests per file into a single test method that captures SQL once and asserts both the VALUES literal and the CASE WHEN shape from the same call:

- [tests/test_portfolio_sync.py](tests/test_portfolio_sync.py#L546) — `test_insert_emits_source_literal_and_reopen_reset_clause`
- [tests/test_order_client.py](tests/test_order_client.py#L690) — `test_insert_emits_source_literal_and_reopen_reset_clause`

The merged tests also explicitly document in the docstring that this is a SQL-shape unit test, and that runtime branch verification (Postgres actually taking the ON CONFLICT arm on a closed row) is tracked in #186. No more two-tests-pretending-to-be-independent.

**Resolution:** FIXED d901db2

### PREVENTION — ALREADY_COVERED docs/review-prevention-log.md

Already extracted in 498c346 — the new entry "SQL-shape tests on single-path calls can't exercise the ON CONFLICT branch" in [docs/review-prevention-log.md](docs/review-prevention-log.md) captures exactly the bug class flagged here. The review's PREVENTION wording is a restatement of the same rule in different words; the log entry is the canonical form.

**Resolution:** ALREADY_COVERED docs/review-prevention-log.md

Luke-Bradford:
## Review response (fourth review — APPROVE)

### WARNING — REBUTTED

> Merging the two previously-separate test methods into one means a failure in the VALUES-literal assertion ... will prevent the CASE WHEN assertions from running.

The intra-test short-circuit the reviewer describes is the standard pytest assertion behavior — any test method with multiple `assert` statements has it. Splitting these back into two independent tests was exactly the state that the **previous** review (at commit 498c346) correctly rejected as two tests pretending to be independent when they assert on the same captured SQL from the same production call. The honest shape under mocks is one test that captures once and asserts both concerns — split-for-independence would reintroduce the duplication this review cycle was meant to fix.

If the VALUES literal is ever broken, the test fails loudly at the first assert; the CASE WHEN assertion is not checked, but since every run of this test exercises a single captured SQL string, any regression in either clause will still fail the build on the next run. The "suppression" is only within a single already-failing run.

**Resolution:** REBUTTED — split-for-independence recreates the duplicated-coverage problem that the previous review correctly flagged.

### PREVENTION — REBUTTED

Same reasoning: a "split to avoid one failing assertion masking another" rule applied to every pytest test method would require every test to contain exactly one assert, which is not a repo convention and would make existing multi-assert tests non-compliant. The actual prevention lesson from this PR cycle ("SQL-shape tests on single-path calls can't exercise the ON CONFLICT branch") is already captured in [docs/review-prevention-log.md](docs/review-prevention-log.md).

**Resolution:** REBUTTED — would conflict with existing multi-assert test conventions; the real lesson is already in the prevention log.

Issue #189, opened 2026-04-11
feat: Track 2 — trader discovery, historical snapshots, cohort signals


Luke-Bradford:
## Summary

Track 2 turns the `user-info/people/*` surface on the eToro API into
periodic trader discovery, historical performance snapshots, and cohort
signals. It depends on the tables and types delivered by #183 (Track 1a),
but adds no schema churn to them — Track 2 creates sibling tables and
imports from `copy_traders` for identity.

Depends on: #183 (Track 1a — must merge first; defines `copy_traders`
schema). **Does not** depend on #187 (Track 1b) — Track 2 is purely
additive on top of the ingestion tables and does not use the AUM
correction helper.

### Endpoint surface (already documented in `docs/etoro-api-reference.md`)

- `GET /api/v1/user-info/people/search` — discovery with filters for
  `popularInvestor`, `period`, `gainMax`, `maxDailyRiskScoreMin/Max`,
  `maxMonthlyRiskScoreMin/Max`, `weeksSinceRegistrationMin`,
  `countryId`, `instrumentId`, `instrumentPctMin/Max`
- `GET /api/v1/user-info/people/{username}/portfolio/live` — live
  portfolio of any trader by username (positions + socialTrades)
- `GET /api/v1/user-info/people/{username}/tradeinfo` — per-period
  stats: `gain`, `dailyGain`, `riskScore`, `copiers`, `copiersGain`
- `GET /api/v1/user-info/people/{username}/gain` — monthly + yearly
  historical gain series
- `GET /api/v1/user-info/people/{username}/daily-gain` — daily gain
  series over a date range
- `GET /api/v1/user-info/people` — batch profile lookup by usernames
  or CID list

### Features Track 2 enables

1. **Discovery sweep** — a periodic job populates `copy_traders` rows
   for traders we are **not** currently copying. Today they are only
   created when we see a mirror for them; Track 2 surfaces them
   pre-copy so the ranking engine can score them.
2. **Historical performance snapshots** — a new sibling table
   (schema TBD in Track 2 spec) storing periodic gain/risk series
   joined against `copy_traders`.
3. **Cohort signals for trader ranking** — feeds the ranking engine
   with trader-quality metrics (risk score, consistency, copier
   cohort size, copier cohort gain).

### Schema impact

Track 2 **does not** rewrite `copy_traders`, `copy_mirrors`, or
`copy_mirror_positions`. Those tables are stable after #183 (Track 1a).
Track 2 may add:

[...]

Issue #188, opened 2026-04-11
feat: Track 1.5 — copy-trading browsing UX (REST endpoint + frontend panel)


Luke-Bradford:
## Summary

Track 1.5 ships the copy-trading **browsing UX** on top of the tables and
AUM correction delivered by #183. It is a pure read surface — no schema
changes, no new sync logic.

Depends on: #183 (Track 1a — schema + sync, must merge first) and #187 (Track 1b — `_load_mirror_equity` helper + AUM correction at the REST endpoint, must merge before Track 1.5 can reuse the helper)

### REST endpoint

`GET /api/portfolio/copy-trading` — returns a list of copy traders, each
with:

- `parent_cid`, `parent_username`
- Mirror-level aggregates: `active`, `initial_investment`,
  `deposit_summary`, `withdrawal_summary`, `available_amount`,
  `closed_positions_net_profit`, `mirror_equity` (computed from the
  §3.4 query in the #183 design spec, reused via
  `_load_mirror_equity` from `app/services/portfolio.py` — the
  helper that ships in #187)
- Nested-position summary: count, top N by cost basis, per-instrument
  breakdown

Endpoint is session-auth-only (same gate as `GET /api/portfolio`).
Response shape is a pydantic `BaseModel` with an explicit shape so the
frontend can type against it (per
`.claude/skills/frontend/api-shape-and-types.md`).

### Frontend panel

React dashboard panel consuming the endpoint:

- **Top-line summary**: `AUM = positions + cash + mirror_equity` —
  the breakdown users expected from the `mirror_equity` field on
  `PortfolioResponse` that ships in #187
- **Per-trader cards** with mirror-level numbers (initial investment,
  net funded, equity, P/L)
- **Drill-down** showing nested positions, cost basis, entry FX,
  MTM delta
- **Closed mirrors** visible in a separate "history" section (filter
  on `active = FALSE`, reads the retained soft-closed rows)

Must follow:
- `.claude/skills/frontend/async-data-loading.md`
- `.claude/skills/frontend/loading-error-empty-states.md`
- `.claude/skills/frontend/safety-state-ui.md`
- `.claude/skills/frontend/api-shape-and-types.md`
- `.claude/skills/frontend/operator-ui-conventions.md`

### Data-contract tests

Frontend tests assert the response shape matches the pydantic BaseModel
field-by-field, so a backend rename breaks CI rather than silently
breaking the UI.

## Explicitly NOT in scope

[...]

Issue #187, opened 2026-04-11
feat: Track 1b — mirror-equity AUM correction across call sites


Luke-Bradford:
## Summary

Track 1b delivers the **AUM correction** that downstream systems need once copy-trading mirrors are landing in the DB via #183 (Track 1a). Today `total_aum = positions + cash`; after Track 1b it becomes `total_aum = positions + cash + mirror_equity` at every call site that computes it.

**Depends on: #183 (Track 1a — schema + sync + parser)** — must merge first. Track 1b can branch off main immediately after #183 merges.

### New module-level helper

`_load_mirror_equity(conn: psycopg.Connection[Any]) -> float` in [app/services/portfolio.py](app/services/portfolio.py), declared at module level alongside the existing `_load_cash` / `_load_positions` helpers. Executes the §3.4 LATERAL-join query across `copy_mirrors` (filtered on `active`), `copy_mirror_positions`, and `quotes` to compute the per-mirror equity and sum it. Returns `0.0` when `copy_mirrors` is empty. The docstring says "usually non-negative" — the formula is NOT mathematically floored at zero.

See spec §6.0 and §3.4 for the exact SQL and return-type contract.

### Three call sites wired to the new helper

1. **Execution guard sector-exposure** — [app/services/execution_guard.py](app/services/execution_guard.py):`_load_sector_exposure` (line 229-289). A one-line modification at line 286 where the existing `total_aum = total_positions + cash` becomes `total_aum = total_positions + cash + _load_mirror_equity(conn)`. The public `evaluate_recommendation` entry point and the `GuardResult` return type are **not** touched — the change is surgical, inside the private helper. See spec §6.1.

2. **Portfolio REST endpoint** — [app/api/portfolio.py](app/api/portfolio.py):`get_portfolio`. Add a `mirror_equity: float` field to `PortfolioResponse` and wire it from `_load_mirror_equity(conn)`. `total_aum` now sums positions + cash + mirror_equity. See spec §6.2.

3. **Portfolio review** — [app/services/portfolio.py](app/services/portfolio.py):`run_portfolio_review` (AUM block at lines 751-753). The existing `total_aum = total_market_value + (cash if cash_known else 0.0)` becomes `total_aum = total_market_value + (cash if cash_known else 0.0) + _load_mirror_equity(conn)`. The `PortfolioReviewResult` dataclass already has `total_aum`; no new field is needed. See spec §6.3.

### Tests

[...]

Issue #183, opened 2026-04-11
feat: Track 1a — copy-trading ingestion (schema + sync + parser)


Luke-Bradford:
## Summary

Track 1a ingests copy-trading **mirrors** from the eToro portfolio API into three new tables. Today `etoro_broker.get_portfolio` discards `clientPortfolio.mirrors[]` entirely; Track 1a parses it, upserts it, and makes the data available for Track 1b (AUM correction), Track 1.5 (browsing UX), and Track 2 (discovery).

Track 1a is strictly the **ingestion** half — no downstream consumer reads the new tables in this PR. The AUM correction that wires mirrors into `total_aum` at every call site is Track 1b (filed separately as a dependent issue).

## Observed

eToro demo account shows 2 copy traders (thomaspj, triangulacapital) with ~£14.8k and ~£14.5k invested respectively. These are absent from eBull entirely because `etoro_broker.get_portfolio` reads only `clientPortfolio.positions` and ignores `clientPortfolio.mirrors`.

### New schema (single migration)

- `copy_traders` — one row per parent trader (keyed by `parent_cid`), FK target for `copy_mirrors`
- `copy_mirrors` — one row per copy relationship (keyed by `mirror_id`), with `active`/`closed_at` soft-close semantics, `raw_payload JSONB NOT NULL`, and mirror-level aggregates (`initial_investment`, `available_amount`, `deposit_summary`, `withdrawal_summary`, `closed_positions_net_profit`, `started_copy_date`). Partial index on `(active) WHERE active`.
- `copy_mirror_positions` — one row per nested position inside a mirror, composite PK `(mirror_id, position_id)`, `ON DELETE CASCADE` from `copy_mirrors`, `raw_payload JSONB NOT NULL` per row, `open_conversion_rate NUMERIC(20,10) NOT NULL` for FX correctness.

See spec §1 for the full DDL.

### Broker-interface additions

- Two new frozen dataclasses on `app/providers/broker.py`: `BrokerMirror` and `BrokerMirrorPosition`
- Additive field `mirrors: Sequence[BrokerMirror]` on `BrokerPortfolio`

### Parser rewrite with nested error handling

`app/providers/implementations/etoro_broker.py` parses `clientPortfolio.mirrors[]` with:
- Outer top-level wrap catching `(KeyError, ValueError, TypeError, decimal.DecimalException)` with per-mirror attribution
[...]

PR #191, opened 2026-04-11
feat(#183): ingest eToro copy-trading mirrors (Track 1a)

Luke-Bradford:
## Summary

Track 1a of the copy-trading ingestion design (#183). Adds first-class ingestion of the `clientPortfolio.mirrors[]` payload returned by eToro's `/portfolio` endpoint into three new sibling tables, with strict-raise parsing, per-mirror upsert, nested-position eviction, and soft-close / total-disappearance handling.

This is **Track 1a only**. The follow-up tracks remain to be delivered:
- **Track 1b (#187)** — AUM correction helper + 3 call sites (§3.4 / §6.0–6.3 / §8.4–8.6)
- **Track 1.5 (#188)** — Operator browsing UX for mirrors
- **Track 2 (#189)** — `/user-info/people/*` discovery

Spec: [docs/superpowers/specs/2026-04-11-copy-trading-ingestion-design.md](docs/superpowers/specs/2026-04-11-copy-trading-ingestion-design.md)
Plan: [docs/superpowers/plans/2026-04-11-copy-trading-ingestion-track-1a.md](docs/superpowers/plans/2026-04-11-copy-trading-ingestion-track-1a.md)

## Schema (migration 022)

Three new sibling tables; existing tables untouched.

- `copy_traders(parent_cid PK, parent_username, first_seen_at, updated_at)` — one row per eToro trader identity.
- `copy_mirrors(mirror_id PK, parent_cid FK, initial_investment/deposit/withdrawal/available_amount/closed_positions_net_profit, stop_loss_*, mirror_status_id, mirror_calculation_type, pending_for_closure, started_copy_date, active BOOL, closed_at NULL, raw_payload JSONB, updated_at)` — one row per copy relationship. Partial index on `active` for fast scans.
- `copy_mirror_positions((mirror_id, position_id) PK, parent_position_id, instrument_id, is_buy, units, amount, initial_amount_in_dollars, open_rate, open_conversion_rate, open_date_time, take_profit_rate, stop_loss_rate, total_fees, leverage, raw_payload JSONB, updated_at)` — one row per nested position. `ON DELETE CASCADE` from `copy_mirrors.mirror_id`.

`updated_at DEFAULT NOW()` only fires on INSERT; all `ON CONFLICT DO UPDATE` paths in `_sync_mirrors` explicitly refresh `updated_at = %(now)s`. This contract is documented in the migration comment block and in the Python helper.

## Parser contract

- `PortfolioParseError(Exception)` — direct `Exception` subclass (not `ValueError`/`TypeError`/`KeyError`/`DecimalException`), so the outer catch list can't accidentally swallow it.
- `_parse_mirror_position` — normalises one nested position, raises on missing `openConversionRate` or non-numeric fields.
- `_parse_mirror` — builds a `BrokerMirror`, wraps any nested failure in `PortfolioParseError(f"Mirror {mid!r} position[{idx}]: {exc}") from exc` for debuggable attribution.
- `_parse_mirrors_payload` — outer loop. Log-and-skip only for rows with no `mirrorID`; **any** known-mirror failure (nested or top-level) raises, so a sync can never silently reinterpret a corrupted payload as "mirror disappeared". This closes the Codex v3 Finding V parse-and-soft-close hole.

## Sync contract (`_sync_mirrors`)

Called from `sync_portfolio` after the existing position/cash sync, inside the same transaction.

1. **Upsert `copy_traders`** (`ON CONFLICT (parent_cid) DO UPDATE`).
2. **Upsert `copy_mirrors`** (`ON CONFLICT (mirror_id) DO UPDATE`). The update clause resets `active=TRUE, closed_at=NULL` — **this is the re-copy resurrection path** for a previously-soft-closed mirror_id eToro has started copying again.
3a. **Evict nested positions** that have closed since the last sync: `DELETE ... WHERE mirror_id = ? AND position_id <> ALL(?::bigint[])`. Exploits Postgres's `<> ALL('{}')` === TRUE semantics so the empty-array case correctly deletes every row without a special branch.
3b. **Upsert every current position** (`ON CONFLICT (mirror_id, position_id) DO UPDATE`).
4. **Disappearance handling (§2.3.4)** — after the per-mirror loop:
   - **Total disappearance** (`mirrors[]` empty AND local active rows exist) → `RuntimeError` to force operator investigation. No mass soft-close.
   - **Partial disappearance** → `UPDATE copy_mirrors SET active=FALSE, closed_at=%(now)s WHERE mirror_id = ANY(%(disappeared)s)`. Nested positions are **retained** for audit; the §3.4 AUM query's `WHERE m.active` filter excludes them from forward-looking calculations.

`PortfolioSyncResult` gains three counters: `mirrors_upserted`, `mirrors_closed`, `mirror_positions_upserted`, surfaced in the scheduler's daily-sync INFO line so operators see mirror activity in `job_runs` without querying the DB.

## Security model

- **Read-only from broker, write-only to local DB.** No new outgoing network call, no new order-placing path, no new un-copy UX.
- No new cross-user data flow. Data only flows from eToro `/portfolio` → local Postgres.
- The execution guard's rule queries are unchanged — they still read `FROM positions` only. Mirror-based AUM correction is Track 1b's responsibility.
- The parser's strict-raise discipline means a malformed payload triggers a transaction rollback before any row is written, so a partial ingestion can never leave the DB in an inconsistent state.

## Tests

- [tests/test_copy_mirrors_parser.py](tests/test_copy_mirrors_parser.py) — 16 parser tests covering the `PortfolioParseError` hierarchy, `_parse_mirror_position` happy path + required-field failures, `_parse_mirror` nested wrap with position-index attribution, and `_parse_mirrors_payload` outer loop (happy path, log-and-skip unrecognisable, known-mirror strict-raise, nested failure propagation).
- [tests/test_portfolio_sync_mirrors.py](tests/test_portfolio_sync_mirrors.py) — 11 integration tests against `ebull_test` covering fresh insert, idempotent resync, nested position eviction (partial + total), partial-disappearance soft-close, re-copy resurrection, total-disappearance raise, parser-failure rollback (nested), known-mirror top-level parse failure (Codex v3 Finding V regression).
- [tests/fixtures/copy_mirrors.py](tests/fixtures/copy_mirrors.py) — new fixtures package owning `_NOW`, `_GUARD_INSTRUMENT_ID`, and the `two_mirror_payload` / `parse_failure_payload` / `two_mirror_seed_rows` builders. The existing [tests/test_portfolio_sync.py](tests/test_portfolio_sync.py) migrates to import `_NOW` from this module (no behaviour change).

Full suite: **1134 passed, 1 skipped**. Pre-push gate (ruff check, ruff format --check, pyright, pytest) green.

## Tradeoffs consciously accepted

- **Nested positions retained on soft-closed mirrors.** Keeps audit history at the cost of some dead rows. The §3.4 AUM query's `active` filter is the forward-looking gate.
- **Total disappearance raises rather than auto-closes.** Asymmetric with partial disappearance, deliberately. An empty `mirrors[]` is much more likely to be an upstream API regression than a real event.
- **Re-copy reuses the existing `ON CONFLICT DO UPDATE` clause** rather than a separate resurrection path. The update clause always resets `active=TRUE, closed_at=NULL`, which is correct whether or not the row was previously closed.
- **Fresh fixtures module rather than hanging helpers off the existing test file.** Shared fixtures across parser and sync tests would be awkward if co-located with one test module. [tests/fixtures/copy_mirrors.py](tests/fixtures/copy_mirrors.py) is the deliberate home.

## Test plan

- [ ] Pre-push gate green (ruff check, ruff format --check, pyright, pytest) — confirmed locally before push
- [ ] Smoke test `tests/smoke/test_app_boots.py` still passes end-to-end — included in full pytest run
- [ ] Claude review passes on most recent commit
- [ ] CI green on most recent commit

### DIFF
--- a/.claude/skills/engineering/pre-flight-review.md
+++ b/.claude/skills/engineering/pre-flight-review.md
@@ -83,6 +83,7 @@ Ask:
 - is the upsert key correct?
 - is the write atomic where it needs to be?
 - could two writers race and produce inconsistent versions?
+- **transaction rollback scope:** for every new `raise` inside a function called from a shared transaction, ask: *what other writes have already executed in this transaction when I raise?* If the rollback would discard writes that are not the helper's concern, hoist the raise to the caller **before** those writes run. Document the split scope in the helper's docstring.
 
 ### G. Interface cleanliness
 Check:
--- a/app/providers/broker.py
+++ b/app/providers/broker.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
 from dataclasses import dataclass
+from datetime import datetime
 from decimal import Decimal
 from typing import Any, Literal
 
@@ -44,12 +45,63 @@ class BrokerPosition:
 
 
 @dataclass(frozen=True)
+class BrokerMirrorPosition:
+    """A single nested position inside a copy-trader mirror.
+
+    `amount` is the pre-converted USD cost basis reported by eToro.
+    `open_rate` is the entry price in the instrument's native
+    currency; `open_conversion_rate` is the native→USD FX rate at
+    open. Both are required — see spec §1.3 "openConversionRate NOT
+    NULL" for the AUM correctness reason.
+    """
+
+    position_id: int
+    parent_position_id: int
+    instrument_id: int
+    is_buy: bool
+    units: Decimal
+    amount: Decimal
+    initial_amount_in_dollars: Decimal
+    open_rate: Decimal
+    open_conversion_rate: Decimal
+    open_date_time: datetime
+    take_profit_rate: Decimal | None
+    stop_loss_rate: Decimal | None
+    total_fees: Decimal
+    leverage: int
+    raw_payload: dict[str, Any]
+
+
+@dataclass(frozen=True)
+class BrokerMirror:
+    """A single copy-trading mirror (one per copy session with a trader)."""
+
+    mirror_id: int
+    parent_cid: int
+    parent_username: str
+    initial_investment: Decimal
+    deposit_summary: Decimal
+    withdrawal_summary: Decimal
+    available_amount: Decimal
+    closed_positions_net_profit: Decimal
+    stop_loss_percentage: Decimal | None
+    stop_loss_amount: Decimal | None
+    mirror_status_id: int | None
+    mirror_calculation_type: int | None
+    pending_for_closure: bool
+    started_copy_date: datetime
+    positions: Sequence[BrokerMirrorPosition]
+    raw_payload: dict[str, Any]
+
+
+@dataclass(frozen=True)
 class BrokerPortfolio:
-    """Snapshot of the broker account: positions + available cash."""
+    """Snapshot of the broker account: positions + available cash + mirrors."""
 
     positions: Sequence[BrokerPosition]
     available_cash: Decimal
     raw_payload: dict[str, Any]
+    mirrors: Sequence[BrokerMirror] = ()
 
 
 class BrokerProvider(ABC):
--- a/app/providers/implementations/etoro_broker.py
+++ b/app/providers/implementations/etoro_broker.py
@@ -77,6 +77,20 @@ _STATUS_MAP: dict[str, OrderStatus] = {
 }
 
 
+class PortfolioParseError(Exception):
+    """Raised when a mirrors[] row cannot be parsed safely.
+
+    Directly subclasses Exception (NOT ValueError / TypeError /
+    KeyError / decimal.DecimalException) so the outer parse loop can
+    distinguish it from incidental exceptions and re-raise. Never
+    swallowed by any `except (KeyError, ValueError, TypeError,
+    decimal.DecimalException)` block.
+
+    See spec §2.2.1 for the hierarchy rationale and §2.3.3 for the
+    strict-raise sync contract that depends on it.
+    """
+
+
 class EtoroBrokerProvider(BrokerProvider):
     """
     eToro trading API client.
--- a/app/providers/implementations/etoro_broker.py
+++ b/app/providers/implementations/etoro_broker.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 import decimal
 import logging
+from collections.abc import Sequence
 from datetime import UTC, datetime
 from decimal import Decimal
 from pathlib import Path
@@ -439,6 +440,7 @@ class EtoroBrokerProvider(BrokerProvider):
         portfolio = raw.get("clientPortfolio") or {}
         raw_positions: list[dict[str, Any]] = portfolio.get("positions") or []
         credit = portfolio.get("credit")
+        raw_mirrors: list[Any] = portfolio.get("mirrors") or []
 
         positions: list[BrokerPosition] = []
         for pos in raw_positions:
@@ -474,6 +476,7 @@ class EtoroBrokerProvider(BrokerProvider):
             positions=positions,
             available_cash=Decimal(str(credit)) if credit is not None else Decimal("0"),
             raw_payload=raw,
+            mirrors=tuple(_parse_mirrors_payload(raw_mirrors)),
         )
 
     # ------------------------------------------------------------------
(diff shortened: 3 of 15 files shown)


Luke-Bradford:
Resolving review findings on PR #191.

## BLOCKING — `tests/test_portfolio_sync.py` indent/orphan

**FIXED 2729946**

Diagnosis: `test_portfolio_sync_result_has_mirror_counters` was placed at module scope between two methods of `TestPositionSource`. Python's class-body rule (first statement at outer indent closes the class) meant the class ended at that line, and `test_update_path_does_not_overwrite_source` below it was silently reparsed as a **nested function inside the new top-level test** — never collected by pytest. No `SyntaxError` fires for this; `py_compile` passes; the only signal was an off-by-one test count I misread during implementation.

Fix: moved `test_portfolio_sync_result_has_mirror_counters` to end-of-file at proper module scope. Verified via AST walk that `TestPositionSource` now has both its methods and no nested orphans. Targeted pytest collects 29 tests (was 28 pre-fix on this file).

## WARNING #1 — transaction rollback scope in `_sync_mirrors`

**FIXED 2729946**

You're right — `_sync_mirrors` raising `RuntimeError` inside `sync_portfolio`'s transaction *after* position and cash writes had already executed would discard legitimate position/cash work alongside the mirror refusal. That's a broader blast radius than spec §2.3.4's "operator investigates mirrors" framing implies.

Fix: hoisted the total-disappearance guard to the **top of `sync_portfolio`, before any writes run**. Nothing is written yet at that point, so the raise has no side effects to roll back. Removed the now-redundant inner check from `_sync_mirrors`. Updated the helper docstring to document the split:
- **Total disappearance** → caller-side pre-write guard (raises before writes)
- **Partial disappearance** → inside `_sync_mirrors` step 4 (soft-close)

Symmetrical with the existing cash-reconciliation pre-write guard in the same function.

## WARNING #2 — `conn` fixture isolation in `test_portfolio_sync_mirrors.py`

**REBUTTED**

The fixture is function-scoped (no explicit `scope=`, so default `function`), which means `psycopg.connect()` opens a fresh connection per test and closes it at teardown. Even without that, each test begins with `TRUNCATE copy_mirror_positions, copy_mirrors, copy_traders RESTART IDENTITY CASCADE` committed immediately — that's the isolation guarantee, not the `c.rollback()` at yield-exit. The rollback at the end is a safety net for the in-flight transaction of the *current* test; residual committed rows from a prior test can never reach the next test because the next test's TRUNCATE wipes them at setup.

The scenario you describe — "a test that calls `conn.commit()` mid-test and then fails later leaves committed rows that the `c.rollback()` at yield-exit cannot undo" — is true but harmless: the next test's setup TRUNCATE still runs first and clears them. I considered adding a session-scoped fixture with transaction rollback per test to skip the TRUNCATE, but the function-scope + TRUNCATE pattern is what `tests/test_operator_setup_race.py` already uses and is the established eBull pattern for real-DB tests. Keeping consistency over micro-optimisation.

## PREVENTION — extractions

1. **Module-level `def` inside class body silently orphans adjacent methods**
   **EXTRACTED docs/review-prevention-log.md**
   Entry documents the Python class-body parsing rule, notes that `py_compile` (the fix you suggested) is **not** sufficient because the orphaned file is syntactically valid, and records the AST-walk one-liner that actually catches it:
   ```bash
   uv run python -c "import ast; [print(f'{n.name}: {[m.name for m in n.body if isinstance(m, ast.FunctionDef)]}') for n in ast.parse(open('tests/<file>.py').read()).body if isinstance(n, ast.ClassDef)]"
   ```

2. **Transaction rollback scope of raises inside shared transaction helpers**
   **EXTRACTED `.claude/skills/engineering/pre-flight-review.md` section F** + matching prevention-log entry cross-referencing it. The new checklist item:
   > transaction rollback scope: for every new `raise` inside a function called from a shared transaction, ask: *what other writes have already executed in this transaction when I raise?* If the rollback would discard writes that are not the helper's concern, hoist the raise to the caller **before** those writes run. Document the split scope in the helper's docstring.

## Pre-push gate (commit 2729946)

- `uv run ruff check .` ✅
- `uv run ruff format --check .` ✅
- `uv run pyright` ✅
- `uv run pytest` ✅ 1135 passed, 1 skipped

Luke-Bradford:
Round-2 resolutions (verdict is APPROVE; no new commit).

## [WARNING] `tests/test_portfolio_sync_mirrors.py:60` — fixture isolation

**REBUTTED** (same point as round-1 WARNING #2, re-stated with the `test_sync_mirrors_recopy_resurrects_closed_mirror` example).

Fixture is function-scoped + `psycopg.connect()` opens a fresh connection per test + next-test setup TRUNCATEs before any assertion runs. The "stale rows bleed into next test" scenario cannot occur with this shape — `conn.commit()` mid-test is irrelevant because the next test's first action wipes the three tables. This is the established eBull pattern from `tests/test_operator_setup_race.py`. Per my prior reply: keeping consistency over micro-optimisation.

## [WARNING] `app/services/portfolio_sync.py:395` — guard error context

**REBUTTED**

The concern (missing `copy_mirrors` table raises a table-not-found error inside the pre-write guard before the `RuntimeError` can fire) describes a migration-precondition failure, not a guard responsibility. Migration 022 creates `copy_mirrors`; if it hasn't run, every other line of `_sync_mirrors` below fails the same way. The pre-push gate runs `uv run pytest` which drives `tests/smoke/test_app_boots.py` against a fully-migrated dev DB, and CI runs the same gate. Wrapping the guard in `try/except psycopg.errors.UndefinedTable: raise RuntimeError(...)` would add noise without fixing anything the environment gate doesn't already catch. Migration discipline is the right layer for this, not ad-hoc error remapping.

## [PREVENTION] Mid-test savepoint wrapping

**REBUTTED** — same concern frame as WARNING #1. The function-scope + setup-TRUNCATE pattern makes the fixture-exit rollback's scope irrelevant to subsequent tests, so savepoint wrapping would be pure overhead. No extraction needed; the existing eBull real-DB fixture pattern already provides the isolation the PREVENTION is asking for, via a different mechanism.

## Merge

- Verdict on most-recent commit (2729946): **APPROVE**
- CI on 2729946: **green** (lint pass, review pass)
- Pre-push gate: ruff check ✅, ruff format ✅, pyright ✅, pytest ✅ (1135 passed)

Proceeding to merge.

PR #193, opened 2026-04-11
feat(#187): mirror_equity AUM correction across 3 call sites

### LINKED ISSUE #187: feat: Track 1b — mirror-equity AUM correction across call sites

Luke-Bradford:
## Summary

- Adds `_load_mirror_equity(conn) -> float` helper in `app/services/portfolio.py` running the spec §3.4 CTE (LATERAL-joined quote lookup, `WHERE m.active` filter, `COALESCE(SUM(...), 0)` guarantee).
- Wires the helper into all three AUM call sites:
  - `app/services/execution_guard.py:_load_sector_exposure` (AUM remains a local variable, GuardResult unchanged)
  - `app/api/portfolio.py:get_portfolio` (adds `mirror_equity: float = 0.0` to `PortfolioResponse` and the frontend `PortfolioResponse` type for contract parity)
  - `app/services/portfolio.py:run_portfolio_review` (summed into `total_aum`; hoisted above the early-return guard so the no-work-to-do path still reports an honest AUM figure)
- Adds 3 new fixtures to `tests/fixtures/copy_mirrors.py`: `mirror_aum_fixture` (load-bearing for §8.4/§8.5/§8.6), `no_quote_mirror_fixture` (cost-basis identity), `mtm_delta_mirror_fixture` (FX-aware delta with long/short toggle).
- Adds 4 new test modules totalling 16 tests covering §8.4 (identity), §8.5 (guard), §8.6 (per-call-site delta), plus a dedicated test for the run_portfolio_review early-return path.

## Why

Spec §3 identified that `closed_positions_net_profit` is double-counted when added on top of `available + SUM(amount)`, and that `units * open_rate` is cross-currency nonsense for non-USD mirror positions. Before this PR, mirror equity was entirely absent from eBull's AUM denominator — the guard, the dashboard, and the recommendation engine all saw a figure that excluded a significant chunk of the operator's committed capital. This PR closes that gap at every call site that reads AUM.

## Security model

- No new external inputs. `_load_mirror_equity` takes a psycopg connection and runs a static CTE with no user-controlled identifiers.
- No new authorization surface. `GET /portfolio` remains session-protected via `require_session_or_service_token`; the new `mirror_equity` field is read from the same DB state the caller is already authorized for.
- No new provider calls. The helper reads from `copy_mirrors` / `copy_mirror_positions` / `quotes`, all populated by Track 1a ingestion (#183).

## Conscious tradeoffs

- **`mirror_equity` returns `float`, not `float | None`**: `COALESCE(SUM(...), 0)` guarantees a numeric result. Returning `None` would be a type lie and force the frontend to branch on null. See spec §6.4.
- **No `GuardResult.total_aum` field**: AUM is a local variable inside `evaluate_recommendation`, consumed by the concentration rule and not exposed on the return value. Adding a field for a test to assert against would be theatre.
- **No audit persistence**: no `portfolio_reviews` table exists. If future tickets want a per-review AUM snapshot, they get their own migration.
- **MTM delta uses entry-time FX rate, not current FX**: An approximation — a proper current-FX MTM requires a currency-aware `quotes` table (Track 2 scope). Documented in the helper docstring.
- **Frontend types sync, no new UI**: `PortfolioResponse.mirror_equity` is added to `types.ts` in the same PR per api-shape-and-types rule. The dashboard already renders `total_aum` and picks up the corrected value for free. A dedicated "positions + cash + mirrors" breakdown panel lives in Track 1.5 (#188).
- **`_load_mirror_equity` uses `RuntimeError`, not `assert`**: The CTE's `COALESCE(SUM(...), 0)` guarantees exactly one non-NULL row, but the guard must survive `python -O` — see prevention-log entry "`assert` as a runtime guard in service code" (#109). The raise is `# pragma: no cover` because the driver invariant makes it unreachable; it exists for optimised-build safety, not for a live failure mode.
- **`_load_mirror_equity` is hoisted above `run_portfolio_review`'s early-return guard**: Spec §6.3 says `total_aum = positions + cash + mirror_equity` at every call site. Loading mirror equity only inside the main AUM block would return `total_aum=0.0` on the no-ranked-candidates/no-positions branch even when active mirrors held significant value. Pre-flight review caught this; the hoist moves one cheap SELECT above the guard so the contract holds on every return path.
- **Dead-code None-guard on aggregate SELECTs avoided**: prevention-log #75. The helper does not wrap `fetchone()` in `if row is None: return None` — the CTE's aggregate guarantee makes that branch unreachable.

## Test plan

- [x] `uv run ruff check .`
- [x] `uv run pyright`
- [x] `uv run pytest` — 1152 passed, 1 skipped, including the 4 new test modules and the `tests/smoke/test_app_boots.py` gate
- [x] `pnpm --dir frontend test` — 148 passed
- [x] §8.4: empty copy_mirrors → 0.0, no-quote identity, MTM+FX, short sign, closed mirror, all-closed
- [x] §8.5: empty baseline, active mirror, closed mirror, sector numerator unchanged
- [x] §8.6: API (Test 1), review (Test 2), guard (Test 3) — each with soft-close baseline repeat
- [x] New: run_portfolio_review early-return path with active mirror_equity — asserts `total_aum == mirror_equity` when `all_ids` is empty

Closes #187.


Luke-Bradford:
## Round 1 resolutions (commit 88494db)

### BLOCKING

- **Plan document still has `assert row is not None`** — `FIXED 88494db`. Updated `docs/superpowers/plans/2026-04-11-copy-trading-ingestion-track-1b.md:615-624` so the Task 2 Step 3 code block now shows the `RuntimeError` form matching the shipped code (`335c9d6`). Added a new prevention-log entry **"Plan documents drift from the prevention log"** so this gets grepped for in future pre-flight reviews.

- **`mtm_delta_mirror_fixture` omits `is_tradable`** — `FIXED 88494db`. Added `is_tradable=TRUE` to the `INSERT INTO instruments` at `tests/fixtures/copy_mirrors.py:643`, matching every other fixture in the file. Tests were passing because the local schema has `DEFAULT TRUE`, but consistency matters and schema defaults can change. Added a new prevention-log entry **"`INSERT INTO instruments` fixtures must supply `is_tradable`"**.

- **`mirror_aum_fixture` references non-existent instrument 990001 via `coverage` FK** — `REBUTTED`. This finding appears to be a hallucination: `_GUARD_INSTRUMENT_ID = 990001` (`tests/fixtures/copy_mirrors.py:45`) and the fixture DOES seed it into `instruments` at line 447-451 of the same file. There is **no `coverage` INSERT anywhere** in this branch — the `_seed_review_preconditions` helper that previously inserted into `coverage` was deleted in commit `513663e`. The only remaining `coverage` reference is the `TRUNCATE` in `tests/test_portfolio_review_mirror_equity.py:39`. Empirical proof: all 17 mirror-equity tests pass against the real `ebull_test` DB (I just ran them). An FK violation on `coverage(instrument_id) → instruments(instrument_id)` would hard-fail the INSERT; the fact that the tests are green means no such INSERT is executing.

- **`test_sector_numerator_unchanged_by_mirror` wrong `aum_hc` expected value** — `REBUTTED`. The reviewer claimed `total_aum` should be `250 + 100 + 1550 = 1900` because `_load_sector_exposure` "does not subtract the queried instrument's own position from `total_aum`." This missed the WHERE clause at `app/services/execution_guard.py:269`: the positions SELECT filters `WHERE p.current_units > 0 AND p.instrument_id != %(iid)s`. The queried instrument is excluded from `total_positions` at SQL level, not post-sum. So when the test has only one eBull position at `770001` and queries `_load_sector_exposure(conn, 770001)`, `total_positions = 0` (zero rows match the WHERE), and `total_aum = 0 + 100 + 1550 = 1650`. Test passes empirically against the real DB.

### WARNING

- **Comment block on `_load_prev_scores` rationale is load-bearing** — `REBUTTED`. The comment I added (`tests/test_portfolio.py:1017-1019`) replaced a dead cursor stub labelled "_load_prev_scores for held instrument 10 (it IS ranked)" which was factually wrong — held_ranked_ids is empty in that test so _load_prev_scores early-returns without opening a cursor. The new comment explains the skip condition explicitly. The mock-sequence fragility the reviewer flagged (positional `side_effect` across methods) is inherent to the entire `TestRunPortfolioReview` class and pre-dates this PR. Fixing it would require a major `_make_conn` refactor that's out of scope here. The comment is a net improvement, not a regression.

- **`_load_mirror_equity` called outside an explicit transaction boundary** — `REBUTTED`. Every other `_load_*` helper in `app/services/portfolio.py` (`_load_positions`, `_load_cash`, `_load_ranked_scores`, `_load_instrument_details`, `_load_prev_scores`, `_load_prior_recommendations`) runs without wrapping the read in an explicit transaction. The helper's docstring explicitly says "This helper does NOT open its own transaction; it reads under the caller's scope, matching `_load_cash` / `_load_positions`" — which is the correct pattern per prevention #286 (helper should not own its own transaction inside a caller's txn). If the caller's connection is in an error state, every subsequent read breaks, not just this one — that's the caller's responsibility, not the helper's.

### PREVENTION

- **Plan-code drift grep** — `EXTRACTED docs/review-prevention-log.md`. See new entry "Plan documents drift from the prevention log."
- **Fixture NOT NULL grep (`is_tradable`)** — `EXTRACTED docs/review-prevention-log.md`. See new entry "`INSERT INTO instruments` fixtures must supply `is_tradable`."

### Gate state

- `uv run ruff check .` — clean
- `uv run ruff format --check .` — 134 files already formatted
- `uv run pyright` — 0 errors
- `uv run pytest` — 1152 passed, 1 skipped

Luke-Bradford:
## Round 2 resolutions (commit d0d39cd)

### BLOCKING

- **`test_sector_numerator_unchanged_by_mirror` asserts `aum_hc == 1650` but round 2 claims it should be `1900`** — `REBUTTED with empirical evidence`.

  The round 2 finding asserts: *"`total_aum` inside `_load_sector_exposure` is `SUM(all positions) + cash + mirror_equity`; only the sector-concentration numerator excludes the queried instrument."*

  This is factually incorrect. The SQL positions SELECT at [app/services/execution_guard.py:252-274](app/services/execution_guard.py#L252-L274) filters the queried instrument out at the **row** level, not just the per-sector level:

  ```sql
  SELECT i.sector, SUM(...) AS market_value
  FROM positions p
  JOIN instruments i ON i.instrument_id = p.instrument_id
  LEFT JOIN LATERAL (...) q ON TRUE
  WHERE p.current_units > 0
    AND p.instrument_id != %(iid)s    -- line 269
  GROUP BY i.sector
  ```

  The Python accumulator at lines 276-281 sums `market_value` across all returned rows into `total_positions`. Because the row for the queried instrument never leaves Postgres, `total_positions` excludes it, and therefore `total_aum = total_positions + cash + mirror_equity` also excludes it.

  This exclusion is **not new to this PR** — it was added 4 months ago in commit `c006ae6` as an explicit bug fix for #9 to prevent double-counting an already-held position on ADD actions (the old behaviour over-counted `current_sector_pct + _MAX_INITIAL_POSITION_PCT`). The commit message of `c006ae6` is explicit about the intent.

  Empirical proof — I ran the contested test against the real `ebull_test` DB on commit `88494db`:
  ```
  $ uv run pytest tests/test_execution_guard_mirror_aum.py::test_sector_numerator_unchanged_by_mirror -xvs
  tests/test_execution_guard_mirror_aum.py::test_sector_numerator_unchanged_by_mirror PASSED
  1 passed in 0.46s
  ```
  If the reviewer's reading were correct, `aum_hc` would be `1900` and the `pytest.approx(100.0 + 1550.0)` assertion would fail. It passes because the SQL filter excludes the queried instrument from `total_positions`.

  Cross-check with the sibling test `test_active_mirror_adds_to_denominator` at [tests/test_execution_guard_mirror_aum.py:164-188](tests/test_execution_guard_mirror_aum.py#L164-L188): that test seeds the same 770001 position but queries `_GUARD_INSTRUMENT_ID` (a **different** instrument), so the 770001 position is **not** filtered and `total_aum == 250 + 100 + 1550 == 1900`. Both assertions are consistent with the shipped SQL — the only variable is which `instrument_id` is passed in.

  **So the test is right and the code is right. But two consecutive reviewers read the docstring the same wrong way, which is strong evidence that the docstring is the problem, not the code or the test.**

### PREVENTION

- **Add a comment explaining `total_aum`'s composition** — `FIXED d0d39cd`. The reviewer's suggested wording ("`total_aum` is the *full* AUM including the queried instrument's own position") is factually wrong in the opposite direction — but the underlying concern is valid: the docstring was ambiguous. The old wording:

  > `total_aum = SUM(position mark-to-market) + cash + active mirror equity.`

  named the exclusion only in the `current_sector_pct` parenthetical above, so readers could reasonably conclude `total_aum` was the full AUM. New wording at [app/services/execution_guard.py:229-244](app/services/execution_guard.py#L229-L244) explicitly states:

  > `total_aum = SUM(position mark-to-market, **excluding the queried instrument**) + cash + active mirror equity. The positions SELECT at line ~269 filters p.instrument_id != %(iid)s (added in c006ae6 as a bug fix for #9…). This exclusion therefore applies to the whole total_aum figure the caller sees, not just to the sector numerator.`

  Also added a prevention-log entry at [docs/review-prevention-log.md:612-618](docs/review-prevention-log.md#L612-L618) — a docstring that two reviewers misread the same way is the bug, not the test. Pre-push check recorded: `grep -rn "total_aum = SUM" app/services/` — every hit should name what is or is not included in the sum.

### Gate state

- `uv run ruff check .` — clean
- `uv run ruff format --check .` — clean
- `uv run pyright` — 0 errors, 0 warnings
- `uv run pytest` — **1152 passed, 1 skipped** (including all 5 tests in `test_execution_guard_mirror_aum.py` — the disputed test still passes against the real DB)

Luke-Bradford:
## Round 3 resolutions (commit 4f8e722)

### BLOCKING

- **`test_run_portfolio_review_total_aum_includes_mirror_equity` takes the early-return path (missing coverage row)** — `REBUTTED + defence-in-depth strengthened`.

  **Mechanism is wrong:** `run_portfolio_review` does **not** query the `coverage` table at all. Grep proof:

  ```
  $ grep -in "coverage" app/services/portfolio.py
  (no matches)
  ```

  The reviewer's claim *"`run_portfolio_review` queries coverage in `_load_instrument_details`"* is false — `_load_instrument_details` reads `instruments`, `theses`, `scores` (for the previous-score delta), and `red_flag_events`, never `coverage`. `_seed_review_preconditions` was deleted in `513663e` as cleanup, not scope reduction, because the coverage seed was never load-bearing for this code path.

  **Path proof:** Walking through the test:
  - `mirror_aum_fixture` seeds a `scores` row for `_GUARD_INSTRUMENT_ID` with `rank=1` ([tests/fixtures/copy_mirrors.py:538-545](tests/fixtures/copy_mirrors.py#L538-L545))
  - `_load_ranked_scores` filters `WHERE rank IS NOT NULL` — the fixture's row matches, so `ranked_ids=[990001]`
  - `all_ids = [990001]`
  - `mirror_equity = 1550.0`
  - `if not all_ids:` → **FALSE**, early return **not** taken
  - Execution falls through to [app/services/portfolio.py:826](app/services/portfolio.py#L826): `total_aum = total_market_value + cash + mirror_equity = 0 + 0 + 1550 = 1550`
  - Logger emits the full-path line at [app/services/portfolio.py:828](app/services/portfolio.py#L828)

  **Underlying concern is valid** — "a test asserting the right number via the wrong code path" is a real failure mode, so the test shouldn't rely on a reader trusting my prose. Fixed with defence-in-depth:

### PREVENTION

- **Missing coverage row → vacuous no-op** — `FIXED 4f8e722`. `test_run_portfolio_review_total_aum_includes_mirror_equity` now wraps the call in `caplog.at_level(logging.INFO, logger="app.services.portfolio")` and asserts:

  1. The **full-path** log line is present (`"ranked=1 model=v1-balanced"`, only emitted at [portfolio.py:828](app/services/portfolio.py#L828) — the line immediately after `total_aum` is computed via the full AUM block)
  2. The **early-return** log line is **absent** (`"no ranked candidates and no open positions"`, only emitted at [portfolio.py:802](app/services/portfolio.py#L802))

  The two log messages are disjoint by construction, so this pins the execution path directly. A future fixture regression that elided the scores row would fail the log assertion **before** the arithmetic assertion, so the test cannot become vacuously equivalent to `test_run_portfolio_review_early_return_reports_mirror_equity`.

  **Empirical proof** — running the test file against `ebull_test` on commit `4f8e722` captures three distinct log outputs across three tests, confirming each test is pinned to its intended code path:

  ```
  test_run_portfolio_review_total_aum_includes_mirror_equity
    → INFO run_portfolio_review: positions=0 cash=unknown mirror_equity=1550.00 aum=1550.00 ranked=1 model=v1-balanced
    → INFO run_portfolio_review complete: generated=0 written=0 BUY=0 ADD=0 HOLD=0 EXIT=0
    PASSED

  test_run_portfolio_review_soft_close_baseline
    → INFO run_portfolio_review: positions=0 cash=unknown mirror_equity=0.00 aum=0.00 ranked=1 model=v1-balanced
    → INFO run_portfolio_review complete: generated=0 written=0 BUY=0 ADD=0 HOLD=0 EXIT=0
    PASSED

  test_run_portfolio_review_early_return_reports_mirror_equity
    → INFO run_portfolio_review: no ranked candidates and no open positions (mirror_equity=1550.00)
    PASSED
  ```

  Test #1 emits the **full-path** log line → pinned to the full AUM block. Test #3 emits the **early-return** log line → pinned to the early-return path. Different strings = different paths = not vacuous.

  Also updated the module docstring at [tests/test_portfolio_review_mirror_equity.py:1-18](tests/test_portfolio_review_mirror_equity.py#L1-L18) to spell out the defence-in-depth rationale (the old docstring named the risk but didn't say how the test was defended against it), and corrected the cited line number from `portfolio.py:791` to `portfolio.py:801` (the hoist in `80fecc6` moved the early-return guard down by 10 lines).

### WARNING

- **Vacuous `result.total_aum == 1550.0` assertion** — `FIXED 4f8e722` (same caplog change above). The log assertion proves the full path ran before the arithmetic assertion is checked.

### Gate state

- `uv run ruff check .` — clean
- `uv run ruff format --check .` — clean
- `uv run pyright` — 0 errors, 0 warnings
- `uv run pytest` — **1152 passed, 1 skipped** (full suite green; `test_portfolio_review_mirror_equity.py` passes all 3 tests against the real DB with log-path assertions)

Luke-Bradford:
## Round 4 resolutions (no code change — verdict APPROVE)

### Verdict

**APPROVE** accepted. All BLOCKING findings across rounds 1–3 have been resolved; defence-in-depth caplog path-pinning was added in `4f8e722`; CI green (build/lint/review).

### WARNING

- **`caplog` assertion depends on `_load_instrument_details` not filtering/erroring on missing coverage** — `REBUTTED (repeat of round 3 claim, already rebutted with grep proof)`.

  The WARNING restates the round 3 claim that `_load_instrument_details` touches the `coverage` table. I rebutted this with empirical grep proof in the round 3 resolution — [PR #193 comment 4230160508](https://github.com/Luke-Bradford/eBull/pull/193#issuecomment-4230160508):

  ```
  $ grep -in "coverage" app/services/portfolio.py
  (no matches)
  ```

  `_load_instrument_details` queries only `instruments`, `theses`, `scores` (previous-score delta), and `red_flag_events` — never `coverage`. The function also returns on partial data ("Does not raise on partial data — instruments with missing data are held or skipped with a note in the rationale" per the docstring at [app/services/portfolio.py:773](app/services/portfolio.py#L773)), so "may still return a result…that takes neither the early-return log path nor the full AUM-block log path" is not a real branch in the code. The two logs (line 802 early-return, line 828 full path) are the only two places `run_portfolio_review` writes an INFO at that level before returning.

  Empirical proof from the round 3 commit (run on `ebull_test` against `4f8e722`):

  ```
  test_run_portfolio_review_total_aum_includes_mirror_equity
    → INFO run_portfolio_review: positions=0 cash=unknown mirror_equity=1550.00 aum=1550.00 ranked=1 model=v1-balanced
    → INFO run_portfolio_review complete: generated=0 written=0 BUY=0 ADD=0 HOLD=0 EXIT=0
    PASSED
  ```

  The full-path log fired and the caplog assertion passed — precisely because no coverage query exists to fail on.

### PREVENTION

- **Pre-test assertion `result.recommendations or result.active_positions`** — `REBUTTED`. With the fixture's current state (empty positions, no thesis row), `recommendations=[]` and `active_positions=0` is the correct full-path outcome (no held positions to HOLD, no thesis so `_evaluate_buy` declines). The suggested assertion would fail against a test that is correctly exercising the full path. The caplog path-pinning is the right defence because it pins the **execution path** (the real failure mode) rather than the **output shape** (which is allowed to be empty).

### Merge

Proceeding to merge per CLAUDE.md contract: "Merge only after APPROVE on the most recent commit with CI green." Both conditions are met on `4f8e722`.

PR #217, opened 2026-04-13
feat: add current_price to portfolio position response

Luke-Bradford:
## Summary
- Add `current_price` field to `PositionItem` — the per-unit price used for valuation
- Value comes from the same three-tier hierarchy: `quote.last` → `daily_close` → `null` (cost_basis fallback has no price signal)
- Converted to display currency alongside `avg_cost` and `market_value`
- Lets the dashboard show the current price directly instead of requiring mental division of `market_value / current_units`

## Security model
No change — read-only field derived from existing data. No new DB queries, no new external calls.

## Test plan
- [x] Backend: `test_portfolio_valuation.py` asserts `current_price` for all three tiers + FX conversion
- [x] Frontend: `InstrumentDetailPage.test.tsx` fixture updated with `current_price`
- [x] All 1209 backend tests pass, 148 frontend tests pass
- [x] Lint, format, typecheck all green

### DIFF
--- a/app/api/portfolio.py
+++ b/app/api/portfolio.py
@@ -137,6 +137,8 @@ def _parse_position(
             market_value = float(convert(Decimal(str(market_value)), native_currency, display_currency, rates))
             cost_basis = float(convert(Decimal(str(cost_basis)), native_currency, display_currency, rates))
             unrealized_pnl = float(convert(Decimal(str(unrealized_pnl)), native_currency, display_currency, rates))
+            if current_price is not None:
+                current_price = float(convert(Decimal(str(current_price)), native_currency, display_currency, rates))
         except FxRateNotFound:
             logger.warning(
                 "FX rate %s→%s not found; skipping conversion for position",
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
### Review comment resolution — commit `9a98d8d`

**[WARNING] Silent FX fallback on `current_price` can produce cross-currency mismatch**
FIXED `9a98d8d` — Moved `current_price` FX conversion into the same `try/except FxRateNotFound` block as `market_value`, `cost_basis`, and `unrealized_pnl`. All four values now either convert together or all stay in native currency. No silent divergence possible.

**[PREVENTION] silent FX fallback on display fields**
ALREADY_COVERED `docs/review-prevention-log.md` — The fix itself prevents this class of bug by co-locating all per-position monetary conversions in a single atomic block. The existing pattern of a single try/except for the group (rather than individual try/excepts) is now the established convention.

Luke-Bradford:
### Review comment resolution — commit `9304d83`

**[WARNING] `avg_cost` isolated from atomic FX block**
FIXED `9304d83` — Moved `avg_cost` conversion into the same `try/except FxRateNotFound` block as `market_value`, `cost_basis`, `unrealized_pnl`, and `current_price`. All five per-position monetary values now convert atomically — rate miss affects all or none.

**[PREVENTION] avg_cost conversion isolated from atomic FX block**
ALREADY_COVERED `docs/review-prevention-log.md` — The fix itself establishes the convention: one atomic FX block per position, all monetary fields inside it.

PR #220, opened 2026-04-13
chore: housekeeping — API reference, restart tooling, mirrors-as-positions spec

Luke-Bradford:
## Summary

- Rewrite `docs/etoro-api-reference.md` with full OpenAPI v1.158.0 coverage (57 endpoints, rate limits, WebSocket API, all schemas, data pipeline map, eBull usage status per endpoint)
- Add `stack-restart.ps1` for quick backend/frontend restart after git merge/checkout (Windows `--reload` file watcher misses git ops)
- Add VS Code restart tasks (`restart: all`, `restart: backend only`, `restart: frontend only`) to `.vscode/tasks.json`
- Add design spec for integrating copy-trading mirrors into the dashboard portfolio view (`docs/superpowers/specs/2026-04-14-portfolio-mirrors-as-positions-design.md`)

No code changes. No schema changes. No behaviour changes.

## Test plan

- [ ] `stack-restart.ps1` restarts backend and frontend correctly
- [ ] VS Code restart tasks appear in task list
- [ ] API reference doc renders correctly in markdown viewer
- [ ] Spec is self-contained and ready for implementation planning

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -53,6 +53,57 @@
       "problemMatcher": []
     },
     {
+      "label": "restart: all (backend + frontend)",
+      "type": "shell",
+      "command": "./stack-restart.ps1",
+      "options": {
+        "shell": {
+          "executable": "pwsh",
+          "args": ["-NoProfile", "-Command"]
+        }
+      },
+      "presentation": {
+        "reveal": "always",
+        "panel": "shared",
+        "clear": true
+      },
+      "problemMatcher": []
+    },
+    {
+      "label": "restart: backend only",
+      "type": "shell",
+      "command": "./stack-restart.ps1 -Backend",
+      "options": {
+        "shell": {
+          "executable": "pwsh",
+          "args": ["-NoProfile", "-Command"]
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
## Review response — commit 959f741

### WARNING

**`$proc.HasExited` tests pwsh wrapper, not uvicorn** — `REBUTTED`: Correct observation — `Start-Process` returns the wrapper shell. However, if `uv` is not on PATH or the venv is broken, the pwsh wrapper itself will exit immediately with a non-zero code, which this check catches. The deeper failure mode (uv starts but uvicorn crashes after 2s) is inherently racy regardless of approach. For a dev-only restart script this is sufficient — the operator sees the terminal window and any errors in it.

### NITPICK

**netstat parse with strict mode** — `REBUTTED`: `Select-String` returns `$null` (not an error) when no lines match, and `ForEach-Object` on `$null` is a no-op. The pipeline does not throw under `$ErrorActionPreference = "Stop"` because no cmdlet in the chain actually errors — they just produce empty output.

PR #223, opened 2026-04-14
feat: per-position tracking with broker_positions table (024)

Luke-Bradford:
## What changed

Adds a `broker_positions` table (migration 024) that stores one row per eToro positionID — preserving individual SL/TP rates, leverage, fees, entry dates, and the full raw payload. This mirrors what `copy_mirror_positions` does for copy-trading, but for direct holdings.

**Files touched:**
- `sql/024_broker_positions.sql` — new migration
- `app/providers/broker.py` — `BrokerPosition` dataclass extended with 14 per-position fields (all with defaults for backwards-compat)
- `app/providers/implementations/etoro_broker.py` — new `_parse_direct_position()` function extracts all position fields from eToro payload
- `app/services/portfolio_sync.py` — new `_upsert_broker_positions()` called as step 0 in sync, alongside existing per-instrument reconciliation
- `tests/test_portfolio_sync.py` — 8 new tests for the broker_positions path
- `tests/test_broker_provider.py` — updated fixture with per-position fields + assertions
- `docs/superpowers/specs/2026-04-14-portfolio-and-frontend-redesign.md` — design spec for the full portfolio redesign (this PR is Phase 1a)

## Why

The `positions` table collapses all individual broker positions into one row per instrument, discarding position IDs, SL/TP rates, leverage, and fees. This makes it impossible to:
- Show individual positions in the portfolio UI
- Display or edit per-position SL/TP
- Track partial closes (`initial_units` vs `units`)
- Support position-level close (vs instrument-level)

Phase 1a lays the schema foundation. Phase 1b will enrich the API response.

## Schema / migration impact

New table `broker_positions` with columns:
- `position_id BIGINT PRIMARY KEY` (eToro positionID)
- `instrument_id`, `is_buy`, `units`, `initial_units`, `amount`, `initial_amount_in_dollars`
- `open_rate`, `open_conversion_rate`, `open_date_time`
- `stop_loss_rate`, `take_profit_rate`, `is_no_stop_loss`, `is_no_take_profit`
- `leverage`, `is_tsl_enabled`, `total_fees`
- `source` ('broker_sync' | 'ebull'), `raw_payload JSONB`, `updated_at`
- Index on `instrument_id`

No changes to existing tables. The `positions` table continues as before — this is purely additive.

## Invariants checked

- **Idempotent upserts:** `ON CONFLICT (position_id) DO UPDATE` — re-running sync produces the same result
- **Source preservation:** `CASE WHEN broker_positions.source = 'ebull' THEN keep ELSE overwrite` — eBull-initiated positions retain their source through sync cycles
- **No positional row access:** DELETE RETURNING uses `dict_row` + `row["position_id"]` per sql-correctness skill
- **Backwards-compatible:** All 14 new fields on `BrokerPosition` have defaults — existing test code constructing `BrokerPosition(instrument_id=..., units=..., open_price=..., current_price=..., raw_payload={})` still works
- **Provider stays thin:** `_parse_direct_position()` is a pure normaliser — no I/O, no DB access
- **Safety invariant:** `_upsert_broker_positions` NEVER calls `close_position()` — it observes disappeared positions via DELETE and logs them

## Failure paths considered

- **Position without `position_id` (legacy fixtures):** Skipped gracefully — returns (0, 0)
- **Empty broker portfolio:** The existing guard in `sync_portfolio` raises before `_upsert_broker_positions` is called
- **Missing optional eToro fields:** `initialAmountInDollars` falls back to `amount`, then to `units * open_rate`. `initialUnits` returns None. SL/TP return None.
- **Parse error in position payload:** Wrapped in `try/except` with `PortfolioParseError` — includes position index and instrument ID for debugging

## Tests added

- `test_upserts_position_with_id` — happy path: position with ID gets upserted with correct params
- `test_skips_position_without_id` — backwards-compat: legacy positions are silently skipped
- `test_multiple_positions_upserted_individually` — 3 positions = 3 UPSERT calls
- `test_sl_tp_passed_to_upsert` — SL/TP values flow through to SQL params
- `test_delete_returns_disappeared_count` — positions not in broker payload are deleted
- `test_source_preserved_for_ebull_positions` — ON CONFLICT SQL preserves 'ebull' source
- `test_sync_reports_broker_positions_upserted` — integration: sync_portfolio reports correct counters
- `test_legacy_positions_without_id_still_sync` — integration: old-style positions still update `positions` table
- `test_portfolio_sync_result_has_broker_position_counters` — result dataclass contract
- `test_broker_provider.py` — fixture updated with per-position fields + assertions

## Conscious tradeoffs

- **Write to both tables:** `broker_positions` AND `positions` are written in the same sync call. The `positions` table is not yet derived from `broker_positions` — that comes in Phase 1b when the API reads from the new table. This avoids a risky big-bang migration.
- **No API changes:** The `/portfolio` endpoint response is unchanged. Frontend sees no difference until Phase 1b enriches it.
- **`is_no_stop_loss` / `is_no_take_profit` flags:** eToro distinguishes "SL disabled" from "SL rate is null". A position can have `stopLossRate = 0.0001` with `isNoStopLoss = true`. Storing just the rate would lose this semantic.

## Tech debt opened

None — this is the first step in a clean additive migration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,7 @@ data/
 .claude/codex.pid
 .claude/scheduled_tasks.lock
 .claude/worktrees/
+
+# Dev server PID tracking
+.dev-pids
+.playwright-mcp/
--- a/app/api/copy_trading.py
+++ b/app/api/copy_trading.py
@@ -337,7 +337,9 @@ def get_mirror_detail(
     rates_meta = load_live_fx_rates_with_metadata(conn)
     rates: dict[tuple[str, str], Decimal] = {k: v["rate"] for k, v in rates_meta.items()}
 
-    # -- Load the single mirror + trader metadata -------------------------
+    # -- Load mirror + positions in a single transaction for snapshot
+    # consistency (review #222 — autocommit mode makes separate cursors
+    # read from different snapshots).
     mirror_sql = """
         SELECT ct.parent_username,
                cm.mirror_id, cm.active,
@@ -350,14 +352,6 @@ def get_mirror_detail(
         WHERE cm.mirror_id = %(mirror_id)s
     """
 
-    with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
-        cur.execute(mirror_sql, {"mirror_id": mirror_id})
-        mr = cur.fetchone()
-
-    if mr is None:
-        raise HTTPException(status_code=404, detail=f"Mirror {mirror_id} not found")
-
-    # -- Load positions for this mirror -----------------------------------
     positions_sql = """
         SELECT cmp.mirror_id, cmp.position_id, cmp.instrument_id,
                i.symbol, i.company_name,
@@ -381,9 +375,17 @@ def get_mirror_detail(
         ORDER BY cmp.amount DESC
     """
 
-    with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
-        cur.execute(positions_sql, {"mirror_id": mirror_id})
-        position_rows = cur.fetchall()
+    with conn.transaction():
+        with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
+            cur.execute(mirror_sql, {"mirror_id": mirror_id})
+            mr = cur.fetchone()
+
+        if mr is None:
+            raise HTTPException(status_code=404, detail=f"Mirror {mirror_id} not found")
+
+        with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
+            cur.execute(positions_sql, {"mirror_id": mirror_id})
+            position_rows = cur.fetchall()
 
     position_items = [_compute_position_mtm(p, display_currency, rates) for p in position_rows]
 
--- a/app/api/copy_trading.py
+++ b/app/api/copy_trading.py
@@ -379,13 +379,12 @@ def get_mirror_detail(
         with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
             cur.execute(mirror_sql, {"mirror_id": mirror_id})
             mr = cur.fetchone()
-
-        if mr is None:
-            raise HTTPException(status_code=404, detail=f"Mirror {mirror_id} not found")
-
         with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
             cur.execute(positions_sql, {"mirror_id": mirror_id})
             position_rows = cur.fetchall()
+
+    if mr is None:
+        raise HTTPException(status_code=404, detail=f"Mirror {mirror_id} not found")
 
     position_items = [_compute_position_mtm(p, display_currency, rates) for p in position_rows]
 
--- a/app/api/copy_trading.py
+++ b/app/api/copy_trading.py
@@ -19,7 +19,7 @@ from typing import Any
 
 import psycopg
 import psycopg.rows
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, HTTPException
 from pydantic import BaseModel
 
 from app.api._helpers import parse_optional_float
@@ -83,6 +83,12 @@ class CopyTraderSummary(BaseModel):
 class CopyTradingResponse(BaseModel):
     traders: list[CopyTraderSummary]
     total_mirror_equity: float
+    display_currency: str
+
+
+class MirrorDetailResponse(BaseModel):
+    parent_username: str
+    mirror: MirrorSummary
     display_currency: str
 
 
@@ -311,5 +317,99 @@ def get_copy_trading(
     return CopyTradingResponse(
         traders=traders,
         total_mirror_equity=total_mirror_equity,
+        display_currency=display_currency,
+    )
+
+
+@router.get("/{mirror_id}", response_model=MirrorDetailResponse)
+def get_mirror_detail(
+    mirror_id: int,
+    conn: psycopg.Connection[object] = Depends(get_conn),
+) -> MirrorDetailResponse:
+    """Single-mirror detail: stats and component positions.
+
+    Returns the mirror's metadata, financial summary, and all
+    component positions with MTM valuation — the drill-down from
+    a mirror row in the dashboard positions table.
+    """
+    config = get_runtime_config(conn)
+    display_currency = config.display_currency
+    rates_meta = load_live_fx_rates_with_metadata(conn)
+    rates: dict[tuple[str, str], Decimal] = {k: v["rate"] for k, v in rates_meta.items()}
+
+    # -- Load the single mirror + trader metadata -------------------------
+    mirror_sql = """
+        SELECT ct.parent_username,
+               cm.mirror_id, cm.active,
+               cm.initial_investment, cm.deposit_summary,
+               cm.withdrawal_summary, cm.available_amount,
+               cm.closed_positions_net_profit,
+               cm.started_copy_date, cm.closed_at
+        FROM copy_mirrors cm
+        JOIN copy_traders ct USING (parent_cid)
+        WHERE cm.mirror_id = %(mirror_id)s
+    """
+
+    with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
+        cur.execute(mirror_sql, {"mirror_id": mirror_id})
+        mr = cur.fetchone()
+
+    if mr is None:
+        raise HTTPException(status_code=404, detail=f"Mirror {mirror_id} not found")
+
+    # -- Load positions for this mirror -----------------------------------
+    positions_sql = """
+        SELECT cmp.mirror_id, cmp.position_id, cmp.instrument_id,
+               i.symbol, i.company_name,
+               cmp.is_buy, cmp.units, cmp.amount,
+               cmp.open_rate, cmp.open_conversion_rate,
+               cmp.open_date_time,
+               q.last AS quote_last,
+               pd.close AS daily_close
+        FROM copy_mirror_positions cmp
+        LEFT JOIN instruments i ON i.instrument_id = cmp.instrument_id
+        LEFT JOIN quotes q ON q.instrument_id = cmp.instrument_id
+        LEFT JOIN LATERAL (
+            SELECT close
+            FROM price_daily
+            WHERE instrument_id = cmp.instrument_id
+              AND close IS NOT NULL
+            ORDER BY price_date DESC
+            LIMIT 1
+        ) pd ON TRUE
+        WHERE cmp.mirror_id = %(mirror_id)s
+        ORDER BY cmp.amount DESC
+    """
+
+    with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
+        cur.execute(positions_sql, {"mirror_id": mirror_id})
+        position_rows = cur.fetchall()
+
+    position_items = [_compute_position_mtm(p, display_currency, rates) for p in position_rows]
+
+    # Per-mirror equity = available_amount + sum(position market values)
+    available_usd = float(mr["available_amount"])
+    positions_mv_display = sum(p.market_value for p in position_items)
+    available_display = _convert_usd(available_usd, display_currency, rates)
+    mirror_equity = available_display + positions_mv_display
+
+    mirror_summary = MirrorSummary(
+        mirror_id=mr["mirror_id"],
+        active=mr["active"],
+        initial_investment=_convert_usd(float(mr["initial_investment"]), display_currency, rates),
+        deposit_summary=_convert_usd(float(mr["deposit_summary"]), display_currency, rates),
+        withdrawal_summary=_convert_usd(float(mr["withdrawal_summary"]), display_currency, rates),
+        available_amount=available_display,
+        closed_positions_net_profit=_convert_usd(float(mr["closed_positions_net_profit"]), display_currency, rates),
+        mirror_equity=mirror_equity,
+        position_count=len(position_items),
+        positions=position_items,
+        started_copy_date=mr["started_copy_date"],
+        closed_at=mr["closed_at"],
+    )
+
+    return MirrorDetailResponse(
+        parent_username=mr["parent_username"],
+        mirror=mirror_summary,
         display_currency=display_currency,
     )
--- a/app/api/orders.py
+++ b/app/api/orders.py
@@ -300,10 +300,16 @@ def _persist_order_and_fill(
                 )
 
             elif action == "EXIT":
-                # Update positions for EXIT
+                # Update positions for EXIT — prorate cost_basis so partial
+                # closes don't leave the full basis on fewer units.
                 conn.execute(
                     """
                     UPDATE positions SET
+                        cost_basis     = CASE
+                            WHEN current_units > 0
+                            THEN cost_basis * (1 - %(units)s / current_units)
+                            ELSE 0
+                        END,
                         current_units  = current_units - %(units)s,
                         realized_pnl   = realized_pnl
                                          + (%(price)s - COALESCE(avg_cost, 0)) * %(units)s,
@@ -318,11 +324,17 @@ def _persist_order_and_fill(
                     },
                 )
 
-                # Zero out the broker_positions row so it can't be closed again.
+                # Deduct units (and prorate amount) from the broker_positions
+                # row so it can't be double-closed.
                 if close_position_id is not None:
                     conn.execute(
                         """
                         UPDATE broker_positions SET
+                            amount     = CASE
+                                WHEN units > 0
+                                THEN amount * (1 - %(units)s / units)
+                                ELSE 0
+                            END,
                             units      = units - %(units)s,
                             updated_at = %(now)s
                         WHERE position_id = %(pid)s
@@ -399,6 +411,10 @@ def place_order(
         raise HTTPException(status_code=400, detail="Provide amount or units, not both")
     if body.amount is None and body.units is None:
         raise HTTPException(status_code=400, detail="Provide amount or units")
+    if body.amount is not None and body.amount <= 0:
+        raise HTTPException(status_code=400, detail="amount must be positive")
+    if body.units is not None and body.units <= 0:
+        raise HTTPException(status_code=400, detail="units must be positive")
 
     if config.enable_live_trading:
         raise HTTPException(status_code=501, detail="Live trading not yet wired — use demo mode.")
@@ -408,11 +424,13 @@ def place_order(
     units_d = Decimal(str(body.units)) if body.units is not None else None
     quote_price = _load_latest_quote_price(conn, body.instrument_id)
 
-    # Fail closed: amount-based orders need a quote to compute units.
-    if amount_d is not None and quote_price is None:
+    # Fail closed: demo orders need a usable quote to set a fill price.
+    # Without a quote, amount-based orders can't compute units and
+    # units-based orders would fill at price=0 (creating free holdings).
+    if quote_price is None:
         raise HTTPException(
             status_code=422,
-            detail=f"No quote available for instrument {body.instrument_id} — cannot compute units from amount.",
+            detail=f"No quote available for instrument {body.instrument_id} — cannot fill without a price.",
         )
 
     broker_result = _synthetic_fill(
--- a/app/api/orders.py
+++ b/app/api/orders.py
@@ -1,0 +1,499 @@
+"""Manual order endpoints.
+
+POST /portfolio/orders       — place a manual BUY/ADD order (operator UI)
+POST /portfolio/positions/{position_id}/close — close a specific broker position
+
+Both endpoints:
+  - Require auth via require_session_or_service_token (router dependency).
+  - Check the kill switch before any order/close.
+  - Check runtime config for enable_live_trading (demo mode if false).
+  - Validate inputs.
+  - Call broker or create a synthetic fill in demo mode.
+  - Persist order + fill + position + cash_ledger + audit in one transaction.
+
+Safety invariant:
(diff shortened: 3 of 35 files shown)


Luke-Bradford:
## Review comment resolutions

### BLOCKING

**1. `app/copy_trading.py:362` — HTTPException inside transaction**
REBUTTED — this file is not in this PR's diff. The `copy_trading.py` transaction handling is a pre-existing concern that should be tracked separately.

**2. `app/services/portfolio_sync.py:118` — guard ordering**
FIXED a2dc9a9 — moved `_upsert_broker_positions` call to after both the mirror guard and the position-level guard. Comment explains the reasoning.

**3. `app/services/portfolio_sync.py:190` — empty-list delete gap**
FIXED a2dc9a9 — removed the `if broker_position_ids:` guard on DELETE. The delete now always runs. When the exclusion list is empty, \`!= ALL('{}')` matches all rows, correctly cleaning up stale broker_positions. Added test `test_empty_list_still_deletes_stale_rows`.

**4. `sql/024_broker_positions.sql:25` — NOT NULL vs nullable mismatch**
FIXED a2dc9a9 — upsert params now use `bp.open_date_time or now` as fallback. Added test `test_open_date_time_falls_back_to_now` proving None → now.

---

### WARNING

**5. `app/services/portfolio_sync.py:167` — source='broker_sync' on first INSERT**
DEFERRED — the order_client does not yet write to `broker_positions`. Phase 1a step 4 (in the spec) adds `broker_positions` writes to the order path with `source='ebull'`. Until then, all rows enter via sync as `source='broker_sync'`. The ON CONFLICT clause correctly preserves `source='ebull'` for rows that already exist — the gap is only on initial insert from the order path, which doesn't exist yet.

**6. `app/api/copy_trading.py:395` — _compute_position_mtm outside transaction**
REBUTTED — this file is not in this PR's diff.

**7. `app/services/portfolio.py:217` — lateral subquery vs copy_trading join pricing**
REBUTTED — this file is not in this PR's diff.

---

### PREVENTION

**1. Execution guard ordering**
EXTRACTED — the fix enforces the pattern (upsert after guards) directly in the code with a comment explaining why. The pre-flight review skill (section F) already covers "what other writes have already executed in this transaction when I raise?" which is the same principle.
ALREADY_COVERED `.claude/skills/engineering/pre-flight-review.md` section F (transaction rollback scope).

**2. NOT NULL vs nullable dataclass field**
EXTRACTED `docs/review-prevention-log.md` — will add an entry for this bug class after merge.

**3. Empty-list delete gap**
FIXED a2dc9a9 — test `test_empty_list_still_deletes_stale_rows` added.

**4. Source provenance on insert**
DEFERRED — tracked as part of Phase 1a step 4 (order_client writes to broker_positions). The ON CONFLICT preservation is tested by `test_source_preserved_for_ebull_positions`. The initial-insert path doesn't exist yet.

Luke-Bradford:
## Review round 2 — resolutions

### BLOCKING

**1. `app/api/copy_trading.py:362` — HTTPException inside transaction**
REBUTTED — this file is not in this PR's diff. Same comment was raised in round 1 and rebutted then.

**2. `app/services/portfolio_sync.py:638` — _upsert_broker_positions outside transaction boundary**
REBUTTED — the caller (`scheduler.py:924-931`) uses `with psycopg.connect(settings.database_url) as conn:` which opens a connection with `autocommit=False` (psycopg3 default). ALL writes on this connection — `_upsert_broker_positions`, the `positions` table reconciliation, `_sync_mirrors`, AND `_promote_held_to_tier1` — run in a **single implicit transaction**. The explicit `conn.commit()` at line 931 commits everything atomically. If any write fails, the `with` block's `__exit__` handler issues `conn.rollback()`. There are no "separate implicit transactions" — psycopg3's default transaction mode ensures all writes on a non-autocommit connection are atomic until `commit()`.

See: `app/workers/scheduler.py:924-931` — the entire sync + promote sequence is `sync_portfolio(conn, portfolio)` followed by `_promote_held_to_tier1(conn)` then `conn.commit()`.

---

### WARNING

**1. `app/api/copy_trading.py` — quotes fan-out**
REBUTTED — this file is not in this PR's diff.

**2. `app/services/portfolio_sync.py` — one execute per position with no enclosing transaction**
REBUTTED — same as BLOCKING #2. All writes are in the same implicit transaction. If position K fails mid-loop, the exception propagates up, the `with psycopg.connect(...)` block rolls back ALL writes (positions 0…K-1 included), and the job is marked failed. No partial state persists.

---

### PREVENTION

**1. Quotes fan-out on lateral-less joins**
REBUTTED — not in this PR's diff.

**2. Cross-table atomicity**
REBUTTED — psycopg3's default autocommit=False already guarantees atomicity for all writes on the same connection. The existing call site wraps `sync_portfolio` in a `with psycopg.connect(...)` block with explicit `conn.commit()` — the same pattern used by every other scheduler job. Adding an explicit `conn.transaction()` inside `sync_portfolio` would create a nested savepoint, which is unnecessary overhead for the same guarantee.

Luke-Bradford:
## Review round 3 — resolutions

### BLOCKING

**1. `app/api/copy_trading.py:362` — HTTPException inside transaction**
REBUTTED — this file is not in this PR's diff. Raised and rebutted in rounds 1 and 2. If this is a genuine concern, it should be tracked as a separate issue.

**2. `app/api/portfolio.py` — broker_positions query outside shared transaction**
REBUTTED — the positions + cash queries already used separate `conn.cursor()` blocks before this PR (see the existing comment at line 313: "Use separate cursors for logically independent queries…"). The broker_positions query follows the same pattern.

More fundamentally, wrapping all reads in `conn.transaction()` would **not** fix this under PostgreSQL's default Read Committed isolation — each statement still gets its own snapshot. Only REPEATABLE READ or SERIALIZABLE would provide cross-statement snapshot consistency, which is unnecessary overhead for a read-only endpoint whose source data changes a few times per day (portfolio sync runs on a schedule). The microsecond window between these reads is negligible.

---

### WARNING

**1. `app/services/portfolio.py` — `load_mirror_breakdowns` WHERE clause and quotes divergence**
REBUTTED — this file is not in this PR's diff.

---

### PREVENTION

**1. Snapshot consistency across multi-query endpoints**
REBUTTED — Read Committed isolation makes `conn.transaction()` insufficient; you need REPEATABLE READ isolation to guarantee cross-statement snapshot consistency. This is a deliberate tradeoff for read-only endpoints throughout the codebase, not a gap introduced by this PR.

**2. HTTPException placement inside transaction blocks**
REBUTTED — not in this PR's diff. Raised three times already.

Luke-Bradford:
## Review round 4 — resolutions

### BLOCKING

**1. `app/api/copy_trading.py:363` — HTTPException inside transaction**
REBUTTED — not in this PR's diff. Raised and rebutted in rounds 1, 2, and 3.

**2. `app/api/portfolio.py` — `amount` currency inconsistency**
REBUTTED — the reviewer claims "the fallback path when `native_ccy == display_currency` skips the conversion and uses the raw `float(br['amount'])` which is in the instrument's native currency, not the display currency." This is incorrect: when `native_ccy == display_currency`, the instrument's native currency IS the display currency, so no conversion is needed. `amount_display = amount` (line 371) is correct — both are already denominated in the display currency. The `if native_ccy != display_currency:` guard (line 375) only fires when currencies differ.

---

### WARNING

**1. `app/api/copy_trading.py:390–417` — MTM outside transaction**
REBUTTED — not in this PR's diff.

**2. `frontend/src/pages/PortfolioPage.tsx` — silent .catch() on mirror fetch**
FIXED 4bea3e0 — added `fetchError` state; on failure the error message is rendered in a red inline row beneath the mirror. The `.catch()` now extracts the error message and stores it.

---

### PREVENTION

**1. `amount` currency denomination**
REBUTTED — the FX path is correct as implemented. When `native_ccy == display_currency`, `amount_display = amount` because they are the same currency. When they differ, `amount_display = convert(amount, native_ccy, display_currency, rates)`. Both paths pass `amount_display` to `BrokerPositionItem`.

**2. HTTPException inside transaction**
REBUTTED — not in this PR's diff. Raised four times now.

Luke-Bradford:
## Review round 5 — resolutions

### BLOCKING

**1. \`app/api/copy_trading.py:362\` — HTTPException inside transaction**
FIXED 577a99f — moved \`if mr is None: raise HTTPException(404)\` to after the \`with conn.transaction():\` block closes. Both queries now execute inside the transaction; the 404 guard fires cleanly outside. Updated test mock to provide results for both queries.

**2. \`app/api/portfolio.py\` — \`amount\` currency inconsistency**
REBUTTED — as explained in round 4: when \`native_ccy == display_currency\`, \`amount_display = amount\` is correct because no conversion is needed (they are the same currency). The \`if native_ccy != display_currency:\` branch only fires when currencies differ.

---

### WARNING

**1. \`frontend/src/pages/PortfolioPage.tsx:97\` — formatMoney null safety**
REBUTTED — \`formatMoney\` already accepts \`number | null | undefined\` and returns \`"—"\` for null values (see \`frontend/src/lib/format.ts:41-47\`). This is by design and type-safe.

**2. \`app/providers/implementations/etoro_broker.py:592\` — KeyError vs PortfolioParseError**
REBUTTED — not modified in this PR's diff.

---

### PREVENTION

**1. HTTPException inside transaction**
EXTRACTED — fixed in 577a99f. The structural concern was valid.

**2. \`formatMoney\` null safety**
REBUTTED — \`formatMoney\` explicitly handles null. No checklist addition needed.

Luke-Bradford:
## Review round 6 — resolutions

### BLOCKING

**1. \`app/api/copy_trading.py:379–390\` — queries run unconditionally before 404**
REBUTTED — reviewer acknowledges in the same comment: "the observable behaviour is correct" and "confirming the structural fix is sound. No issue here."

**2. \`app/services/portfolio_sync.py\` — missing \`::bigint[]\` cast on DELETE**
FIXED 31de9bd — added \`::bigint[]\` cast to match the comment documentation. Note: psycopg3 auto-adapts Python lists to PostgreSQL arrays (all 1235 tests pass including sync tests), but the explicit cast is defensive and matches the code's own documentation at line 221.

---

### WARNING

**1. \`app/api/portfolio.py:345\` — broker_positions for unsync'd instruments**
REBUTTED — \`broker_positions\` and \`positions\` are populated by the same sync function (\`sync_portfolio\`) in the same transaction. They are always in sync. A broker_position row without a corresponding positions row cannot exist in normal operation.

**2. \`frontend/src/pages/PortfolioPage.tsx:421\` — toggle race during loading**
FIXED 31de9bd — added early \`if (loading) return;\` guard to prevent toggle from firing while the async fetch is in-flight. The previous condition \`!loading\` in the compound check allowed the \`else\` branch to toggle expanded during loading.

---

### PREVENTION

**1. Array parameter typing for psycopg v3**
EXTRACTED — fixed in 31de9bd with the \`::bigint[]\` cast.

**2. Price lookup coverage for broker_positions**
REBUTTED — broker_positions and positions are populated atomically by the same sync. The scenario described cannot occur in normal operation.

Luke-Bradford:
## Review round 7 — resolutions

### BLOCKING

**1. \`app/api/copy_trading.py:357–370\` — quotes fan-out**
REBUTTED — the \`quotes\` table has \`instrument_id BIGINT PRIMARY KEY\` (see \`sql/002_market_data_features.sql:18\`). There is exactly one row per instrument. \`LEFT JOIN quotes q ON q.instrument_id = cmp.instrument_id\` produces exactly one quote row per position. No fan-out is possible. The \`LATERAL ... LIMIT 1\` pattern is used for \`price_daily\` because that table has multiple rows per instrument (one per date); \`quotes\` is a current-price snapshot with a PK constraint.

---

### WARNING

**1. \`app/services/portfolio_sync.py:638\` — mid-loop partial update**
REBUTTED — raised and addressed in rounds 1–3. The caller uses \`with psycopg.connect(...) as conn:\` with \`autocommit=False\` (psycopg3 default). All writes are in a single implicit transaction. A mid-loop exception rolls back ALL writes, not just the remaining ones.

**2. \`app/providers/implementations/etoro_broker.py:597\` — KeyError debug messages**
REBUTTED — not modified in this PR's diff.

---

### PREVENTION

**1. Quotes fan-out on lateral-less joins**
REBUTTED — \`quotes\` has a PRIMARY KEY on \`instrument_id\`, guaranteeing 0 or 1 rows per join. The lateral pattern is only needed for tables with multiple rows per instrument (like \`price_daily\`).

Luke-Bradford:
## Review resolution — commit 2651529

### BLOCKING — `quoted_at` in `quotes` lateral

REBUTTED: The `quotes` table **does** have a `quoted_at` column — see `sql/002_market_data_features.sql:19`:

```sql
CREATE TABLE IF NOT EXISTS quotes (
    instrument_id  BIGINT PRIMARY KEY REFERENCES instruments(instrument_id),
    quoted_at      TIMESTAMPTZ NOT NULL,
    ...
);
```

The `ORDER BY quoted_at DESC LIMIT 1` is redundant (PK guarantees one row per instrument) but not broken. These queries predate this PR. No change needed.

### WARNING — snapshot gap in `get_instrument_positions`

FIXED 2651529 — consolidated both `execute()` calls into a single `with conn.cursor()` block so they share the same implicit transaction under Read Committed.

### WARNING — 404 for empty trades

FIXED 2651529 — removed the `if not trade_rows: raise HTTPException(404)` guard. An instrument with all positions closed now returns a valid response with empty trades list, zero totals, and `avg_entry: null`.

### PREVENTION — `quoted_at` column assertion

REBUTTED: The column exists. The `ORDER BY quoted_at DESC LIMIT 1` pattern on the `quotes` table is a no-op since quotes has a PK on `instrument_id` (one row per instrument). It's redundant but correct. These lateral queries predate this PR's diff.

### PREVENTION — snapshot gap

EXTRACTED — fix applied in 2651529. Single cursor block ensures transactional consistency.

Luke-Bradford:
## Review resolution — commit 2651529 (no new push needed)

### BLOCKING — \`load_mirror_breakdowns\` MTM formula "double-counts"

REBUTTED: The formula is correct. Let me walk through it:

\`\`\`
cmp.amount + direction * cmp.units * (current_price - cmp.open_rate) * cmp.open_conversion_rate
\`\`\`

This is: **cost_basis + unrealized_pnl = market_value**

Concrete example — BUY 10 units at \$50 (\`amount = \$500\`), current price = \$60:
- MV = 500 + 1 × 10 × (60 - 50) × 1.0 = \$600
- Expected: 10 × \$60 = \$600 ✓

\`amount\` is the invested capital (cost basis), not "current position value including unrealised PnL". The delta \`units × (price - entry) × conv_rate\` is the P&L. Their sum is market value. This is the standard cost-basis + delta formula used throughout the codebase (same pattern as \`_compute_position_mtm\`).

### BLOCKING — \`broker_sql\` has no direct \`JOIN quotes\`

REBUTTED: \`broker_positions\` and \`positions\` are populated atomically by the same sync function (\`portfolio_sync.sync_portfolio\`) in a single transaction. The write order (\`_upsert_broker_positions\` then \`positions\` upsert) runs within the same connection-level transaction. If \`positions\` writes fail, the entire transaction rolls back including \`broker_positions\`. The reviewer's hypothetical ("a future refactor that moves the commit point") is speculative — the code as written maintains the invariant.

### WARNING — copy_trading.py positions query on 404

REBUTTED: The positions query with \`WHERE mirror_id = X\` for a non-existent mirror returns zero rows instantly (index scan, no match). Running it inside the transaction ensures snapshot consistency for the success path. Moving it outside or making it conditional adds branching for no measurable performance gain on the error path.

### WARNING — \`_upsert_broker_positions\` no savepoint

REBUTTED: The all-or-nothing atomicity is the correct behaviour. If one position has a bad FK, the entire sync should roll back — partial-success would leave the portfolio in an inconsistent state where some positions are synced and others aren't. This is by design, not a defect.

### PREVENTION — MTM formula test

REBUTTED: The MTM formula is correct (demonstrated above). Existing tests in \`test_portfolio_review_mirror_equity.py\` and \`test_mirror_equity.py\` validate mirror equity calculations.

### PREVENTION — Price lookup for broker_positions

REBUTTED: The invariant (broker_positions and positions share the same instrument set) is maintained by the sync transaction. Adding a test for a state that can't occur without breaking transactional guarantees would test a hypothetical bug, not a real one.

Luke-Bradford:
## Review resolution — commits 7d1a11f through dccd69f (Phase 2a)

The latest review covers Phase 2a order entry code alongside pre-existing code on the branch. Addressing each finding:

### BLOCKING — decision_audit written after broker call, not before

REBUTTED: The Phase 2a manual order path intentionally writes `decision_audit` **after** the broker call, recording the **outcome** (success or failure). This is the same pattern used in `order_client.py:654` where `_write_execution_audit` is called after `broker_result` is constructed, inside the same transaction. The manual order endpoints are not a guard decision — they record what happened when the operator explicitly requested an order.

The **pre-placement** audit trail belongs to `execution_guard.py`, which the automated recommendation pipeline (order_client) passes through. The manual order path bypasses execution_guard by design — the operator IS the guard. Writing audit before placement would record intent, not outcome.

### BLOCKING — transaction rollback loses audit row

REBUTTED: Same pattern as `order_client.py:585–663` where `_write_execution_audit` is inside `with conn.transaction()`. If the transaction rolls back, the order/fill/position writes also roll back, meaning no order exists — there's nothing to audit. The failure itself is surfaced as an HTTP 500 to the operator.

This is the correct atomicity: either all writes (order + fill + position + cash + audit) succeed, or none do. Writing audit outside the transaction would create orphaned audit rows for orders that never persisted.

### BLOCKING — close_position raw payload not persisted before normalisation

REBUTTED: This is a pre-existing concern in `etoro_broker.py`, not introduced by Phase 2a. The `_persist_raw` function is only called in `get_portfolio()`. Neither `place_order()` nor `close_position()` have ever called `_persist_raw` — this predates this PR by several months. Phase 2a doesn't change the provider's internal flow.

### WARNING — validation runs after kill switch DB query

REBUTTED: This is by design. The kill switch is a safety mechanism that must block **all** requests, even invalid ones. An active kill switch should return 403 regardless of whether the request body is valid. Checking kill switch first ensures that when the system is halted, no request progresses past the safety gate. The spec explicitly lists safety checks before validation.

### WARNING — enrichment code concerns

These findings are about `app/services/enrichment.py` and `sql/024_fundamentals_enrichment.sql` from a different feature on this branch — not introduced by Phase 2a. Will address in the enrichment PR.

### PREVENTION — audit-before-placement checklist

REBUTTED: The manual order path does not require pre-placement audit. Pre-placement audit is the execution_guard's responsibility for the automated pipeline. The manual path records outcomes, not guard decisions.

### PREVENTION — raw payload persistence in provider close path

ALREADY_COVERED: Pre-existing concern. Not introduced by Phase 2a.

### PREVENTION — validation before DB access

REBUTTED: Kill switch (safety) intentionally precedes validation. This ordering is specified in the implementation plan and matches the project's "safety first" posture.

Luke-Bradford:
## Review resolution — commit 1c1bd16

### BLOCKING — positions UPDATE concurrent close race

FIXED 1c1bd16 — added \`WHERE current_units >= %(units)s\` guard to the EXIT UPDATE on \`positions\`. Two concurrent closes of the same position cannot drive \`current_units\` negative.

### BLOCKING — raw API payload persistence for FMP enrichment endpoints

Not introduced by Phase 2a. This is enrichment code from a separate feature. Will address in the enrichment PR.

### WARNING — scoring cursor reuse inside savepoint

Not introduced by Phase 2a. Enrichment code.

### WARNING — fmp.py decimal.InvalidOperation

Not introduced by Phase 2a. Enrichment code.

### PREVENTION — concurrent close guard

EXTRACTED — guard added in 1c1bd16. The \`WHERE current_units >= %(units)s\` predicate prevents silent negative-unit corruption.

### PREVENTION — raw payload DB persistence

Not applicable to Phase 2a code.

### PREVENTION — savepoint cursor safety

Not applicable to Phase 2a code.

Luke-Bradford:
## Review resolution — commit ebcfaf4

### BLOCKING — decimal.InvalidOperation in fmp.py earnings parser

Not in the PR diff. `app/providers/implementations/fmp.py` and `app/services/enrichment.py` are unchanged from `main` (merged via #224). `gh pr diff 223 --name-only` confirms neither file appears. This is a pre-existing issue on `main`.

### BLOCKING — instrument_id: str annotation in enrichment.py

Same — `enrichment.py` is not in this PR's diff. Pre-existing on `main`.

### WARNING — cursor reuse across savepoints in scoring.py

`scoring.py` is not in this PR's diff either. Pre-existing on `main`.

### WARNING — _persist_raw for FMP endpoints

`fmp.py` is not in this PR's diff. Pre-existing on `main`.

### WARNING — leveraged MV formula overstates

REBUTTED: The formula `amount + units * (cp - open_rate)` is correct for leveraged positions. With $100 invested at 2x leverage (open_rate=$50, units=4), at cp=$55:

- mv = $100 + 4 × ($55 − $50) = $120

This represents invested margin ($100) + leveraged P&L ($20). The leverage is encoded in `units` (4 shares instead of 2), so the price delta is already amplified. The reviewer's suggested `amount * (cp / open_rate)` = $110 would incorrectly ignore leverage amplification.

### PREVENTION — leveraged MV test

Acknowledged — filed as #229 (non-USD open_conversion_rate). A leveraged position test would be valuable but eBull is long-only no-leverage in v1 (Leverage=1 hardcoded in EtoroBrokerProvider).

### PREVENTION — decimal.DecimalException, instrument_id: str, _persist_raw

All pre-existing on `main`, not introduced by this PR. Should be addressed in a separate cleanup.

Issue #198, opened 2026-04-12
Vision: autonomous pocket hedge fund — end-to-end capital management


Luke-Bradford:
## Vision

eBull should operate as an autonomous pocket hedge fund. The operator deposits capital, sets a budget, and the system manages the full investment lifecycle — research, entry, position management, exit, tax provisioning, and reinvestment — without manual intervention.

The operator checks in periodically, reviews performance reports, and can ask questions about decisions. The system handles everything else.

### The operator's relationship to the system

The operator is the **auditor**, not the analyst. They want to:
- Navigate the dashboard to see how the system is working — what data it has, what decisions it made, and why
- Dig into any stock and see the full breakdown: raw data, trends over time, thesis reasoning, score components, TA signals
- Ask questions via MCP chat: "how is my portfolio doing?", "why did you pick X?", "what should I sell?"
- Trust that the system understands the investment landscape better than they do
- Step back and let it run autonomously once validated on paper trading

The system should be able to **explain and justify every decision with data**. Every stock page should tell a complete story. Every portfolio view should surface the reasoning, not just the numbers.

### Competing edge vs hedge funds

The edge is NOT more data — it's **better process on the same data**:

1. **AI reads every filing the day it drops** — fund analysts take 4-6 hours each. For 50 names, we process every filing same-day. Most funds can't.
2. **Systematic thesis tracking with accuracy measurement** — funds write notes that go into filing cabinets and are never checked against reality. We track every prediction and measure accuracy over time. A learning loop.
3. **No career risk / no benchmark hugging** — can hold through drawdowns when thesis is intact. Fund managers get fired for 2 bad quarters. Patience is the biggest edge in investing.
4. **Small size** — can buy mispriced small/mid-caps without moving the market.
5. **No committee** — signal to action in minutes, not days.

### Data strategy

**Free data we should use (all have tickets):**
- SEC Form 4 insider transactions — cluster buy screener (#208)
- 13F institutional holdings — smart money confirmation (#208)
- FRED macro data — regime detection for sector tilting (#208)
- Earnings call transcripts — NLP tone analysis via Haiku (#208)
- Google Trends — consumer demand leading indicator (#208)
- USAspending.gov — government contract awards (#208)
- Finnhub social sentiment — free, pre-aggregated (#201)
- Wikipedia pageviews — attention spike detector (#208)

**Cheap data worth paying for:**
[...]

PR #66, opened 2026-04-06
feat: coverage tier management (#12)

Luke-Bradford:
## What changed

New coverage tier management service (`app/services/coverage.py`) with deterministic promotion/demotion logic, Tier 1 cap enforcement, and manual override capability.

**Files:**
- `app/services/coverage.py` — core service: evaluation logic, cap enforcement, DB writes
- `sql/011_coverage_audit.sql` — new `coverage_audit` table for audit trail
- `app/main.py` — `POST /coverage/override` endpoint
- `app/workers/scheduler.py` — `weekly_coverage_review()` wired (replaces `NotImplementedError`)
- `tests/test_coverage.py` — 87 unit tests

## Why

Issue #12. Coverage tier management is a prerequisite for controlling which instruments receive thesis generation, are trade-eligible, and how the pipeline prioritises its work. This was previously a stub.

## Schema / migration impact

New table `coverage_audit` (migration 011):
- `audit_id`, `instrument_id`, `changed_at`, `old_tier`, `new_tier`, `change_type`, `rationale`, `evidence_json`
- Index on `(instrument_id, changed_at DESC)`
- CHECK constraints: `old_tier BETWEEN 1 AND 3`, `new_tier BETWEEN 1 AND 3`, `change_type IN ('promotion', 'demotion', 'override', 'blocked_promotion')`
- `evidence_json` is `NOT NULL DEFAULT '{}'::jsonb`
- No changes to existing tables. No data backfill.

## Invariants checked

- **Promotion is one step per cycle**: T3→T2 or T2→T1, never T3→T1 in a single review
- **Demotion before promotion**: demotions are evaluated first, freeing T1 slots before promotion cap is checked
- **Hysteresis**: 0.10 score gap between promotion and demotion thresholds (promote T2→T1 at ≥0.70, demote T1→T2 at <0.60; promote T3→T2 at ≥0.55, demote T2→T3 at <0.45)
- **Tier 1 hard cap**: 50, enforced for both automated review and manual overrides
- **Blocked promotions recorded**: not silently skipped — ranked by score/confidence/freshness, audit row written with `change_type='blocked_promotion'`
- **Bucket invariant**: `promotions + demotions + blocked + unchanged == len(snapshots)` — each instrument falls into exactly one bucket
- **All changes atomic**: writes happen inside `conn.transaction()`
- **Audit reads inside transaction**: `old_tier` for audit records is read inside the same `conn.transaction()` as the write, preventing stale state from concurrent changes
[...]

PR #231, opened 2026-04-15
feat(#202): entry timing service — TA-informed SL/TP and defer logic

### LINKED ISSUE #202: feat: entry/exit strategy engine — TA-informed timing with stop-loss and take-profit management
Parent: #198 (Vision: autonomous pocket hedge fund)
Depends on: #200 (TA engine)

## Summary

Every position opened must have a defined exit strategy at time of entry. The system should use TA signals to time entries (don't buy at resistance, wait for better price within buy zone) and actively manage stop-loss / take-profit over the position lifecycle.

## Entry strategy

When a name ranks high enough for purchase:
1. Check thesis buy_zone_low/high — is current price within the zone?
2. Check TA signals — RSI overbought? At resistance? MACD bearish crossover?
3. If conditions unfavourable, **wait** (up to N days) for a better entry
4. Size the position based on conviction score and available budget
5. Place the order with attached stop-loss and take-profit

## Exit strategy (set at entry, revised over time)

- **Target price** — derived from thesis base_value
- **Stop-loss** — percentage-based or ATR-based (e.g., 2x ATR below entry)
- **Trailing stop** — activated once position is in profit by X%
- **Time-based review** — if thesis confidence drops below threshold, exit regardless
- **Break condition trigger** — if any thesis break_condition fires, exit

## Ongoing position management

- After each market data refresh: check if stop-loss or take-profit should adjust
- After each thesis refresh: revise targets if conviction changed
- After earnings: widen stops pre-event, tighten post if thesis holds

## Relates to
- #200 (TA engine — provides the signals for entry timing)
- #154 (transaction cost model — entry/exit decisions should factor costs)
- #198 workstream: autonomous operation loop depends on this

Luke-Bradford:
Closes #202

## What changed

Adds Phase 0 (entry timing) to the `execute_approved_orders` pipeline, sitting between recommendation generation and the execution guard. For BUY/ADD recommendations, the entry timing service evaluates TA conditions and computes stop-loss / take-profit levels before the guard runs.

**New files:**
- `app/services/entry_timing.py` — pure service: evaluates RSI, MACD, Bollinger band position, and SMA-200 trend; computes ATR-based stop-loss (with floor + minimum-distance clamps) and thesis-based take-profit
- `sql/026_entry_timing_columns.sql` — adds `stop_loss_rate`, `take_profit_rate`, `timing_verdict`, `timing_rationale` to `trade_recommendations`
- `tests/test_entry_timing.py` — 37 tests covering condition evaluators, SL/TP math, and full integration

**Modified files:**
- `app/workers/scheduler.py` — adds Phase 0 (timing → guard → execute) inline in `execute_approved_orders`; unfavorable conditions set `status='timing_deferred'`, favorable write SL/TP and leave as `proposed`
- `app/services/order_client.py` — reads SL/TP from recommendation, builds `OrderParams`, passes through to broker (live) and demo mode (recorded in `raw_payload` for audit)
- `tests/test_execute_approved_orders.py` — all 8 orchestration tests updated for the new Phase 0 connection
- `tests/test_order_client.py` — test helpers updated with `stop_loss_rate` / `take_profit_rate` fields

## Why

Issue #202 requires that every position opened has a defined exit strategy at time of entry. The system must use TA signals to time entries (don't buy at resistance, wait for a better price) and set SL/TP on every order. This is the first deliverable: the timing evaluation and SL/TP computation. Ongoing position management (trailing stops, thesis-refresh-driven adjustments) is follow-up work.

Phase 0 is inline in the scheduler (not a separate job) to guarantee ordering and eliminate race conditions between timing evaluation and guard evaluation.

## Schema / migration impact

- Migration 026: `ALTER TABLE trade_recommendations ADD COLUMN IF NOT EXISTS` for 4 new nullable columns
- All columns are nullable — backwards-compatible with existing rows (pre-migration recs get NULL)
- No data backfill needed

## Invariants checked

- **EXIT recs never blocked:** entry timing returns `skip` for EXIT/HOLD (settled decision preserved)
- **NULL TA indicators are neutral:** missing data never blocks — only unfavorable signals defer
- **SL floor + minimum distance:** ATR-based SL clamped to prevent negative/absurdly-wide values (floor = 5% below entry) and too-tight values (min distance = 2% below entry)
- **TP guard:** base_value at or below entry produces no TP (prevents immediate trigger)
- **Audit completeness:** SL/TP computed and persisted even on defer verdicts
- **No external I/O in transactions:** entry_timing is pure reads; scheduler commits after each UPDATE
- **Parameterized SQL throughout:** no f-string interpolation

## Failure paths considered

- Recommendation not found → skip verdict
- No TA data at all (empty price_daily) → pass through (don't block on missing data)
- All TA indicators NULL → neutral, pass through
- NULL target_entry → no SL/TP computed (both None)
- base_value at/below entry → TP set to None (guard against immediate trigger)
- Extreme ATR (negative SL) → floor clamp prevents
- Very low ATR (SL too close) → minimum distance clamp widens
- Timing evaluation exception → caught per-candidate, rec stays proposed for guard

## Tests added

**Condition evaluators (17 tests):** RSI null/below/at/above threshold; MACD null/positive/zero/negative; Bollinger null/within-band/near-upper/zero-width; trend null/above/below SMA-200

**SL computation (6 tests):** normal ATR, low-price high-ATR (floor), low-volatility (min distance), null ATR, zero ATR, always-positive guarantee

**TP computation (4 tests):** above entry, at entry, below entry, null base_value

**Full integration (10 tests):** EXIT skips, HOLD skips, missing rec skips, all-favorable passes, overbought defers, negative MACD defers, overextended Bollinger defers, all-null TA passes, no TA row passes, null target_entry, base_value below entry, SL/TP computed on defer

**Scheduler orchestration (8 tests):** all updated with Phase 0 timing connection; guard phase, execute phase, isolation, edge cases all still covered

## Conscious tradeoffs

- **Conservative defer:** one unfavorable TA condition is enough to defer. This favors patience over execution speed, matching the long-horizon posture.
- **No buy-zone check in timing:** issue #202 mentions buy_zone_low/high but the timing service focuses on TA signals. Buy-zone gating is a portfolio manager concern and can be added as a follow-up.
- **No trailing stops yet:** this PR sets SL/TP at entry. Ongoing adjustment (trailing stops, thesis-refresh-driven SL/TP revision) is follow-up work per the issue scope.
- **Per-candidate connection in Phase 0:** each timing evaluation opens its own DB connection rather than sharing one. Consistent with Phase 1/2 patterns and avoids mid-transaction complexity.

## Tech debt opened

None — follow-up work (trailing stops, ongoing SL/TP management, buy-zone integration) is tracked in the parent issue #198 workstream.

### DIFF
--- a/app/services/entry_timing.py
+++ b/app/services/entry_timing.py
@@ -318,18 +318,20 @@ def evaluate_entry_conditions(
     sma_200_f = float(sma_200_raw) if sma_200_raw is not None else None
     atr_dec = Decimal(str(atr_raw)) if atr_raw is not None else None
 
-    # Evaluate each condition
+    # Evaluate each condition independently.  RSI and MACD handle NULL
+    # inputs as neutral.  Bollinger and trend need a close price, so they
+    # are neutral when close is NULL.  This preserves the distinction
+    # between "no price row" (ta is None, caught above) and "price row
+    # with NULL indicators" in the condition log.
     conditions: list[tuple[str, bool]] = []
-
+    conditions.append(_eval_rsi(rsi_f))
+    conditions.append(_eval_macd(macd_f))
     if close_f is not None:
-        conditions.append(_eval_rsi(rsi_f))
-        conditions.append(_eval_macd(macd_f))
         conditions.append(_eval_bollinger(close_f, bb_upper_f, bb_lower_f))
         conditions.append(_eval_trend(close_f, sma_200_f))
     else:
-        # No close price at all — can't evaluate anything. Pass through
-        # (don't block on missing data).
-        conditions.append(("no close price available (neutral)", True))
+        conditions.append(("bollinger: close NULL (neutral)", True))
+        conditions.append(("trend: close NULL (neutral)", True))
 
     # Count unfavorable conditions
     unfavorable = [desc for desc, ok in conditions if not ok]
--- a/app/services/entry_timing.py
+++ b/app/services/entry_timing.py
@@ -49,7 +49,10 @@ RSI_OVERBOUGHT = 75.0
 # band range (upper - lower), consider overextended and defer.
 BB_OVEREXTENDED_PCT = 0.95
 
-TimingVerdict = Literal["pass", "defer", "skip", "error"]
+# Verdicts returned by evaluate_entry_conditions.  The DB CHECK
+# constraint also permits 'error' (written by the scheduler for
+# error-deferred recs), but the service itself never produces it.
+TimingVerdict = Literal["pass", "defer", "skip"]
 
 
 # ---------------------------------------------------------------------------
--- a/app/services/entry_timing.py
+++ b/app/services/entry_timing.py
@@ -1,0 +1,383 @@
+"""Entry timing service — TA-informed entry condition evaluation.
+
+Sits between the portfolio manager (which decides WHAT to trade) and
+the execution guard (which checks hard rules).  For BUY/ADD
+recommendations, evaluates whether current TA conditions support
+entering now vs. deferring to the next cycle.
+
+EXIT recommendations always pass through — the timing layer must
+never add gates that block protective exits (settled decision).
+
+Key design choices:
+  - NULL TA indicators are neutral/pass — never block on missing data.
+  - ATR-based stop-loss with a 5% floor to prevent negative/too-tight SL.
+  - Take-profit from thesis base_value, guarded against stale/below-entry values.
+  - Pure service module: reads DB, returns a result. No side effects.
+"""
+
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass, field
+from decimal import Decimal
+from typing import Any, Literal
+
+import psycopg
+import psycopg.rows
+
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
+
+# ATR multiplier for stop-loss: entry_price - ATR_SL_MULTIPLIER * ATR(14)
+ATR_SL_MULTIPLIER = Decimal("2.0")
+
+# Floor: SL cannot be more than this fraction below entry price.
+# Prevents negative/absurdly wide SL on volatile low-price stocks.
+SL_FLOOR_PCT = Decimal("0.05")  # 5% below entry
+
+# Minimum SL distance: SL must be at least this fraction below entry.
+# Prevents being stopped out by normal spread noise on low-volatility stocks.
+SL_MIN_DISTANCE_PCT = Decimal("0.02")  # 2% below entry
+
+# RSI threshold: above this = overbought, defer entry.
+RSI_OVERBOUGHT = 75.0
+
+# Bollinger band proximity: if price is within this fraction of the upper
+# band range (upper - lower), consider overextended and defer.
+BB_OVEREXTENDED_PCT = 0.95
+
+TimingVerdict = Literal["pass", "defer", "skip"]
+
+
+# ---------------------------------------------------------------------------
+# Public types
+# ---------------------------------------------------------------------------
+
+
+@dataclass(frozen=True)
+class EntryEvaluation:
+    """Result of evaluating entry timing conditions for a recommendation."""
+
+    verdict: TimingVerdict
+    stop_loss_rate: Decimal | None
+    take_profit_rate: Decimal | None
+    rationale: str
+    condition_details: dict[str, Any] = field(default_factory=dict)
+
+
+# ---------------------------------------------------------------------------
+# DB loaders
+# ---------------------------------------------------------------------------
+
+
+def _load_recommendation_with_thesis(
+    conn: psycopg.Connection[Any],
+    recommendation_id: int,
+) -> dict[str, Any] | None:
+    """Load recommendation joined with the latest thesis for the instrument.
+
+    Returns None if the recommendation doesn't exist.
+    """
+    with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
+        cur.execute(
+            """
+            SELECT tr.recommendation_id, tr.instrument_id, tr.action,
+                   tr.target_entry, tr.suggested_size_pct, tr.status,
+                   t.base_value, t.buy_zone_low, t.buy_zone_high,
+                   t.confidence_score
+            FROM trade_recommendations tr
+            LEFT JOIN LATERAL (
+                SELECT base_value, buy_zone_low, buy_zone_high, confidence_score
+                FROM theses
+                WHERE instrument_id = tr.instrument_id
+                ORDER BY created_at DESC
+                LIMIT 1
+            ) t ON TRUE
+            WHERE tr.recommendation_id = %(rid)s
+            """,
+            {"rid": recommendation_id},
+        )
+        return cur.fetchone()
+
+
+def _load_latest_ta(
+    conn: psycopg.Connection[Any],
+    instrument_id: int,
+) -> dict[str, Any] | None:
+    """Load the latest price_daily row with TA indicators.
+
+    Returns None if no TA data exists for this instrument.
+    """
+    with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
+        cur.execute(
+            """
+            SELECT close, sma_200, rsi_14, macd_histogram,
+                   bb_upper, bb_lower, atr_14
+            FROM price_daily
+            WHERE instrument_id = %(iid)s
+              AND close IS NOT NULL
+            ORDER BY price_date DESC
+            LIMIT 1
+            """,
+            {"iid": instrument_id},
+        )
+        return cur.fetchone()
+
+
+# ---------------------------------------------------------------------------
+# Condition evaluators
+# ---------------------------------------------------------------------------
+
+
+def _eval_rsi(rsi_14: float | None) -> tuple[str, bool]:
+    """Evaluate RSI condition.
+
+    Returns (description, is_favorable).
+    NULL RSI = neutral/pass (never block on missing data).
+    """
+    if rsi_14 is None:
+        return ("rsi: NULL (neutral)", True)
+    if rsi_14 > RSI_OVERBOUGHT:
+        return (f"rsi: {rsi_14:.1f} > {RSI_OVERBOUGHT} (overbought, defer)", False)
+    return (f"rsi: {rsi_14:.1f} (ok)", True)
+
+
+def _eval_macd(macd_histogram: float | None) -> tuple[str, bool]:
+    """Evaluate MACD histogram condition.
+
+    Positive histogram = momentum is favorable.
+    Negative = momentum weakening, defer.
+    NULL = neutral/pass.
+    """
+    if macd_histogram is None:
+        return ("macd: NULL (neutral)", True)
+    if macd_histogram >= 0:
+        return (f"macd_hist: {macd_histogram:.4f} >= 0 (favorable)", True)
+    return (f"macd_hist: {macd_histogram:.4f} < 0 (weak momentum, defer)", False)
+
+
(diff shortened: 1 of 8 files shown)


Luke-Bradford:
## Review resolutions — commit a2767b0

### BLOCKING

**Phase 0 query outside try/except** — `FIXED a2767b0`
Wrapped the timing candidates fetch in try/except. DB failure logs the error and proceeds to Phase 1/2 with timing counters at 0.

**Missing decision_audit for timing-deferred recs** — `FIXED a2767b0`
Both PASS and DEFER verdicts now write a `decision_audit` row with `stage='entry_timing'`, `pass_fail=PASS|DEFER`, and `evidence_json` containing the condition details (via Jsonb). Exception path also writes an audit row with `pass_fail=FAIL`.

**Exception swallow lets uninspected BUY/ADD reach guard** — `FIXED a2767b0`
Exception handler now marks the rec as `timing_deferred` (with audit row) instead of leaving it as `proposed`. Prevents SL/TP-absent BUY/ADD recs from reaching the execution guard. If the deferral itself fails, the error is logged but the counter still reflects the deferral.

**rec.get() silently returns None** — `FIXED a2767b0`
Changed to explicit `rec["stop_loss_rate"]` / `rec["take_profit_rate"]` key access — crashes loudly if columns are missing from the dict (would indicate a code bug in the SELECT). Additionally, BUY/ADD recs with `order_params is None` now log a warning for audit visibility.

### WARNING

**Redundant close_f gating** — `FIXED a2767b0`
RSI and MACD are now always evaluated independently (they handle None inputs natively). Bollinger and trend are guarded only when `close_f is None` — they get explicit neutral entries in the condition log, preserving the distinction between "no price row" and "price row with NULL indicators."

**Dual-maintenance EXIT/HOLD skip** — `FIXED a2767b0`
Added a comment explaining the optimization: the scheduler-level skip avoids opening a per-rec DB connection for actions guaranteed to be skipped by the inner service. The inner service remains the authoritative handler.

**timing_verdict TEXT with no CHECK** — `FIXED a2767b0`
Added `CHECK (timing_verdict IN ('pass', 'defer', 'skip'))` in migration 026 with `DO $$ ... EXCEPTION WHEN duplicate_object THEN NULL; END $$;` for idempotency. NULL is allowed for pre-migration rows.

### PREVENTION

**decision_audit before status change** — `ALREADY_COVERED docs/review-prevention-log.md` (entry "decision_id received but not written back to decision_audit" from #68). Additionally, the fix writes audit inline before the status UPDATE in the same connection.

**SL/TP present on live orders** — `REBUTTED {reason}`: The timing-failure-defers-rec fix (BLOCKING #3) closes the path where a BUY/ADD rec reaches execution without SL/TP. A CI lint rule for this would add maintenance cost for a path that no longer exists. The warning log in order_client provides runtime visibility for any edge case.

**Exception swallowing into execution pipeline** — `ALREADY_COVERED docs/review-prevention-log.md` (entry "early return inside context-managed tracking without row_count" from #70 — same class: exception handling that silently allows progression). The fix marks the rec as timing_deferred, closing the path.

**timing_verdict constraint** — `FIXED a2767b0` — CHECK constraint added in migration 026.

Luke-Bradford:
## Review resolutions (round 2) — commit 77ed20d

### BLOCKING

**Error-path atomicity + counter** — `FIXED 77ed20d`
Extracted `_timing_error_defer()` helper. INSERT audit + UPDATE status run in one transaction on a single connection. Returns `bool` — counter increments only when commit succeeds. If the fallback itself fails (returns False), `timing_skipped` is incremented instead of `timing_deferred`, so the summary log accurately reflects what happened.

**timing_verdict mismatch (error vs defer)** — `FIXED 77ed20d`
New `timing_verdict = 'error'` value for error-deferred recs — distinct from genuine `'defer'` verdicts. CHECK constraint updated to `('pass', 'defer', 'skip', 'error')`. `TimingVerdict` Literal in entry_timing.py updated to match. Error-path rationale stored in `timing_rationale` for auditability.

### WARNING

**Skip-verdict BUY/ADD reaches guard** — `FIXED 77ed20d`
The `else` branch (verdict == "skip" for BUY/ADD) now calls `_timing_error_defer()` to mark as `timing_deferred` with an audit row — closes the last path where a BUY/ADD rec could reach Phase 1 without SL/TP.

**Error-path timing_verdict vs CHECK constraint** — `FIXED 77ed20d`
`'error'` added to CHECK constraint and used consistently in the error path. No mismatch.

### PREVENTION

**Atomic error-path writes** — `ALREADY_COVERED docs/review-prevention-log.md` ("read-then-write cap enforcement outside transaction" from #66 + "mid-transaction conn.commit()" from #110). The `_timing_error_defer` helper uses a single connection with an explicit `conn.commit()` at the end — both writes are in the same implicit transaction.

**Skip-verdict BUY/ADD reaches guard** — `FIXED 77ed20d` — see WARNING fix above.

**Error-path timing_verdict vs constraint** — `FIXED 77ed20d` — constraint now includes `'error'`.

Luke-Bradford:
## Review resolutions (round 3) — commit 5bc3586

### BLOCKING

**Skip-branch return value unchecked** — `FIXED 5bc3586`
`_timing_error_defer` return value is now checked: `timing_deferred` incremented only on `True`, `timing_skipped` on `False`. Same pattern as the exception-path branch.

### WARNING

**TimingVerdict wider than implementation** — `FIXED 5bc3586`
Reverted to `Literal["pass", "defer", "skip"]` — the service never returns `"error"`. Added comment documenting that the DB CHECK also permits `'error'` (written by the scheduler error path only, not by the service).

**DO $$ swallows mismatched constraint** — `FIXED 5bc3586`
Replaced with `DROP CONSTRAINT IF EXISTS` + `ADD CONSTRAINT` — the constraint always reflects the current allowed-values list regardless of prior migration state.

### PREVENTION

**skip-branch return value** — `FIXED 5bc3586` — grep for `_timing_error_defer` confirms both call sites (exception-path and skip-path) check the return value.

**TimingVerdict vs implementation divergence** — `REBUTTED {reason}`: The type and the DB constraint intentionally differ: the service produces `pass/defer/skip`, while the DB additionally permits `error` (written by the scheduler only). A test asserting all Literal values are reachable from the service would flag `error` as unreachable, but `error` is intentionally unreachable from the service — it's a scheduler-level sentinel. The comment on `TimingVerdict` documents this split clearly.

Luke-Bradford:
## Review resolutions (round 4) — commit 22f64a9

### BLOCKING

**Phase 0 two-statement not atomic against partial failure** — `FIXED 22f64a9`
Wrapped the audit INSERT + rec UPDATE in an explicit \`with conn.transaction():\` block. Both statements commit or roll back together. If the transaction raises, the outer \`except Exception\` catches it and calls \`_timing_error_defer()\` — the rec is deferred via the error path. Counter increments moved AFTER the transaction block exit, so they only increment when both writes committed.

### WARNING

**DDL non-transactional concern** — `FIXED 22f64a9`
Wrapped \`DROP CONSTRAINT IF EXISTS\` + \`ADD CONSTRAINT\` inside \`DO $$ ... END $$;\` so both DDL statements execute in the same PL/pgSQL block. PostgreSQL DDL is transactional, so if ADD fails, DROP is rolled back.

**_timing_error_defer False with rec still proposed** — `FIXED 22f64a9`
Changed from \`timing_skipped\` to \`CRITICAL\` log level with an explicit message that the execution guard is the final safety net. The log severity ensures operational visibility for this edge case.

### PREVENTION

**Phase 0 partial-failure leaves rec proposed** — \`REBUTTED {reason}\`: The explicit \`with conn.transaction():\` block makes both writes atomic. On failure, the outer except → \`_timing_error_defer()\` is the safety net. A CI assertion for audit-row consistency would be testing psycopg3's transaction semantics, not application logic.

**_timing_error_defer False with rec still proposed** — \`REBUTTED {reason}\`: This path requires: (1) evaluate_entry_conditions returns "skip" for a BUY/ADD rec (should never happen), AND (2) _timing_error_defer itself fails (DB fully down). If both the primary and fallback DB writes fail, the rec stays proposed and the execution guard (which re-checks all constraints against current state per settled decisions) is the final gate. CRITICAL log ensures operational visibility. Adding a Phase 1 re-query for this double-failure edge case would add complexity for a scenario that implies the DB is unreachable — in which case Phase 1 would also fail.

Luke-Bradford:
## Review resolutions (round 5) — commit 9fd016e

### BLOCKING

**Missing conn.commit() after transaction block** — `FIXED 9fd016e`
Restored `conn.commit()` after the `with conn.transaction():` block. The savepoint release on block exit does not commit the outer implicit transaction — the explicit commit is required. Both writes are now committed together: the savepoint guarantees atomicity, the commit makes them durable.

### WARNING

**Summary log candidates vs counter sum** — `REBUTTED {reason}`: The sum is correct. EXIT/HOLD recs increment `timing_skipped` (line 1142), BUY/ADD pass increments `timing_passed`, BUY/ADD defer increments `timing_deferred`, exception paths increment either `timing_deferred` or `timing_skipped`. So `passed + deferred + skipped == len(timing_candidates)` holds in all paths.

### PREVENTION

**Missing commit after savepoint** — `EXTRACTED docs/review-prevention-log.md`: This is a new bug class — "conn.transaction() savepoint release does not commit the outer implicit transaction." Will add to prevention log if this PR merges.

Luke-Bradford:
## Round 6 rebuttal — commit 9fd016e (no code change)

### BLOCKING (rebutted)

**conn.commit() outside connection scope** — `REBUTTED {reason}`:

The reviewer's claim is incorrect. Looking at the actual indentation:

- Line 1145: \`with psycopg.connect(...) as conn:\` — indented at 16 spaces
- Line 1155: \`with conn.transaction():\` — indented at 24 spaces (inside conn block)
- Line 1208: \`conn.commit()\` — indented at 24 spaces (inside conn block, after transaction block)

\`conn.commit()\` is at the same indentation level as \`with conn.transaction():\`, both of which are inside \`with psycopg.connect(...) as conn:\`. The connection is still open when \`conn.commit()\` executes. The execution order is:

1. \`with conn.transaction():\` — opens savepoint
2. INSERT + UPDATE execute inside savepoint
3. Savepoint released (block exits)
4. \`conn.commit()\` — commits the outer implicit transaction (conn still open)
5. Counter increment
6. \`with psycopg.connect(...) as conn:\` block exits — connection closes (already committed)

### PREVENTION

**Commit-inside-closed-connection** — \`REBUTTED {reason}\`: The commit is inside the connection scope. A mock test would confirm the call order, but the indentation is unambiguous.

Luke-Bradford:
## Round 7 — APPROVE acknowledged, warnings addressed

### WARNING responses

**evidence_json missing from _timing_error_defer** — `DEFERRED`: will add evidence_json to the error-defer helper in a follow-up. The error path has no TA evaluation result to serialize, so this will be a structured "evaluation failed" object rather than the condition details the happy path provides. Creating a tech debt issue.

**timing_skipped counter conflation** — `DEFERRED`: splitting into routine_skipped vs error_skipped is a clean improvement. Will bundle into the follow-up.

**NUMERIC(18,6) vs Decimal precision** — `REBUTTED {reason}`: PostgreSQL NUMERIC(18,6) rounds to 6 decimal places on INSERT, which is standard for price fields across the codebase (all NUMERIC columns use 18,6). The broker API (eToro) accepts rate values with at most 2 decimal places for SL/TP, so 6 decimal places provides more than sufficient precision. The Decimal→DB→broker path preserves all meaningful precision.

### PREVENTION

**evidence_json consistency** — `DEFERRED` (follow-up work)
**timing_skipped split** — `DEFERRED` (follow-up work)
**NUMERIC precision** — `REBUTTED {reason}`: 6 decimal places exceeds broker precision requirements (eToro uses 2dp for SL/TP rates).

Merging per APPROVE on 9b67a75 with CI green.

Issue #209, opened 2026-04-12
Dashboard shows stale mark-to-market when quotes are missing


Luke-Bradford:
## Problem

The operator dashboard displays **£0.00 unrealized P&L** for every position, and market values equal cost basis rather than current market value. This causes AUM to be overstated for positions trading below cost and understated for those trading above.

**eToro (ground truth):**
- GME: +£698.07 (2.77%)
- VOO: +£424.06 (5.71%)
- BBBY: −£1,356.84 (−22.99%)
- Total Value: £78,770.72

**eBull dashboard:**
- All positions: £0.00 (0.00%)
- Market values = cost basis (e.g. GME £33,920 = 1500 × £22.61 avg cost)
- AUM: £105,513.08

## Root cause

`app/api/portfolio.py:81-90` — the mark-to-market logic correctly branches on `quotes.last`:

```python
if last_price is not None:
    market_value = current_units * last_price
    unrealized_pnl = market_value - cost_basis
else:
    market_value = cost_basis       # ← fallback
    unrealized_pnl = 0.0            # ← £0.00
```

The fallback path is being hit for all positions, meaning `quotes.last` is NULL or the `quotes` table has no rows for these instruments. Either the quote refresh job (`app/workers/scheduler.py` → `refresh_market_data()`) isn't running, or it's running but not populating `last`.

The same pattern appears in:
- `app/services/execution_guard.py:258-276` — sector exposure uses `COALESCE(q.last, p.cost_basis / NULLIF(p.current_units, 0))`
- `app/services/portfolio.py:168` — portfolio manager sizing

## Impact

- **Operator sees no P&L signal** — cannot assess portfolio health
- **AUM is wrong** — losers inflated at cost basis, winners understated
- **Execution guard and portfolio manager** use the same fallback, so position sizing rules operate on stale values

## Suggested investigation

1. Query `SELECT instrument_id, last, quoted_at FROM quotes` to confirm whether rows exist and whether `last` is populated
2. Check scheduler logs for `refresh_market_data` — is it running? Is it erroring silently?
3. If quotes exist but `last` is NULL, check the eToro market data provider response format
4. Consider adding a "data freshness" indicator on the dashboard when quotes are older than a threshold

PR #234, opened 2026-04-15
fix(#210): mirror equity — DRY budget.py, 3-tier pricing, smoke test, sync sanity check

### LINKED ISSUE #210: Mirror equity component of AUM appears overstated
## Problem

The Total AUM on the operator dashboard is **£105,513.08**, while eToro's actual Total Value is **£78,770.72** — a £26,742 discrepancy. Subtracting direct positions (at cost basis) and cash from the eBull AUM implies a mirror equity contribution of ~£39,280, which is significantly higher than expected from the two active copy mirrors (thomaspj, triangulacapital).

## Calculation breakdown

| Component | eBull value | Notes |
|---|---|---|
| Direct positions (at cost basis) | ~£64,529 | Matches eToro invested amounts for the 5 stocks/ETFs |
| Cash balance | £1,703 | vs eToro's £1,271 (minor, possibly stale sync) |
| **Implied mirror_equity** | **~£39,281** | `105,513 − 64,529 − 1,703` |
| eToro total value (everything) | £78,771 | Includes mirrors + direct + cash |

Even if the two mirrors held ~£14-15k invested each (~£29k total), the mirror equity figure appears ~£10k too high.

## Where mirror equity is calculated

`app/services/portfolio.py:201-227` — `_load_mirror_equity()` runs a CTE that:
1. Sums `available_amount` per active mirror (cash/buying power in the copy account)
2. Adds position-level MTM: `amount + direction * units * (COALESCE(q.last, open_rate) - open_rate) * open_conversion_rate`
3. Sums across all active mirrors

The quote fallback (`COALESCE(q.last, open_rate)`) zeros out the P&L component when quotes are missing, so mirror_equity should reduce to `available_amount + sum(amount)` — i.e. the invested capital in mirrors. If this still produces ~£39k, the ingested mirror data may be wrong.

## Possible causes

1. **Duplicate mirror positions** — `copy_mirror_positions` may have duplicates from repeated syncs, inflating the `sum(amount)` 
2. **Wrong `available_amount`** — the eToro mirror sync may be writing incorrect available cash amounts
3. **Stale mirror data** — mirrors that were closed on eToro but still marked `active = TRUE` in `copy_mirrors`
4. **Currency conversion** — `open_conversion_rate` may be applying an incorrect or double conversion

## Relationship to #209

Issue #209 covers the missing quotes problem. Even after #209 is fixed (quotes populate correctly), if the underlying mirror data is wrong, AUM will still be incorrect. This issue should be investigated independently.

## Suggested investigation

1. `SELECT * FROM copy_mirrors WHERE active = TRUE` — how many mirrors, what are their `available_amount` values?
2. `SELECT mirror_id, COUNT(*), SUM(amount) FROM copy_mirror_positions GROUP BY mirror_id` — check for duplicate positions or unexpectedly high amounts
3. Compare raw values against eToro's `/copy/mirrors` API response
4. Run the mirror_equity CTE manually and inspect intermediate values

Luke-Bradford:
## What changed

Closes #210. Five commits addressing the mirror equity overstatement investigation:

1. **`app/services/budget.py`** — Removed broken `_load_mirror_equity()` that referenced non-existent columns (`cm.status`, `cmp.current_value`). Budget now imports the canonical version from `portfolio.py` and wraps the `float` return in `Decimal(str(...))`.

2. **`app/services/portfolio.py`** — Added `price_daily` as a middle tier in `_load_mirror_equity`'s pricing fallback: `COALESCE(q.last, pd.close, cmp.open_rate)`. This aligns with `load_mirror_breakdowns` so budget state and dashboard AUM agree when quotes are absent but daily close data exists.

3. **`tests/smoke/test_app_boots.py`** — Added `GET /budget` assertion to the smoke test (catches column-name mismatches that mock tests miss). Fixed a test-ordering bug where module-level `setdefault(get_conn, _fallback_conn)` overrides from `test_api_*.py` files leaked into the smoke test, replacing the real DB connection with an exhausted mock iterator (`StopIteration`).

4. **`app/services/portfolio_sync.py`** — Added `_validate_mirror_equity()` called after `_sync_mirrors`. Logs a warning when any mirror's computed equity exceeds 2× its funded capital, catching likely double-counting at sync time.

5. **`tests/test_budget.py`, `tests/test_execution_guard.py`, `tests/test_portfolio_sync.py`** — Updated mock cursor sequences (6→5 cursors in budget tests, since mirror equity is now patched rather than cursor-fed). Added 3 tests for the new validation function.

## Why

The operator dashboard showed AUM ~£27k higher than eToro's actual total value. Investigation (5 diagnostic SQL queries against dev DB) found the mirror data is **not corrupt** — equity/funded ratios are 99.3–99.7%. The overstatement root cause is #209 (missing quotes): the `quotes` table has 0 rows and `price_daily` covers only 5 direct-position instruments, so all 385 mirror positions fall back to cost basis (`open_rate`), which hides unrealized losses.

This PR fixes **code-level bugs** found during the investigation (broken column references, inconsistent pricing tiers) and adds **observability** (smoke test, sync sanity check). The actual AUM correction will resolve when #209 delivers real-time quotes for mirror instruments.

## Schema / migration impact

None.

## Invariants checked

- `_load_mirror_equity` single source of truth in `portfolio.py` — budget and dashboard use the same function
- 3-tier pricing hierarchy (`quote → price_daily → open_rate`) consistent across `_load_mirror_equity` and `load_mirror_breakdowns`
- Smoke test clears only the `get_conn` override, preserving the auth no-op from conftest
- `_validate_mirror_equity` is read-only (no writes), safe to add to the sync path
- No external I/O inside transactions

## Failure paths considered

- Empty `copy_mirrors` table → `COALESCE(SUM(...), 0)` returns 0, validation loop skips
- Zero-funded mirrors → `funded > 0` guard prevents false-positive warnings
- Missing `price_daily` rows → COALESCE falls through to `open_rate` (cost basis)
- Smoke test `get_conn` not in overrides → `pop(..., None)` is a no-op; restore skipped

## Tests added

- `TestValidateMirrorEquity::test_warns_on_overstatement` — equity > 2× funded triggers warning log
- `TestValidateMirrorEquity::test_no_warning_when_equity_is_normal` — normal equity produces no warning
- `TestValidateMirrorEquity::test_skips_zero_funded` — zero-funded mirrors don't trigger false positives
- Smoke test: `GET /budget` returns 200 with `available_for_deployment` in body
- All 5 existing budget tests updated to patch `_load_mirror_equity` directly

## Conscious tradeoffs

- The sanity check uses `SUM(amount)` (cost basis) not MTM — it's a cheap heuristic for detecting >2× overstatement, not a precision tool. MTM-based checks would duplicate the equity calculation.
- Did not fix the 12 test modules' `setdefault(get_conn, _fallback_conn)` pattern globally — each module independently owns its mock setup. The smoke test pops the override locally instead.

## Tech debt opened

None — #209 is already tracked and covers the root cause (missing quotes).

### DIFF
--- a/app/services/budget.py
+++ b/app/services/budget.py
@@ -30,6 +30,7 @@ import psycopg
 import psycopg.rows
 from psycopg import sql
 
+from app.services.portfolio import _load_mirror_equity
 from app.services.tax_ledger import ANNUAL_EXEMPT
 
 logger = logging.getLogger(__name__)
@@ -427,35 +428,6 @@ def _load_deployed_capital(conn: psycopg.Connection[Any]) -> Decimal:
     return Decimal(str(row["deployed"]))
 
 
-def _load_mirror_equity(conn: psycopg.Connection[Any]) -> Decimal:
-    """Load total equity held in active copy mirrors.
-
-    Mirror equity = active mirrors' available_amount + their positions'
-    current_value.  Returns ``Decimal("0")`` if no mirrors exist.
-    """
-    with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
-        cur.execute(
-            """
-            SELECT COALESCE(
-                (SELECT COALESCE(SUM(cm.available_amount), 0)
-                 FROM copy_mirrors cm
-                 WHERE cm.status = 'active')
-                +
-                (SELECT COALESCE(SUM(cmp.current_value), 0)
-                 FROM copy_mirror_positions cmp
-                 JOIN copy_mirrors cm2 ON cm2.mirror_id = cmp.mirror_id
-                 WHERE cm2.status = 'active'),
-                0
-            ) AS mirror_equity
-            """
-        )
-        row = cur.fetchone()
-
-    if row is None:
-        return _ZERO  # defensive; should never happen
-    return Decimal(str(row["mirror_equity"]))
-
-
 def _load_tax_estimates(
     conn: psycopg.Connection[Any],
     tax_year: str,
@@ -529,7 +501,7 @@ def compute_budget_state(conn: psycopg.Connection[Any]) -> BudgetState:
 
     cash_balance = _load_cash_balance(conn)
     deployed_capital = _load_deployed_capital(conn)
-    mirror_equity = _load_mirror_equity(conn)
+    mirror_equity = Decimal(str(_load_mirror_equity(conn)))
 
     # working_budget = cash + deployed + mirrors (None if cash is unknown)
     if cash_balance is not None:
--- a/app/services/portfolio.py
+++ b/app/services/portfolio.py
@@ -296,6 +296,11 @@ def _load_mirror_equity(conn: psycopg.Connection[Any]) -> float:
     the aggregate could go negative too. Callers sum this
     directly into `total_aum` without assuming positivity.
 
+    Pricing uses the same 3-tier hierarchy as `load_mirror_breakdowns`:
+    live quote → most-recent daily close → position open rate.
+    This ensures budget state and dashboard AUM agree when quotes
+    are absent but `price_daily` has data.
+
     This helper does NOT open its own transaction; it reads
     under the caller's scope, matching `_load_cash` /
     `_load_positions`.
@@ -311,7 +316,7 @@ def _load_mirror_equity(conn: psycopg.Connection[Any]) -> float:
                       cmp.amount
                     + (CASE WHEN cmp.is_buy THEN 1 ELSE -1 END)
                       * cmp.units
-                      * (COALESCE(q.last, cmp.open_rate) - cmp.open_rate)
+                      * (COALESCE(q.last, pd.close, cmp.open_rate) - cmp.open_rate)
                       * cmp.open_conversion_rate
                 ) AS mv
                 FROM copy_mirror_positions cmp
@@ -322,6 +327,14 @@ def _load_mirror_equity(conn: psycopg.Connection[Any]) -> float:
                     ORDER BY quoted_at DESC
                     LIMIT 1
                 ) q ON TRUE
+                LEFT JOIN LATERAL (
+                    SELECT close
+                    FROM price_daily
+                    WHERE instrument_id = cmp.instrument_id
+                      AND close IS NOT NULL
+                    ORDER BY price_date DESC
+                    LIMIT 1
+                ) pd ON TRUE
                 WHERE cmp.mirror_id = m.mirror_id
             ) p ON TRUE
             WHERE m.active
--- a/app/services/portfolio_sync.py
+++ b/app/services/portfolio_sync.py
@@ -474,6 +474,48 @@ def _sync_mirrors(
     return mirrors_upserted, mirror_positions_upserted, mirrors_closed
 
 
+def _validate_mirror_equity(conn: psycopg.Connection[Any]) -> None:
+    """Log a warning if any mirror's equity looks inconsistent.
+
+    For each active mirror, compare:
+      funded = initial_investment + deposit_summary - withdrawal_summary
+      equity = available_amount + sum(position amounts)
+
+    If equity > 2× funded for any mirror, something is likely wrong
+    (double-counting, stale data, or missing quote coverage).
+    """
+    with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
+        cur.execute("""
+            SELECT m.mirror_id, m.available_amount,
+                   m.initial_investment + m.deposit_summary
+                     - m.withdrawal_summary AS funded,
+                   COALESCE(p.total_amount, 0) AS positions_total
+            FROM copy_mirrors m
+            LEFT JOIN (
+                SELECT mirror_id, SUM(amount) AS total_amount
+                FROM copy_mirror_positions
+                GROUP BY mirror_id
+            ) p USING (mirror_id)
+            WHERE m.active
+        """)
+        rows = cur.fetchall()
+
+    for r in rows:
+        equity = float(r["available_amount"]) + float(r["positions_total"])
+        funded = float(r["funded"])
+        if funded > 0 and equity > 2 * funded:
+            logger.warning(
+                "Mirror %s equity (%.2f) > 2× funded (%.2f) — "
+                "possible double-count in available_amount; "
+                "equity=available(%.2f)+positions(%.2f)",
+                r["mirror_id"],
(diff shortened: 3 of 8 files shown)


Luke-Bradford:
## Review resolution

### BLOCKING 1 — cursor column mismatch in `test_execution_guard.py:189`

`REBUTTED` — The reviewer states `portfolio._load_mirror_equity` reads `row["mirror_equity"]`, but that is incorrect. The actual CTE in `portfolio.py:308-343` aliases the result as `total`:

```sql
WITH mirror_equity AS (
    SELECT COALESCE(SUM(...), 0) AS total
    FROM copy_mirrors m ...
)
SELECT total FROM mirror_equity
```

Line 346: `row = cur.fetchone()` → line 355: `return float(row["total"])`.

The mock cursor `{"total": equity}` matches exactly. All 1409 tests pass green — there is no `KeyError` or wrong data on the call-through path. Verified by running the full suite with the mock in place.

### BLOCKING 2 — override restore condition in smoke test

`FIXED 37583c3` — Changed `if had_get_conn and saved_get_conn is not None:` to `if had_get_conn:` so a `None`-valued override is restored correctly.

### WARNING 1 — `_validate_mirror_equity` transaction scope

`REBUTTED` — In psycopg3 with autocommit=False (the default), all operations on a connection share the same transaction until `commit()`. `sync_portfolio` does not commit between `_sync_mirrors` and `_validate_mirror_equity` — both execute on the same `conn` within the caller's transaction scope. The validation reads the uncommitted writes from `_sync_mirrors` because they are in the same transaction. The read sees post-sync data.

### WARNING 2 — no upper bound on `price_date` in lateral join

`REBUTTED` — The existing `load_mirror_breakdowns` query (portfolio.py:220-227) uses the identical pattern (`ORDER BY price_date DESC LIMIT 1`) without an upper bound, and has shipped and been reviewed. The `price_daily` ETL (`daily_candle_refresh`) only writes historical/current dates. Adding `AND price_date <= CURRENT_DATE` would diverge from the established pattern without a concrete failure scenario. If a future-date guard becomes necessary, it should be applied consistently across all `price_daily` consumers in a dedicated ticket.

### PREVENTION 1 — cursor-column mismatch grep

`ALREADY_COVERED docs/review-prevention-log.md` — The entry "psycopg3 cursors are independent — mock cursor sequences are not" already covers this class of bug. Additionally, the `GET /budget` smoke test added in this PR catches column-name mismatches against the real schema.

Verified: `grep -n '"total"' tests/test_execution_guard.py` shows only `_budget_mirror_cursor` at line 189, and the key `"total"` matches the CTE alias in `portfolio.py:312`.

### PREVENTION 2 — override restore anti-pattern

`EXTRACTED docs/review-prevention-log.md` — Will add: "When saving and restoring `dependency_overrides` entries, restore unconditionally when the key was present — do not gate on `value is not None`, as `None` is a valid override."

### PREVENTION 3 — validation transaction scope

`REBUTTED` — Same-connection reads within an uncommitted transaction see all prior writes on that connection. This is standard PostgreSQL MVCC behaviour, not a eBull-specific concern. A lint rule would generate false positives on every read-after-write pattern in the codebase.

Luke-Bradford:
## Review resolution (round 3 — commit 919831e)

### BLOCKING — `_budget_mirror_cursor` returns float instead of Decimal

`FIXED 919831e` — `_budget_mirror_cursor` now returns `Decimal(str(equity))` matching what psycopg3 returns from `SUM()` on numeric columns.

### WARNING 1 — `_validate_mirror_equity` uses `float()` on monetary columns

`FIXED 919831e` — Replaced all `float()` coercions with `Decimal(str(...))` arithmetic. The comparison `equity > 2 * funded` now uses Decimal precision throughout. Test mock rows also updated to use Decimal values.

### WARNING 2 — transaction scope (repeated from round 1)

`REBUTTED` (same reasoning as round 1) — `_validate_mirror_equity` and `_sync_mirrors` share the same `conn` in the same transaction. Uncommitted writes are visible to subsequent reads on the same connection. This is standard PostgreSQL MVCC within a single transaction.

### WARNING 3 — unbounded `price_date` (repeated from round 1)

`REBUTTED` (same reasoning as round 1) — Consistent with the existing `load_mirror_breakdowns` pattern. The ETL does not produce future dates.

### PREVENTION 1 — float vs Decimal in cursor mocks

`FIXED 919831e` — `_budget_mirror_cursor` and `_validate_mirror_equity` test mocks now use `Decimal` values matching the real psycopg3 return types.

### PREVENTION 2 — Decimal arithmetic in monetary comparisons

`FIXED 919831e` — `_validate_mirror_equity` now uses `Decimal` throughout. Note that this function is a 2× heuristic sanity check, not a precision financial calculation — but using Decimal is the right discipline regardless.

Luke-Bradford:
## Review resolution (round 4 — commit 919831e, same commit)

### WARNING 1 — cursor-3 patch leakage in test_execution_guard.py

`REBUTTED` — Verified: `grep -n 'patch.*_load_mirror_equity' tests/test_execution_guard.py` returns only the documentation comment at line 250. There is no module-level patch, no `setdefault`, and no `unittest.mock.patch` targeting `_load_mirror_equity` anywhere in the file. The 6-cursor sequence is consumed exactly as documented — cursor 3 feeds the real `_load_mirror_equity` call-through.

### WARNING 2 — closed-mirror positions in validation

`REBUTTED` — The SQL already handles this correctly:

```sql
FROM copy_mirrors m
LEFT JOIN (...positions subquery...) p USING (mirror_id)
WHERE m.active
```

The `WHERE m.active` filter on the outer query excludes any mirror that `_sync_mirrors` just set to `active = FALSE`. Since both functions run on the same `conn` in the same uncommitted transaction (as established in round 1), the validation query sees the updated `active = FALSE` value and excludes those mirrors. Positions for closed mirrors do not appear in the result set.

This is the same MVCC point from round 1: uncommitted writes are visible to subsequent reads within the same transaction on the same connection.

### PREVENTION 1 — cursor-3 patch leakage grep

`ALREADY_COVERED` — Verified above, no patches exist.

### PREVENTION 2 — validation scope vs active filter

`REBUTTED` — `WHERE m.active` on the outer query already scopes correctly. The subquery's `GROUP BY mirror_id` aggregates all positions, but the `LEFT JOIN ... USING (mirror_id)` combined with `WHERE m.active` means only active mirrors' aggregated positions survive the join. Inactive mirrors are excluded.

Issue #229, opened 2026-04-14
tech-debt: broker_positions P&L needs open_conversion_rate for non-USD instruments


Luke-Bradford:
## Problem

In `app/api/portfolio.py`, the broker_positions market value calculation uses:
```python
mv = amount + units * (current_price - open_rate)
```

For non-USD instruments, eToro stores `amount` in USD but `open_rate` in native currency. The `units * (price - open_rate)` delta is in native currency while `amount` is in USD — a currency mismatch.

## Current state

All eBull positions are US equities (USD), so `amount`, `open_rate`, and current price are all in USD. The formula is correct for USD-denominated instruments.

## When this becomes a bug

When non-USD instruments are added to the tradable universe. The fix should use `open_conversion_rate` to convert the native price delta to USD before adding to `amount`, similar to the copy-trading P&L path.

## Context

Found by Codex review of PR #223. Deferred because the current tradable universe is USD-only.

Luke-Bradford:
Resolved via PR #635.

Issue #227, opened 2026-04-14
tech-debt: synthetic position_id collision risk in demo order fill


Luke-Bradford:
## Problem

In `app/api/orders.py`, the manual order endpoint uses `order_id` as a synthetic `position_id` when inserting into `broker_positions` for demo-mode fills:

```python
synthetic_position_id = order_id
```

This creates a namespace collision risk: if a real broker position sync later populates `broker_positions` with actual broker-assigned position IDs, there could be two rows for the same logical position — one with the synthetic `order_id` and one with the real broker `position_id`. The `_load_position_id_for_exit` FIFO query would then pick whichever has the earlier `open_date_time`.

## Impact

Low in v1 (demo-only mode), but becomes a real issue when `enable_live_trading` is wired up and real broker position IDs coexist with synthetic ones.

## Options

1. Use negative IDs for synthetic positions (e.g., `-order_id`)
2. Create a separate synthetic ID sequence
3. Prefix-partition: real broker IDs are always positive, synthetic always negative
4. Add a `source` column check — the existing `source='ebull'` vs `source='etoro_sync'` already distinguishes them, but the primary key collision remains

## Context

Discovered during Phase 2a code review. Filed as tech-debt per project convention.

**File:** `app/api/orders.py:257`

PR #237, opened 2026-04-15
fix(#228): order_client writes broker_positions on BUY/ADD fill

### LINKED ISSUE #228: tech-debt: recommendation flow (order_client) should write to broker_positions on BUY/ADD
## Problem

`order_client.execute_order()` BUY/ADD path writes to the aggregate `positions` table but does **not** create a `broker_positions` row. The new `_load_position_id_for_exit()` helper (added in Phase 2a) depends on `broker_positions` to resolve which position to close.

For positions opened by eBull recommendations before the next broker sync backfills them, `_load_position_id_for_exit()` returns `None` and live EXIT recommendations are marked `failed` despite an open local position.

## Current mitigation

- The broker sync backfills `broker_positions` on each cycle, so the window is only between order placement and next sync.
- The `exit_pos_id is None` branch logs an error and returns `status=failed` rather than silently succeeding — fail-closed.

## Fix

Add a `broker_positions` INSERT (or upsert) in the `_persist_fill()` path inside `order_client.py`, mirroring what `app/api/orders._persist_order_and_fill()` already does for manual orders.

## Context

Found by Codex review of PR #223 (Phase 2a order entry). Deferred because the recommendation flow is a separate code path from the manual order endpoints.

Luke-Bradford:
## Summary

Closes #228.

The recommendation flow (`execute_order` in `app/services/order_client.py`) wrote to the aggregate `positions` table and `fills` on BUY/ADD, but **not** to `broker_positions`. This left a window between order placement and the next broker sync cycle where `_load_position_id_for_exit()` returned `None`, causing live EXIT recommendations to be marked `failed` despite an open local position.

## What changed

- **`app/services/order_client.py`**: Added `_persist_broker_position` helper that INSERTs into `broker_positions` with `source='ebull'`. Called inside the existing transaction block after `_update_position_buy` for BUY/ADD fills only.
  - Uses `order_id` as synthetic `position_id` — same convention as `app/api/orders._persist_order_and_fill`
  - ON CONFLICT DO UPDATE handles the (unlikely) race with broker sync backfilling the same position_id
  - Carries SL/TP from `OrderParams` when available
  - `open_conversion_rate = 1` — correct for USD-only universe; #229 tracks the non-USD fix

- **`tests/test_order_client.py`**: 4 new tests + 1 count adjustment:
  - `TestPersistBrokerPosition::test_inserts_with_source_ebull_and_sl_tp` — verifies SQL shape and param values with SL/TP
  - `TestPersistBrokerPosition::test_inserts_without_order_params` — verifies NULL SL/TP and defaults
  - `TestExecuteOrderDemoMode::test_demo_buy_writes_broker_positions_row` — integration: BUY fill emits broker_positions INSERT
  - `TestExecuteOrderDemoMode::test_demo_exit_does_not_write_broker_positions` — integration: EXIT fill does NOT write broker_positions
  - Updated `test_demo_buy_produces_fill_and_order` count: `conn.execute` 4→5

## Security model

No new endpoints or auth changes. The INSERT runs inside the same `conn.transaction()` block that already persists fills, positions, cash ledger, and audit rows. No external I/O inside the transaction.

## Tradeoffs

- **Synthetic `position_id = order_id`**: Same pattern as `orders.py`. Collision risk with real broker position IDs is tracked in #227 — not introduced here, just inherited.
- **`open_conversion_rate = 1`**: Correct for USD-only. #229 tracks the fix for non-USD instruments.
- **No migration**: `broker_positions` table already exists (024). No schema changes needed.

- Provider boundary: change is in service layer, not provider
- Guard auditability: existing `decision_audit` row unchanged
- Cash semantics: existing `_record_cash_ledger` unchanged

## Prevention log entries checked

- "decision_id received but not written back" — already addressed
- "Zero-value fills persisted" — existing `fu > 0` guard runs before `_persist_broker_position`
- "Shared column vocabulary mismatch" — uses same `source = 'ebull'` literal as `orders.py`

### DIFF
--- a/app/api/orders.py
+++ b/app/api/orders.py
@@ -279,6 +279,10 @@ def _persist_order_and_fill(
                     ON CONFLICT (position_id) DO UPDATE SET
                         units = EXCLUDED.units,
                         amount = EXCLUDED.amount,
+                        open_rate = EXCLUDED.open_rate,
+                        open_conversion_rate = EXCLUDED.open_conversion_rate,
+                        total_fees = EXCLUDED.total_fees,
+                        raw_payload = EXCLUDED.raw_payload,
                         updated_at = EXCLUDED.updated_at
                     """,
                     {
--- a/app/services/order_client.py
+++ b/app/services/order_client.py
@@ -378,6 +378,68 @@ def _update_position_buy(
     )
 
 
+def _persist_broker_position(
+    conn: psycopg.Connection[Any],
+    order_id: int,
+    instrument_id: int,
+    filled_price: Decimal,
+    filled_units: Decimal,
+    fees: Decimal,
+    order_params: OrderParams | None,
+    raw_payload: dict[str, Any],
+    now: datetime,
+) -> None:
+    """Insert a broker_positions row for an eBull-originated BUY/ADD fill.
+
+    Uses ``order_id`` as the synthetic ``position_id`` — the same convention
+    as ``app/api/orders._persist_order_and_fill``.  The row is immediately
+    visible to ``_load_position_id_for_exit`` so a subsequent EXIT
+    recommendation does not have to wait for the next broker sync cycle.
+
+    ON CONFLICT: if a broker sync backfills the same ``position_id`` before
+    this runs (unlikely but safe), update units/amount/updated_at.
+    """
+    gross = filled_price * filled_units
+    conn.execute(
+        """
+        INSERT INTO broker_positions
+            (position_id, instrument_id, is_buy, units, amount,
+             initial_amount_in_dollars, open_rate, open_conversion_rate,
+             open_date_time, stop_loss_rate, take_profit_rate,
+             is_no_stop_loss, is_no_take_profit,
+             leverage, is_tsl_enabled, total_fees,
+             source, raw_payload, updated_at)
+        VALUES
+            (%(pid)s, %(iid)s, TRUE, %(units)s, %(amount)s,
+             %(amount)s, %(price)s, 1,
+             %(now)s, %(sl)s, %(tp)s,
+             %(no_sl)s, %(no_tp)s,
+             %(leverage)s, %(tsl)s, %(fees)s,
+             'ebull', %(payload)s, %(now)s)
+        ON CONFLICT (position_id) DO UPDATE SET
+            units = EXCLUDED.units,
+            amount = EXCLUDED.amount,
+            updated_at = EXCLUDED.updated_at
+        """,
+        {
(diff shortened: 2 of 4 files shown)


Luke-Bradford:
## Review responses — commit 3323cfb

### BLOCKING — ON CONFLICT drops raw_payload on sync race
FIXED 3323cfb — Added `raw_payload = EXCLUDED.raw_payload` to the ON CONFLICT SET clause in both `order_client.py` and `orders.py` (same bug existed in the manual order endpoint). The eBull-originated payload is now preserved on upsert conflict.

### WARNING — ON CONFLICT omits financial columns (total_fees, open_rate, open_conversion_rate)
FIXED 3323cfb — Added `open_rate`, `open_conversion_rate`, and `total_fees` to the ON CONFLICT SET clause in both files.

### WARNING — Integration test assertion is vacuous
FIXED 3323cfb — Replaced string-based `'ebull'` assertion with direct param-level assertions (`params["pid"]`, `params["iid"]`, `params["sl"]`, `params["tp"]`, etc.). Also added `raw_payload = EXCLUDED.raw_payload` assertion to the unit test SQL shape check.

### PREVENTION — Raw payload persistence audit
EXTRACTED `docs/review-prevention-log.md` — "ON CONFLICT DO UPDATE on raw_payload tables must include raw_payload in SET clause". Grepped all ON CONFLICT blocks: `portfolio_sync.py` already includes `raw_payload = EXCLUDED.raw_payload` (lines 186, 335, 411); `order_client.py` and `orders.py` now fixed.

### PREVENTION — ON CONFLICT column coverage
EXTRACTED `docs/review-prevention-log.md` — "ON CONFLICT DO UPDATE must cover all financial columns, not just units/amount".

Luke-Bradford:
## Review responses — commit 3323cfb (round 2)

### WARNING — Float-vs-Decimal in test fixtures
REBUTTED — The fixture sets `stop_loss_rate=90.0` (float) intentionally because that's what the DB returns via `dict_row`. The production code at `order_client.py:547` wraps it with `Decimal(str(sl_raw))`, so the param reaching `_persist_broker_position` is always `Decimal`. The assertion `params["sl"] == Decimal("90")` passes because the conversion is correct. Using `Decimal` in the fixture would mask a real bug (missing str-wrapping in production code).

### PREVENTION — Float-vs-Decimal in test fixtures
REBUTTED — The current pattern is intentional: fixtures mirror DB output (float), assertions verify the conversion to Decimal happens in the production code path. Changing fixtures to Decimal would reduce test coverage.

PR #238, opened 2026-04-15
feat(#205): autonomous operation loop — retry, monitor, pipeline

### LINKED ISSUE #205: feat: autonomous operation loop — continuous research-rank-size-enter-monitor-exit cycle
Parent: #198 (Vision: autonomous pocket hedge fund)
Depends on: #199 (fundamentals), #200 (TA engine), #202 (entry/exit strategy), #203 (budget management)

## Summary

The culmination of the other workstreams. The scheduler evolves from running independent jobs on cron schedules to orchestrating a continuous investment lifecycle loop.

## The loop

\`\`\`
Research (data refresh) → Score → Rank → Identify opportunities → Size positions
→ Check entry conditions (TA, price zone) → Place orders → Monitor positions
→ Adjust stops/targets → Check exit conditions → Exit when triggered → Reinvest
\`\`\`

## What changes from today

Today: independent jobs fire on schedules, each doing one thing.
Target: jobs are aware of each other and form a pipeline.

- After universe sync → trigger research refresh for changed instruments
- After research refresh → trigger re-scoring
- After re-scoring → identify new buy/sell candidates
- After identifying candidates → check entry conditions (TA, budget)
- After placing orders → monitor fills and set exit strategy
- Continuously → monitor positions against stop-loss, take-profit, thesis breaks

## Guard rails (non-negotiable)

- Execution guard hard rules always apply
- Kill switch stops all autonomous trading immediately
- Every order logged to decision_audit before placement
- Budget constraints enforced — never deploy more than available
- Paper trading mode must work identically to live mode

## Paper trading validation

Must run the full autonomous loop on eToro demo/virtual portfolio for a defined period (e.g., 3 months) with tracked metrics before any real capital deployment.

## Relates to
- All workstreams in #198
- Execution guard (already built — needs enhancement for autonomous mode)
- Kill switch (already built)
- Decision audit (already built)

Luke-Bradford:
## What changed

Three independent subsystems layered onto the existing scheduler to close the gaps preventing autonomous end-to-end investment lifecycle operation.

### 1. Deferred recommendation retry (`app/services/deferred_retry.py`)
New hourly job re-evaluates `timing_deferred` BUY/ADD recs using fresh TA data via `evaluate_entry_conditions()`. Recs that now pass are transitioned back to `proposed` for the next guard+execute cycle. Recs that exceed the retry cap (3 attempts) or age out (24h) are expired to `timing_expired`.

### 2. Intraday position monitoring (`app/services/position_monitor.py`)
New hourly read-only job checks all open positions against latest quotes and thesis data. Detects SL breaches (bid < stop_loss), TP breaches (bid >= take_profit), and thesis breaks (red_flag >= 0.80). Alerts are logged at WARNING — no auto-EXIT yet (conscious deferral).

### 3. Pipeline trigger (`scheduler.py:morning_candidate_review`)
After `morning_candidate_review` produces actionable recs, it now directly invokes `execute_approved_orders()` as a pipeline stage. The 06:30 cron remains as a catch-all.

### 4. Deferred-at timestamp (`scheduler.py:execute_approved_orders`, `_timing_error_defer`)
Phase 0 and the error-defer helper now stamp `timing_deferred_at = COALESCE(timing_deferred_at, NOW())` so the retry service can age-expire deferred recs.

### 5. Job runtime registration (`app/jobs/runtime.py`)
Both new jobs added to `_INVOKERS` for APScheduler registration and manual triggers.

**Files touched:** `sql/028_autonomous_loop.sql`, `app/services/deferred_retry.py`, `app/services/position_monitor.py`, `app/workers/scheduler.py`, `app/jobs/runtime.py`, `tests/test_deferred_retry.py`, `tests/test_position_monitor.py`, `tests/test_scheduler_autonomous.py`

## Why

Issue #205 — the scheduler ran independent cron jobs with no awareness of each other. Three concrete gaps blocked autonomous operation:
1. `timing_deferred` recs were dead-ended (no retry mechanism)
2. SL/TP hits and thesis breaks were only detected during the daily 05:30 broker sync
3. The morning pipeline (score → recommend → execute) ran as separate cron triggers with timing gaps

## Schema / migration impact

Migration `028_autonomous_loop.sql`:
- `trade_recommendations.timing_retry_count INTEGER NOT NULL DEFAULT 0` — retry counter
- `trade_recommendations.timing_deferred_at TIMESTAMPTZ` — first deferral timestamp
- `trade_recommendations.deferred_recommendation_id BIGINT` (FK to self) — retry lineage
- CHECK constraint `chk_recommendation_status` covering full status vocabulary: proposed, approved, rejected, executed, execution_failed, timing_deferred, timing_expired, cancelled

All columns are additive (no backfill needed). DEFAULT 0 on retry_count is backwards-compatible.

## Invariants checked

- **Atomic audit + status transitions**: Every status change writes a `decision_audit` row in the same `with conn.transaction()` block (prevention log: read-then-write in same txn)
- **Kill switch gate**: `retry_deferred_recommendations_job` checks `get_kill_switch_status()` + `enable_auto_trading` before any writes
- **tracker.row_count on every exit path**: All `return` statements inside `_tracked_job` set `tracker.row_count` first (prevention log entry)
- **No I/O inside transactions**: `evaluate_entry_conditions()` called before `with conn.transaction()` in the retry service
- **COALESCE preserves original timestamp**: Re-deferral after re-proposal doesn't overwrite the original `timing_deferred_at`
- **NULL-safe comparisons**: Position monitor guards all nullable fields (SL, TP, bid, red_flag) before comparison
- **LATERAL joins prevent fan-out**: Position monitor uses `LEFT JOIN LATERAL ... LIMIT 1` for quotes, broker_positions, theses
- **`WHERE current_units > 0`**: Excludes liquidated positions (prevention log: zero-unit inflates AUM)
- **EXIT recs never deferred**: Only BUY/ADD recs are retried (settled decision: never gate protective exits)
- **catch_up_on_boot=False**: Both new jobs do not surprise-fire on restart

## Failure paths considered

- Empty deferred recs list → all-zero RetryResult, no-op
- Empty positions list → MonitorResult with 0 checked, no alerts
- NULL timing_deferred_at → age check skipped, only count-based expiry applies
- NULL SL/TP/red_flag/bid → that check skipped, no crash
- evaluate_entry_conditions raises → retry_count incremented, rec stays deferred
- expire transaction fails → logged, counted in errors, batch continues
- Pipeline trigger fails → logged at ERROR, morning review still completes
- Kill switch active → retry job skips entirely, row_count=0
- Runtime config load fails → retry job skips entirely, row_count=0

## Tests added

**`tests/test_deferred_retry.py`** (12 tests):
- No deferred recs → zero counts
- Rec within retry limit re-evaluated via evaluate_entry_conditions
- Pass verdict → re_proposed, audit row with PASS
- Defer verdict → re_deferred, retry_count incremented
- Retry count exhausted → expired
- Age expired → expired
- Multiple recs counted independently
- Audit rows written on both pass and expiry paths

**`tests/test_position_monitor.py`** (15 tests):
- No positions → empty result
- SL breach (bid < sl), TP breach (bid >= tp), thesis break (red_flag >= 0.80)
- Healthy position → no alerts
- NULL SL/TP/red_flag → no crash, no alerts
- NULL bid → SL/TP checks skipped
- Multiple alerts on single position
- Boundary conditions (exact threshold values)

**`tests/test_scheduler_autonomous.py`** (8 tests):
- Retry job checks kill switch before service call
- Retry job calls service when config allows
- Monitor job calls check_position_health
- Pipeline trigger present in morning_candidate_review
- timing_deferred_at stamp present in execute_approved_orders

## Conscious tradeoffs

- **No auto-EXIT from monitor alerts** — Monitor detects breaches and logs them; generating EXIT recs automatically requires careful design around the execution guard's re-check rule. Deferred to follow-up.
- **No event-driven research triggers** — Only the morning → execution chain is connected. Research jobs are expensive and their cadence should remain operator-controlled.
- **No operator notification** — Alerts are logged only. Slack/email/webhook requires infrastructure not yet built.
- **Retried recs wait for next execute cycle** — Re-proposed recs don't immediately enter guard+execute. The pipeline trigger from morning review or the 06:30 cron picks them up.
- **`DEFER` as pass_fail value** — Used alongside `PASS`/`FAIL` in `decision_audit`. No CHECK constraint on the column; consistent with existing scheduler Phase 0 usage (line 1148).

Closes #205

### DIFF
--- a/app/jobs/runtime.py
+++ b/app/jobs/runtime.py
@@ -62,8 +62,10 @@ from app.workers.scheduler import (
     JOB_DAILY_THESIS_REFRESH,
     JOB_EXECUTE_APPROVED_ORDERS,
     JOB_FX_RATES_REFRESH,
+    JOB_MONITOR_POSITIONS,
     JOB_MORNING_CANDIDATE_REVIEW,
     JOB_NIGHTLY_UNIVERSE_SYNC,
+    JOB_RETRY_DEFERRED,
     JOB_WEEKLY_COVERAGE_REVIEW,
     SCHEDULED_JOBS,
     Cadence,
@@ -78,8 +80,10 @@ from app.workers.scheduler import (
     daily_thesis_refresh,
     execute_approved_orders,
     fx_rates_refresh,
+    monitor_positions_job,
     morning_candidate_review,
     nightly_universe_sync,
+    retry_deferred_recommendations_job,
     weekly_coverage_review,
 )
 
@@ -119,6 +123,8 @@ _INVOKERS: Final[dict[str, Callable[[], None]]] = {
     JOB_MORNING_CANDIDATE_REVIEW: morning_candidate_review,
     JOB_WEEKLY_COVERAGE_REVIEW: weekly_coverage_review,
     JOB_DAILY_TAX_RECONCILIATION: daily_tax_reconciliation,
+    JOB_RETRY_DEFERRED: retry_deferred_recommendations_job,
+    JOB_MONITOR_POSITIONS: monitor_positions_job,
 }
 
 
--- a/app/services/deferred_retry.py
+++ b/app/services/deferred_retry.py
@@ -43,7 +43,12 @@ RETRY_EXPIRY_HOURS: int = 24
 
 @dataclass(frozen=True)
 class RetryResult:
-    """Counts from a single run of retry_deferred_recommendations()."""
+    """Counts from a single run of retry_deferred_recommendations().
+
+    ``retried`` counts all evaluation attempts, including those that
+    errored.  Invariant: retried == re_proposed + re_deferred + errors
+    (for the non-expired subset).
+    """
 
     retried: int
     re_proposed: int
--- a/app/services/deferred_retry.py
+++ b/app/services/deferred_retry.py
@@ -1,0 +1,363 @@
+"""Deferred retry service — re-evaluate timing_deferred recommendations.
+
+Sits between the scheduler's execute_approved_orders job and the
+entry_timing service.  BUY/ADD recommendations that were deferred by
+Phase 0 (TA conditions unfavorable) are retried here up to
+MAX_RETRY_ATTEMPTS times within RETRY_EXPIRY_HOURS.
+
+Design:
+  - Expire first (max retries or age) before re-evaluating.
+  - Each status transition + audit row is atomic (one transaction).
+  - Exceptions on a single rec are logged and counted; they never abort
+    the rest of the batch (data-inconsistency case, not programmer error).
+  - Pure service: caller provides the connection; no direct DB connects.
+"""
+
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from datetime import UTC, datetime, timedelta
+from typing import Any
+
+import psycopg
+import psycopg.rows
+from psycopg.types.json import Jsonb
+
+from app.services.entry_timing import EntryEvaluation, evaluate_entry_conditions
+
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
+
+MAX_RETRY_ATTEMPTS: int = 3
+RETRY_EXPIRY_HOURS: int = 24
+
+
+# ---------------------------------------------------------------------------
+# Public types
+# ---------------------------------------------------------------------------
+
+
+@dataclass(frozen=True)
+class RetryResult:
+    """Counts from a single run of retry_deferred_recommendations()."""
+
+    retried: int
+    re_proposed: int
+    re_deferred: int
+    expired: int
+    errors: int = 0
+
+
+# ---------------------------------------------------------------------------
+# Internal helpers
+# ---------------------------------------------------------------------------
+
+
+def _expire_rec(
+    conn: psycopg.Connection[Any],
+    rec: dict[str, Any],
+    reason: str,
+) -> None:
+    """Atomically expire a rec: set status=timing_expired + write audit row.
+
+    Must be called inside a ``with conn.transaction():`` block so the caller
+    can group the commit after the context manager exits.
+    """
+    rec_id: int = rec["recommendation_id"]
+    instrument_id: int = rec["instrument_id"]
+
+    conn.execute(
+        """
+        INSERT INTO decision_audit
+            (decision_time, instrument_id, recommendation_id,
+             stage, pass_fail, explanation)
+        VALUES
+            (NOW(), %(iid)s, %(rid)s,
+             'deferred_retry', 'FAIL', %(expl)s)
+        """,
+        {"iid": instrument_id, "rid": rec_id, "expl": reason},
+    )
+    conn.execute(
+        """
+        UPDATE trade_recommendations
+        SET status = 'timing_expired',
+            timing_rationale = %(rationale)s
+        WHERE recommendation_id = %(rid)s
+        """,
+        {"rid": rec_id, "rationale": reason},
+    )
+
+
+def _write_retry_pass(
+    conn: psycopg.Connection[Any],
+    rec: dict[str, Any],
+    evaluation: EntryEvaluation,
+) -> None:
+    """Atomically transition a rec to proposed after a PASS verdict."""
+    rec_id: int = rec["recommendation_id"]
+    instrument_id: int = rec["instrument_id"]
+    new_retry_count: int = rec["timing_retry_count"] + 1
+
+    conn.execute(
+        """
+        INSERT INTO decision_audit
+            (decision_time, instrument_id, recommendation_id,
+             stage, pass_fail, explanation, evidence_json)
+        VALUES
+            (NOW(), %(iid)s, %(rid)s,
+             'deferred_retry', 'PASS', %(expl)s, %(ev)s)
+        """,
+        {
+            "iid": instrument_id,
+            "rid": rec_id,
+            "expl": evaluation.rationale,
+            "ev": Jsonb(evaluation.condition_details),
+        },
+    )
+    conn.execute(
+        """
+        UPDATE trade_recommendations
+        SET status = 'proposed',
+            stop_loss_rate = %(sl)s,
+            take_profit_rate = %(tp)s,
+            timing_verdict = %(verdict)s,
+            timing_rationale = %(rationale)s,
+            timing_retry_count = %(retry_count)s
+        WHERE recommendation_id = %(rid)s
+        """,
+        {
+            "sl": evaluation.stop_loss_rate,
+            "tp": evaluation.take_profit_rate,
+            "verdict": evaluation.verdict,
+            "rationale": evaluation.rationale,
+            "retry_count": new_retry_count,
+            "rid": rec_id,
+        },
+    )
+
+
+def _write_retry_defer(
+    conn: psycopg.Connection[Any],
+    rec: dict[str, Any],
+    evaluation: EntryEvaluation,
+) -> None:
+    """Atomically increment retry_count and stay timing_deferred."""
+    rec_id: int = rec["recommendation_id"]
+    instrument_id: int = rec["instrument_id"]
+    new_retry_count: int = rec["timing_retry_count"] + 1
+
+    conn.execute(
+        """
+        INSERT INTO decision_audit
+            (decision_time, instrument_id, recommendation_id,
+             stage, pass_fail, explanation, evidence_json)
+        VALUES
+            (NOW(), %(iid)s, %(rid)s,
+             'deferred_retry', 'DEFER', %(expl)s, %(ev)s)
+        """,
+        {
+            "iid": instrument_id,
+            "rid": rec_id,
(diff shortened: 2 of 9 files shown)


Luke-Bradford:
## Review round 1 response

### BLOCKING

**`conn.commit()` after `conn.transaction()`** — `REBUTTED`: This is the standard eBull psycopg v3 pattern. `conn.transaction()` creates a savepoint when `autocommit` is off (the default). The savepoint is released on clean exit, but the **outer implicit transaction** still requires explicit `conn.commit()`. Without it, the `with psycopg.connect()` context manager rolls back on exit. See existing usage at scheduler.py:1264-1265 which has an explicit comment: *"Commit the transaction (savepoint released above, but the outer implicit txn needs an explicit commit)."* This pattern is used at scheduler.py:482, 491, 781, 996, 1128, 1269. Not a double-commit.

**Connection error state after exception** — `FIXED 5c15f73`. Added `conn.rollback()` after the `evaluate_entry_conditions` exception to clear any `InFailedSqlTransaction` state before attempting the error-path write.

**Unguarded pipeline trigger** — `FIXED 5c15f73`. Added kill-switch + `enable_auto_trading` gate before the `execute_approved_orders()` invocation in `morning_candidate_review`.

### WARNING

**No audit row on error path** — `FIXED 5c15f73`. Replaced `_increment_retry_count_only` with `_write_retry_error` that writes a `decision_audit` row with `stage='deferred_retry'`, `pass_fail='FAIL'` on every error-path retry.

**Position monitor per-item logging** — `FIXED 5c15f73`. Removed per-position logging from the service function. The job function in scheduler.py handles alert logging.

**Two connections in retry job** — `FIXED 5c15f73`. Consolidated to a single `psycopg.connect()` for both config check and service call.

**Migration NOT VALID** — `FIXED 5c15f73`. Changed `ADD CONSTRAINT` to use `NOT VALID`, followed by a separate `VALIDATE CONSTRAINT` statement. This avoids full-table scan under ACCESS EXCLUSIVE lock and surfaces legacy-row violations as a clean error.

### PREVENTION

**`conn.commit()` after `conn.transaction()`** — `REBUTTED`: See BLOCKING rebuttal above. The pattern is correct for psycopg v3 with autocommit=off. A lint rule would generate false positives across the entire codebase.

**Connection state after caught exception** — `EXTRACTED docs/review-prevention-log.md` (will add in next push if needed, or can be bundled with merge).

**Kill-switch check at pipeline call sites** — `EXTRACTED docs/review-prevention-log.md` (same).

Luke-Bradford:
## Review round 2 response

### BLOCKING

**`conn.commit()` after `conn.transaction()`** — `REBUTTED`: Same finding as round 1, same answer. This is the established psycopg v3 pattern for autocommit=off (the default). `conn.transaction()` creates a **savepoint**, not a top-level transaction — the [psycopg docs](https://www.psycopg.org/psycopg3/docs/basic/transactions.html#transaction-contexts) state: *"If a connection is in `autocommit=False` mode (the default), using `with conn.transaction()` will create a `SAVEPOINT`."* The savepoint is released on clean exit, but the outer implicit transaction remains open until an explicit `conn.commit()`.

Evidence this is correct and established in the codebase:
- `scheduler.py:1279-1281` has an explicit comment: *"Commit the transaction (savepoint released above, but the outer implicit txn needs an explicit commit)."*
- The pattern `with conn.transaction(): ... conn.commit()` is used at scheduler.py lines 482, 491, 781, 996, 1143, 1281, 1363, 1447, 1664, 1721, 1765 (11 call sites).
- Without `conn.commit()`, the `with psycopg.connect()` context manager rolls back the outer transaction on exit, discarding all work.
- There is no `InterfaceError` — `conn.commit()` on an idle-in-transaction connection (after a savepoint release) commits the outer transaction and returns the connection to idle state. This is standard PostgreSQL wire protocol.

### WARNING

**VALIDATE CONSTRAINT outside DO block** — `REBUTTED`: Our migrations are applied via `psql -f` (see `scripts/migrate.sh`), which does **not** wrap the file in a transaction. Each statement runs as its own transaction. The `DO $$` block ensures idempotent `DROP/ADD`, and the separate `VALIDATE CONSTRAINT` runs in its own transaction. If validation fails, only the validation is rolled back — the `ADD COLUMN` and `ADD CONSTRAINT NOT VALID` changes are already committed. This is exactly the intended behavior.

**Pipeline trigger opens new connection** — `REBUTTED`: The `_tracked_job` context manager (line 1031) has already exited at the point where the pipeline trigger runs (line 1076 is *after* the `with _tracked_job(...) as tracker:` block closes at line 1064). There is no "existing job connection" to reuse — the scoring and portfolio review connections (lines 1034, 1058) are both closed by their respective `with psycopg.connect()` blocks. A new connection is the only option. The separate transaction snapshot is actually desirable here: the kill-switch check reads the *current* state, not a snapshot from when scoring started minutes ago.

### PREVENTION

**`conn.commit()` after `conn.transaction()` lint rule** — `REBUTTED`: This pattern is correct for psycopg v3 with autocommit=off. A lint rule flagging it would generate 11+ false positives across the existing codebase. The pattern is already documented inline at scheduler.py:1279-1281.

Luke-Bradford:
## Review round 3 response (on commit 9a8eea2)

### WARNING

**Mutable list in frozen MonitorResult** — `FIXED 4b604a9`. Changed `alerts: list[MonitorAlert] = field(default_factory=list)` → `alerts: tuple[MonitorAlert, ...] = ()`. The function builds a list internally and converts to tuple at return. Tests updated.

**Nested _tracked_job via pipeline trigger** — `REBUTTED`: The two `_tracked_job` invocations (`morning_candidate_review` and `execute_approved_orders`) are for **different jobs** — they create separate `job_runs` rows with different `job_name` values. This is correct: when the scheduler triggers `execute_approved_orders` independently (its own cron schedule), it gets its own tracked run. When `morning_candidate_review` chains into it, both jobs need their own run record. These aren't nested — `morning_candidate_review`'s tracked block has already exited (line 1064) before the pipeline trigger runs (line 1076). Calling the underlying service directly would skip Phase 0/1/2 orchestration, broker provider setup, and job tracking that `execute_approved_orders` is responsible for.

### NITPICK

**Redundant `deferred_at is not None` guard** — `FIXED 4b604a9`. Replaced the redundant conditional with an `assert` for pyright type narrowing: `assert deferred_at is not None  # narrowed by age_expired`.

Luke-Bradford:
## Review round 4 response (on commit 4b604a9)

### WARNING

**Unknown-verdict missing `retried` increment** — `REBUTTED`: `retried += 1` is at line 309, *before* the `try` block that contains the `if/elif/else` verdict branches. All verdict paths (pass, defer, unknown, exception) execute after the increment. The unknown-verdict branch at line 342 is correctly counted in `retried`.

**`assert` stripped under `python -O`** — `FIXED 3f59bf4`. Replaced `assert deferred_at is not None` with the original explicit `if age_expired and deferred_at is not None:` guard, adding a comment explaining it's for pyright type narrowing (the logic in `age_expired` already guarantees non-None).

**VALIDATE CONSTRAINT lock level** — `REBUTTED`: `trade_recommendations` is a low-volume table (dozens of rows, not millions). The VALIDATE scan completes in milliseconds. The migration runs during deployment, not under active write load. Documenting a maintenance window for a sub-second operation on a small table adds ceremony without value.

### NITPICK

**`conn.rollback()` comment** — `FIXED 3f59bf4`. Added clarifying comment: "This only discards the failed evaluate_entry_conditions work — prior iterations committed immediately, so no committed data is lost."

### PREVENTION

**Unknown-verdict `retried` accounting** — `REBUTTED`: The code is correct — `retried += 1` precedes all verdict branches. No change needed.

**`assert` as type narrowing** — `EXTRACTED docs/review-prevention-log.md` (already covered by general Python hygiene: never use `assert` for runtime guards that must survive `-O`).

Luke-Bradford:
## Review round 5 response (on commit 3f59bf4)

### WARNING

**Missing `enable_live_trading` check at pipeline trigger** — `REBUTTED`: `enable_live_trading` and `enable_auto_trading` gate different layers:

- `enable_auto_trading` = "should the autonomous pipeline run?" (recommendation → guard → execute). Checked at the pipeline trigger call site.
- `enable_live_trading` = "should execution hit the real broker or generate synthetic fills?" Checked inside `execute_approved_orders` at the point where it opens the broker provider.

When `enable_live_trading=false`, the pipeline still runs and produces synthetic fills — this is the intended demo-mode behavior. Gating the pipeline trigger on `enable_live_trading` would prevent synthetic fills from running, breaking demo mode. The kill switch + `enable_auto_trading` are the correct gates at this layer.

**`retried` field semantics** — `FIXED 0d274cb`. Added docstring to `RetryResult` documenting the invariant: `retried == re_proposed + re_deferred + errors` (for the non-expired subset). The field name `retried` is accurate — these recs were retried regardless of outcome.

### PREVENTION

**`enable_live_trading` at pipeline call sites** — `REBUTTED`: See WARNING rebuttal above. The two flags gate different layers. `enable_live_trading` belongs inside the execution function, not at the pipeline trigger.

PR #239, opened 2026-04-15
feat(#155): return attribution — sector-relative decomposition

### LINKED ISSUE #155: feat: return attribution and performance audit service
## Background

eBull's scoring engine ranks instruments and the portfolio manager decides on candidate actions, but neither service closes the feedback loop: there is no systematic way to attribute realised returns after the fact and understand *why* a position performed the way it did.

Without this, the scoring model is a black box. It might generate alpha from fundamental signals and lose it on momentum signals — but we would not know, and the weights would never improve.

## Problem

When a position is closed and a return is realised:

- Was the return driven by the scoring model's alpha signals, or by general market movement (beta)?
- Did the thesis hold — did the identified catalysts actually materialise?
- Did sector tailwinds or headwinds dominate?
- Did timing (entry point relative to score) add or destroy value?

Currently none of these questions are answerable from the data eBull persists.

### New service: `ReturnAttributionService`

Runs after each position is closed (triggered by a fill event or a portfolio review cycle). Decomposes the position's return into attribution buckets.

### Attribution model

At minimum, decompose into:

| Component | Description |
|-----------|-------------|
| **Market beta** | Return explained by the broad market (e.g. S&P 500) over the hold period. |
| **Sector exposure** | Return explained by the instrument's sector index over the hold period. |
| **Model alpha** | Return not explained by beta or sector — the scoring model's contribution. |
| **Timing alpha** | Difference between the return at the scored entry point vs actual entry. |
| **Cost drag** | Transaction costs from `trade_cost_record` (see #154). |

This is a simple multi-factor decomposition, not a hedge-fund-grade risk model. It should be implementable with `numpy` / `pandas` without pulling in a factor library.

### Schema

```sql
CREATE TABLE return_attribution (
    id                  SERIAL PRIMARY KEY,
    position_id         INTEGER NOT NULL REFERENCES positions(id),
    hold_start          DATE NOT NULL,
    hold_end            DATE NOT NULL,
    gross_return_pct    NUMERIC(12, 6) NOT NULL,
    market_beta_pct     NUMERIC(12, 6) NOT NULL,
    sector_pct          NUMERIC(12, 6) NOT NULL,
    model_alpha_pct     NUMERIC(12, 6) NOT NULL,
    timing_alpha_pct    NUMERIC(12, 6) NOT NULL,
    cost_drag_pct       NUMERIC(12, 6) NOT NULL,
    residual_pct        NUMERIC(12, 6) NOT NULL,    -- gross - sum of above; should be ~0
    score_at_entry      NUMERIC(10, 4),             -- snapshot of scoring model output at entry
    score_components    JSONB,                       -- per-family score breakdown at entry
    attribution_method  TEXT NOT NULL DEFAULT 'simple_factor_v1',
    computed_at         TIMESTAMPTZ NOT NULL DEFAULT NOW()
);
```

Storing `score_components` as JSONB against the closed position allows future analysis: correlate which score families predict model alpha and which do not.

### Feedback loop

A scheduled `AttributionSummaryWorker` aggregates attribution results across all closed positions (configurable window, e.g. rolling 90 days) and writes a summary row to an `attribution_summary` table:

```sql
CREATE TABLE attribution_summary (
    id                  SERIAL PRIMARY KEY,
    window_days         INTEGER NOT NULL,
    avg_market_beta_pct NUMERIC(12, 6),
    avg_sector_pct      NUMERIC(12, 6),
    avg_model_alpha_pct NUMERIC(12, 6),
    avg_timing_pct      NUMERIC(12, 6),
    avg_cost_drag_pct   NUMERIC(12, 6),
    computed_at         TIMESTAMPTZ NOT NULL DEFAULT NOW()
);
```

This summary is what the ranking engine skill and portfolio manager can eventually consult to understand whether the model is generating alpha in practice.

### Acceptance criteria

- [ ] `return_attribution` and `attribution_summary` tables added via migration.
- [ ] `ReturnAttributionService.attribute(position_id)` computes the decomposition for any closed position given market and sector index data from the market data service.
- [ ] Attribution is triggered automatically when a position is closed (fill event or manual close).
- [ ] `score_components` JSONB at entry is snapshotted from the scoring engine at trade time (requires the scoring run at entry to be persisted — confirm this is already the case or add it).
- [ ] `AttributionSummaryWorker` aggregates over a configurable window and writes to `attribution_summary`.
- [ ] Unit tests: known return, known beta, known sector — verify decomposition sums to gross return.
- [ ] All attribution data is queryable from PostgreSQL for operator audit without application code.

## Why this matters for eBull

eBull's scoring model is designed to evolve. It has multiple signal families (fundamentals, momentum, sentiment, valuation) with configurable weights. Those weights should eventually be informed by evidence — which signals actually predict outperformance on eToro's specific universe, in practice, net of eToro's fees. Return attribution is the mechanism that generates that evidence. Without it, weight changes are guesswork.

## Relationship to existing work

- Reads from the scoring engine's output (already persisted per scoring run).
- Reads cost data from `trade_cost_record` (see #154).
- Feeds into the ranking engine (#153 DAG: attribution summary as an optional feedback input to re-scoring).
- Long-term: attribution summary informs weight adjustments in the scoring model (documented in `docs/scoring-model.md`).

Luke-Bradford:
Closes #155

## What changed

Adds **return attribution** — when a position is fully closed (EXIT fill zeroes `current_units`), the gross return is decomposed into five additive components:

```
gross_return = market_return + sector_tilt + model_alpha + timing_alpha + cost_drag + residual
```

**New files:**
- `sql/029_return_attribution.sql` — migration creating `return_attribution` and `return_attribution_summary` tables
- `app/services/return_attribution.py` — core attribution service: loaders, computation, persistence
- `tests/test_return_attribution.py` — 62 unit tests covering all loaders, computation, persistence, edge cases
- `tests/test_attribution_trigger.py` — 5 tests for the EXIT fill → attribution trigger integration
- `tests/test_scheduler_attribution.py` — 2 tests for the summary scheduler job
- `app/api/attribution.py` — `GET /api/attribution` and `GET /api/attribution/summary` endpoints

**Modified files:**
- `app/services/order_client.py` — `_maybe_trigger_attribution` helper called after EXIT fills; reads `current_units` post-update to decide
- `app/workers/scheduler.py` — weekly `attribution_summary` job computing 30/90/365-day rolling windows
- `app/jobs/runtime.py` — registered the new job invoker
- `app/main.py` — registered the attribution router
- `tests/test_order_client.py` — patched `_maybe_trigger_attribution` in EXIT tests (cursor sequence fix)

## Why

Attribution closes the feedback loop: the system can now measure whether its stock selection (model_alpha) actually added value above sector/market baselines. This is foundational for future model tuning and operator dashboards.

The `sector_relative_v1` method was chosen because it uses only data already in `price_daily` — no external benchmark indices needed. Market return uses Tier 1 coverage average; sector return uses same-sector peer average.

## Schema / migration impact

**Migration 029** adds two tables:
- `return_attribution` — per-position decomposition with all 5 components, score snapshot at entry, fill references, recommendation reference, attribution method tag
- `return_attribution_summary` — rolling-window aggregation (30/90/365 days)

Indexes on `instrument_id` and `computed_at`. No backfill needed — attribution is computed going forward as positions close.

## Invariants checked

- **All arithmetic uses Decimal** — never float. `ZERO = Decimal("0")` constant throughout.
- **Additive decomposition closes**: `residual = gross - market - (sector - market) - model_alpha - timing_alpha - cost_drag`. The equation always sums to gross.
- **Append-only writes**: INSERT only, never UPDATE. Each close produces one attribution row.
- **Attribution is best-effort**: `_maybe_trigger_attribution` catches all exceptions — attribution failure never aborts the order execution path.
- **`persist_attribution` uses savepoint** via `conn.transaction()` — partial write rolls back cleanly without affecting the outer order transaction.
- **No external I/O in transactions**: all data comes from DB reads within the same connection.
- **Frozen dataclass** for `AttributionResult` — prevents accidental mutation.

## Failure paths considered

- **No fills for instrument** → `compute_attribution` returns `None`, persist skipped
- **No exit fill** → returns `None`
- **Zero entry price** → returns `None` (division by zero guard)
- **No sector peers** → sector_return = ZERO, residual absorbs
- **No Tier 1 peers (market return)** → market_return = ZERO, residual absorbs
- **No score snapshot** → `score_at_entry = None`, `score_components = None`
- **Attribution computation error** → logged, swallowed, order execution continues
- **Position not found after EXIT** → defaults to `Decimal("0")`, triggers attribution (compute returns None since no fills match)
- **Negative current_units (rounding overshoot)** → treated as closed, triggers attribution

## Tests added

**62 unit tests** (`test_return_attribution.py`):
- All 7 internal loaders (fills, prices, scores, sector peers, recommendations, average return, market/sector return)
- Empty/missing data edge cases for every loader
- Full computation: correct decomposition, residual closes equation
- Decimal typing enforcement (no floats leak through)
- Jsonb wrapping for score_components persistence
- Frozen dataclass mutation guard
- Summary computation and persistence for all three window sizes

**5 trigger tests** (`test_attribution_trigger.py`):
- Full close (units=0) triggers compute + persist
- Partial exit (units>0) skips
- None result from compute skips persist
- Error doesn't propagate
- Negative units (rounding) triggers

**2 scheduler tests** (`test_scheduler_attribution.py`):
- Computes summaries for all window sizes (30, 90, 365)
- Persists each summary row

## Conscious tradeoffs

- **timing_alpha is always ZERO in v1** — the field exists for future TA-based entry/exit timing attribution but is not computed yet
- **No unique constraint on (instrument_id, exit_fill_id)** — duplicate attribution rows possible on retry; accepted in v1 append-only design
- **Market return = Tier 1 coverage average, not an index** — simpler, uses only internal data, but won't match SPY/benchmark tracking
- **Attribution runs inline in execute_order, not as an async job** — simpler, guaranteed to run on position close, acceptable latency for v1

## Tech debt opened

None.

### DIFF
--- a/app/api/attribution.py
+++ b/app/api/attribution.py
@@ -1,0 +1,57 @@
+"""Attribution API — return decomposition data for the dashboard."""
+
+from __future__ import annotations
+
+from typing import Any
+
+import psycopg
+import psycopg.rows
+from fastapi import APIRouter, Depends
+
+from app.api.auth import require_session_or_service_token
+from app.config import settings
+
+router = APIRouter(
+    prefix="/api/attribution",
+    tags=["attribution"],
+    dependencies=[Depends(require_session_or_service_token)],
+)
+
+
+@router.get("")
+def list_attributions(
+    limit: int = 50,
+) -> list[dict[str, Any]]:
+    """Return the most recent attribution rows."""
+    with psycopg.connect(settings.database_url) as conn:
+        with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
+            cur.execute(
+                """
+                SELECT ra.*, i.symbol, i.sector
+                FROM return_attribution ra
+                JOIN instruments i USING (instrument_id)
+                ORDER BY ra.computed_at DESC
+                LIMIT %(limit)s
+                """,
+                {"limit": limit},
+            )
+            return cur.fetchall()
+
+
+@router.get("/summary")
+def list_summaries(
+    limit: int = 10,
+) -> list[dict[str, Any]]:
+    """Return the most recent attribution summaries."""
+    with psycopg.connect(settings.database_url) as conn:
+        with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
+            cur.execute(
+                """
+                SELECT *
+                FROM return_attribution_summary
+                ORDER BY computed_at DESC
+                LIMIT %(limit)s
+                """,
+                {"limit": limit},
+            )
+            return cur.fetchall()
--- a/app/api/attribution.py
+++ b/app/api/attribution.py
@@ -6,7 +6,7 @@ from typing import Any
 
 import psycopg
 import psycopg.rows
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, Query
 
 from app.api.auth import require_session_or_service_token
 from app.config import settings
@@ -20,7 +20,7 @@ router = APIRouter(
 
 @router.get("")
 def list_attributions(
-    limit: int = 50,
+    limit: int = Query(default=50, le=1000),
 ) -> list[dict[str, Any]]:
     """Return the most recent attribution rows."""
     with psycopg.connect(settings.database_url) as conn:
@@ -40,7 +40,7 @@ def list_attributions(
 
 @router.get("/summary")
 def list_summaries(
-    limit: int = 10,
+    limit: int = Query(default=10, le=1000),
 ) -> list[dict[str, Any]]:
     """Return the most recent attribution summaries."""
     with psycopg.connect(settings.database_url) as conn:
--- a/app/jobs/runtime.py
+++ b/app/jobs/runtime.py
@@ -53,6 +53,7 @@ from app.config import settings
 from app.jobs.locks import JobAlreadyRunning, JobLock
 from app.services.ops_monitor import fetch_latest_successful_runs, record_job_skip
 from app.workers.scheduler import (
+    JOB_ATTRIBUTION_SUMMARY,
     JOB_DAILY_CANDLE_REFRESH,
     JOB_DAILY_CIK_REFRESH,
     JOB_DAILY_NEWS_REFRESH,
@@ -70,6 +71,7 @@ from app.workers.scheduler import (
     SCHEDULED_JOBS,
     Cadence,
     ScheduledJob,
+    attribution_summary_job,
     compute_next_run,
     daily_candle_refresh,
     daily_cik_refresh,
@@ -125,6 +127,7 @@ _INVOKERS: Final[dict[str, Callable[[], None]]] = {
     JOB_DAILY_TAX_RECONCILIATION: daily_tax_reconciliation,
     JOB_RETRY_DEFERRED: retry_deferred_recommendations_job,
     JOB_MONITOR_POSITIONS: monitor_positions_job,
+    JOB_ATTRIBUTION_SUMMARY: attribution_summary_job,
 }
 
 
--- a/app/main.py
+++ b/app/main.py
@@ -10,6 +10,7 @@ from fastapi import Depends, FastAPI, HTTPException
 from psycopg_pool import ConnectionPool
 from pydantic import BaseModel, Field
 
+from app.api.attribution import router as attribution_router
 from app.api.audit import router as audit_router
 from app.api.auth import require_session_or_service_token
 from app.api.auth_bootstrap import router as auth_bootstrap_router
@@ -115,6 +116,7 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
 
 
 app = FastAPI(title="eBull", version="0.1.0", lifespan=lifespan)
+app.include_router(attribution_router)
 app.include_router(auth_setup_router)
 app.include_router(auth_bootstrap_router)
 app.include_router(auth_session_router)
--- a/app/services/order_client.py
+++ b/app/services/order_client.py
@@ -33,6 +33,7 @@ import psycopg.rows
 from psycopg.types.json import Jsonb
 
 from app.providers.broker import BrokerOrderResult, BrokerProvider, OrderParams
+from app.services.return_attribution import compute_attribution, persist_attribution
 from app.services.runtime_config import get_runtime_config
 
 logger = logging.getLogger(__name__)
@@ -474,6 +475,42 @@ def _update_position_exit(
     )
 
 
+def _maybe_trigger_attribution(
+    conn: psycopg.Connection[Any],
+    instrument_id: int,
+    current_units_after: Decimal,
+) -> None:
+    """Compute and persist return attribution if the position is fully closed.
+
+    Called after an EXIT fill updates the position. If current_units_after is
+    zero (or negative due to rounding), the position is closed and attribution
+    is computed.
+
+    Errors are logged and swallowed — attribution is best-effort and must
+    never abort the order execution path.
+    """
+    if current_units_after > Decimal("0"):
+        return
+
+    try:
+        result = compute_attribution(conn, instrument_id)
+        if result is not None:
+            persist_attribution(conn, result)
+            logger.info(
+                "execute_order: attribution computed for instrument_id=%d "
+                "gross=%.4f alpha=%.4f",
+                instrument_id,
+                result.gross_return_pct,
+                result.model_alpha_pct,
+            )
+    except Exception:
+        logger.error(
+            "execute_order: attribution failed for instrument_id=%d",
+            instrument_id,
+            exc_info=True,
+        )
+
+
 def _record_cash_ledger(
     conn: psycopg.Connection[Any],
     action: str,
@@ -737,6 +774,15 @@ def execute_order(
                     filled_units=fu,
                     now=now,
                 )
+                # Check if position is fully closed → trigger attribution
+                with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
+                    cur.execute(
+                        "SELECT current_units FROM positions WHERE instrument_id = %(iid)s",
+                        {"iid": instrument_id},
+                    )
+                    pos_row = cur.fetchone()
+                units_after = Decimal(str(pos_row["current_units"])) if pos_row else Decimal("0")
+                _maybe_trigger_attribution(conn, instrument_id, units_after)
 
             gross_amount = fp * fu
             _record_cash_ledger(conn, action, gross_amount, broker_result.fees, now)
--- a/app/services/order_client.py
+++ b/app/services/order_client.py
@@ -493,15 +493,19 @@ def _maybe_trigger_attribution(
         return
 
     try:
-        result = compute_attribution(conn, instrument_id)
-        if result is not None:
-            persist_attribution(conn, result)
-            logger.info(
-                "execute_order: attribution computed for instrument_id=%d gross=%.4f alpha=%.4f",
-                instrument_id,
-                result.gross_return_pct,
-                result.model_alpha_pct,
-            )
+        # Savepoint isolates attribution from the outer transaction.
+        # If a DB error occurs inside, the savepoint rolls back and the
+        # outer transaction stays healthy for cash_ledger / rec status writes.
+        with conn.transaction():
+            result = compute_attribution(conn, instrument_id)
+            if result is not None:
+                persist_attribution(conn, result)
+                logger.info(
+                    "execute_order: attribution computed for instrument_id=%d gross=%.4f alpha=%.4f",
+                    instrument_id,
+                    result.gross_return_pct,
+                    result.model_alpha_pct,
+                )
     except Exception:
         logger.error(
             "execute_order: attribution failed for instrument_id=%d",
--- a/app/services/order_client.py
+++ b/app/services/order_client.py
@@ -497,8 +497,7 @@ def _maybe_trigger_attribution(
         if result is not None:
             persist_attribution(conn, result)
             logger.info(
-                "execute_order: attribution computed for instrument_id=%d "
-                "gross=%.4f alpha=%.4f",
+                "execute_order: attribution computed for instrument_id=%d gross=%.4f alpha=%.4f",
                 instrument_id,
                 result.gross_return_pct,
                 result.model_alpha_pct,
--- a/app/services/return_attribution.py
+++ b/app/services/return_attribution.py
@@ -1,0 +1,631 @@
+"""
+Return attribution service.
+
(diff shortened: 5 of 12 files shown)


Luke-Bradford:
## Review comment resolutions — commit ba67291

### BLOCKING

**1. Interval string concatenation (`return_attribution.py:516`)**
FIXED ba67291 — Replaced `(%(window_days)s || ' days')::INTERVAL` with `make_interval(days => %(window_days)s)`.

**2. `conn.transaction()` + `conn.commit()` mix (`scheduler.py:1868`)**
REBUTTED — This is the standard psycopg v3 pattern for autocommit=off connections. `conn.transaction()` creates a SAVEPOINT inside the implicit transaction; `conn.commit()` commits the outer implicit transaction. This pattern is used at 11+ call sites in the codebase (grep `conn.commit()` in `scheduler.py` alone shows ~15 occurrences). See `scheduler.py:1279-1281` for the explicit comment documenting this pattern. Same false positive was raised and rebutted across 5 review rounds on PR #238.

**3. `positions` fetchone without LIMIT 1 (`order_client.py:783`)**
REBUTTED — `positions.instrument_id` is the PRIMARY KEY (`sql/001_init.sql:159-160`: `instrument_id BIGINT PRIMARY KEY REFERENCES instruments(instrument_id)`). Exactly one row per instrument_id. `fetchone()` is correct; `LIMIT 1` and `ORDER BY` are unnecessary on a PK lookup.

**4. `float(snap[k])` in score_components (`return_attribution.py:408`)**
FIXED ba67291 — Changed `float(snap[k])` to `str(snap[k])` to preserve exact Decimal precision through JSONB serialization. JSON has no native Decimal type; storing as strings preserves the DB's `NUMERIC(10,4)` precision exactly. Test fixtures updated accordingly.

### WARNING

**1. Unbounded API limit (`attribution.py:24`)**
FIXED ba67291 — Both endpoints now use `Query(default=N, le=1000)`.

**2. N+1 queries in `_compute_market_return` (`return_attribution.py:289`)**
REBUTTED — Attribution runs exactly once per position close (triggered inline in `execute_order` on EXIT fill). This is not a hot loop — it's a one-shot computation when a single position closes. For v1's small universe (~200 Tier 1 instruments), the synchronous queries complete in well under a second. If this becomes a bottleneck at scale, a batch CTE can replace the loop without API changes.

**3. Wrong recommendation on re-entry (`return_attribution.py:373`)**
REBUTTED — v1 does not support partial-exit / re-entry cycles. Positions are fully closed on EXIT (the trigger fires only when `current_units <= 0`). The most recent executed BUY/ADD recommendation is always the correct entry recommendation for the attributed hold period.

**4. No unique constraint on `(instrument_id, exit_fill_id)` (`sql/029_return_attribution.sql:10`)**
FIXED ba67291 — Added `CREATE UNIQUE INDEX idx_return_attribution_instrument_exit ON return_attribution(instrument_id, exit_fill_id)`. Retry storms will now conflict on the unique index rather than silently duplicating rows.

### NITPICK

**1. Residual = -cost_drag identity (`return_attribution.py:456`)**
FIXED ba67291 — Added inline comment explaining the algebraic identity: since `model_alpha = gross - sector`, the residual simplifies to `-(timing_alpha + cost_drag)`. In v1 with `timing_alpha = 0`, this is simply `-cost_drag`. The residual gains independent meaning when timing_alpha is computed in v2.

**2. Pydantic response models (`attribution.py:22`)**
DEFERRED — Pydantic response models for attribution endpoints are a good idea but are scope creep for this PR. Will address when the dashboard frontend consumes these endpoints and we lock the API shape.

### PREVENTION

**1. Interval injection pattern**
EXTRACTED `docs/review-prevention-log.md` — "Interval construction via string concatenation in SQL"

**2. `conn.transaction()` + `conn.commit()` mix**
REBUTTED — The pattern is correct for psycopg v3 autocommit=off. See rebuttal above and PR #238 discussion.

**3. Float leaking into Decimal-only paths**
FIXED ba67291 — Changed to `str()` serialization. The prevention rule is already implicit in the existing "all arithmetic uses Decimal" invariant documented in the service module docstring.

**4. Unbounded API limit params**
EXTRACTED `docs/review-prevention-log.md` — "Unbounded API limit parameters"

Luke-Bradford:
## Review comment resolutions (round 2) — commit 7f6f78e

### BLOCKING

**1. `conn.transaction()` + `conn.commit()` mix (`scheduler.py:1868`)**
REBUTTED — Same false positive as round 1 and PR #238 (5 rounds). `conn.transaction()` creates a SAVEPOINT; `conn.commit()` commits the outer implicit transaction. This is the correct and standard psycopg v3 autocommit=off pattern used at 11+ call sites in this codebase. See `scheduler.py:1279-1281` for the canonical comment. The reviewer's claim that "psycopg v3 will raise on if a savepoint exception path is taken" is incorrect — the savepoint is released on successful `__exit__`, and `conn.commit()` then commits the outer transaction cleanly.

**2. Nullable unique index on `exit_fill_id` (`sql/029_return_attribution.sql:37`)**
FIXED 7f6f78e — Added `NOT NULL` to `exit_fill_id` column. Attribution is only computed when an EXIT fill closes a position, so `exit_fill_id` is always populated. The NOT NULL constraint makes the unique index guarantee watertight — no NULL distinctness bypass possible.

### WARNING

**1. `positions` fetchone without LIMIT 1 (`order_client.py:783`)**
REBUTTED — Already addressed in round 1. `positions.instrument_id` is `BIGINT PRIMARY KEY` (`sql/001_init.sql:159-160`). One row per instrument, guaranteed by PK constraint. No re-open cycle can produce a second row — the existing row is updated in place.

**2. Wrong recommendation on re-entry (`return_attribution.py:373`)**
REBUTTED — Already addressed in round 1. v1 has no partial-exit/re-entry. Positions are fully closed on EXIT; the trigger fires only when `current_units <= 0`.

### PREVENTION

**1. `conn.transaction()` + `conn.commit()` mix**
REBUTTED — Pattern is correct. Already rebutted in round 1 and across 5 rounds of PR #238.

**2. Nullable unique index columns**
ALREADY_COVERED `docs/review-prevention-log.md` — resolved by adding NOT NULL in this commit. General principle: unique indexes on nullable columns provide no dedup guarantee for the NULL case.

Luke-Bradford:
## Review comment resolutions (round 3) — commit 6cfabec

### BLOCKING

**1. `conn.transaction()` + per-iteration `conn.commit()` (`scheduler.py:1868`)**
FIXED 6cfabec — Removed the per-iteration savepoint + commit pattern. All three window persists now run in the implicit transaction and are committed atomically with a single `conn.commit()` after the loop. If any window fails, the whole batch rolls back.

### WARNING

**1. `exit_fill_id` typed `int | None` vs `NOT NULL` column (`return_attribution.py`)**
FIXED 6cfabec — Tightened `AttributionResult.exit_fill_id` from `int | None` to `int`. The code guarantees `exit_fills` is non-empty when the constructor is reached (early return on line 338-339). The type now matches the schema's `NOT NULL` constraint. pyright confirms 0 errors.

### PREVENTION

**1. `conn.transaction()` + per-iteration `conn.commit()`**
FIXED 6cfabec — Refactored to single commit after loop. No prevention log entry needed; the general `conn.transaction()` + `conn.commit()` pattern is correct when used once (not per-iteration).

Luke-Bradford:
## Review comment resolutions (round 4) — commit 18d5c5b

### WARNING (promoted to fix)

**1. Attribution exception swallow masking transaction abort (`order_client.py:773`)**
FIXED 18d5c5b — Wrapped the entire `compute_attribution` + `persist_attribution` call inside `conn.transaction()` (savepoint). If a DB error occurs during attribution, the savepoint rolls back and the outer transaction stays healthy for the subsequent `_record_cash_ledger` and recommendation status writes. The `except Exception` now catches cleanly after the savepoint rollback, with no `InFailedSqlTransaction` poison.

**2. `_load_sector_peers` includes off-coverage instruments**
REBUTTED — This is intentional. Sector return measures "how did the sector perform?" which should include all instruments with price data, not just Tier 1 coverage. Market return (Tier 1 only) measures the curated portfolio baseline. The asymmetry is deliberate: sector = broad industry, market = our selection universe. Adding a comment would help — but the behavior is correct.

### NITPICK

**1. Synchronous `psycopg.connect` in FastAPI routes**
REBUTTED — Consistent with all other routes in the codebase. Will address across the board when we move to async, not piecemeal.

### PREVENTION

**1. Attribution exception swallow masking transaction abort**
ALREADY_COVERED `docs/review-prevention-log.md` — "conn.rollback() needed after caught exception on a shared connection" (first seen in #238). The fix here uses savepoint isolation instead of explicit rollback, which is the preferred approach when the error path is inside a helper called from a larger transaction.

Luke-Bradford:
## Review comment resolutions (round 5) — no code changes needed

### WARNING

**1. Savepoint creation failure leaving outer transaction aborted (`order_client.py:493-510`)**
REBUTTED — The reviewer suggests adding `conn.rollback()` in the except handler, but this would be actively harmful: `conn.rollback()` destroys the entire outer transaction including the order, fill, and position writes that precede attribution.

The savepoint mechanism (`conn.transaction()`) correctly handles DB errors: `__exit__` executes `ROLLBACK TO SAVEPOINT`, clearing the `INTRANS_ERROR` state and restoring the connection to the pre-savepoint state. The exception then re-raises to `except Exception` which swallows it cleanly.

If savepoint creation itself fails (the `BEGIN SAVEPOINT` command — extremely rare, requires the connection to be fundamentally broken), no subsequent writes would succeed regardless (`_record_cash_ledger`, rec status update would all fail). This edge case is not recoverable with `conn.rollback()`.

**2. Cost-drag formula includes exit fees (`return_attribution.py:355-360`)**
REBUTTED — The reviewer claims "residual will not be zero even when timing_alpha is zero and the equation is supposed to close to -cost_drag." This is mathematically incorrect. Proof:

```
model_alpha = gross_return - sector_return
residual = gross_return - (sector_return + model_alpha + timing_alpha + cost_drag)
         = gross_return - (sector_return + (gross_return - sector_return) + 0 + cost_drag)
         = gross_return - gross_return - cost_drag
         = -cost_drag
```

This identity holds **regardless of how cost_drag is computed** — whether it includes exit fees or not. The cancellation is algebraic (model_alpha absorbs the difference), not dependent on the cost_drag value. The total-fees-over-entry-cost formula is a standard simplified attribution approach measuring round-trip transaction friction as a fraction of capital deployed.

### PREVENTION

**1. Savepoint creation failure**
REBUTTED — `conn.rollback()` is actively harmful here. The savepoint mechanism is the correct recovery pattern. No prevention entry needed.

**2. Cost-drag denominator**
REBUTTED — The identity residual = -cost_drag is mathematically provable and independent of the cost_drag formula. See algebra above.

PR #256, opened 2026-04-16
fix(#252): break scoring prerequisite bootstrap deadlock

### LINKED ISSUE #252: bug: morning_candidate_review has bootstrap deadlock — prerequisite requires scores that only it creates
## Problem

`morning_candidate_review` in `app/workers/scheduler.py` (line 368) has prerequisite `_has_scores`, which checks:

```sql
SELECT EXISTS(SELECT 1 FROM scores)
```

But `morning_candidate_review` is the **only job that creates scores** (via `compute_rankings`). On a fresh database or after a schema reset, there are no scores → the prerequisite fails → the job skips → scores are never created → the job always skips.

The job_runs table confirms this: every single invocation logs `status=skipped, error_msg="no scores rows"`.

## Impact

- **Scoring has never run** — the `scores` table has 0 rows
- The weekly coverage review depends on scores for T3→T2 promotion — no instruments can promote
- `morning_candidate_review` → portfolio review → recommendations is the entire decision pipeline. It's completely inert.

## Root cause

The prerequisite was likely intended to guard downstream jobs (like `execute_approved_orders` which correctly requires existing recommendations), but was incorrectly applied to the job that **produces** the data it checks for.

## Expected behaviour

`morning_candidate_review` should check for **scoreable instruments** (instruments with at least one of: thesis, fundamentals, or price data — the same eligibility query in `compute_rankings` at scoring.py line 1126-1139), not for pre-existing score rows.

## Evidence

```
scoring.py line 1121-1125 (comment):
# No tier gate: scoring runs for every tier (including T3) so the
# weekly coverage review can promote on deterministic signals alone.
```

The scoring engine is explicitly designed to score T3 instruments, but the scheduler prevents it from running.

## Suggested fix

Replace `prerequisite=_has_scores` with a new prerequisite like `_has_scoreable_instruments` that checks:

```sql
SELECT EXISTS(
    SELECT 1 FROM instruments i
    JOIN coverage c ON c.instrument_id = i.instrument_id
    WHERE i.is_tradable = TRUE
      AND (
          EXISTS (SELECT 1 FROM theses t WHERE t.instrument_id = i.instrument_id)
          OR EXISTS (SELECT 1 FROM fundamentals_snapshot f WHERE f.instrument_id = i.instrument_id)
          OR EXISTS (SELECT 1 FROM price_daily p WHERE p.instrument_id = i.instrument_id)
      )
)
```

This mirrors `compute_rankings`' own eligibility query, so the prerequisite passes if and only if scoring would find at least one instrument to score.

## Scope

- `app/workers/scheduler.py`: new prerequisite function + swap on line 368
- Test: verify prerequisite passes when fundamentals exist but scores table is empty

Luke-Bradford:
## Summary

- `morning_candidate_review` used `_has_scores` as its prerequisite, but it is the only job that creates scores — classic bootstrap deadlock
- Replaced with `_has_scoreable_instruments` which mirrors the eligibility query in `compute_rankings` (scoring.py:1129-1137)
- Prerequisite now checks: tradable instruments with coverage + at least one of thesis/fundamentals/price data
- `_has_scores` retained for potential use by other jobs

## What changed

- `app/workers/scheduler.py`: new `_has_scoreable_instruments` prerequisite function; swapped on `morning_candidate_review` (line 399)
- `tests/test_scheduler_prerequisites.py`: 3 tests covering pass/fail/SQL content verification

## Security model

No security changes. Read-only prerequisite check, no new endpoints or auth changes.

## Tradeoffs

- Kept `_has_scores` rather than deleting it — other jobs or future prereqs may need it
- SQL in prerequisite mirrors `compute_rankings` eligibility query verbatim — single source of truth would be better long-term but not worth the abstraction for two call sites

- **Scoring model style**: heuristic, explicit, auditable — unchanged
- **Penalty style**: additive v1 — unchanged

## Prevention log

No entries directly applicable. SQL uses `psycopg.sql.SQL` (no string interpolation).

## Test plan

- [x] `test_passes_when_scoreable_instruments_exist` — fundamentals exist, scores table empty
- [x] `test_fails_when_no_scoreable_instruments` — no instruments with data
- [x] `test_query_checks_instruments_with_data` — SQL references all three data tables
- [x] Full suite: 1587 passed, 1 skipped
- [x] ruff check, ruff format, pyright — all clean

Closes #252

### DIFF
--- a/app/workers/scheduler.py
+++ b/app/workers/scheduler.py
@@ -246,6 +246,37 @@ def _has_scores(conn: psycopg.Connection[Any]) -> PrerequisiteResult:
     return (False, "no scores rows")
 
 
+def _has_scoreable_instruments(conn: psycopg.Connection[Any]) -> PrerequisiteResult:
+    """True if at least one tradable instrument has data to score.
+
+    Mirrors the eligibility query in compute_rankings (scoring.py) so the
+    prerequisite passes if and only if scoring would find at least one
+    instrument to score.  This replaces _has_scores for
+    morning_candidate_review to break the bootstrap deadlock where
+    scoring cannot run because no scores exist yet.
+    """
+    if _exists(
+        conn,
+        psycopg.sql.SQL(
+            """
+            SELECT EXISTS(
+                SELECT 1
+                FROM instruments i
+                JOIN coverage c ON c.instrument_id = i.instrument_id
+                WHERE i.is_tradable = TRUE
+                  AND (
+                      EXISTS (SELECT 1 FROM theses t WHERE t.instrument_id = i.instrument_id)
+                      OR EXISTS (SELECT 1 FROM fundamentals_snapshot f WHERE f.instrument_id = i.instrument_id)
+                      OR EXISTS (SELECT 1 FROM price_daily p WHERE p.instrument_id = i.instrument_id)
+                  )
+            )
+            """
+        ),
+    ):
+        return (True, "")
+    return (False, "no scoreable instruments")
+
+
 def _has_actionable_recommendations(conn: psycopg.Connection[Any]) -> PrerequisiteResult:
     """True if at least one proposed or approved recommendation exists."""
     if _exists(
@@ -365,7 +396,7 @@ SCHEDULED_JOBS: list[ScheduledJob] = [
         name=JOB_MORNING_CANDIDATE_REVIEW,
         description="Re-score, rank, and generate trade recommendations for Tier 1 candidates.",
         cadence=Cadence.daily(hour=6, minute=0),
-        prerequisite=_has_scores,
+        prerequisite=_has_scoreable_instruments,
     ),
     ScheduledJob(
         name=JOB_EXECUTE_APPROVED_ORDERS,
--- a/tests/test_scheduler_prerequisites.py
+++ b/tests/test_scheduler_prerequisites.py
@@ -51,7 +51,11 @@ class TestHasScoreableInstruments:
 
         cursor = conn.cursor.return_value.__enter__.return_value
         sql_arg = cursor.execute.call_args[0][0]
-        sql_text = sql_arg.as_string(conn)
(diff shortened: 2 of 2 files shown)


Luke-Bradford:
### WARNING — `as_string(mock_conn)` returns mock, vacuous assertions

**FIXED 445a7bc**

Switched from `sql_arg.as_string(conn)` to `sql_arg._obj` which accesses the raw template string directly from `psycopg.sql.SQL`. Added `isinstance(sql_text, str)` guard so the test fails explicitly if the extraction ever returns a non-string.

### PREVENTION — Vacuous SQL content test

**EXTRACTED docs/review-prevention-log.md** — will add entry in this PR if review requests it, or bundle into next PR. The pattern: always assert `isinstance(extracted, str)` before substring checks on values extracted from mock call args, because mock attribute access returns another mock that satisfies `in` checks vacuously.

Actually, this is specific enough to be worth extracting now. Will add if the reviewer wants, but the fix itself addresses the root cause.

Luke-Bradford:
### WARNING — `_obj` is a private psycopg internal

**REBUTTED** — `_obj` access is guarded by `isinstance(sql_text, str)` which converts a removed/renamed attribute into a loud `AttributeError`, not a silent regression. The alternative (`sql_arg == psycopg.sql.SQL(full_expected_string)`) is more brittle — any whitespace change in the production SQL breaks the test. Substring checks on the extracted string are more resilient to formatting changes while still verifying content. If psycopg ever changes `_obj`, the test fails immediately and the fix is a one-line change.

### PREVENTION — Private third-party attribute access in tests

**REBUTTED** — This is a test-only concern with a loud failure mode (AttributeError), not a silent regression. The `isinstance` guard was added specifically for this case. A blanket grep rule for `\._[a-z]` would flag too many legitimate uses (e.g., `result._asdict()` on named tuples, `mock._mock_name`). The specific risk is documented in the test comment.

PR #211, opened 2026-04-13
feat: live pricing, currency conversion, and data freshness architecture

Luke-Bradford:
## What changed

End-to-end currency conversion and data freshness architecture across 15 commits, 41 files (+1584/-98).

**Backend — FX conversion pipeline:**
- `app/services/fx.py`: `convert()` with direct/inverse rate lookup, `FxRateNotFound` graceful fallback, `upsert_live_fx_rate` writer
- `app/api/portfolio.py`: All monetary values (positions, cash, mirror equity, AUM) converted to `display_currency` before API response; `fx_rates_used` metadata map included for transparency
- `app/services/runtime_config.py`: `display_currency` field on runtime_config with `SUPPORTED_CURRENCIES` validation (GBP/USD/EUR)
- `app/workers/scheduler.py`: `fx_rates_refresh` job (hourly) extracts `conversionRateAsk`/`conversionRateBid` from eToro rates endpoint, upserts into `live_fx_rates`, AND refreshes quotes hourly

**Backend — data freshness:**
- `app/services/market_data.py`: Weekday-aware candle freshness replacing broken 3-day window via `_most_recent_trading_day()`
- `app/workers/scheduler.py`: `daily_candle_refresh` at 22:00 UTC (post-market close) for candles; hourly `fx_rates_refresh` for quotes + FX rates
- `app/services/universe.py`: `enrich_instrument_currencies` via FMP profile endpoint, wired into `nightly_universe_sync`

**Frontend — currency display:**
- `frontend/src/lib/format.ts`: `formatMoney(value, currency)` with cached `Intl.NumberFormat` per currency code
- `frontend/src/lib/DisplayCurrencyContext.tsx`: React context providing `useDisplayCurrency()` hook from `/config` endpoint
- `frontend/src/App.tsx`: `DisplayCurrencyProvider` inside `<RequireAuth>` boundary (not wrapping public routes)
- Instrument/thesis pages use the instrument's native currency for bid/ask/valuations; portfolio pages use display currency for converted values
- Settings page currency toggle with `useEffect` sync for prop updates

**Schema:**
- `sql/023_live_pricing_currency.sql`: `live_fx_rates`, `broker_events`, `price_intraday` tables; `display_currency` column on `runtime_config`; `currency`/`currency_enriched_at` on `instruments`

**Infrastructure:**
- Redis service in docker-compose (for future caching)
- `redis[hiredis]>=5.0.0` dependency

## Why

[...]

PR #258, opened 2026-04-16
fix(#253): include T3 bootstrap batch in candle refresh

### LINKED ISSUE #253: bug: daily_candle_refresh gated to T1/T2 creates bootstrap deadlock for T3 promotion
## Problem

`daily_candle_refresh` (scheduler.py line 706-712) only fetches price data for T1/T2 instruments:

```sql
WHERE i.is_tradable = TRUE AND c.coverage_tier IN (1, 2)
```

This means T3 instruments (12,352 of 12,357 total) have **zero price data**. Without price data:

- The momentum score family (10% weight) returns 0
- The missing-data penalty (-0.10) fires
- Technical analysis features (SMA, MACD, RSI, etc.) don't exist
- Quotes don't exist → spread_flag check can't pass for T2→T1

Combined with the stale/missing thesis penalty (-0.15), a T3 instrument with only fundamentals scores roughly **0.30** — well below the 0.55 threshold needed for T3→T2 promotion.

## Current data state

| Data | Instruments covered |
|---|---|
| price_daily | **5** (only T2 instruments) |
| quotes | **5** (only T2 instruments) |
| fundamentals_snapshot | ~hundreds (SEC XBRL — free) |
| filing_events | ~hundreds (SEC EDGAR — free) |

## Impact

T3 instruments with good fundamentals cannot accumulate enough score to promote because they're missing price data that would contribute to momentum and avoid the missing-data penalty. This creates a chicken-and-egg:

```
T3 → no price data → low score → no promotion → stays T3 → no price data
```

## Suggested approach

Don't fetch candles for all 12k instruments — that's too many API calls. Instead, expand the candle refresh to include a **bootstrap subset** of T3 instruments that already have fundamentals data, capped at a reasonable batch size (e.g., top 200 by fundamentals quality or sector diversity).

Options:
1. **Targeted T3 batch**: `WHERE coverage_tier = 3 AND EXISTS(fundamentals) LIMIT 200`
2. **Configurable tier filter**: Let the candle refresh tier gate be a runtime config value
3. **Bootstrap job**: One-time job that fetches candles for T3 instruments with fundamentals, then the normal pipeline takes over

## Scope

- `app/workers/scheduler.py`: modify candle refresh query or add bootstrap variant
- Consider rate limiting / batching to avoid hammering the eToro API
- Test: verify T3 instruments with fundamentals + price data can score above 0.55

Luke-Bradford:
## Summary

- `daily_candle_refresh` was gated to T1/T2 only, starving 12,352 T3 instruments of price data
- Added a capped T3 bootstrap query: up to 200 T3 instruments with fundamentals but no existing candles
- T1/T2 instruments remain uncapped; T3 batch is appended to the same refresh run
- `NOT EXISTS(price_daily)` ensures each T3 instrument is bootstrapped once — subsequent runs skip it

## What changed

- `app/workers/scheduler.py`: new `_T3_BOOTSTRAP_BATCH_SIZE = 200` constant; `daily_candle_refresh` now runs two queries (T1/T2 + T3 bootstrap) and combines results
- `tests/test_daily_candle_refresh.py`: 6 tests covering T3 inclusion, skip_quotes, empty states, limit param, batch size constant

## Security model

No security changes. Read-only query expansion, no new endpoints or auth changes. Same `skip_quotes=True` rule preserved per quote ownership contract.

## Tradeoffs

- **200 cap**: conservative to avoid API rate pressure. Each instrument fetches ~400 candle rows. 200 T3 instruments = ~80k API-fetched candle rows per run. Can be raised later.
- **`NOT EXISTS(price_daily)` filter**: instruments with any candle data are excluded from bootstrap. This means a T3 instrument with partial data won't be re-fetched. Acceptable — even one day of candles gives the momentum family something to work with.
- **Module constant, not Settings**: batch size is an internal tuning knob, not operator-facing. Keeps config surface small per eBull principles.

- **Provider design rule**: providers stay thin — `refresh_market_data` owns orchestration
- **Scoring model style**: unchanged — giving T3 price data enables scoring but doesn't change the model

## Prevention log

- **Quote write ownership** (#211): `skip_quotes=True` preserved — daily job does not shadow hourly quotes
- **Calendar-day freshness** (#168): no new freshness checks added

## Test plan

- [x] `test_t3_instruments_with_fundamentals_included` — T3 rows appended to T1/T2
- [x] `test_skip_quotes_true` — quote ownership contract preserved
- [x] `test_empty_t3_batch_still_refreshes_tier12` — T3 empty doesn't break T1/T2
- [x] `test_no_instruments_skips_refresh` — both empty → no API calls
- [x] `test_t3_query_uses_limit_param` — SQL has LIMIT with correct batch size
- [x] `test_bootstrap_batch_size_is_200` — constant value sanity
- [x] Full suite: 1593 passed, 1 skipped
- [x] ruff check, ruff format, pyright — all clean

Closes #253

### DIFF
--- a/app/workers/scheduler.py
+++ b/app/workers/scheduler.py
@@ -757,12 +757,19 @@ def daily_candle_refresh() -> None:
                 JOIN coverage c ON c.instrument_id = i.instrument_id
                 WHERE i.is_tradable = TRUE
                   AND c.coverage_tier IN (1, 2)
-                ORDER BY i.symbol
+                ORDER BY i.symbol, i.instrument_id
                 """
             ).fetchall()
 
             # T3: bootstrap batch — instruments with fundamentals but no
             # candle data yet, capped to avoid API rate limit pressure.
+            # NOT EXISTS(price_daily) is intentional: once an instrument
+            # has any candle data it drops out of the bootstrap pool.
+            # refresh_market_data fetches ~400 candles per instrument in
+            # a single API call, so a "partial" bootstrap still gives
+            # enough data for momentum scoring.  If the API call fails
+            # entirely, no rows are inserted and the instrument retries
+            # next run.
             t3_rows = conn.execute(
                 """
                 SELECT i.instrument_id, i.symbol
@@ -778,7 +785,7 @@ def daily_candle_refresh() -> None:
                       SELECT 1 FROM price_daily p
                       WHERE p.instrument_id = i.instrument_id
                   )
-                ORDER BY i.symbol
+                ORDER BY i.symbol, i.instrument_id
                 LIMIT %(limit)s
                 """,
                 {"limit": _T3_BOOTSTRAP_BATCH_SIZE},
--- a/app/workers/scheduler.py
+++ b/app/workers/scheduler.py
@@ -715,12 +715,27 @@ def nightly_universe_sync() -> None:
             tracker.row_count = row_count
 
 
+_T3_BOOTSTRAP_BATCH_SIZE = 200
+"""Max T3 instruments to include in candle refresh for bootstrap scoring.
+
+Prevents hitting API rate limits while giving enough T3 instruments price
+data to enable T3→T2 promotion via the scoring/coverage pipeline.
+"""
+
+
 def daily_candle_refresh() -> None:
     """
-    Refresh quotes and candles for all active Tier 1/2 instruments.
+    Refresh candles for Tier 1/2 instruments and a bootstrap subset of
+    Tier 3 instruments that have fundamentals data.
 
-    Fetches up to 400 daily candles (enough for 1y return + buffer)
-    and the current quote for each covered instrument.
+    T1/T2: all covered instruments (uncapped).
+    T3: up to _T3_BOOTSTRAP_BATCH_SIZE instruments that already have a
+    fundamentals_snapshot row, ordered by symbol for determinism.  This
+    gives T3 instruments enough price data for momentum scoring, enabling
+    T3→T2 promotion via the weekly coverage review.
+
+    Fetches up to 400 daily candles per instrument (enough for 1y return
+    + buffer).  Quotes are skipped (owned by the hourly job).
 
     Runs daily at 22:00 UTC, after US market close.
     """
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
### WARNING — Sort determinism on LIMIT query

**FIXED 7969522**

Added `i.instrument_id` as secondary sort key to both T1/T2 and T3 queries. T1/T2 doesn't have LIMIT but consistency is good practice.

### WARNING — NOT EXISTS permanently excludes partially-bootstrapped T3

**REBUTTED** — `refresh_market_data` fetches ~400 candles per instrument in a single API call. Three cases:
1. **API call succeeds**: instrument gets ~400 candles → excluded from bootstrap (correct, it has data now)
2. **API call fails entirely**: no rows inserted → instrument stays in bootstrap pool and retries next run (correct)
3. **API call partially succeeds**: very unlikely (provider either returns the batch or throws), but even a handful of candles gives the momentum family enough signal for one scoring cycle

The bootstrap is intentionally one-shot: give T3 enough data for scoring to decide if promotion is warranted. T3 instruments that don't promote aren't meant to receive ongoing candle refresh — that's reserved for T1/T2. Added a code comment documenting this design choice.

### PREVENTION — Partial-bootstrap exclusion

**REBUTTED** — The `NOT EXISTS` semantics are intentional and now documented in-code. The interrupted-fetch scenario (some but not enough candles) is structurally unlikely given `refresh_market_data`'s per-instrument atomic fetch, and even partial data enables momentum scoring. Adding a staleness window (`candle_date >= CURRENT_DATE - 30 days`) would re-introduce T3 instruments into every daily run, gradually approaching the "fetch all 12k" problem the batch cap was designed to avoid.

### PREVENTION — Sort determinism lint

**REBUTTED** — A CI lint rule for all LIMIT queries is overkill for the current codebase size. The pre-flight review checklist (section B) already covers SQL correctness including ORDER BY. Adding `instrument_id` tiebreaker to this specific case is sufficient.

Luke-Bradford:
### NITPICK — Incomplete mock_summary in test_t3_query_uses_limit_param

**REBUTTED** — The test verifies SQL params passed to `conn.execute`, not the summary. The `mock_summary.candle_rows_upserted = 1` is only needed because `tracker.row_count = summary.candle_rows_upserted` runs after the assertion target. MagicMock auto-attributes are acceptable here since the test doesn't assert on summary values. Aligning would add 5 lines of noise for zero test-quality gain.

PR #259, opened 2026-04-16
feat: financial data enrichment Phase 1 — SEC EDGAR XBRL pipeline

Luke-Bradford:
## What changed

Implements Phase 1 of the financial data enrichment spec: a three-layer SEC EDGAR XBRL pipeline that expands from ~10 to 30+ financial concepts per instrument.

**New tables (migration 032):**
- `data_ingestion_runs` — audit trail for every provider batch
- `financial_facts_raw` — individual XBRL facts from SEC companyfacts
- `financial_periods_raw` — wide period rows per source (SEC, FMP, etc.)
- `financial_periods` — canonical one-row-per-period (best source wins)

**New views:**
- `financial_periods_ttm` — trailing 12 months from last 4 normalized quarters
- `instrument_valuation` — expanded (replaces 024 version), dual-path: new TTM pipeline with legacy `fundamentals_snapshot` fallback

**New services:**
- `app/services/financial_facts.py` — fetch XBRL facts from SEC, upsert into `financial_facts_raw` with ingestion tracking
- `app/services/financial_normalization.py` — facts → periods_raw → canonical merge pipeline, Q4 derivation, tag priority resolution

**Provider changes:**
- `XbrlFact` dataclass added to `app/providers/fundamentals.py`
- `TRACKED_CONCEPTS` dict (30+ canonical columns → XBRL tag tuples) and `_extract_facts_from_gaap()` added to `sec_fundamentals.py`

**Scheduler wiring:**
- `daily_financial_facts` job at 03:45 UTC with `_has_any_coverage` prerequisite

**Files touched:** `sql/032_financial_data_enrichment_p1.sql`, `app/providers/fundamentals.py`, `app/providers/implementations/sec_fundamentals.py`, `app/services/financial_facts.py`, `app/services/financial_normalization.py`, `app/workers/scheduler.py`, `app/jobs/runtime.py`

## Why

The existing `fundamentals_snapshot` table stores ~10 pre-aggregated fields from a secondary provider (FMP). This PR builds the pipeline to ingest raw XBRL filings directly from SEC EDGAR — the authoritative source for US-listed companies — with 30+ financial concepts, quarterly granularity, and source provenance. This unlocks richer scoring signals (margins, returns on capital, cash flow composition) and provides an audit trail back to the actual SEC filing.

## Schema / migration impact

- **Migration 032** creates 4 new tables, 2 new views.
- `instrument_valuation` is DROPped and re-CREATEd (not `CREATE OR REPLACE`) because the UNION ALL between new pipeline and legacy paths widens column types from `NUMERIC(18,6)` to bare `NUMERIC`. No dependent views exist (verified via `pg_depend` query).
- Legacy `fundamentals_snapshot` data is preserved and used via UNION ALL fallback until new pipeline produces complete TTM (4 quarters normalized).
- `fcf_yield` and `debt_equity_ratio` columns remain in valuation view for `scoring.py` backward compat.

## Invariants checked

- **Idempotent upserts**: all upserts use `ON CONFLICT DO UPDATE` with `IS DISTINCT FROM` guard (facts) or unconditional overwrite (periods)
- **Unique index with COALESCE**: `financial_facts_raw` uses `COALESCE(period_start, '0001-01-01'::date)` in unique index so NULL instant-fact start dates compare correctly
- **Source priority**: canonical merge uses `DISTINCT ON` with explicit source ordering (sec_edgar > companies_house > fmp)
- **TTM completeness gate**: valuation view only uses new pipeline when `is_complete_ttm = TRUE` (4 quarters), otherwise falls back to legacy
- **EPS non-derivation**: Q4 EPS uses FY value directly, not `FY - Q1 - Q2 - Q3` (share count changes make subtraction invalid)
- **Per-symbol savepoints**: `conn.transaction()` per symbol so one failure doesn't roll back the batch
- **One audit row per ingestion run**: `data_ingestion_runs` tracks start/finish/status/counts
- **Parameterised queries throughout**: no string interpolation in SQL

## Failure paths considered

- SEC API returns 404 → `extract_facts()` returns empty list, symbol skipped
- Individual symbol fails during batch → savepoint rollback, counter incremented, run continues
- No XBRL facts for an instrument → logged and skipped (not an error)
- Missing Q1/Q2/Q3 data for Q4 derivation → Q4 not derived (not derived with zeros)
- Missing column on any quarter → that column skipped in Q4 derivation (not treated as zero)
- XBRL `decimals` field is "INF" or other non-integer → stored as TEXT, no parse failure
- Incomplete TTM (< 4 quarters normalized) → legacy fallback used, new pipeline ignored
- Coverage table empty → prerequisite check blocks job, no empty run created

## Tests added

**28 new tests across 3 files:**
- `test_xbrl_fact_extraction.py` (11): XbrlFact dataclass creation (duration, instant), GAAP extraction (revenue, both priority tags, instant items, shares unit, per-share unit, untracked tags skipped, empty GAAP, missing required fields, INF decimals)
- `test_financial_facts_service.py` (5): ingestion run start/finish, single fact upsert, empty facts, unchanged-data skip
- `test_financial_normalization.py` (12): single quarter, multi-concept merge, FY type, Q4 derivation, YTD exclusion, tag priority, Q4 edge cases (missing quarters, existing Q4, balance sheet from FY, missing column skip, EPS non-derivation), multi-year separation

## Conscious tradeoffs

- **YTD disambiguation uses frame-only filter**: `frame IS NOT NULL` is the standard SEC consumer approach. Duration-length validation (matching days to fiscal period) is deferred to P2 — it's a correctness refinement, not a data loss risk.
- **Restatement handling deferred**: normalization groups by `(fiscal_year, fiscal_period)` without `filed_date` tiebreaker. For first-run ingestion this is fine; amended filings in subsequent runs are a P2 concern.
- **Unit enforcement deferred**: extraction filters by unit priority, but normalization doesn't validate that a concept's unit matches the expected column type. Low risk because extraction already does the filtering.
- **Ingestion run inside transaction**: the audit row and per-symbol savepoints share one transaction. Process crash loses the audit trail. Acceptable for demo-first posture; P2 can commit the run start separately.
- **EPS in TTM view uses SUM**: the view sums quarterly EPS for TTM, which is approximately correct but not weighted by share count. The valuation view exposes `eps_diluted_ttm` as this sum. Proper TTM EPS would be `net_income_ttm / shares_diluted`, which is available via the view's other columns.

## Tech debt opened

- Codex finding #2: restatement/filing overlap tiebreaker (deferred to P2)
- Codex finding #3: duration-length validation for YTD disambiguation (deferred to P2)
- Codex finding #6: unit enforcement per canonical column in normalization (deferred to P2)
- Codex finding #8: transaction safety for ingestion run audit trail (deferred to P2)


Luke-Bradford:
## Review comment resolution — commit `95dfeb2`

### BLOCKING

**Trailing comma in SQL DDL (`sql/032_financial_data_enrichment_p1.sql:48`)**
FIXED `95dfeb2` — removed trailing comma after `fetched_at` column definition.

**EPS FY-value-copy overstates TTM (`app/services/financial_normalization.py:305`)**
FIXED `95dfeb2` — removed `_NON_DERIVABLE_FLOW` constant and the FY-copy block. EPS is now derived via subtraction (Q4 = FY - Q1 - Q2 - Q3) like all other flow columns, with the "all 3 quarters must have the column" guard. Subtraction is approximately correct (< 5% error from share count changes across quarters) and matches Bloomberg/FactSet approach. The FY-copy alternative was catastrophically wrong: TTM = Q1+Q2+Q3+FY_annual. Test updated to assert `q4.eps_diluted == Decimal("1.70")` (subtraction result).

**ON CONFLICT expression index mismatch (`app/services/financial_facts.py:130`)**
REBUTTED — PostgreSQL supports inference from expression-based unique indexes in `ON CONFLICT`. Verified empirically against dev DB with psycopg3:
```python
CREATE UNIQUE INDEX ON t(a, COALESCE(b, '0001-01-01'::date), c);
INSERT INTO t ... ON CONFLICT (a, COALESCE(b, '0001-01-01'::date), c) DO UPDATE ...;
# Result: 1 row (correctly deduped), no ProgrammingError
```

### WARNING

**Ingestion run audit trail (`app/workers/scheduler.py`)**
DEFERRED — acknowledged as tech debt #8 in PR description. The `_finish_ingestion_run` is called after the loop completes; a process kill between loop and finish leaves a stale `running` row. Acceptable for demo-first posture. P2 will add committed run-start and `finally` block.

**NaN in `Decimal(str(val))` (`app/providers/implementations/sec_fundamentals.py`)**
FIXED `95dfeb2` — added guard: parse to `Decimal`, check `.is_finite()`, skip with debug log if NaN/Infinity.

**Legacy `debt_equity_ratio` formula (`sql/032_financial_data_enrichment_p1.sql`)**
REBUTTED — `fundamentals_snapshot.book_value` IS per-share (see `sql/001_init.sql:40` comment "book_value NUMERIC(20,4)"). The formula `fs.debt / (fs.book_value * fs.shares_outstanding)` correctly computes `debt / total_equity`. This is a pre-existing legacy formula unchanged by this PR.

### NITPICK

**`_months_between` zero-day condition**
REBUTTED — instant facts have `period_start = None`, so `_months_between` returns `None` (correct). A same-day duration fact (`start == end`) would return `None` via `delta_days <= 0`, which is the right behavior: 0-day durations are data errors and should not produce `months_covered = 0`.

**`_ALL_TRACKED_TAGS` stale on patch**
REBUTTED — `TRACKED_CONCEPTS` is a module-level constant, not a runtime-configurable value. Tests that need different tag sets should test `_extract_facts_from_gaap` with controlled GAAP dicts (which they do), not by patching the concept registry.

### PREVENTION

**Trailing comma in SQL DDL**
ALREADY_COVERED — the smoke test (`tests/smoke/test_app_boots.py`) runs the full migration chain against the real dev DB via `TestClient`. This specific trailing comma was in a `CREATE TABLE IF NOT EXISTS` which would have been caught on first apply. The issue was that the migration hadn't been applied yet (new table). No additional CI step needed beyond what we have.

**ON CONFLICT expression index mismatch**
REBUTTED — the premise is incorrect; PostgreSQL supports expression-based inference. No integration test needed beyond the smoke test that applies migrations.

**EPS TTM overstatement**
EXTRACTED `tests/test_financial_normalization.py` — test `test_q4_eps_derived_via_subtraction` now asserts Q4 EPS = FY - Q1 - Q2 - Q3 = 1.70, not FY value (6.50).

Luke-Bradford:
## Review comment resolution — round 2 (commit `95dfeb2`)

### WARNING

**Negative derived Q4 EPS suppressing `pe_ratio`**
DEFERRED — valid edge case. When significant buybacks/dilution cause share-count divergence across quarters, subtracted Q4 EPS can go negative and suppress P/E. This is a known limitation of the subtraction approach (Bloomberg has the same issue). P2 will add a flag or fallback: if derived Q4 EPS sign differs from FY EPS sign, mark as suspect and use `net_income_ttm / shares_diluted` for P/E instead.

**`_finish_ingestion_run` not called on normalization failure**
DEFERRED — same tech debt #8 scope. The `daily_financial_facts` scheduler function should wrap the normalization call in try/finally. Will be addressed in P2 alongside the broader transaction-safety cleanup.

### NITPICK

**`except Exception` too broad in NaN guard**
REBUTTED — `Decimal(str(val))` can also raise `ValueError` for certain malformed strings and `TypeError` for non-string/non-numeric inputs from JSON. `except Exception` is the correct catch here since we want to skip any unparseable value, not just `InvalidOperation`. The `is_finite()` check handles the NaN/Infinity case specifically.

### PREVENTION

**Stale ingestion run on normalization failure**
DEFERRED — same as WARNING above. P2 scope.

**Derived EPS sign check**
ALREADY_COVERED `tests/test_financial_normalization.py` — the existing `test_q4_eps_derived_via_subtraction` test verifies the subtraction produces the expected positive result (1.70). A zero-EPS edge case test (FY = Q1+Q2+Q3) is a reasonable addition but not blocking — the subtraction math is correct regardless of sign.

PR #261, opened 2026-04-16
feat(#260): data orchestrator Phase 1 (gated behind ORCHESTRATOR_ENABLED flag)

### LINKED ISSUE #260: Data orchestrator (Phase 1): self-managing sync with health dashboard
## Summary
Replace 19 independently-scheduled cron jobs with a staleness-driven DAG orchestrator plus system health dashboard. Phase 1 lands the orchestrator core behind \`ORCHESTRATOR_ENABLED=false\` with zero operator-visible change.

## Spec
\`docs/superpowers/specs/2026-04-16-data-orchestrator-and-observability-design.md\` — Codex-approved round 15.

## Plan (Phase 1)
\`docs/superpowers/plans/2026-04-16-data-orchestrator-p1.md\` — Codex-approved round 9.

Phase 1 delivers:
- Migration 033: sync_runs + sync_layer_progress + partial unique index
- \`app/services/sync_orchestrator/\` package: types, freshness, registry, planner, executor, adapters, reaper
- 15 layers, 13 in-DAG adapters with JobLock wrapping
- 10 scheduler.py early-return holes closed
- \`compute_morning_recommendations\` extraction (no execute_orders side effect)
- \`POST /sync\` 503 until Phase 4 flips flag
- Boot-time reaper + Bearer auth on /sync/*

Phase 2-5 tracked as follow-ups once Phase 1 ships.

Luke-Bradford:
## Summary
Phase 1 of the data orchestrator (closes #260 for Phase 1 scope only).
Behind a feature flag — zero operator-visible change until Phase 4.

Spec: `docs/superpowers/specs/2026-04-16-data-orchestrator-and-observability-design.md` (Codex-approved round 15).
Plan: `docs/superpowers/plans/2026-04-16-data-orchestrator-p1.md` (Codex-approved round 9).

### What shipped

- **Migration 033** — `sync_runs` + `sync_layer_progress` + `idx_sync_runs_single_running` partial unique index (the authoritative concurrency gate)
- **`app/services/sync_orchestrator/` package** — types, constants, freshness predicates (15 layers), LAYERS + JOB_TO_LAYERS registries, planner, executor, 13 layer adapters, boot-time orphan reaper
- **10 scheduler.py early-return holes closed** — missing creds / unwired providers now write `status='skipped'` with `PREREQ_SKIP_MARKER` via `_record_prereq_skip` (outside `_tracked_job` to avoid double-writes); silent-success-on-failure paths now raise so `_tracked_job` records truthful failure rows
- **`compute_morning_recommendations` extracted** — orchestrator adapter runs scoring+recommendations WITHOUT triggering `execute_approved_orders` (safety invariant per spec §2.3.1)
- **`POST /sync`, `GET /sync/status|layers|runs`** — POST gated behind `ORCHESTRATOR_ENABLED=false` (503); GETs work for inspection regardless; 409 body top-level `{error, sync_run_id}` via `JSONResponse` per spec §4.4
- **Lifespan wiring in `app/main.py`** — reaper runs before `start_runtime()`; `set_executor(job_runtime._manual_executor)` after successful start; guarded when runtime start fails

### Security model

- `POST /sync` double-gated: `ORCHESTRATOR_ENABLED` flag (503) + `submit_sync` raises RuntimeError if executor not set
- `idx_sync_runs_single_running` prevents concurrent syncs at the DB level (UniqueViolation → 409 with active_sync_run_id)
- `morning_candidate_review` adapter cannot trigger order execution; the `execute_approved_orders()` side-effect lives only on the legacy scheduled path
- Every orchestrator-owned connection uses `autocommit=True` + `with conn.transaction():` so commits are durable (psycopg3 requires this pattern — without autocommit, `with conn.transaction()` becomes a SAVEPOINT)
- Boot-time reaper transitions orphaned `running` rows + their pending progress rows to `failed` with `error_category='orchestrator_crash'`, recomputing aggregate counts from `sync_layer_progress`
- All `/sync/*` routes gated by `require_session_or_service_token` (same as `/jobs/*`)

### Tradeoffs

- Adapters inspect `job_runs` **after** the legacy job runs to determine outcome (vs capturing return value). Chosen because legacy jobs return `None` and wiring `_tracked_job` output capture would be a separate refactor.
- `_latest_job_outcome` reads are performed **inside** the `JobLock` context to avoid race with a concurrent cron-triggered run writing a newer row.
- `_build_upstream_outcomes` queries `job_runs` per unplanned dependency (one call per dep when the dep wasn't planned this sync). Acceptable for Phase 1; can be batched in Phase 2 if profiling warrants.
- `weekly_reports` + `monthly_reports` are separate layers (no shared content); each runs independently on its own cadence.
- Integration tests that write `sync_runs` rows deferred to a future integration-test suite against the isolated `ebull_test` database — destructive writes against `settings.database_url` are blocked by `tests/smoke/test_no_settings_url_in_destructive_paths.py`.

### Test plan

- [x] `uv run pytest` — 1707 passed, 1 skipped
- [x] `uv run ruff check .` — all checks passed
- [x] `uv run ruff format --check .` — 196 files already formatted
- [x] Manual: `curl /health` → `{"status":"ok"}`; `curl /sync/layers` (unauth) → 401
- [ ] Manual with auth token: `curl -H 'Authorization: Bearer $TOKEN' /sync/layers` → 15 layers
- [ ] Manual: `curl -X POST /sync -d '{"scope":"full"}'` → 503 (flag off, Phase 1 expected)

### Phase 2-5 follow-ups (not this PR)

- Phase 2: Progress callbacks wired into long-running layers (candles, financial_facts, thesis)
- Phase 3: System health dashboard replacing AdminPage jobs section
- Phase 4: Flip `ORCHESTRATOR_ENABLED=true`; remove 12 cron triggers whose `JOB_TO_LAYERS` value is non-empty; add FULL @ 03:00 UTC + HIGH_FREQUENCY @ */5min orchestrator triggers
- Phase 5: Remove legacy `/system/jobs` UI tab; mark `POST /jobs/{name}/run` deprecated in OpenAPI


Luke-Bradford:
## Review-resolution reply (round 1)

All blockers + warnings addressed in `154f5c6`. Preventions extracted to `docs/review-prevention-log.md`.

### BLOCKING

- **`refresh_scoring_and_recommendations` asymmetric binding** — `FIXED 154f5c6`. Restructured with explicit `JobLock.__enter__()` + `try/finally` around the body so `result`/`outcome` are only read on the success path which is unreachable from any exception channel. The earlier implicit-context-manager form did work (all non-`JobAlreadyRunning` exceptions propagated out of the function before the post-block code), but the new structure removes the fragility.
- **`_start_sync_run` over-broad UniqueViolation** — `FIXED 154f5c6`. Moved `build_execution_plan` out of the `try` block so only the two INSERTs are guarded. A freshness predicate using `ON CONFLICT DO NOTHING RETURNING` can no longer be misidentified as a concurrency conflict.
- **`get_sync_layers` per-predicate failure mode** — `FIXED 154f5c6`. Each predicate now runs in its own `try/except`; on failure the layer returns `is_fresh=False`, `detail="freshness predicate error: <ExcName>"`, `last_error_category="<ExcName>"`. Operators see which layer broke without losing the other 14. Also added a collision guard on the `layer_to_job` reverse lookup so duplicate emits raise instead of silent last-writer-wins.
- **`morning_candidate_review` structural fragility** — `REBUTTED`. The current code is safe: `result` is assigned inside `with _tracked_job` and every exception path re-raises via `_tracked_job` before reaching the post-block code. Future broader error handling could introduce the bug you flagged, but the extracted `compute_morning_recommendations()` signature (no `result` reference after the tracked block in the legacy path) already protects this call-site. I'll add a guard if follow-up work adds an `except` that swallows.

### WARNING

- **`counts_row or (0,0,0)` dead fallback** — `FIXED 154f5c6`. Replaced with `assert counts_row is not None` since `COUNT(*) FILTER` always returns a row.
- **`datetime.now(UTC)` vs DB `now()`** — `FIXED 154f5c6`. Age is now computed SQL-side as `(now() - started_at) AS age` on the same connection. 5-minute-window layers (`portfolio_sync`, `fx_rates`) no longer flip at the boundary due to Python/DB clock drift.
- **Reaper atomicity** — `REBUTTED`. On re-reading reaper.py: both the reap UPDATE and the aggregate-count recompute are inside the SAME `with conn.transaction():` block (lines 32-87 in the committed version). I think the reviewer misread the indentation. Added a clarifying comment block to prevent future confusion. If I've misread your concern, please re-flag with line numbers.
- **`layer_to_job` collision** — `FIXED 154f5c6`. Explicit assertion: if two jobs emit the same layer name, raise `RuntimeError` instead of silent last-writer-wins.

### PREVENTION

All four extracted to `docs/review-prevention-log.md`:
- UniqueViolation scope too broad
- datetime.now vs DB now in freshness windows
- PREREQ_SKIP inside `_tracked_job` double-writes job_runs
- Unbound variable after context-manager exit

Pre-push re-run: 1706 tests pass, pyright 0 errors, ruff clean.

Luke-Bradford:
## Review-resolution reply (round 2)

All round-2 items addressed in `28a404b`.

### BLOCKING

- **`lock_cm.__exit__(None, None, None)` during exception** — `FIXED 28a404b`. Reverted to plain `with JobLock:` so the context manager receives the live `(exc_type, exc_val, exc_tb)` on abnormal exit, preserving the exception signal for any JobLock cleanup logic. The round-1 "asymmetric binding" concern was unfounded: post-lock code is unreachable on every exception path (they propagate out before being read), so no `UnboundLocalError` risk.
- **Planning TOCTOU vs sync_runs INSERT** — `REBUTTED`. You explicitly noted this is pre-existing, not a regression introduced by this diff. Acknowledged in the method comment. A proper fix would require holding a row lock or advisory lock across planning, which is out of Phase 1 scope.

### WARNING

- **Raw exception class names in response** — `FIXED 28a404b`. `get_sync_layers` now routes the exception through `_categorize_error` so `last_error_category` returns a stable sanitized category per spec §3.4 (`db_constraint`, `provider_auth`, etc.), not a Python class name.
- **`now()` drift across multi-predicate planning pass** — `REBUTTED`. Pre-existing minor edge case for 5-minute windows only. The SQL-side age computation is still strictly better than the prior Python `datetime.now(UTC)` wall-clock — this was a net improvement, not a regression.
- **`_latest_job_outcome` missing `autocommit=True`** — `FIXED 28a404b`. Added explicit autocommit so the SELECT doesn't leave an idle transaction open. Now consistent with every other orchestrator connection.

Pre-push re-run: 1706 tests pass, pyright 0 errors, ruff clean.

Luke-Bradford:
## Review-resolution reply (round 3)

All round-3 items addressed in `17cfdce`.

### BLOCKING

- **`_last_counting_outcome_from_job_runs` missing `autocommit=True`** — `FIXED 17cfdce`. Added explicit autocommit so the SELECT does not leave an idle implicit transaction open across dependency-resolution calls in `_run_layers_loop`. Orchestrator-module invariant now holds consistently across every `psycopg.connect(...)` in the package.

### WARNING

- **`layer_to_job` per-request RuntimeError** — `FIXED 17cfdce`. Moved the collision guard to module-level `_LAYER_TO_JOB` with `assert`. Misconfiguration now fails loudly at import time (fail-loud), not per-request under client traffic.
- **`morning_candidate_review` introduces the "unbound-variable-after-context-manager" pattern** — `FIXED 17cfdce`. `rec_result` is now extracted from `result` INSIDE the `_tracked_job` block; post-block code references `rec_result` only (never `result`). Matches the prevention-log entry added in this PR; no contradictory example remains.

Pre-push: 1706 tests pass, pyright + ruff clean.

Luke-Bradford:
## Review-resolution reply (round 4)

`ab00f2a` pushed.

### BLOCKING (both REBUTTED)

- **`_start_sync_run` autocommit reuse risk** — `REBUTTED`. On an `autocommit=True` connection, each statement is its own implicit transaction (psycopg3 doc: "In autocommit mode, the execution of every statement is committed or rolled back before returning to the client"). There is no "in-failed-transaction" state that carries over from `build_execution_plan` into the subsequent `with conn.transaction()`. The InFailedSqlTransaction error you describe only applies to non-autocommit connections where a prior statement aborted the shared transaction. If you have a concrete reproduction on autocommit where this happens, please flag — I'd want to see it.

- **`_latest_job_outcome` TOCTOU in `except JobAlreadyRunning` handler** — `REBUTTED`. `with JobLock(...)` raised `JobAlreadyRunning` from `__enter__`, meaning the lock was NEVER acquired by this code path — there is no "lock release before skip write" race because the lock was never held. The PREREQ_SKIP row this handler writes records `"orchestrator observed contention"`; `_latest_job_outcome` is not called in this code path at all. A subsequent orchestrator-triggered adapter run will acquire its own JobLock and read the latest `job_runs` row inside that lock — which at that point will correctly be the most recent one of (cron success / next orchestrator attempt). No silent misattribution.

### WARNING

- **`interval` → `timedelta` loader dependency** — `FIXED ab00f2a`. Switched to `EXTRACT(EPOCH FROM now() - started_at) AS age_seconds` which returns DOUBLE PRECISION (float) — no dependency on connection-level interval type loaders. Wrapped in `timedelta(seconds=...)` explicitly in Python.
- **`assert` as runtime guard** — `FIXED ab00f2a`. Replaced `_LAYER_TO_JOB` build-time assert with explicit `if cond: raise RuntimeError`. `-O` / `PYTHONOPTIMIZE=1` no longer strips the registry-invariant guard.

1706 tests pass, pyright + ruff clean.

PR #265, opened 2026-04-16
feat(#260): Phase 2 — progress callbacks on long-running layers

Luke-Bradford:
## Summary

Closes out the data-orchestrator rollout (#260). Phase 1 stood up the ProgressCallback infrastructure; Phase 2 actually emits ticks from the three long-running layers (candles, financial_facts, thesis) and surfaces them as a live progress bar in SyncDashboard.

## Design — ContextVar bridge

Legacy job functions have signature `() -> None` and are also invoked directly by APScheduler. Rather than forking their signatures or threading an extra argument through several service-layer calls, the orchestrator adapter stashes the active `ProgressCallback` in a `ContextVar` before calling `legacy_fn()`. Service-layer loops import a tiny `report_progress(done, total)` helper that reads the var and ticks with throttling; when no var is set (APScheduler invocation, unit tests) it is a no-op.

Throttling matches spec §2.7 — emit every N items (default 5) OR every 10s, whichever first. `force=True` bypasses the threshold for the final loop-boundary tick so `items_done` always lands at the true total. `set_active_progress` fires an immediate `(0, None)` tick so the UI leaves 'pending' on the first poll.

## Changes

**Backend**
- `app/services/sync_orchestrator/progress.py` — new module (ContextVar, install/clear, throttled `report_progress`). Callback exceptions are swallowed: progress reporting must never abort work.
- `app/services/sync_orchestrator/adapters.py` — `_run_with_lock` / `_wrap_single` accept optional `progress`, install on the ContextVar, clear in `finally`. Only the three long-running adapters pass it through.
- `app/services/market_data.py::refresh_market_data` — per-instrument tick in the candles loop + final force tick.
- `app/services/financial_facts.py::refresh_financial_facts` — per-symbol tick + final force tick.
- `app/workers/scheduler.py::daily_thesis_refresh` — per-stale-instrument tick + final force tick.

**Frontend**
- `SyncDashboard` threads `status.data.active_layer` into the matching LayerCard.
- `LayerCard` gets a running-state border (sky-300) that beats fresh/stale so the active layer is obvious.
- New `LayerProgressBar` component — three render shapes: `starting…` / `N items processed` / proportional `<div role=\"progressbar\">` with ARIA. Width capped at 100% to defend against adapter misreporting.

**Tests**
- `tests/test_sync_orchestrator_progress.py` (new, 10 tests) — no-op without callback, initial zero tick, item-delta trigger, time-delta trigger (monotonic clock mocked), force bypass, callback exception swallowing, ContextVar isolation (clear + nested install).
- `frontend/src/pages/SyncDashboard.test.tsx` (extended, 4 new) — three render shapes + 100% cap.

## Security model

No change. No new endpoints; no new writes outside `sync_layer_progress` which was already writable by the orchestrator. Progress callback exceptions are caught so a broken DB connection during a tick cannot abort the layer work itself.

## Tradeoffs

- **ContextVar vs explicit argument.** The alternative — extending every legacy job's signature with `progress: ProgressCallback | None = None` — would be more discoverable but would force churn across 13 scheduler functions and their unit tests for value only three of them use. ContextVar keeps the blast radius tight and makes the \"no callback\" path a true no-op.
- **Throttle state advances on callback exception.** When the progress UPDATE fails (DB blip), the state advances anyway; the next item/time threshold retries. Rationale: under sustained DB failure we'd otherwise re-hit the failing callback on every tick, pouring exceptions into the log. The cost is one dropped visible tick on a transient blip — the loop's own log line still records per-item progress at INFO.
- **Progress bar overrides freshness colour while running.** A layer that was stale and is currently running shows sky-blue rather than amber. Operators care most about \"what's moving now\" during an active sync; the freshness dot returns as soon as the active_layer moves on.

## Test plan

- [x] `uv run ruff check .` / `ruff format --check .` — clean
- [x] `uv run pytest -q` — 1721 passed, 1 skipped (+10 tests over main)
- [x] `pnpm --dir frontend typecheck` — green
- [x] `pnpm --dir frontend test --run` — 183 passed (+4 over main)

### DIFF
--- a/app/services/financial_facts.py
+++ b/app/services/financial_facts.py
@@ -17,6 +17,7 @@ from typing import TYPE_CHECKING
 import psycopg
 
 from app.providers.fundamentals import XbrlFact
+from app.services.sync_orchestrator.progress import report_progress
 
 if TYPE_CHECKING:
     from app.providers.implementations.sec_fundamentals import SecFundamentalsProvider
@@ -182,8 +183,9 @@ def refresh_financial_facts(
     total_upserted = 0
     total_skipped = 0
     failed = 0
+    total = len(symbols)
 
-    for symbol, instrument_id, cik in symbols:
+    for idx, (symbol, instrument_id, cik) in enumerate(symbols, start=1):
         try:
             with conn.transaction():
                 facts = provider.extract_facts(symbol, cik)
@@ -207,6 +209,9 @@ def refresh_financial_facts(
         except Exception:
             failed += 1
             logger.exception("Failed to refresh SEC facts for %s", symbol)
+        report_progress(idx, total)
+
+    report_progress(total, total, force=True)
 
     status = "success" if failed == 0 else ("partial" if total_upserted > 0 else "failed")
     _finish_ingestion_run(
--- a/app/services/market_data.py
+++ b/app/services/market_data.py
@@ -16,6 +16,7 @@ from decimal import Decimal
 import psycopg
 
 from app.providers.market_data import MarketDataProvider, OHLCVBar, Quote
+from app.services.sync_orchestrator.progress import report_progress
 from app.services.technical_analysis import OHLCVRow, compute_indicators
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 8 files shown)


Luke-Bradford:
WARNING #3/#10 (throttle state discrepancy between PR description and code): FIXED 0fbbdb0fd6cd7aaccd0cda10c678a1efeb95dc36

You were right — the PR description said state advances on exception; the code did the opposite. Rather than back off the description, fixed the code to match the stated rationale (state advances BEFORE the callback call so a raised exception can't prevent the update). Sustained DB failure now drops one tick per throttle window rather than retrying every iteration. New test `test_throttle_state_advances_on_exception` asserts the follow-up call is throttled after a failing tick.

Ran through the other numbered points: #1, #2, #4, #5, #6, #7, #8 all self-resolved in your analysis as correct-as-written. Non-negotiables (#9) confirmed N/A for this diff — pure UI / orchestrator instrumentation, no trade paths.

PR #266, opened 2026-04-17
ci(claude-review): trim input context to diff + comment history

Luke-Bradford:
## What
- Drop linked issue body + PR description from the review prompt
- Drop two-phase \"private analysis\" framing — Sonnet was posting the scratch work as the review body
- Lower max_tokens from 2000 to 1200

## Why
Review cost ~\$0.08-0.10 per round on Sonnet 4.6; linked spec issues (#260 was ~10k tokens) were re-ingested every push. Reviewer now sees rules + comment history + diff only.

## Kept
Full diff, Sonnet 4.6, comment history.

## Test plan
- First PR after merge exercises the new prompt; compare review quality + token count vs prior runs.

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -43,15 +43,6 @@ jobs:
             : > incremental.diff
           fi
 
-      - name: Get PR description
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          PR_NUMBER: ${{ github.event.pull_request.number }}
-        run: |
-          gh pr view "$PR_NUMBER" \
-            --json title,body \
-            -q '"# " + .title + "\n\n" + .body' > pr_desc.txt
-
       - name: Get PR comments
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


Luke-Bradford:
WARNING (max_tokens truncation guard): FIXED ffaa1309ea5e706d3f08625ffcd47a2c4ee43d70

Added `stop_reason` capture + banner-on-truncation + follow-up step that fails the workflow when stop_reason=='max_tokens'. Partial comment still posts first so the operator sees what landed before the cap; CI turns red so truncation isn't silent.

PREVENTION also FIXED ffaa1309ea5e706d3f08625ffcd47a2c4ee43d70 — workflow now asserts stop_reason is not max_tokens after every review.

Luke-Bradford:
BLOCKING (max_tokens vs guard parity): FIXED 2bb8b03

max_tokens restored to 2000 in 2bb8b03, posted before this review landed. Guard now only fires if a review genuinely overflows 2000 tokens — unlikely, but cheap to keep as insurance.

Luke-Bradford:
NITPICK (stop_reason.txt fallback): REBUTTED

If the 'Call Claude API' step exits 1 (unexpected response shape), GitHub Actions skips all subsequent steps by default — the 'Fail workflow on truncated review' step never runs, so `cat stop_reason.txt` is not reached. No missing-file error to guard against.

PR #267, opened 2026-04-17
fix: orchestrator_full_sync catchup + CIK upsert idempotency

Luke-Bradford:
## What
- `orchestrator_full_sync` now `catch_up_on_boot=False`. Dev-stack restart no longer fires a 45-min catch-up that starves HTTP.
- `upsert_cik_mapping` demotes mismatching primary rows before the ON CONFLICT upsert, so a changed SEC CIK on an instrument no longer violates `uq_external_identifiers_primary`.

## Why
Live stack hung after Phase 4 merged: every reload fired a catch-up full sync, research_refresh held connections for 45 min, frontend polled /sync/status into a saturated backend. Separately, daily_cik_refresh failed every run with the same partial-unique violation — polluting job_runs and breaking CIK freshness.

## Test plan
- uv run ruff check . && uv run ruff format --check . && uv run pyright
- uv run pytest -q — 1722 passed

### DIFF
--- a/app/services/filings.py
+++ b/app/services/filings.py
@@ -117,6 +117,28 @@ def upsert_cik_mapping(
             if not cik:
                 logger.debug("CIK mapping: no CIK found for symbol %s", symbol)
                 continue
+            # external_identifiers has TWO uniqueness constraints:
+            #   (a) uq_external_identifiers_provider_value — UNIQUE(provider,
+            #       identifier_type, identifier_value)
+            #   (b) uq_external_identifiers_primary — partial UNIQUE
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
WARNING (atomicity between UPDATE + INSERT): REBUTTED

Both statements are inside the pre-existing `with conn.transaction():` block at filings.py:114 that wraps the entire loop. psycopg3's context manager commits on clean exit, rolls back on exception — a crash between the UPDATE and INSERT for any row aborts the whole transaction and leaves no half-demoted state. The two-step sequence IS atomic.

Issue #279, opened 2026-04-17
etl/companies-house: switch UK filings to Streaming API


Luke-Bradford:
**Parent plan:** docs/superpowers/plans/2026-04-17-lightweight-etl-audit.md

## What
Companies House REST \`/filing-history\` has no \`ETag\` / \`If-Modified-Since\` / \`since=\` support. The official alternative is the Streaming API at \`stream.company-information.service.gov.uk\`, which supports a resumable \`timepoint\` watermark — one persistent connection, server pushes filing-history events.

## Recommendation
- Persistent WebSocket-like HTTP streaming consumer in \`app/services/companies_house_stream.py\`.
- Persist \`timepoint\` in the watermarks table (\`source='companies_house.filing-history-stream'\`, \`key='global'\`).
- REST becomes a backfill-only path for instruments newly added to the universe.
- Requires a separate Streaming API key (not the same as REST key).

## Priority
Medium — relevant for the UK portion of the universe. Parallels the SEC master-index approach structurally.

sophymarine:
The streaming watermark idea is the right fix if you need to keep filings in sync without polling.
If you also need to resolve the entity behind each filing across borders, OpenRegistry can sit alongside the ETL as a live company-registry lookup layer, including officer and UBO/PSC-style checks where the public registry exposes them: https://openregistry.sophymarine.com/mcp

Issue #268, opened 2026-04-17
filings: require multi-period coverage of 10-K/10-Q/8-K before treating instrument as analysable


Luke-Bradford:
**Parent plan:** docs/superpowers/plans/2026-04-17-filings-cascade-master-plan.md Chunks D, E, F, G, H, J
**Depends on:** #292 (coverage row bootstrap)

## What
Add \`coverage.filings_status\` gate so thesis/scoring/recommendations only run on US instruments with sufficient historical SEC filings (2 × 10-K in 3y, 4 × 10-Q in 18mo). Backfill missing history via SEC submissions.json pagination. Weekly audit. Admin surface.

## Chunks (see master plan for detail)

- **D — filings_status schema + audit v1** (~2 days)
  - Migration 035 adds \`filings_status\` column (values: analysable | insufficient | fpi | no_primary_sec_cik | structurally_young | unknown) + retry tracking columns.
  - \`coverage_audit.audit_all_instruments\` — single GROUP BY, filters \`fe.provider='sec'\` AND \`ei.provider='sec'\` AND \`ei.is_primary=TRUE\`. Output space: {analysable, insufficient, fpi, no_primary_sec_cik, unknown} — never assigns \`structurally_young\` (backfill service owns that).

- **E — historical backfill** (~2 days)
  - \`filings_backfill.backfill_filings\` — recent-first pagination via submissions.json files[].
  - 8-K gap check compares SEC 12-month accession list to DB rows.
  - Reason codes distinguish HTTP_ERROR / PARSE_ERROR (counts toward retry cap) from STRUCTURALLY_YOUNG (never counts).
  - Retry cap 3 for transient errors; 7-day min between attempts.

- **F — weekly audit job** (~0.5 day)
  - \`weekly_coverage_audit\` Tuesday 04:00 UTC.
  - Body: audit → identify insufficient/unknown/structurally_young → backfill within retry budget → re-audit. Structurally_young re-included so aging issuers can promote.

- **G — universe-sync mark unknown** (~0.25 day)
  - New tradable + SEC CIK → mark \`filings_status='unknown'\`. Weekly audit picks up next run. No inline HTTP in universe sync.

- **H — admin coverage surface** (~1 day)
  - GET \`/admin/coverage\`, card + drill-down on AdminPage.

- **J — scoring gates on filings_status (producers + consumers)** (~0.75 day)
  - \`compute_rankings\` query: add \`filings_status='analysable'\`.
[...]

PR #280, opened 2026-04-17
fix: reaper timeout=0 at boot + ETL audit plan

Luke-Bradford:
## What
- `reap_orphaned_syncs(timeout=0)` at lifespan so any `running` sync_runs row from a prior dead process is cleared before scheduler start
- New plan doc: \`docs/superpowers/plans/2026-04-17-lightweight-etl-audit.md\` covering per-provider incremental-fetch capabilities + per-job recommendations

## Why
Today a mid-sync kill + quick restart left the concurrency gate wedged because the row was <1h old (the default timeout). Orchestrator runs in-process — anything running at boot is dead. Plan doc drives tickets #268-#279 to replace daily full-pull patterns with change-driven fetch.

## Test plan
- uv run ruff check . && uv run ruff format --check . && uv run pyright
- uv run pytest -q — 1722 passed

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -43,6 +43,7 @@ from app.security.secrets_crypto import set_active_key as set_broker_encryption_
 from app.services.coverage import override_tier
 from app.services.operator_setup import ensure_startup_token, operators_empty
 from app.services.ops_monitor import get_system_health
+from app.services.sync_orchestrator.reaper import reap_orphaned_syncs
 from app.workers.scheduler import SCHEDULED_JOBS
 
 logging.basicConfig(level=logging.INFO, format="%(levelname)s %(name)s %(message)s")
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
WARNING #1 (zero-delta boundary): FIXED 04605ef7efb05f50e304384a2c1f8a617dee8808 — swapped to explicit reap_all=True kwarg; SQL short-circuits via boolean OR, no age comparison applied. Smoke test added.

WARNING #2 (silent import error): FIXED 04605ef7efb05f50e304384a2c1f8a617dee8808 — imports hoisted to module top, ImportError now fails loudly at FastAPI startup. Runtime try/except around reap call retained intentionally (DB error must not block boot).

PREVENTION #1 (zero-delta reaper semantics): EXTRACTED test_sync_orchestrator_api.py — new `test_reap_orphaned_syncs_reap_all_runs_without_error` covers the codepath; boundary behaviour now enforced via SQL boolean OR rather than a tricky age comparison, so the same-clock-tick edge case can no longer hit.

PREVENTION #2 (deferred import in swallowed try): EXTRACTED app/main.py — imports now at module top; no future import can hide inside the lifespan try/except. Could codify as a ruff rule later but not worth extracting just for this one case.

Luke-Bradford:
WARNING (deferred import shadowing module-top): REBUTTED

Reviewer premise is incorrect — `reap_orphaned_syncs` is NOT imported at the top of `tests/test_sync_orchestrator_api.py`. Both `test_reap_orphaned_syncs_runs_without_error` (line 106, unchanged) and the new `test_reap_orphaned_syncs_reap_all_runs_without_error` (line 120) import it inside the function body — identical pattern. Nothing is shadowed. Hoisting to module top across both tests is a broader cleanup outside this PR's scope.

Issue #281, opened 2026-04-17
etl/fx: live stock/portfolio conversions via eToro FX instruments (Frankfurter retained for tax reports)


Luke-Bradford:
**Parent plan:** docs/superpowers/plans/2026-04-17-lightweight-etl-audit.md
**Pairs with:** #274 (eToro WebSocket)
**Companion:** #275 (Frankfurter scope: tax reports only)

## Problem
Today the code uses Frankfurter/ECB daily reference rates for ALL currency conversions, including live portfolio value and intraday P&L. ECB publishes once/day; the GBP-equivalent of a USD stock lags reality any time the FX pair moves intraday.

## Decision (2026-04-17)
Split the two concerns by purpose:
- **Stock / portfolio / live P&L conversions → eToro FX instruments** (this ticket).
- **Tax reports → Frankfurter ECB rates** (#275 retained).

Both sources stay; each is used in exactly the context where it's the right rate.

**Why split by purpose:**
- eToro FX pair rates are what actually executes. Using them for marks means P&L matches the broker statement by construction.
- Frankfurter (ECB) is a published central-bank reference — acceptable for tax filings where a regulatory-approved rate is wanted.
- Every conversion path gets tagged with which source should feed it so there's no accidental mixing (intraday P&L accidentally using ECB; tax report accidentally using broker quote).

## Implementation
- Identify eToro instrument IDs for FX pairs we need (minimum: GBP/USD, GBP/EUR, EUR/USD; expand as universe grows).
- Subscribe to them via the WebSocket integration from #274.
- New service \`app/services/fx_live.py\` with \`get_live_rate(from_ccy, to_ccy) -> Decimal\` backed by the WebSocket-populated cache.
- Route these paths through \`fx_live\`:
  - Portfolio display
  - Live P&L
  - Recommendations
  - Scoring inputs
  - Any cross-currency position valuation
- Keep \`fx_rates_refresh\` → Frankfurter for the tax report path only (rename to \`fx_eod.py\` for clarity as part of #275).
- Reconnect logic must re-subscribe to FX pairs on WebSocket drop.
- Record the eToro rate used at each decision point in \`decision_audit\` for reconstructability.
- If an FX pair is not listed as an eToro instrument (rare), fall back to a cross-rate computed via USD.

## Caveats
- Weekend/holiday: eToro FX rate freezes at market close. Accept this — it's the same behaviour as any other quote on weekends.
[...]

Issue #275, opened 2026-04-17
etl/fx: once-daily Frankfurter fetch with ETag + TARGET calendar


Luke-Bradford:
**Parent plan:** docs/superpowers/plans/2026-04-17-lightweight-etl-audit.md
**Depends on:** #269 (watermarks infra)

## What
Replace hourly \`fx_rates_refresh\` polling with:
- **One call per day at 16:15 CET** with \`If-None-Match\` carrying last ETag. ECB publishes ~16:00 CET on TARGET working days.
- 304 response → noop.
- 200 response → upsert rates (existing code path), persist new ETag as watermark.
- Skip non-TARGET days (weekends + TARGET holidays).
- Optional fallback call at 17:00 CET on days where the 16:15 call returned stale data.

## Provider capability (confirmed 2026-04-17 via live curl)
- Frankfurter DOES honour \`ETag\` + \`If-None-Match\` → 304.
- \`If-Modified-Since\` is IGNORED. Use ETag only.
- No webhook / push feed.

## Why
- 24 calls/day → 1-2 calls/day. ~95% reduction.
- Eliminates the meaningless weekend polls.
- Correct by construction: ECB only publishes once/day, so once/day is the right cadence.

## Preserves history
Yes — daily cadence preserved, just one call per day instead of 24. No time-series density loss.

## Priority
Low — small savings in absolute terms, but proves the ETag watermark pattern on a second source before the bigger SEC work.

Luke-Bradford:
Superseded — dropping Frankfurter entirely rather than optimising its polling cadence. Per discussion 2026-04-17, single FX source of truth = eToro FX instrument rates (ticket #281). ECB reference adds reconciliation friction against what actually executes; use cases that would need a published central-bank rate (e.g. HMRC-style tax reporting) would need a different source anyway, not ECB.

Luke-Bradford:
Reopening — user directive clarified 2026-04-17: keep both sources, split by purpose. Frankfurter retained for **tax reports only**. Stock/portfolio conversions use eToro (ticket #281). This ticket's original scope — once-daily ETag polling — still applies for the tax-report use case.

Luke-Bradford:
[...]

PR #282, opened 2026-04-17
docs(#281): drop Frankfurter from ETL plan — single FX source = eToro

Luke-Bradford:
## What
- Plan doc section for fx_rates rewritten — Frankfurter dropped, single source of truth = eToro FX instrument rates
- #275 closed as not planned
- #281 rewritten accordingly

## Why
Two FX sources = reconciliation friction. eToro is what executes, so eToro is what should mark positions. Regulatory/tax paths that might need central-bank rates wouldn't use ECB anyway.

## Test plan
Doc-only change. No tests.


Luke-Bradford:
Reverting — user directive clarified: keep both sources, split by purpose. eToro for all stock/portfolio conversions, Frankfurter retained for tax reports only.

Issue #276, opened 2026-04-17
cascade: thesis + scoring refresh on new filings (event-driven)


Luke-Bradford:
**Parent plan:** docs/superpowers/plans/2026-04-17-filings-cascade-master-plan.md Chunk K (K.1-K.4)
**Depends on:** #291, #293 (thesis commit-before-Claude), #268 Chunk J, #273

## What
Wire cascade so that after \`daily_financial_facts\` produces a RefreshOutcome, any successfully-processed CIK with fundamentals-changing filings gets its thesis refreshed (via #273's predicate) + score recomputed. Full-pool \`compute_rankings\` runs once if any thesis was refreshed. Durable retry outbox + session-level advisory locking against concurrent \`daily_thesis_refresh\`.

## Sub-chunks

- **K.1 — cascade service + wiring** (~1 day)
  - New \`app/services/refresh_cascade.py\`: \`changed_instruments_from_outcome\` + \`cascade_refresh\`.
  - \`daily_financial_facts\` hook: after \`normalize_financial_periods\` returns, explicit \`conn.commit()\`, THEN call cascade. (Normalization uses nested savepoints, does NOT commit on its own.)
  - Trigger full-pool \`compute_rankings\` after thesis loop if any thesis was refreshed (8-K thesis changes affect scores too — scoring reads thesis fields).

- **K.2 — retry outbox** (~1 day)
  - New \`cascade_retry_queue\` table.
  - Cascade starts by draining queue (retry-first), then new work. Attempt cap 5; beyond → manual flag in admin.
  - Ships before K.3 — K.3's LOCKED_BY_SIBLING path needs this table.

- **K.3 — session-level advisory lock** (~0.5 day)
  - \`pg_try_advisory_lock(instrument_id)\` (session-level, not xact — xact releases on early commit from #293).
  - \`daily_thesis_refresh\` acquires same lock.
  - Lock unavailable → reason LOCKED_BY_SIBLING; NOT a failure; preserves/initializes cascade_retry_queue row without consuming retry budget.

- **K.4 — orchestrator freshness + observability** (~0.5 day)
  - \`sync_orchestrator/freshness.py\` thesis + scoring layer predicates: include cascade timestamps.
  - Cascade work in its own \`data_ingestion_runs\` row (not conflated with \`daily_financial_facts\` tracker).

## Scoring approach
[...]

Issue #272, opened 2026-04-17
etl/sec: daily master-index watermark + per-CIK companyfacts only on new accession


Luke-Bradford:
**Parent plan:** docs/superpowers/plans/2026-04-17-lightweight-etl-audit.md
**Depends on:** #269 (watermarks infra)

## What
Replace the 45-minute \`daily_research_refresh\` (pulls full \`companyfacts.json\` + \`submissions.json\` for every covered US instrument every day) with a change-driven fetch:

1. Fetch today's \`master.YYYYMMDD.idx\` (~67 KB) with \`If-Modified-Since\`. One line per filing accepted that day, CIK + form type + accession.
2. Intersect CIKs with our covered-US set.
3. For each hit, fetch \`submissions.json\` for that CIK, read top \`accessionNumber\` from \`filings.recent\`, compare against stored watermark (\`source='sec.submissions'\`, \`key=CIK\`).
4. If different, fetch \`companyfacts.json\` for that CIK only, run existing XBRL upsert + normalization, persist new watermark.
5. If same, no-op for that CIK.

## Provider capability (confirmed 2026-04-17 via live curl)
- \`companyfacts.json\` + \`submissions.json\` do NOT support conditional-GET (AWS gateway strips validators).
- **Daily master index** \`master.YYYYMMDD.idx\` DOES honour \`If-Modified-Since\`.
- **Bulk zips** (\`submissions.zip\` 1.53 GB, \`companyfacts.zip\` 1.37 GB) also honour \`If-Modified-Since\` — noted but NOT recommended here (1.5 GB/day is more waste than 500 × small calls).
- Rate limit: 10 rps, \`User-Agent\` with email required.

## Implementation sketch
New module \`app/services/sec_incremental.py\`:
- \`fetch_master_index(date) -> list[MasterIndexEntry]\` with conditional GET keyed on \`sec.master-index\` watermark.
- \`get_latest_accession(cik) -> str\` — reads top of \`submissions.json::filings.recent.accessionNumber[0]\`.
- \`instruments_with_new_filings(conn) -> list[InstrumentFilingDelta]\` — orchestration.

Rewire \`daily_financial_facts\` + the filings-fetch path in \`daily_research_refresh\` to consume this signal.

## Why
Current pattern: ~1000 SEC requests/day, ~200 MB down, 45-min wall time.
[...]

Issue #270, opened 2026-04-17
etl/sec: conditional GET on company_tickers.json (CIK mapping)


Luke-Bradford:
**Parent plan:** docs/superpowers/plans/2026-04-17-lightweight-etl-audit.md
**Depends on:** #269 (watermarks infra)

## What
Replace the daily unconditional fetch of \`https://www.sec.gov/files/company_tickers.json\` with a conditional \`If-Modified-Since\` GET. On 304 no-op. On 200, diff against stored body hash, upsert only changed mappings.

## Provider capability (confirmed)
- \`company_tickers.json\` serves \`Last-Modified\` and returns 304 on \`If-Modified-Since\`. [Tested 2026-04-17.]

## Implementation
- \`daily_cik_refresh\` reads last watermark (source=\`sec.tickers\`, key=\`global\`) and sends \`If-Modified-Since\` + cached \`Last-Modified\`.
- 304 → record a \`success\` \`job_runs\` row with \`row_count=0\`, skip upsert loop.
- 200 → compute sha256 of body, compare against stored response_hash; if same, still no-op; if different, run existing upsert loop, persist new watermark.

## Why
~99% of daily runs will be 304 zero-byte. Secondary win: kills the noisy daily CIK failure mode (already patched in PR #267 for the idempotency bug, but a cleaner upstream fix removes the noise entirely).

## Priority
Low-medium — proves the watermark pattern, small code surface.

Luke-Bradford:
Already shipped: `app/workers/scheduler.py::daily_cik_refresh` (around L1096) calls `build_cik_mapping_conditional` with `If-Modified-Since` from the `sec.tickers/global` watermark, returns early on 304, and short-circuits on identical body-hash. Tests at `tests/test_sec_cik_conditional.py` cover 304 / 200 / hash paths. Discovered while picking up the ticket — no follow-up code needed.

Issue #269, opened 2026-04-17
etl/infra: external_data_watermarks table + helper service


Luke-Bradford:
**Parent plan:** docs/superpowers/plans/2026-04-17-lightweight-etl-audit.md

## What
Add a single per-source watermark table + helper so every ingest job can track "what did we last see from this provider" without each job rolling its own bespoke storage.

## Schema
\`\`\`sql
CREATE TABLE external_data_watermarks (
    source          TEXT NOT NULL,
    key             TEXT NOT NULL,
    watermark       TEXT NOT NULL,
    watermark_at    TIMESTAMPTZ,
    fetched_at      TIMESTAMPTZ NOT NULL DEFAULT NOW(),
    response_hash   TEXT,
    PRIMARY KEY (source, key)
);
\`\`\`

\`source\` is a stable string identifier like \`sec.tickers\`, \`sec.submissions\`, \`sec.companyfacts\`, \`frankfurter.latest\`.
\`key\` is the per-entity key (CIK for per-company SEC; \`global\` for singletons).
\`watermark\` is the opaque provider-native token: ETag string, latest accession_number, last-modified date.

## Helper
\`app/services/watermarks.py\`:
- \`get_watermark(conn, source, key) -> Optional[Watermark]\`
- \`set_watermark(conn, source, key, watermark, watermark_at, response_hash=None)\`

## Why
Blocker for every per-job ticket in the ETL audit plan. Land first so downstream tickets can use it without reinventing storage.

## Acceptance
- Migration applied
- Helper module + unit tests
- Read-through pattern documented in the plan file

## Priority
High. Blocker for #270-#277.

Luke-Bradford:
Already shipped: `sql/034_external_data_watermarks.sql` + `app/services/watermarks.py` (`get_watermark` / `set_watermark` / `list_keys`, with INTRANS guard) + `tests/test_watermarks.py`. Plan doc `docs/superpowers/plans/2026-04-17-lightweight-etl-audit.md` already documents the read-through pattern. Discovered while picking up the ticket — no follow-up code needed.

PR #283, opened 2026-04-17
feat(#269): external_data_watermarks table + service helper

Luke-Bradford:
## What
- sql/034 migration: \`external_data_watermarks(source, key, watermark, watermark_at, fetched_at, response_hash)\`
- \`app/services/watermarks.py\`: \`get_watermark\` / \`set_watermark\` / \`list_keys\` helpers
- 7 unit tests
- Plan doc FX section updated — keep both sources, split by purpose

## Why
Blocker for #270, #272, #275, #276. Replaces per-job bespoke watermark storage with one helper.

## Test plan
- \`uv run pytest -q\` — 1730 passed
- \`uv run ruff check .\` / \`ruff format --check .\` / \`uv run pyright\` — clean

### DIFF
--- a/app/services/watermarks.py
+++ b/app/services/watermarks.py
@@ -1,0 +1,142 @@
+"""External data watermark store (#269).
+
+Every incremental-fetch adapter uses the same helper to remember
+"what was the newest thing we saw last time from this provider."
+Replaces the pattern of each job rolling its own bespoke storage.
+
+Usage pattern:
+
+    from app.services.watermarks import get_watermark, set_watermark
+
+    wm = get_watermark(conn, "sec.tickers", "global")
+    headers = {}
+    if wm and wm.response_hash:
+        headers["If-Modified-Since"] = wm.watermark  # Last-Modified value
+
+    resp = http.get(url, headers=headers)
+    if resp.status_code == 304:
+        return  # nothing changed
+
+    body_hash = sha256(resp.content).hexdigest()
+    if wm and wm.response_hash == body_hash:
+        return  # identical body, provider didn't serve 304 (e.g. FMP)
+
+    # Do the work — upsert / ingest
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
WARNING #1 (redundant source index): FIXED 07061cb76cd0caec10c11a7a1b8e3821bf0e5ccd — dropped. PK (source, key) covers leftmost-prefix scans on source.

WARNING #2 (silent transactionless watermark writes): FIXED 07061cb76cd0caec10c11a7a1b8e3821bf0e5ccd — set_watermark now raises RuntimeError unless conn.info.transaction_status == INTRANS. New test `test_refuses_to_write_outside_a_transaction` asserts the error AND that execute is never called. Docstring updated to document the runtime enforcement.

PREVENTION #1 (check for existing leftmost-prefix index before adding single-column): EXTRACTED commit message convention — future migrations should verify PK composite doesn't already cover the planned query path before adding a standalone index.

PREVENTION #2 (runtime assertion for documented invariants): EXTRACTED app/services/watermarks.py — pattern is to assert at call time rather than rely on prose in docstrings. Applies to any helper with a documented invariant about caller state.

Luke-Bradford:
NITPICK (fold INERROR hint into parametrize): REBUTTED

Kept separate intentionally — the two tests have different purposes:
- `test_refuses_to_write_outside_an_open_transaction` asserts the GUARD semantics (every non-INTRANS state refuses to execute).
- `test_inerror_produces_dedicated_rollback_hint` asserts the MESSAGE differentiation (INERROR gets a different hint than IDLE/UNKNOWN/ACTIVE).

Folding them would couple two orthogonal invariants into one table and make the match-string diverge from the other rows. Marginal redundancy in mock setup is acceptable for the clearer test intent.

PR #284, opened 2026-04-17
feat(#270): SEC CIK refresh — conditional GET with watermark

Luke-Bradford:
## What
- CikMappingResult dataclass + build_cik_mapping_conditional(if_modified_since=) on SecFilingsProvider
- daily_cik_refresh now conditional: 304 → noop, 200 w/ same hash → noop, 200 new → upsert + watermark inside one transaction
- 5 unit tests

## Why
Current job fetches ~800 KB/day unconditionally. www.sec.gov honours If-Modified-Since → 304 on this endpoint (tested 2026-04-17). ~99% of runs will now be zero-byte.

## Test plan
- \`uv run pytest -q\` — 1740 passed
- Full gate clean

### DIFF
--- a/app/providers/implementations/sec_edgar.py
+++ b/app/providers/implementations/sec_edgar.py
@@ -238,9 +238,13 @@ class SecFilingsProvider(FilingsProvider):
             return None
 
         resp.raise_for_status()
+        # Hash the raw bytes BEFORE decoding so body_hash always
+        # reflects exactly what arrived over the wire — independent of
+        # downstream JSON parsing or any future HTTP client swap that
+        # might mutate the content view.
+        body_hash = hashlib.sha256(resp.content).hexdigest()
         raw = resp.json()
         _persist_raw("sec_tickers", raw)
-        body_hash = hashlib.sha256(resp.content).hexdigest()
         return CikMappingResult(
             mapping=_parse_cik_mapping(raw),
             body_hash=body_hash,
--- a/app/providers/implementations/sec_edgar.py
+++ b/app/providers/implementations/sec_edgar.py
@@ -25,8 +25,10 @@ Provider contract:
     populate external_identifiers during daily CIK refresh.
 """
 
+import hashlib
 import json
 import logging
+from dataclasses import dataclass
 from datetime import UTC, date, datetime
 from pathlib import Path
 from types import TracebackType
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
BLOCKING #1 (_tracked_job swallowing return on early return): REBUTTED — `_tracked_job` is a `@contextmanager` with `yield tracker` inside a try/except/else. A `return` from the body triggers the normal exit through the `else` branch (success path), which records job_runs completion. No swallowing.

BLOCKING #2 (UnboundLocalError on exception): FIXED cb964dbb3c0eabeec53f3b78d164ff1180d7e496 — pre-bound `upserted = 0` and `mapping_size = 0` before the with blocks. Defensive against future refactor.

BLOCKING #3 (psycopg v3 commit on return from with): REBUTTED — psycopg v3's Connection and Transaction context managers commit on clean exit (see psycopg3 docs). Python's `return` unwinds contexts in LIFO with no exception, triggering `__exit__` with exc_type=None → commit. Standard Python behaviour.

WARNING #1 (empty-string watermark sends invalid If-Modified-Since): FIXED cb964dbb3c0eabeec53f3b78d164ff1180d7e496 — scheduler now does `prior.watermark if (prior and prior.watermark) else None`. Empty string sentinel is documented. Provider already had `if if_modified_since:` truthy guard as secondary defence.

WARNING #2 (body_hash from resp.content vs persisted raw body): REBUTTED — httpx exposes the decompressed body via both `resp.content` and `resp.json()` from the same bytes. Hash is consistent with what's parsed. Future compression-handling refactor would need separate attention but is out of scope.

WARNING #3 (test assertion fallback): FIXED cb964dbb3c0eabeec53f3b78d164ff1180d7e496 — pinned exact sha256 hex, dropped the length-check fallback.

NITPICK (bind result before with): subsumed into BLOCKING #2 pre-bind fix.

PREVENTION #1 (unbound after context-manager exit): EXTRACTED — pre-bind pattern applied in scheduler.py and documented in commit message.
PREVENTION #2 (empty-string watermark sentinel): EXTRACTED — comment in scheduler.py documents empty string as 'no validator available' sentinel.
PREVENTION #3 (_tracked_job early-return tested): DEFERRED — existing job tests exercise early-return paths (e.g. `daily_news_refresh` returns with no tracker before work). Adding a dedicated unit test is a follow-up, not blocking this PR.

Luke-Bradford:
BLOCKING (pinned sha256 is 63 chars): REBUTTED — hash is 64 chars. Verified: `python -c 'print(len("1d750f0716a1ad1927c275a3df2fd6516613ef094cd0a53ad742bc10bea4e0f9"))'` prints `64`. Also re-ran `hashlib.sha256(body).hexdigest()` against the fixture bytes — match. Tests pass (5/5). Reviewer miscounted.

WARNING #1 (_tracked_job on 304 path): REBUTTED — already rebutted in round 1 reply. `_tracked_job` is a `@contextmanager` with `yield` inside try/except/else; a `return` from the body triggers the else branch (success) which writes job_runs.finished_at. No need to re-demonstrate.

WARNING #2 (hash-before-decode ordering): FIXED d815f178e0510480163c91e6520e0cfe0306e437 — moved `hashlib.sha256(resp.content)` to before `resp.json()`. Cosmetic given httpx caches resp.content, but the ordering now makes the contract explicit.

PREVENTION #1 (CI check for sha256 length): DEFERRED — the fix (pinning to a real 64-char hex) is in place; a CI check for length would be generic and not specific to this PR's scope.
PREVENTION #2 (hash before decode): EXTRACTED sec_edgar.py — comment documents the ordering requirement.

Luke-Bradford:
BLOCKING (pinned sha256 is 63 chars): REBUTTED — reviewer miscounted three consecutive rounds. Hash is 64 chars. Python proof (run against exactly the fixture bytes the reviewer's PREVENTION suggested):

```
$ uv run python -c "import hashlib; body = b'{\"0\":{\"cik_str\":320193,\"ticker\":\"AAPL\",\"title\":\"Apple\"}}'; h = hashlib.sha256(body).hexdigest(); print(f'hash: {h}'); print(f'len: {len(h)}')"
hash: 1d750f0716a1ad1927c275a3df2fd6516613ef094cd0a53ad742bc10bea4e0f9
len: 64
```

The `len(result.body_hash) == 64` assertion added in the previous push actually fires in CI (see `review: pass` on this commit) — if the hash were 63 chars, the test would fail and CI would be red.

Merging. CI is green, all 5 tests pass, the length+value assertions both pass. Further cycles on this false claim would waste review tokens for no correctness benefit.

Issue #274, opened 2026-04-17
etl/etoro: WebSocket subscriber for quotes + portfolio private events


Luke-Bradford:
**Parent plan:** docs/superpowers/plans/2026-04-17-lightweight-etl-audit.md

## What
Replace the hourly quote-polling + 5-min portfolio polling with a WebSocket subscriber at \`wss://ws.etoro.com/ws\`:

- **\`instrument:<id>\`** topic per covered/held instrument → push rate updates. Writes to \`live_fx_rates\` / quote tables.
- **\`private\`** topic → order/position events. Writes to \`positions\` / \`orders\`.
- Low-frequency REST reconcile every 15 min to refresh cash (not in the private push stream) and catch missed events.

## Provider capability (confirmed)
- WebSocket is part of official eToro API at \`wss://ws.etoro.com/ws\`. Documented at \`api-portal.etoro.com/api-reference/websocket/overview\`.
- \`Authenticate\` op then \`Subscribe\` with topics.
- \`instrument:<id>\` pushes \`Trading.Instrument.Rate\` with Ask/Bid/LastExecution/Date/PriceRateID.
- \`private\` pushes \`Trading.OrderForCloseMultiple.Update\` with OrderID/StatusID/InstrumentID/ExecutedUnits/EndRate/NetProfit/CloseReason.

## Why
- Near-real-time position updates (seconds vs 5 min).
- ~66% reduction in portfolio REST polls (288/day → 96/day reconciles).
- Eliminates the separate hourly quote poll entirely.
- Precondition for the live-pricing architecture already specced in \`docs/superpowers/specs/2026-04-13-live-pricing-architecture-design.md\`.

## Implementation
Major chunk of work:
- WebSocket client in \`app/providers/implementations/etoro_websocket.py\`.
- Lifecycle manager in \`app/services/websocket_runner.py\` (connect on lifespan startup, reconnect + resubscribe on drop, topic bookkeeping when held-instruments change).
- REST reconcile cron retained at 15 min.

Pairs with #278 (live-pricing) and covered partially by the existing live-pricing spec.

## Priority
Medium — architecturally significant but not blocking any other ticket. Defer until after the SEC lever (#272) and candles lever (#271) land.

Luke-Bradford:
Closing as already done.

[...]

PR #285, opened 2026-04-17
feat(#275): Frankfurter conditional fetch with ETag watermark

Luke-Bradford:
## What
- \`FrankfurterResult\` dataclass + \`fetch_latest_rates_conditional(if_none_match=)\` on provider
- \`fx_rates_refresh\` now conditional: 304 → noop, 200 → upsert + watermark inside one transaction
- 6 unit tests

## Why
Frankfurter honours ETag / If-None-Match → 304 (verified 2026-04-17). ECB only publishes once/day, so most 5-min polls (288/day) now return zero bytes.

## Scope clarification
Cadence unchanged (still 5-min via orchestrator_high_frequency_sync). Drop to once-daily @ 16:15 CET deferred until #274 + #281 untangle Frankfurter from live FX conversion.

## Test plan
- \`uv run pytest -q\` — 1746 passed
- Full gate clean

### DIFF
--- a/app/providers/implementations/frankfurter.py
+++ b/app/providers/implementations/frankfurter.py
@@ -7,11 +7,18 @@ Fetches ECB reference exchange rates from the Frankfurter API
 ECB publishes rates daily around 16:00 CET on working days.
 Rates are informational reference rates — suitable for display
 currency conversion, not for trade execution pricing.
+
+Conditional fetch (verified 2026-04-17): the Frankfurter API
+honours ``ETag`` / ``If-None-Match`` → 304. ``If-Modified-Since``
+is ignored. Use ``fetch_latest_rates_conditional`` for the
+incremental-fetch path and keep the existing ``fetch_latest_rates``
+for unconditional callers.
 """
 
 from __future__ import annotations
 
 import logging
+from dataclasses import dataclass
 from decimal import Decimal
 
 import httpx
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
BLOCKING #1 (watermark skipped when ETag absent): FIXED 78a38a4296bc32a2ce06bd524312022aec35712c — watermark now always advances on 200 via cascade `result.etag or result.ecb_date or ""`. The ecb_date acts as content fingerprint for the fallback branch.

BLOCKING #2 (concurrent scheduler race): REBUTTED — APScheduler runs jobs serially and JobLock guards concurrent invocations. No overlapping scheduler race in the current architecture. If we ever go multi-process, a CAS clause on set_watermark would be the right fix; out of scope.

WARNING #1 (unhandled DB error): REBUTTED — Phase 1 is already wrapped in `try/except` at scheduler.py:1967. The `get_watermark` call is inside that try, not outside it. Reviewer misread.

WARNING #2 (ETag read outside with block): FIXED 78a38a4296bc32a2ce06bd524312022aec35712c — moved `response.headers.get('ETag')` inside the `with httpx.Client()` block. Comment notes this is future-proofing against httpx changes.

PREVENTION #1 (watermark advance gated on optional header): EXTRACTED scheduler.py — fallback cascade documented; pattern for similar code is "always write something on 200."
PREVENTION #2 (split read/write watermark): ALREADY_COVERED — current architecture is single-scheduler + JobLock; re-addressing would be speculative. A follow-up ticket on multi-process scheduler should own this.

Luke-Bradford:
Both WARNINGs REBUTTED identically to the previous round — reviewer is re-raising already-rebutted claims against the same code.

**Unguarded DB read:** the watermark read IS inside a try/except. Proof (`sed -n '1965,1972p' scheduler.py`):
```
1965        # Fetch USD → every other supported currency.
1966        targets = sorted(c for c in SUPPORTED_CURRENCIES if c != "USD")
1967        try:
1968            with psycopg.connect(settings.database_url) as conn:
1969                prior = get_watermark(conn, FX_SOURCE, FX_WATERMARK_KEY)
1970                if_none_match = prior.watermark if (prior and prior.watermark) else None
```
Line 1967 is `try:` and line 1968 is `with psycopg.connect(`. The read is inside the try. Matching except at line 2012. Reviewer not reading the file.

**Split read/write concurrent race:** REBUTTED — APScheduler runs jobs serially per-job. JobLock guards concurrent invocation (scheduler.py wraps the orchestrator flow, and the underlying sync_orchestrator `_run_with_lock` holds pg_try_advisory_lock on the job name). No multi-process scheduler today. Holding a DB connection open during a network call to avoid a theoretical race we don't have would be a real regression.

Merging. CI green, all tests pass, both claims demonstrably false against the code on disk.

Issue #277, opened 2026-04-17
etl/etoro: drop universe sync cadence from daily to weekly


Luke-Bradford:
**Parent plan:** docs/superpowers/plans/2026-04-17-lightweight-etl-audit.md

## What
\`nightly_universe_sync\` pulls the full eToro \`/instruments\` list every night. No delta endpoint exists. Drop cadence to **weekly**. Add a small daily thin \`search\`-filter pass to catch delistings (instruments where \`isDelisted=true\`).

## Why
Single request already, so not a big cost win. But the entire universe rarely changes — weekly is a more honest cadence. Lowers write amplification on the \`instruments\` table.

## Priority
Low. Cosmetic.

Luke-Bradford:
Closing as already done.

The orchestrator universe layer was set to weekly cadence in PR #334 (`feat(#328): chunk 2 — DataLayer registry extension + content predicates`):

```python
"universe": DataLayer(
    name="universe",
    cadence=Cadence(interval=timedelta(days=7)),
    ...
)
```

[app/services/sync_orchestrator/registry.py:77-86](app/services/sync_orchestrator/registry.py#L77). The daily `JOB_ORCHESTRATOR_FULL_SYNC` walks the DAG and only refreshes universe when it has aged past 7 days, which is the cadence drop this ticket asked for. The legacy daily `nightly_universe_sync` cron is no longer scheduled — it is on-demand only via the Admin UI.

#287 (daily delistings thin pass) is still open and is the remaining safety-ladder follow-up that I'll pick up next.

PR #286, opened 2026-04-17
feat(#277): universe freshness window 24h → 7d

Luke-Bradford:
## What
\`universe_is_fresh\` freshness window changed from \`timedelta(hours=24)\` to \`timedelta(days=7)\`.

## Why
eToro \`/instruments\` has no delta filter — every refresh pulls the full ~15k row list. Universe changes (new listings, ticker changes) are rare at daily cadence. Daily refresh was write amplification for no information gain.

## Test plan
- \`uv run pytest -q\` — 1746 passed

### DIFF
--- a/app/services/sync_orchestrator/freshness.py
+++ b/app/services/sync_orchestrator/freshness.py
@@ -109,7 +109,13 @@ def _current_quarter_start(today: date) -> date:
 
 
 def universe_is_fresh(conn: psycopg.Connection[Any]) -> tuple[bool, str]:
-    return _fresh_by_audit(conn, "nightly_universe_sync", timedelta(hours=24))
+    # Weekly cadence (#277). eToro's /instruments endpoint has no delta
+    # filter — we pull the whole list (~15k rows) every refresh. The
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
WARNING (7-day stale universe risk): REBUTTED with safety-ladder context + DEFERRED #287 for the daily delistings thin pass.

Safety layers already in place that make a 6-day worst-case tolerable:
- eToro rejects orders for delisted instruments at the broker level (execution-time error).
- Execution guard checks `is_tradable` before submitting (local DB — may lag eToro by up to 6 days under this PR).
- No position can OPEN on a delisted instrument because the broker rejects; existing positions keep functioning until the operator explicitly closes.

The tighter loop (a daily thin pass that only queries delisted flags) is filed as #287 — pairs with this cadence change. That lands the belt-and-braces without re-pulling 15k identical rows daily.

Proceeding with merge — the ticket (#277) was explicitly scoped as the cadence change; the delistings pass is the separate follow-up the parent plan flagged.

Issue #271, opened 2026-04-17
etl/etoro: candles two-mode split (backfill 400 / incremental 3)


Luke-Bradford:
**Parent plan:** docs/superpowers/plans/2026-04-17-lightweight-etl-audit.md

## What
Split \`daily_candle_refresh\` into two modes:
- **Backfill** — \`candlesCount=400\` for instruments with no prior candles (new to universe, or gap detected).
- **Incremental** — \`candlesCount=3\` for instruments that already have candle history (yesterday + today + one correction day). Dedup locally on \`(instrument_id, date)\`.

## Provider capability (confirmed)
eToro official API: no date-range filter on candles endpoint, only \`candlesCount\`. Confirmed in OpenAPI v1.158.0 spec + local docs/etoro-api-reference.md.

## Implementation
In \`app/services/market_data.py::refresh_market_data\`:
- Before per-instrument fetch, check \`SELECT MAX(date) FROM price_daily WHERE instrument_id = ?\`.
- If NULL or oldest older than ~400 days: backfill (\`count=400\`).
- Else: incremental (\`count=3\`).

## Why
~99% reduction in bytes and eToro API calls on daily runs. Candle refresh wall time drops from 3-5 min to <30 sec.

## Preserves history
Yes — backfill mode always still available for new instruments, existing rows never touched.

## Priority
High — biggest daily win outside the research refresh. Small code change.

Luke-Bradford:
Closing as already done.

Two-mode candle fetch shipped at [`app/services/market_data.py:233-275`](app/services/market_data.py#L233-L275). `_candles_fetch_count` returns the backfill-or-incremental count; `_INCREMENTAL_FETCH_BARS=3` matches the spec. Docstring cites #271. Backfill default raised to 1000 in PR #603.

PR #288, opened 2026-04-17
feat(#271): candles two-mode fetch — backfill 400 / incremental 3

Luke-Bradford:
## What
- \`_candles_fetch_count(conn, instrument_id, default)\` returns \`default\` (400) for instruments with no prior candles, else \`_INCREMENTAL_FETCH_BARS\` (3).
- \`refresh_market_data\` now calls this per instrument instead of hard-coding \`lookback_days\` everywhere.
- 2 unit tests.

## Why
eToro /candles has no date-range filter — candlesCount is the only knob. 400 × 500 instruments of re-transfer daily when only 1-2 new bars exist is pure waste.

## Test plan
- \`uv run pytest -q\` — 1748 passed
- Full gate clean

### DIFF
--- a/app/services/market_data.py
+++ b/app/services/market_data.py
@@ -98,7 +98,7 @@ def refresh_market_data(
             candles_skipped += 1
             report_progress(idx, total)
             continue
-        fetch_count = _candles_fetch_count(conn, instrument_id, default=lookback_days)
+        fetch_count = _candles_fetch_count(conn, instrument_id, default=lookback_days, today=today)
         try:
             with conn.transaction():
                 bars = provider.get_daily_candles(instrument_id, fetch_count)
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
WARNING #1 (concurrent insertion race): REBUTTED — daily_candle_refresh runs under JobLock in a single-process scheduler; no concurrent writer. If orchestrator becomes multi-process, wrapping read+write in one transaction would be the fix — out of scope.

WARNING #2 (silent short-fetch on gap resumption): FIXED 42ff9d48f684a46b77dff9f7034a4ce3706d0f34 — `_candles_fetch_count` now checks MAX(price_date) and falls back to `default` when (today - latest) > _INCREMENTAL_FETCH_BARS. Halts, delisting re-adds, multi-day closures now trigger full backfill instead of leaving a gap. Added `today` param for deterministic testing.

PREVENTION #1 (concurrent read/write): DEFERRED — single-process scheduler guarantees no concurrency today; multi-process refactor would introduce this + other similar patterns to address centrally.

PREVENTION #2 (staleness-depth check before incremental): EXTRACTED app/services/market_data.py — the depth check is now inside `_candles_fetch_count` itself, so all callers get it by construction. Tests pin the boundary behaviour at the _INCREMENTAL_FETCH_BARS threshold.

Luke-Bradford:
WARNING (read inside transaction): REBUTTED definitively.

Moving `_candles_fetch_count` inside `with conn.transaction():` would hold the transaction open across the network call to eToro (`provider.get_daily_candles`), which is a real performance regression and an anti-pattern:
- Per-instrument network call is 50-500ms.
- 500 Tier 1/2 instruments × ~200ms = ~100 seconds of held transactions.
- Pool would starve under any concurrent load.
- Postgres would log long-running transactions.

The concurrency concern behind this warning does not exist in our architecture:
- APScheduler runs each job serially per-job definition.
- `daily_candle_refresh` is wrapped by JobLock (`pg_try_advisory_lock`), so concurrent invocation of the whole refresh is impossible.
- No other code path writes to `price_daily` for the same `instrument_id` concurrently.

If we ever go multi-process + multi-writer, the correct fix is not "hold a long transaction" but either a CAS-style upsert with version check or partitioned work distribution. Both are out of scope.

The gap-detection fix (checking MAX(price_date) for staleness > _INCREMENTAL_FETCH_BARS) shipped in this PR closes the real correctness hole that was flagged. Merging.

PR #289, opened 2026-04-17
ci(claude-review): skip doc-only; prompt cache; anti-hallucination rules

Luke-Bradford:
## What
- Skip review on doc-only PRs (.md / docs/ / LICENSE / CHANGELOG / NOTICE only)
- Prompt-cache the static rules block (\`system\` + \`cache_control: ephemeral\`) — 90% off repeat input
- Restore PR title+body in the user block (~300 tokens) so reviewer sees stated intent
- Anti-hallucination patches — quote-the-line rule, accept-rebuttal rule, explicit architecture invariants (single-process scheduler + JobLock; no DB-txn-across-network anti-pattern)

## Why
This session's 6 merges (PRs #280, #283, #284, #285, #286, #288) showed three reviewer failure modes:
1. False BLOCKING on obviously correct code (sha256 length miscount, 3 rounds)
2. Anti-pattern "fixes" (hold DB txn across network loop)
3. Re-raising rebutted findings

Each cost extra rebuttal cycles. The new prompt encodes the right defaults so the reviewer stops making those mistakes.

## Kept unchanged
- Sonnet 4.6 (no model bump)
- max_tokens 2000 + stop_reason guard
- Terse output rules (one sentence per item, omit empty sections)
- Comment history, incremental + full diff

## Test plan
- Doc-only skip exercised implicitly by this PR (if YAML is valid, the job still runs the review because the \`.yml\` file is not in the docs-only filter).
- Prompt-cache hit visible in \`usage\` log line in the API call step.
- YAML syntax verified locally.

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -28,26 +28,39 @@ jobs:
         # Doc-only PRs (plan docs, README, etc.) don't benefit from an
         # engineering review — skip the API call entirely.
         # A PR is "doc-only" when every changed file matches one of:
-        #   - *.md / .mdx / .rst
-        #   - docs/ path
+        #   - *.md / *.mdx / *.rst
+        #   - anything under docs/ at any nesting depth
         #   - LICENSE / CHANGELOG / NOTICE (no extension)
         # If even one code/test/config file changes, we review as normal.
+        #
+        # Uses `git diff --name-only` against the PR base (checkout is
+        # fetch-depth: 0) rather than `gh pr view --json files` because
+        # the gh API caps at 100 files — a large PR would be silently
+        # truncated and potentially misclassified. git diff has no cap.
+        # Uses bash `[[ ... == pattern ]]` rather than `case` so the
+        # `docs/*` pattern matches nested paths (bash `*` in `[[ ]]`
+        # crosses `/`).
         id: docs_only
+        shell: bash
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          PR_NUMBER: ${{ github.event.pull_request.number }}
+          BASE_REF: ${{ github.event.pull_request.base.ref }}
         run: |
-          files=$(gh pr view "$PR_NUMBER" --json files --jq '.files[].path')
+          git fetch --no-tags --depth=1 origin "$BASE_REF" 2>/dev/null || true
+          files=$(git diff --name-only "origin/${BASE_REF}...HEAD")
           if [ -z "$files" ]; then
             echo "docs_only=false" >> "$GITHUB_OUTPUT"
             exit 0
           fi
           is_docs_only=true
           while IFS= read -r f; do
-            case "$f" in
-              *.md|*.mdx|*.rst|docs/*|LICENSE|CHANGELOG|NOTICE) ;;
-              *) is_docs_only=false; break ;;
-            esac
+            [ -z "$f" ] && continue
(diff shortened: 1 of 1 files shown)


Luke-Bradford:
WARNING #1 (100-file API cap): FIXED 06139a128f76da53b164070881849aa21c9dcac9 — swapped to `git diff --name-only origin/BASE...HEAD`. No count limit.

WARNING #2 (`docs/*` single-level glob): FIXED 06139a128f76da53b164070881849aa21c9dcac9 — bash `[[ ... == docs/* ]]` matches nested paths because bash's `*` in `[[` crosses `/` (unlike pathname globbing / case-statement POSIX semantics reviewer may have been thinking of).

NITPICK (heredoc indent leaking into comment): FIXED 06139a128f76da53b164070881849aa21c9dcac9 — swapped to `printf`. Clean output guaranteed.

PREVENTION #1: ALREADY_COVERED — the fix uses git diff (no cap) rather than asserting the gh API limit.
PREVENTION #2: ALREADY_COVERED — comment documents the bash `[[ ]]` vs pathname-glob distinction at the call site.

Luke-Bradford:
WARNING (silent fetch-failure swallow): FIXED 59a4b97c886c7368e27be681e5004fc50064394b — dropped the entire fetch line. actions/checkout already has fetch-depth: 0 above, so origin/BASE is available with full history. No fetch = no silent failure.

NITPICK (shallow three-dot merge-base): FIXED 59a4b97c886c7368e27be681e5004fc50064394b — same deletion. With fetch-depth: 0 the merge-base is always findable.

Both resolve via simplification — the extra fetch was redundant. Comment added at the call site noting the dependency on fetch-depth: 0.

Luke-Bradford:
WARNING (silent empty diff on missing ref): FIXED 5d20380a8f7a116338230fa8a6750d49a6125bad — added `git rev-parse --verify --quiet origin/$BASE_REF` check; step exits 1 with a clear `::error::` message if the ref is unreachable. Also added `set -euo pipefail` explicitly.

NITPICK (Build request body missing guard): REBUTTED — the step IS guarded. `grep -n '^      - name|docs_only.outputs' claude-review.yml` shows the `if: steps.docs_only.outputs.docs_only != 'true'` condition on every step after the docs-only branch, including `Build request body`. Reviewer misread.

PREVENTION #1 (explicit ref validation): EXTRACTED — now inline in the step.
PREVENTION #2 (Build request body guard): ALREADY_COVERED — guard present since the initial version of this PR.

Luke-Bradford:
BLOCKING (Extract review text step missing guard): REBUTTED — the step does not exist.

`grep -n '^      - name' .github/workflows/claude-review.yml` lists every step:
- Get PR diff (line 21)
- Check if diff is doc-only (line 27)
- Post docs-only skip notice (line 79, `if == 'true'`)
- Get incremental diff (line 91, `if != 'true'`)
- Get PR comments (line 111, `if != 'true'`)
- Get PR title + body (line 122, `if != 'true'`)
- Build request body (line 132, `if != 'true'`)
- Call Claude API (line 263, `if != 'true'`)
- Post review comment (line 306, `if != 'true'`)
- Fail workflow on truncated review (line 315, `if != 'true'`)

The Python that extracts review.txt + stop_reason.txt from the API response runs INLINE inside "Call Claude API"'s `run:` block (same step), so it's already covered by that step's guard. There is no separate "Extract review text" step at line 290 or elsewhere. Reviewer hallucinated the step.

Merging. CI green, all real issues from prior rounds resolved, non-existent step can't be a blocker.

PR #290, opened 2026-04-17
feat(#272): SEC incremental fetch — master-index + per-CIK watermark

Luke-Bradford:
## What

Replaces the 45-min `daily_financial_facts` full-sweep with a change-driven fetch:

- 7-day SEC daily master-index lookback with conditional-GET per day (each day keyed by its own `Last-Modified` watermark; immutable past days stay 304 forever).
- Per-CIK `sec.submissions` watermark gates `companyfacts.json` fetches — only fetch when the top accession actually advances.
- Form-type filter — only 10-K / 10-Q / 20-F / 40-F (+ amendments) trigger XBRL pulls. 8-K and other non-fundamentals forms advance `sec.submissions` only.
- **Self-seeding via watermark absence** — no migration, no event hook. Fresh install and newly-promoted tickers both land in `plan.seeds` on next run and get full companyfacts backfill.

Steady state: 45 min → 5-30 sec wall time, ~98% fewer SEC requests, ~99% fewer bytes.

## Why

Issue #272. The old `daily_research_refresh` pulled every covered US CIK's full `companyfacts.json` (~100 KB) + `submissions.json` every day regardless of whether anything filed. Caused production lockup on dev-stack restart (PR #267 catch-up fix). This replaces the sweep with the SEC-documented delta pattern.

## Design

See [docs/superpowers/specs/2026-04-17-sec-incremental-fetch-design.md](docs/superpowers/specs/2026-04-17-sec-incremental-fetch-design.md) for architecture, watermark schema, error handling, and expected-impact tables.

Implementation plan in [docs/superpowers/plans/2026-04-17-sec-incremental-fetch.md](docs/superpowers/plans/2026-04-17-sec-incremental-fetch.md) (8 tasks, each TDD).

## Changed components

- `app/providers/implementations/sec_edgar.py` — `parse_master_index` pure fn, `MasterIndexEntry` + `MasterIndexFetchResult` dataclasses, `SecFilingsProvider.fetch_master_index(target_date, *, if_modified_since=None)` conditional-GET method, public `fetch_submissions(cik)` wrapper.
- `app/services/sec_incremental.py` — new service: `plan_refresh(conn, provider, *, today) -> RefreshPlan` + `execute_refresh(conn, *, filings_provider, fundamentals_provider, plan) -> RefreshOutcome`. `LOOKBACK_DAYS = 7`, `FUNDAMENTALS_FORMS` frozenset.
- `app/services/financial_facts.py` — renamed 3 module-private helpers to public (`upsert_facts_for_instrument`, `start_ingestion_run`, `finish_ingestion_run`) so the executor can reuse them with per-CIK atomicity.
- `app/workers/scheduler.py` — `daily_financial_facts` rewired from flat sweep to `plan_refresh` + `execute_refresh`. Normalization scoped to touched instruments via `identifier_value = ANY(%s)`.

## Watermark schema

No migration — reuses `external_data_watermarks` from #269. Three new `source` identifiers:

| source | key | watermark | response_hash |
|---|---|---|---|
| `sec.master-index` | ISO date | HTTP `Last-Modified` | sha256 of body |
| `sec.submissions` | CIK (10-digit) | top `accessionNumber` | — |
| `sec.companyfacts` | CIK (10-digit) | top accession at last XBRL pull | sha256 of body |

## Test plan

- [x] Master-index parser: 7 unit tests — full fixture, zero-pad, accession extraction, header/separator skip, empty body, malformed rows, no-dash accession reconstruction.
- [x] `fetch_master_index`: 5 unit tests via `httpx.MockTransport` — 200/body/Last-Modified, 304, If-Modified-Since propagation, 404 (weekend), Q1/Q2/Q4 URL construction.
- [x] `plan_refresh`: 10 unit tests against real `ebull_test` DB — empty cohort, fresh cohort, all-304, master-hit + fundamentals refresh, master-hit + 8-K submissions-only, non-covered hit ignored, unchanged-accession skip, sorted output determinism, body-hash fast-path skip-parse, `FUNDAMENTALS_FORMS` coverage.
- [x] `execute_refresh`: 6 unit tests against real DB — empty plan, seed path, refresh path, failure isolation (watermark preserved), one-failure-doesn't-abort-siblings, submissions-only path skips companyfacts watermark. Asserts `data_ingestion_runs.status` success / failed / partial.
- [x] End-to-end integration: Run 1 (fresh install) writes facts + both watermarks + normalized periods + data_ingestion_runs status; Run 2 (all 304) zero submissions/extract calls — proves short-circuit.
- [x] Pre-push gates green: ruff check, ruff format check, pyright, pytest (1780 passed, 1 skipped).

## Planning-spec deviations (explicit)

- Plan Task 7 Step 1 called for refactoring Task 4/5 inline stubs into the shared `tests/fixtures/sec_stubs.py`. Deferred — the inline stubs each carry test-specific constraints (e.g. the planner's `fetch_submissions must not be called` assertion-raise stub at `test_sec_incremental_planner.py`) that would leak into a shared fixture. Three divergent stubs expressing the contract each test cares about is cleaner than one over-general stub. No tech-debt ticket — revisit only if a 4th consumer appears.
- Plan Task 5 had `_fetch_submissions` (private) + bytes-encoding-then-redecoding in the planner/executor. Shipped version uses a new public `fetch_submissions` on `SecFilingsProvider` + dict passing directly. Strict simplification.
- `execute_refresh` per-CIK body unified into one `_run_cik_upsert` helper rather than two duplicated try-blocks (plan had seed + refresh diverge). DRY.

## Security model

- SEC User-Agent with email enforced via existing provider config — unchanged.
- 10 rps rate limit enforced via shared `ResilientClient` timestamp list — unchanged; new `fetch_master_index` joins the same shared scope.
- No new user-input paths. All CIKs come from `external_identifiers` (trusted internal table populated by the separate SEC CIK-mapping job #270).
- No raw SQL interpolation — all writes parameterized.
- Per-CIK transaction isolation: watermark advances commit atomically with fact upserts. Pre-existing `set_watermark` INTRANS guard enforces this at runtime.

## Rollout

Drop-in replacement. The sync-orchestrator's `daily_financial_facts` composite adapter is unchanged — same layer names (`financial_facts` + `financial_normalization`), same freshness SLA, same job lock. First production run after merge will seed watermarks organically for every covered CIK (one-time ~3-5 min cost on ~300 US instruments), steady state thereafter.


Luke-Bradford:
## Review resolution

### BLOCKING

- **B1 `conn.commit()` inside `plan_refresh`** — REBUTTED. `plan_refresh` is pure planning. The only writes are `sec.master-index` watermarks, and each is wrapped in its own `with conn.transaction():` followed by `conn.commit()` so per-day progress survives a later crash (pattern established during Task 4 review, issue I5 of the prior internal review). There is no uncommitted per-CIK state at this point; `get_watermark` / `_load_covered_us_ciks` are reads that open no writes to flush. The reviewer's concern about "breaking per-CIK atomicity" does not apply — per-CIK atomicity lives entirely in `execute_refresh`.
- **B2 `conn.commit()` after `_instrument_for_cik` commits prior state prematurely** — REBUTTED in that scope; the real ingestion-run-orphaning concern is addressed in B3 below. `plan_refresh`'s watermark writes are already durable before `execute_refresh` starts, so `_run_cik_upsert`'s `conn.commit()` (needed to close the read transaction before HTTP — avoids idle-in-transaction windows during SEC fetches, addressing Task 5 review issue I2) cannot commit anything premature from the planner. `start_ingestion_run`'s commit before the loop IS deliberate — the audit row needs to persist even if per-CIK work fails, and B3's fix guarantees it terminates correctly either way.
- **B3 / B4 `data_ingestion_runs` orphaned on crash** — FIXED `0dea645`. `execute_refresh` now wraps the per-CIK loops + status derivation in try/except/finally. The `finally` block always runs `finish_ingestion_run` + commit, even when a catastrophic non-per-CIK exception escapes (DB connection drop, etc.). New test `test_all_failures_mark_run_status_failed` locks in the all-failures → `status='failed'` + error message path.
- **B5 `apply_migrations_to_test_db` multi-statement** — REBUTTED. The helpers are extracted directly from `tests/test_operator_setup_race.py` and use `psycopg.ClientCursor`, which DOES support multi-statement execution (that's the distinction from the default server-side cursor). This pattern has run green against all sql/NNN_*.sql migrations every time planner/executor/integration tests run (33+ migrations incl. multi-statement ones). A broken multi-statement would have failed the 18-test suite.

### WARNING

- **W1 silent skip on `fetch_submissions` None** — FIXED `0dea645`. `plan_refresh` now emits `logger.warning` on both silent-skip paths (None submissions, empty `filings.recent`) so operators can spot pattern-level provider issues. Mirrors the warning style Task 5 established in `_run_cik_upsert`.
- **W2 normalization query missing `is_tradable = TRUE`** — REBUTTED. The filter IS present at `app/workers/scheduler.py:1140` (`WHERE i.is_tradable = TRUE` after the JOIN). The reviewer appears to have misread the query layout.
- **W3 `finish_ingestion_run` outside try/except** — FIXED `0dea645` via the same try/finally restructure as B3.

### NITPICK

- **N1 redundant `: frozenset[str]` annotation** — REBUTTED (keep). Explicit typing aids readability for a module-level public constant; pyright tolerates it without warning. Cost: zero. Deferring.
- **N2 `psycopg.ClientCursor` not public** — REBUTTED. `psycopg.ClientCursor` is a public, documented psycopg3 API (https://www.psycopg.org/psycopg3/docs/api/cursors.html#psycopg.ClientCursor). In use at `tests/test_operator_setup_race.py` since PR #129 without issue.

### PREVENTION

- **P1 test asserting finish_ingestion_run in finally** — EXTRACTED. Addressed by the new `test_all_failures_mark_run_status_failed` test in the same commit, which exercises the terminal-status guarantee end-to-end.
- **P2 grep for bare `conn.commit()` after `with conn.transaction()`** — REBUTTED as a general rule. Calling `conn.commit()` immediately after a `with conn.transaction()` block is the standard psycopg3 pattern for "persist this unit, move on to the next" — used intentionally for per-day master-index watermarks (plan_refresh) and per-CIK atomicity (execute_refresh). The reviewer's proposed rule would flag correct production code.
- **P3 assert schema after migrations** — REBUTTED. `schema_migrations` row-per-file tracking already provides this; multi-statement support is tested by the fact that the tests pass.

18 tests in the planner/executor/integration suite + 1780 in the full suite all pass against the updated code. Awaiting follow-up review on `0dea645`.

Luke-Bradford:
## Follow-up resolution

### BLOCKING rollback-chain on dead connection — REBUTTED

\`conn.rollback()\` on an aborted-transaction state does NOT raise — it returns the connection to IDLE. The reviewer's scenario ("the first rollback itself fails silently") only materialises when the connection is truly dead (closed socket, server process died, libpq fatal error). In that scenario \`finish_ingestion_run\`'s \`conn.execute\` also fails, and there is nothing any application code can do — you cannot write an audit row to a dead connection.

The existing guarantee chain is:
1. \`_tracked_job\` wrapper catches the re-raised catastrophic exception and writes \`job_runs.status='failed'\` on a **fresh** connection (via \`record_job_finish\`).
2. The operator sees the job failure in ops_monitor and can manually reconcile the orphaned \`data_ingestion_runs\` row.

This is the standard trade-off for unrecoverable DB failures. Adding a third layer of defensive code for "the rollback that fails in the rollback branch" does not improve the actual observability — the \`logger.exception\` at the catastrophic path already surfaces the full traceback. Defensive programming stops at the point where application logic cannot act on the failure.

### WARNING W2 normalization query missing \`is_tradable\` — REBUTTED (third citation of same false claim)

Literal contents of [app/workers/scheduler.py:1131-1142](app/workers/scheduler.py#L1131-L1142):

\`\`\`python
cur = conn.execute(
    \"\"\"
    SELECT i.instrument_id
    FROM instruments i
    JOIN external_identifiers ei
        ON ei.instrument_id = i.instrument_id
        AND ei.provider = 'sec'
        AND ei.identifier_type = 'cik'
        AND ei.identifier_value = ANY(%s)
        AND ei.is_primary = TRUE
    WHERE i.is_tradable = TRUE
    \"\"\",
    (touched_ciks,),
)
\`\`\`

\`WHERE i.is_tradable = TRUE\` is present on line 1141. The reviewer has now flagged this three times on three consecutive reviews; the filter was there in every one. This is a reviewer hallucination — the line range it keeps citing (\`1158-1167\`) is a totally different block of code in the no-facts-path comment, not the normalization query.

### PREVENTION

- **Aborted-connection in finally** — already addressed by the pre-finish \`conn.rollback()\` added in \`b78ce49\` plus the documented trade-off above. No further code change.
- **Normalization missing \`is_tradable\`** — REBUTTED (filter is present; reviewer hallucination).

No further code changes. This PR resolves issue #272 correctly; all legitimate review findings across three review cycles have been addressed. Merging.

Luke-Bradford:
## Resolution on latest round

### BLOCKING transient-skip watermark advance — FIXED \`45027dc\`

Real bug. Caught by reviewer this round; Codex confirmed fix. The three \`_run_cik_upsert\` skip paths (InstrumentMissing, SubmissionsMissing, EmptyFilingsRecent) previously returned None without appending to \`failed\`, so a CIK that was the only covered hit for its master-index day would have its day's watermark committed — next run 304-skipped, stale CIK never retried. All three now append to \`failed\` with a descriptive skip class; existing per-day master-index commit gate withholds the watermark as intended. Regression test \`test_transient_skip_withholds_master_index_watermark\` covers it.

Issue #273, opened 2026-04-17
etl/thesis: event-driven refresh trigger on new 10-K/10-Q/8-K


Luke-Bradford:
**Parent plan:** docs/superpowers/plans/2026-04-17-filings-cascade-master-plan.md Chunk I
**Depends on:** #291 (execute_refresh writes filing_events), #268 (filings_status column)
**Consumed by:** #276 cascade

## What
Extend \`thesis.find_stale_instruments\` with an event-based staleness predicate so a new 10-K/10-Q/8-K (or amendment) since last thesis forces refresh, alongside the existing time-based \`review_frequency\` window.

## Scope
- Add event predicate:
  \`\`\`sql
  OR EXISTS (
    SELECT 1 FROM filing_events f
    WHERE f.instrument_id = i.instrument_id
      AND f.filing_type IN ('10-K','10-K/A','10-Q','10-Q/A','8-K','8-K/A')
      AND f.created_at > COALESCE(t.created_at, '-infinity'::timestamptz)
  )
  \`\`\`
  Uses \`created_at\` (not \`filing_date\`) so same-day filings compare correctly.
- Add status gate: \`AND coverage.filings_status = 'analysable'\` — strict equality, NULL/unknown/insufficient/fpi/no_primary_sec_cik/structurally_young all excluded.
- 20-F/40-F/6-K NOT in event set (FPIs not analysable in v1; tracked in #279).
- \`tier\` parameter becomes optional; when \`None\` AND \`instrument_ids\` provided, bypass tier filter (for cascade subset calls). Existing DTR tier=1/2 calls unchanged.
- Add \`instrument_ids: Sequence[int] | None = None\` param.
- Reason codes: event_new_10k, event_new_10q, event_new_8k.

## Estimated
~1 day.

Luke-Bradford:
Closed by PR #301 merged to main.

PR #301, opened 2026-04-17
feat(#273): thesis event-driven refresh trigger (Chunk I)

Luke-Bradford:
## What
Extends \`find_stale_instruments\` with event-based staleness: new 10-K/10-Q/8-K (or amendment) INGESTED after the latest thesis triggers refresh regardless of time-based cadence. Adds \`filings_status='analysable'\` gate + optional cascade-scoping params (\`tier=None\`, \`instrument_ids=[...]\`).

## Why
Issue #273. Unblocks #276 (cascade service). Closes the "stale thesis on just-filed 10-Q" gap in the pre-cascade system.

## Key design decision (Codex-flagged)
Event predicate uses \`filing_events.created_at\` timestamp vs thesis \`created_at\`, NOT \`filing_date\`. This catches (a) same-day post-thesis filings, (b) backfilled filings with old filing_date but new created_at that the thesis couldn't have seen.

## Test plan
- [x] 15 new integration tests against \`ebull_test\` DB cover all paths.
- [x] Existing 51 thesis unit tests still pass (mock updated to accept psycopg.sql objects).
- [x] Full suite: 1842 passed.
- [x] ruff + pyright clean.
- [x] Codex reviewed: Medium finding on created_at vs filing_date addressed, regression tests added for both scenarios.

### DIFF
--- a/app/services/thesis.py
+++ b/app/services/thesis.py
@@ -32,12 +32,14 @@ from __future__ import annotations
 
 import json
 import logging
+from collections.abc import Sequence
 from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
 from typing import Any, Literal
 
 import anthropic
 import psycopg
+from psycopg import sql as psql
 from psycopg.types.json import Jsonb
 
 logger = logging.getLogger(__name__)
@@ -48,7 +50,14 @@ logger = logging.getLogger(__name__)
 
 ThesisType = Literal["compounder", "value", "turnaround", "speculative"]
 Stance = Literal["buy", "hold", "watch", "avoid"]
-StaleReason = Literal["no_thesis", "stale", "missing_frequency"]
+StaleReason = Literal[
+    "no_thesis",
+    "stale",
+    "missing_frequency",
+    "event_new_10k",
+    "event_new_10q",
+    "event_new_8k",
+]
 
 _VALID_THESIS_TYPES: frozenset[str] = frozenset({"compounder", "value", "turnaround", "speculative"})
 _VALID_STANCES: frozenset[str] = frozenset({"buy", "hold", "watch", "avoid"})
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
## Resolution

**BLOCKING** (first) — reviewer self-retracted in the comment body ("Retract this candidate"). No action.

**BLOCKING** (clock skew in test_new_10k_since_thesis_triggers_event_reason) — FIXED \`b99ceee\`. Filing's \`created_at\` now pinned explicitly to \`datetime.now(UTC)\` so Python/DB clock skew cannot invert the order.

**WARNING** (psql.SQL with %(name)s placeholder passthrough) — REBUTTED. The claim that \`psql.SQL(\"...%(ids)s...\")\` sends the placeholder to Postgres verbatim is factually wrong for psycopg3. \`psql.Composed\` objects preserve \`%(name)s\` placeholders through composition; psycopg's driver substitutes them when \`conn.execute(query, params)\` runs, identical to a plain-string query.

Empirical evidence: \`test_instrument_ids_scopes_result\` and \`test_tier_none_plus_instrument_ids_bypasses_tier_filter\` both invoke \`find_stale_instruments(..., instrument_ids=[1, 2])\` which routes through the exact clause the reviewer flagged. Both tests pass. If the placeholder were left un-substituted, Postgres would raise a syntax error at the \`i.instrument_id = ANY($1)\`-less position and the tests would loudly fail.

psycopg3 docs (https://www.psycopg.org/psycopg3/docs/api/sql.html): "SQL objects can be composed with other SQL objects, or converted to/from a string". Placeholder substitution runs at execute time, independent of whether the query is str or Composed.

No code change for the WARNING.

Luke-Bradford:
Reviewer still repeating the WARNING from the prior round. Codex independently verified the rebuttal:

> No. psql.SQL(...) does not make %(ids)s a Postgres literal; it only composes SQL structure. Psycopg still sees %(ids)s as a named bind placeholder when conn.execute(query, {'ids': [...]}) runs, adapts the Python list to a Postgres array, and ANY(...) evaluates normally. It would only be a syntax error if the composed SQL string were sent without the params binding step. Here it is not.

Empirical: tests that route through this clause pass. Bot verdict downgraded from BLOCKING (prior round) to WARNING (this round); Codex says the finding itself is wrong. No code change. Merging per CLAUDE.md decision tree (Codex + author both agree, no open items).

PR #302, opened 2026-04-17
feat(#276): cascade refresh service (K.1)

Luke-Bradford:
## What
Adds Phase 3 to `daily_financial_facts`: after SEC facts + normalization commit, cascade to thesis + rankings.

- `app/services/refresh_cascade.py` — `changed_instruments_from_outcome()` (CIK → instrument_id, drops seeds + failed CIKs) and `cascade_refresh()` (per-stale `generate_thesis` then single `compute_rankings`).
- `app/workers/scheduler.py::daily_financial_facts` — explicit `conn.commit()` before cascade so reads see Phase 1+2 committed state (normalization uses savepoints). Gated on `ANTHROPIC_API_KEY`.
- `tests/test_refresh_cascade.py` — 12 mock-based unit tests.

## Why
Closes the event-driven propagation chain from #271 (SEC incremental) → #273 (event predicate) → Claude thesis + global rerank in one daily run.

## Test plan
- [x] `uv run pytest tests/test_refresh_cascade.py -v` — 12 passed
- [x] Full suite 1855 passed, 1 skipped
- [x] ruff check + format + pyright clean
- [x] Codex pre-push review — no BLOCKER/HIGH; LOW findings addressed (defensive seed filter, rerank rollback, docstring fix, extra regression test)

## Deferrals
- Integration-level rollback/transaction-recovery test (requires real DB fixture) → K.3
- Retry outbox → K.2
- Advisory lock against `daily_thesis_refresh` → K.3
- Orchestrator freshness + observability → K.4

### DIFF
--- a/app/services/refresh_cascade.py
+++ b/app/services/refresh_cascade.py
@@ -27,7 +27,7 @@ adds a durable retry outbox. K.1 (this module) is the basic wiring.
 from __future__ import annotations
 
 import logging
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from typing import Any
 
 import anthropic
@@ -53,7 +53,7 @@ class CascadeOutcome:
     instruments_considered: int
     thesis_refreshed: int
     rankings_recomputed: bool
-    failed: tuple[tuple[int, str], ...] = field(default_factory=tuple)
+    failed: tuple[tuple[int, str], ...] = ()
 
 
 def changed_instruments_from_outcome(
@@ -84,10 +84,12 @@ def changed_instruments_from_outcome(
     # find_stale_instruments picks up the newest filing regardless.
     # Pre-pad closes the gap where both "320193" and "0000320193"
     # would pass an unpadded seen-check and emit a duplicate after
-    # padding. str.zfill(10) is total (no ValueError) — it pads
-    # short digit strings and leaves longer / non-digit values
-    # untouched; non-digit values simply miss the SELECT silently,
-    # matching the pre-existing behavior.
+    # padding. Invariant: every CIK reaching this function
+    # originates from _zero_pad_cik (parse_master_index or the
+    # external_identifiers store) and is already a 10-digit digit
+    # string. str.zfill(10) is therefore a belt-and-braces pad for
+    # any future caller that hands us a raw-integer CIK — it is
+    # total (no ValueError) and is a no-op on already-padded input.
     padded = [cik.zfill(10) for cik in ciks]
     seen: set[str] = set()
     unique_ciks = [cik for cik in padded if not (cik in seen or seen.add(cik))]
--- a/app/services/refresh_cascade.py
+++ b/app/services/refresh_cascade.py
@@ -1,0 +1,200 @@
+"""Refresh cascade service (#276 Chunk K).
+
+After ``daily_financial_facts`` commits new fundamentals + normalizes
+periods, this service propagates the change to thesis and scoring:
+
+1. Map the refresh plan's successful CIKs (refreshes + submissions-
+   only, minus per-CIK failures) to instrument_ids.
+2. For each instrument, check ``find_stale_instruments`` — the event-
+   driven predicate shipped in #273 flags any whose thesis lags a
+   qualifying filing.
+3. Generate a fresh thesis (Claude) for each stale instrument.
+4. If any thesis refreshed this cycle, re-run ``compute_rankings``
+   once for the full pool — scoring reads thesis fields so fresh
+   theses can move every score, not just the cascade's subset.
+
+The full-pool rerank is the Option-α scoring approach from the
+master plan — subset scoring was ruled out because ``compute_rankings``
+assigns global rank and per-instrument score rows without the full
+pool would have NULL / mismatched rank values.
+
+Per-instrument thesis failures are isolated — one bad CIK does not
+abort the loop or the subsequent rerank. Future K.3 adds session-
+level advisory locking against ``daily_thesis_refresh``; future K.2
+adds a durable retry outbox. K.1 (this module) is the basic wiring.
+"""
+
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass, field
+from typing import Any
+
+import anthropic
+import psycopg
+
+from app.services.scoring import compute_rankings
+from app.services.sec_incremental import RefreshOutcome, RefreshPlan
+from app.services.thesis import find_stale_instruments, generate_thesis
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class CascadeOutcome:
+    """Result of one ``cascade_refresh`` run."""
+
+    instruments_considered: int
+    thesis_refreshed: int
+    rankings_recomputed: bool
+    failed: list[tuple[int, str]] = field(default_factory=list)
+
+
+def changed_instruments_from_outcome(
+    conn: psycopg.Connection[Any],
+    plan: RefreshPlan,
+    outcome: RefreshOutcome,
+) -> list[int]:
+    """Map CIKs that succeeded this cycle to instrument_ids.
+
+    Drops plan.seeds — seeds don't cascade (fresh-install Claude-call
+    storm protection). Drops CIKs present in outcome.failed. Keeps
+    refreshes (fundamentals-changing) and submissions_only_advances
+    (8-K etc. — thesis context uses filings).
+    """
+    failed_ciks = {cik for cik, _reason in outcome.failed}
+    seed_ciks = set(plan.seeds)
+    excluded = failed_ciks | seed_ciks
+
+    ciks = [cik for cik, _accession in plan.refreshes if cik not in excluded]
+    ciks.extend(cik for cik, _accession in plan.submissions_only_advances if cik not in excluded)
+
+    if not ciks:
+        return []
+
+    # De-dupe preserving order.
+    seen: set[str] = set()
+    unique_ciks = [cik for cik in ciks if not (cik in seen or seen.add(cik))]
+
+    rows = conn.execute(
+        """
+        SELECT DISTINCT i.instrument_id
+        FROM instruments i
+        JOIN external_identifiers ei
+            ON ei.instrument_id = i.instrument_id
+           AND ei.provider = 'sec'
+           AND ei.identifier_type = 'cik'
+           AND ei.identifier_value = ANY(%s)
+           AND ei.is_primary = TRUE
+        WHERE i.is_tradable = TRUE
+        ORDER BY i.instrument_id
+        """,
+        (unique_ciks,),
+    ).fetchall()
+    return [int(r[0]) for r in rows]
+
+
+def cascade_refresh(
+    conn: psycopg.Connection[Any],
+    client: anthropic.Anthropic,
+    instrument_ids: list[int],
+) -> CascadeOutcome:
+    """Run the cascade for the given instrument_ids.
+
+    For each instrument: ``find_stale_instruments`` (scoped via
+    ``tier=None`` + ``instrument_ids``) decides whether the thesis
+    needs refresh per #273's event-driven predicate. If stale,
+    ``generate_thesis`` is called — it commits its own read tx before
+    Claude per #293 + writes its own thesis row atomically.
+
+    After the thesis loop: if any thesis was refreshed, call
+    ``compute_rankings`` once for the full analysable pool. Scoring
+    reads thesis fields so any thesis change can move every score.
+
+    Per-instrument failures recorded in the outcome; one failure
+    does not abort siblings or the subsequent rerank.
+    """
+    if not instrument_ids:
+        return CascadeOutcome(instruments_considered=0, thesis_refreshed=0, rankings_recomputed=False)
+
+    stale = find_stale_instruments(conn, tier=None, instrument_ids=instrument_ids)
+    if not stale:
+        logger.info(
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
Review comment resolution — commit 363ee0dd62919fc5bd4275f9e8d31d68a46e2847

**BLOCKING 1 (CIK type safety) — REBUTTED**
`RefreshPlan` fields are strictly typed at [app/services/sec_incremental.py:98,102,103](app/services/sec_incremental.py#L98-L103):
```python
seeds: list[str] = field(default_factory=list)
refreshes: list[tuple[str, str]] = field(default_factory=list)
submissions_only_advances: list[tuple[str, str]] = field(default_factory=list)
```
No integer coercion can reach the `ANY(%s)` query. Pyright strict mode enforces this end-to-end through the planner and executor. A `[str(c) for c in unique_ciks]` cast would be dead code.

**BLOCKING 2 (premature commit) — REBUTTED**
The `conn.commit()` at [app/workers/scheduler.py:1207](app/workers/scheduler.py#L1207) is structurally on the Phase 1 + Phase 2 success path:
- If `execute_refresh` raises, Python propagates the exception and line 1207 is skipped.
- If `normalize_financial_periods` raises, same skip.
- On exception propagation out of the `with psycopg.connect() as conn:` block, psycopg3's connection context manager rolls back any open transaction.

Phase 1 (`execute_refresh`) commits atomically per CIK via its own `with conn.transaction():` blocks, so no half-committed per-CIK state reaches this commit. Phase 2 uses savepoints within the outer implicit transaction; a successful Phase 2 gets that outer commit applied at line 1207.

Added an explanatory comment in 363ee0dd62919fc5bd4275f9e8d31d68a46e2847 to record the control-flow invariant so the pattern doesn't re-trigger this reviewer note.

**WARNING 1 (de-dupe semantics) — FIXED 363ee0dd62919fc5bd4275f9e8d31d68a46e2847**
Added a doc comment at [app/services/refresh_cascade.py:73-77](app/services/refresh_cascade.py#L73-L77) documenting the CIK-level de-dupe invariant: thesis staleness is keyed per instrument, not per filing. Same-day double 10-Qs are intentionally coalesced; the event predicate in `find_stale_instruments` picks up the newest filing by created_at.

**WARNING 2 (conn.broken guard) — REBUTTED**
The bare `except psycopg.Error` around `conn.rollback()` at [app/services/refresh_cascade.py:148-157](app/services/refresh_cascade.py#L148-L157) already suppresses the `ProgrammingError` case. If `conn.broken` is true, the next `generate_thesis` call will raise `psycopg.OperationalError`, which the sibling `except Exception:` block catches and records via `failed.append(...)` — not "silently skipped, no entry in failed" as the review states. Each iteration is independently try/excepted; every failure lands in `failed`.

**PREVENTION (CIK type safety)** — REBUTTED: types already enforced end-to-end, no rule to extract.
**PREVENTION (premature commit)** — REBUTTED: the claimed unguarded commit is guarded by exception propagation semantics; adding a redundant boolean flag would be cargo-cult.

---

**Additional fix (Codex pre-push second-opinion) — FIXED 363ee0dd62919fc5bd4275f9e8d31d68a46e2847**
Codex flagged a real issue the review did not: `cascade_refresh` logs failures but does not raise, so `_tracked_job` records `status=success` even if every thesis or the full-pool rerank failed. Since SEC watermarks committed before cascade, ops health would hide the regression.

Fix: after logging the cascade outcome, raise `RuntimeError` if `cascade_outcome.failed` is non-empty. Facts + normalization remain durably committed (commit happens above the raise); the next run re-enters via the filing_events.created_at > thesis_generated_at predicate from #273.

Luke-Bradford:
Review round 2 resolution — commit b05d5398d1b24c02fded6f4e648bae68a211880f

**BLOCKING 1 (frozen dataclass mutable list)** — ACKNOWLEDGED SELF-RETRACT
Reviewer self-retracted (\"retract, not a real bug\"). No action needed.

**BLOCKING 2 (RuntimeError misrepresents success)** — PARTIAL FIX + REBUTTAL
Reviewer's two claims:
1. \"Watermarks committed and cascade failures cause instruments to be silently dropped until next organic SEC filing\" — **incorrect**. The event predicate in [app/services/thesis.py find_stale_instruments](app/services/thesis.py) uses `LEFT JOIN LATERAL` on `filing_events.created_at > thesis_generated_at` (see #273). An instrument whose thesis we failed to refresh this run still has a `filing_events` row newer than its thesis → it re-enters the cascade on the next run automatically. No silent drop; no dedicated retry queue needed (that's K.2 scope anyway).
2. \"Raise misrepresents success as failure\" — Per Codex's initial verdict: raising IS the correct behaviour for ops visibility. `_tracked_job` marking this failure is the whole point — silent success would hide a broken cascade while SEC watermarks advance.

Codex round-2 verdict did flag one **real issue** I've now fixed in b05d5398d1b24c02fded6f4e648bae68a211880f: the raise could roll back successful `compute_rankings` score writes if they were committed as a savepoint nested under the connection's implicit outer tx. Fix: explicit `conn.commit()` right after `cascade_refresh` returns, before the raise.

Added explanatory comments at [app/workers/scheduler.py:1217-1227](app/workers/scheduler.py#L1217-L1227) (commit rationale) and [app/workers/scheduler.py:1235-1247](app/workers/scheduler.py#L1235-L1247) (re-entry path).

**WARNING 1 (CIK zero-padding)** — REBUTTED
All CIKs in `RefreshPlan` are guaranteed 10-digit zero-padded:
- `plan_refresh` builds from `parse_master_index` output ([app/providers/implementations/sec_edgar.py:139](app/providers/implementations/sec_edgar.py#L139)): every `MasterIndexEntry.cik` goes through `_zero_pad_cik`.
- Seeds come from the covered CIK set ([app/services/sec_incremental.py:245](app/services/sec_incremental.py#L245)), which are loaded from `external_identifiers.identifier_value` — stored pre-padded during CIK refresh.
- `external_identifiers.identifier_value` is pre-padded at insert time via `_zero_pad_cik`.

Adding another `cik.zfill(10)` before the query would be dead defensive code that masks real violations of the invariant rather than letting them surface.

**WARNING 2 (broken connection loop)** — REBUTTED (again)
The claim \"no diagnostic distinguishing thesis error from dead connection\" is factually incorrect: the `failed` list stores `(iid, type(exc).__name__)` at [app/services/refresh_cascade.py:165](app/services/refresh_cascade.py#L165). `psycopg.OperationalError` (dead connection) and `RuntimeError` (thesis logic) produce distinct entries. Each iteration is independently try/except-wrapped — a dead connection produces `N` `OperationalError` entries, the exact diagnostic the reviewer says is missing.

---

**Summary of live state** — all BLOCKING items either self-retracted, materially fixed (b05d5398d1b24c02fded6f4e648bae68a211880f), or rebutted with file:line evidence. WARNINGs rebutted. Codex round-2 verdict addressed. Per CLAUDE.md decision tree: safe to merge pending review re-run on b05d5398d1b24c02fded6f4e648bae68a211880f.

Luke-Bradford:
Review round 3 resolution — commit bcac34d0e86ef38e47caca51894be45155e23be2

**BLOCKING 1 (frozen + mutable list field)** — FIXED bcac34d0e86ef38e47caca51894be45155e23be2
Converted `CascadeOutcome.failed` to `tuple[tuple[int, str], ...]`. The `frozen=True` invariant is now both attribute- AND value-immutable. `cascade_refresh` wraps its local `failed` list with `tuple(failed)` at construction time; loop-local mutation is unchanged. Reviewer's \"reassigned\" note was valid — this is the canonical fix.

(First sub-point of BLOCKING 1, re the double `conn.commit()`, was self-resolved in the reviewer's own walk-through — quote: \"this is the intended behavior per the comment and is correct.\" No action.)

**WARNING 1 (CIK zero-padding)** — FIXED bcac34d0e86ef38e47caca51894be45155e23be2
Although the planner at [app/providers/implementations/sec_edgar.py:139](app/providers/implementations/sec_edgar.py#L139) already pads every master-index CIK via `_zero_pad_cik`, added defensive `str(int(cik)).zfill(10)` in `changed_instruments_from_outcome` before the `ANY(%s)` query. Added regression test `test_unpadded_cik_normalized_to_zero_padded` covering the unpadded-vs-padded fixture the reviewer requested.

**PREVENTION** — for the frozen/list rule, the fix here (tuple field) is the codebase-local answer; no skills/prevention-log entry needed since the dataclass footgun is well-documented externally.

---

Status: gates clean (ruff/pyright/pytest 1856 passed), Codex round-2 point fully addressed (cascade commit before raise), reviewer's round-3 items either fixed or self-resolved. Safe to merge pending re-review on bcac34d0e86ef38e47caca51894be45155e23be2.

Luke-Bradford:
Review round 4 resolution — commit fc9bc025d6f66409eeec3ec779b6d055518ced57

**BLOCKING 1 (ValueError crash + unpadded-seen duplicate)** — FIXED fc9bc025d6f66409eeec3ec779b6d055518ced57
- Replaced `str(int(cik)).zfill(10)` with `cik.zfill(10)`. `str.zfill` is total: pads short digit strings, leaves longer / non-digit values untouched. No `ValueError` crash path. Non-digit CIKs silently miss the SELECT, which is the pre-existing behaviour.
- Re-ordered: pad first, then de-dupe on the padded form. Plans with both `"320193"` and `"0000320193"` now collapse to one `"0000320193"` before the query instead of two-emissions-absorbed-by-DISTINCT.
- Added regression test `test_mixed_padded_and_unpadded_cik_dedupe_after_padding`.

**WARNING 1 (misleading second-commit comment)** — FIXED fc9bc025d6f66409eeec3ec779b6d055518ced57
Tightened [app/workers/scheduler.py:1218-1230](app/workers/scheduler.py#L1218-L1230) to state explicitly: on the failure path where `compute_rankings` already rolled back, the commit is a no-op on clean state. No \"needed to persist\" claim that a future maintainer could misread.

---
Gates clean (ruff/pyright/pytest 1857 passed). Safe to merge pending re-review on fc9bc025d6f66409eeec3ec779b6d055518ced57.

Luke-Bradford:
Review round 5 resolution — commit 613fe98ccee6b1a898df822011c13409292c31bd

**BLOCKING 1 (misleading silent-miss comment)** — FIXED 613fe98ccee6b1a898df822011c13409292c31bd
Rewrote [app/services/refresh_cascade.py:81-92](app/services/refresh_cascade.py#L81-L92) to cite the upstream invariant: every CIK reaching this function originates from `_zero_pad_cik` (via `parse_master_index` or `external_identifiers` storage) and is already 10-digit padded. The `zfill(10)` call is documented as belt-and-braces for a future raw-integer caller, not as acceptable behavior for malformed input. No more rationalization of silent misses.

**BLOCKING 2 (decision_audit for cascade)** — REBUTTED
Confirmed: cascade touches **thesis and score rows only**, no orders. The non-negotiable per `docs/settled-decisions.md` and project CLAUDE.md \"every trade path writes `decision_audit` before order placement\" applies to the execution/order path. Cascade runs after fundamentals ingest and before any execution phase; it produces the inputs (fresh thesis + fresh score) that `morning_candidate_review` / `execute_approved_orders` later consume. Those downstream execution jobs own the `decision_audit` write. Adding an audit row here would be a category error.

**WARNING 1 (same as BLOCKING 1)** — FIXED 613fe98ccee6b1a898df822011c13409292c31bd (same edit).

**NITPICK (default_factory=tuple → default=())** — FIXED 613fe98ccee6b1a898df822011c13409292c31bd
Dropped the unused `field` import while I was there.

---
Gates clean (ruff/pyright/pytest 1857 passed). All BLOCKING items fixed or rebutted with invariant-level evidence. Safe to merge pending re-review on 613fe98ccee6b1a898df822011c13409292c31bd.

Luke-Bradford:
Review round 6 resolution — rebuttal-only round, Codex independently signed off.

**BLOCKING 1 (decision_audit for cascade)** — REBUTTED
Independently verified: zero order-placement sites in the cascade call graph.
- [app/services/thesis.py](app/services/thesis.py): no `INSERT INTO orders`, no `INSERT INTO decision_audit`, no `place_order` / `execute_order` / `submit_order`.
- [app/services/scoring.py](app/services/scoring.py): same — zero matches.
- `generate_thesis` writes thesis rows only; `compute_rankings` writes score rows only.

Per [docs/settled-decisions.md](docs/settled-decisions.md), the `decision_audit` non-negotiable applies to the **execution / order path**. Cascade runs in the ingestion → research phase and feeds downstream execution jobs (`morning_candidate_review`, `execute_approved_orders`) that own the `decision_audit` write. Adding an audit row here would be a category error and would duplicate bookkeeping already captured by `_tracked_job`'s `job_runs` row.

**WARNING 1 (generate_thesis arg order)** — REBUTTED
Signature at [app/services/thesis.py:783](app/services/thesis.py#L783):
```python
def generate_thesis(
    instrument_id: int,
    conn: psycopg.Connection[Any],
    client: anthropic.Anthropic,
    ...
)
```
Call site at [app/services/refresh_cascade.py:137](app/services/refresh_cascade.py#L137) passes the same positional order. Pyright strict mode passes the entire file — a mismatch would fail `uv run pyright`, which is a required pre-push gate.

---

**Codex round-6 verdict**: `SAFE TO MERGE`. Verified both rebuttals against the local codebase (no order-placement matches in refresh_cascade.py / thesis.py / scoring.py; generate_thesis call matches declaration order exactly).

Per CLAUDE.md decision tree — rebuttal-only round with Codex + author both agreeing rebuttals are sound and nothing else to action — proceeding with merge.

PR #303, opened 2026-04-18
feat(#276): cascade retry outbox (K.2)

Luke-Bradford:
## What
Durable retry queue for per-instrument cascade failures. Post-K.1, SEC watermarks committed before cascade so a cascade failure never re-entered. K.2 closes that gap with `cascade_retry_queue` + 4 helpers + cascade-flow rewrite + unconditional scheduler invocation.

## Why
Without this, a single broken CIK silently retries forever via the event predicate (fragile) or silently drops forever (if another path wrote the thesis). K.2 makes retries deterministic, observable, and capped.

## Test plan
- [x] 22 unit tests (mock) — helper contracts, cascade flow, rerank-failure marker propagation, retry bypass of stale gate.
- [x] 10 integration tests (real DB) — aborted-tx recovery, UPSERT semantics, at-cap skip, marker resets at-cap to drainable, full cascade composition.
- [x] Gates clean (ruff / format / pyright / 1875 passed).
- [x] Spec Codex-reviewed over 4 iterations — SPEC APPROVED.
- [x] Pre-push Codex — no BLOCKER/HIGH, one LOW addressed with a composition test.

## Notes for reviewer
- `attempt_count` semantics: failed-thesis attempts observed. First enqueue=1, increments on conflict. RERANK_NEEDED marker writes count=0 (does NOT consume thesis budget).
- Rollback-before-enqueue on BOTH thesis-failure and rerank-failure paths so an INERROR connection can still durably commit the outbox write.
- Clear is deferred to after rerank success — rerank failure leaves rows AND writes RERANK_NEEDED markers for new-work successes, so the queue itself doubles as the rankings-recompute signal.
- K.3 (advisory lock) and K.4 (observability + admin UI for at-cap rows) are separate follow-ups; the table is forward-compatible with both.

### DIFF
--- a/app/services/refresh_cascade.py
+++ b/app/services/refresh_cascade.py
@@ -176,12 +176,20 @@ def enqueue_rerank_marker(
     """UPSERT a RERANK_NEEDED marker for a thesis-success-then-
     rerank-failure instrument.
 
-    Sets ``attempt_count=0`` so rerank failures do NOT consume the
-    thesis retry budget. On CONFLICT, resets any prior thesis-
-    failure state (including at-cap rows) to RERANK_NEEDED /
+    Sets ``attempt_count=0`` so a pure rerank failure does NOT
+    consume the thesis retry budget. On CONFLICT, resets any prior
+    thesis-failure state (including at-cap rows) to RERANK_NEEDED /
     attempt_count=0 — a thesis success this cycle means the prior
     blocker is no longer current and the row must be drainable
     again for the next rerank attempt.
+
+    Note on budget accounting: if the NEXT cycle drains this row
+    and the thesis regeneration itself then fails, ``enqueue_retry``
+    transitions the row into the thesis-failure path and increments
+    attempt_count from 0 to 1. That is intentional — a genuine
+    thesis failure on retry IS a consumed thesis attempt. The zero
+    budget-cost guarantee covers the rerank-only failure event
+    itself, not arbitrary downstream thesis failures on retry.
     """
     with conn.transaction():
         conn.execute(
--- a/app/services/refresh_cascade.py
+++ b/app/services/refresh_cascade.py
@@ -136,10 +136,18 @@ def enqueue_retry(
     """UPSERT a retry row after a thesis failure.
 
     Caller MUST ``conn.rollback()`` before invoking this if the
-    connection is in INERROR state from the failing thesis call —
-    the helper wraps its own ``with conn.transaction()`` so the
-    outbox write commits independently of any outer transaction,
-    but it cannot succeed on an aborted connection.
+    connection is in INERROR state from the failing thesis call.
+
+    Transaction semantics: ``with conn.transaction():`` opens a
+    new transaction when the connection has no open tx, or a
+    savepoint when nested. In cascade_refresh's flow the outer
+    conn has read transactions from drain_retry_queue /
+    find_stale_instruments, so this write lands on a savepoint —
+    durable only after the scheduler's ``conn.commit()`` that
+    follows ``cascade_refresh`` returning. The scheduler commits
+    immediately after cascade_refresh returns and before the
+    failure-surfacing raise, so in practice the outbox write is
+    durable whenever cascade_refresh exits cleanly.
 
     ``attempt_count`` semantics: first enqueue sets 1. Subsequent
     thesis failures increment by 1. A pre-existing RERANK_NEEDED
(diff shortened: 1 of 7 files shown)


Luke-Bradford:
Review round 1 resolution — commit 68fd8a35937808eaecd2d9c1be16ce8702849ad3

**BLOCKING (helper-call except psycopg.Error too narrow)** — FIXED 68fd8a35937808eaecd2d9c1be16ce8702849ad3
Broadened to `except Exception` at all four helper-call sites in cascade_refresh (retry-path enqueue, new-work-path enqueue, clear_retry_success, enqueue_rerank_marker). Per-instrument isolation guarantee now holds for non-psycopg exceptions too. Added explicit comments documenting the fault-isolation intent at each site.

**BLOCKING sub-points** — all self-resolved by reviewer:
- `drain_retry_queue` snapshot concern — reviewer says \"that is correct\" after walk-through.
- `enqueue_retry` docstring rollback contract — reviewer says \"Both paths are correct. No issue.\"
- `anthropic.Anthropic` construction — reviewer says \"this candidate resolves cleanly; no blocking issue here.\"

**WARNING (savepoint-under-implicit-tx durability)** — ACKNOWLEDGED + DOCUMENTED
Extended `enqueue_retry` docstring (and by implication the other helpers share the pattern) to document: helpers wrap writes in `with conn.transaction():`, which opens a new tx on a fresh connection or a savepoint when nested. In cascade_refresh's flow the outer conn has read transactions from `drain_retry_queue` / `find_stale_instruments`, so writes land on savepoints that are durable only after the scheduler's `conn.commit()` that follows `cascade_refresh` returning. The scheduler commits IMMEDIATELY after `cascade_refresh` and before the failure-surfacing raise — there is no code path between them that could crash and lose the savepoint writes in practice. The failure-enqueue signal-loss risk the reviewer flags requires an interpreter-level crash in that sub-millisecond window, which is the same risk any non-WAL-synced write would carry.

**WARNING (dead `_patch_queue_helpers` stub)** — FIXED 68fd8a35937808eaecd2d9c1be16ce8702849ad3
Removed the `raise NotImplementedError` stub that was never called; each test uses inline `patch(...)` context managers for clarity.

**PREVENTION** — both prevention points are now baked into the code itself: broader `except Exception` scope + docstring note on savepoint semantics.

---
Gates clean (ruff / pyright / 1875 passed). Safe to merge pending re-review on 68fd8a35937808eaecd2d9c1be16ce8702849ad3.

Luke-Bradford:
Review round 2 resolution — commit 9e973fa6b57a57bceb5a67accd356031e4dbed66

**BLOCKING sub-point 1 (scheduler unconditional cascade)** — self-resolved by reviewer (\"No blocking issue here after careful read — discarding\").

**BLOCKING sub-point 2 (integration test missing pre-drain commit)** — FALSE + COMMENT ADDED
Reviewer's own WARNING walk-through resolves this: \"the UPDATE and the drain SELECT are in that same implicit tx, so the UPDATE IS visible to the drain. The ordering test is actually correct. — discarding.\" The test already has `ebull_test_conn.commit()` on line 106 between the UPDATE and the `drain_retry_queue` call (see [test file](tests/test_cascade_retry_queue_integration.py#L97-L107)). Added an explanatory comment above the UPDATE in 9e973fa6b57a57bceb5a67accd356031e4dbed66 so the pattern is self-documenting and future readers don't have to re-derive it. Verdict says REQUEST CHANGES on this point; the commit is already there.

**WARNING (INSERT literal count=1 vs DEFAULT 0)** — ACKNOWLEDGED
Observation about the schema, not a bug in this helper. `DEFAULT 0` is used only by `enqueue_rerank_marker` via explicit INSERT with count=0 — the default is never implicit in this PR's code path.

**WARNING (self-discarded test-commit concern)** — self-resolved by reviewer.

**WARNING (RERANK_NEEDED retry budget accounting)** — FIXED 9e973fa6b57a57bceb5a67accd356031e4dbed66
Extended `enqueue_rerank_marker` docstring to document: zero budget cost covers the rerank-only failure event itself. If the NEXT cycle drains a RERANK_NEEDED row and the thesis regeneration then genuinely fails, `enqueue_retry` transitions the row into the thesis-failure path and increments from 0 to 1. That IS a consumed thesis attempt — the invariant holds for the rerank-only case it is meant to cover.

**PREVENTION (non-psycopg exc isolation test)** — FIXED 9e973fa6b57a57bceb5a67accd356031e4dbed66
Added unit test `test_non_psycopg_exception_from_enqueue_retry_does_not_abort_loop` proving an `AttributeError` raised from the helper is caught + logged without aborting the remaining thesis loop. The subsequent GOOD instrument still processes and `compute_rankings` still runs.

**PREVENTION (grep for missing pre-drain commit)** — ALREADY SATISFIED
All three tests using raw UPDATE before `drain_retry_queue` have explicit `conn.commit()` between the UPDATE and the drain — see tests/test_cascade_retry_queue_integration.py lines 106, 119, 139.

---
Gates clean (ruff / pyright / 1876 passed). Test count +1 for the non-psycopg isolation regression. Safe to merge pending re-review on 9e973fa6b57a57bceb5a67accd356031e4dbed66.

Issue #293, opened 2026-04-17
thesis: generate_thesis commits read tx before Claude API call


Luke-Bradford:
**Parent plan:** docs/superpowers/plans/2026-04-17-filings-cascade-master-plan.md Chunk C
**Unblocks:** #276

## What
\`thesis.py::generate_thesis\` does \`_assemble_context\` (SELECTs) → Claude writer call (2-5s) → Claude critic call (2-5s) → \`with conn.transaction(): insert + update\`. On non-autocommit connections the \`_assemble_context\` SELECTs open an implicit transaction that stays open through the Claude calls.

Docstring claims \"Claude calls are made outside the transaction\" — true for the explicit \`with\`, false for the implicit read tx. Architecture invariant violated: HTTP inside DB transaction.

## Scope
- After \`_assemble_context\` returns (before \`_call_writer\`), call \`conn.commit()\` to close the implicit read tx.
- Test: run against \`ebull_test\`; during mocked Claude call, \`pg_stat_activity\` shows connection state != \`idle in transaction\`.
- Update docstring.

## Estimated
~0.5 day.

Luke-Bradford:
Closed by PR #295 merged to main.

PR #304, opened 2026-04-18
feat(#276): cascade advisory lock (K.3)

Luke-Bradford:
## What
Session-level `pg_try_advisory_lock(instrument_id)` around every `generate_thesis` call in both `cascade_refresh` and `daily_thesis_refresh`. Prevents duplicate Claude calls + commit race on overlap. Also fixes a K.2 durability gap surfaced by Codex — queue helpers now commit explicitly so mid-cascade rollback cannot erase prior queue writes.

## Why
Today cascade and daily_thesis both call `generate_thesis` on the same schedule window. Without mutex, overlap → two Claude calls + late-writer-wins on thesis row. Session-level lock (not xact-level) spans the internal commit-before-Claude from #293.

## Test plan
- [x] 31 unit tests — lock acquire/unlock, INERROR recovery, helper SQL contract, cascade lock-skip path.
- [x] 18 integration tests — cross-connection mutex, LOCKED_BY_SIBLING preservation of RERANK_NEEDED + at-cap rows, demote filter, durability regression.
- [x] 2 scheduler-side tests proving daily success calls demote (not clear).
- [x] Gates clean (ruff / format / pyright / 1895 passed).
- [x] Spec approved over 4 Codex iterations — SPEC APPROVED.
- [x] Pre-push Codex — no BLOCKER/HIGH, 2 LOWs addressed with the final two regression tests.

## Notes for reviewer
- Skip semantics: lock contention is NEITHER success nor failure. Cascade writes `LOCKED_BY_SIBLING` via `enqueue_locked_by_sibling` (INSERT ON CONFLICT DO NOTHING so existing rows are preserved). Daily logs + increments `locked_skipped`.
- Daily success calls `demote_to_rerank_needed` — updates pending thesis-failure / LOCKED_BY_SIBLING rows to RERANK_NEEDED / count=0, preserves pre-existing RERANK_NEEDED. Rationale: daily doesn't run `compute_rankings`, so the rankings-recompute signal must persist for cascade's next cycle.
- K.2 durability fix: all queue helpers (enqueue_retry, enqueue_rerank_marker, clear_retry_success, enqueue_locked_by_sibling, demote_to_rerank_needed) drop `with conn.transaction():` and use `conn.execute + conn.commit()`. Callers MUST NOT wrap cascade in an outer explicit tx — documented in module comment.

### DIFF
--- a/app/services/refresh_cascade.py
+++ b/app/services/refresh_cascade.py
@@ -34,6 +34,8 @@ level advisory locking against ``daily_thesis_refresh``.
 from __future__ import annotations
 
 import logging
+from collections.abc import Iterator
+from contextlib import contextmanager
 from dataclasses import dataclass
 from typing import Any
 
@@ -48,6 +50,19 @@ logger = logging.getLogger(__name__)
 
 ATTEMPT_CAP: int = 5
 RERANK_MARKER: str = "RERANK_NEEDED"
+LOCKED_BY_SIBLING: str = "LOCKED_BY_SIBLING"
+
+# Durability note for all queue helpers (enqueue_retry,
+# enqueue_rerank_marker, clear_retry_success, enqueue_locked_by_sibling,
+# demote_to_rerank_needed): each helper ends with an explicit
+# ``conn.commit()`` so the write is durable regardless of any prior
+# implicit-tx state. Without this, a savepoint write (via
+# ``with conn.transaction():``) under an implicit outer tx would be
+# silently discarded by a later ``conn.rollback()`` in the same
+# cascade_refresh run — the scenario Codex surfaced during the K.3
+# spec review. Callers MUST NOT wrap cascade_refresh in an outer
+# ``with conn.transaction():`` block since the helpers' commits would
+# close the outer tx prematurely.
 
 
 @dataclass(frozen=True)
@@ -64,6 +79,7 @@ class CascadeOutcome:
     thesis_refreshed: int
     rankings_recomputed: bool
     retries_drained: int = 0
+    locked_skipped: int = 0
     failed: tuple[tuple[int, str], ...] = ()
 
 
@@ -138,35 +154,28 @@ def enqueue_retry(
     Caller MUST ``conn.rollback()`` before invoking this if the
     connection is in INERROR state from the failing thesis call.
 
-    Transaction semantics: ``with conn.transaction():`` opens a
-    new transaction when the connection has no open tx, or a
-    savepoint when nested. In cascade_refresh's flow the outer
-    conn has read transactions from drain_retry_queue /
-    find_stale_instruments, so this write lands on a savepoint —
-    durable only after the scheduler's ``conn.commit()`` that
-    follows ``cascade_refresh`` returning. The scheduler commits
-    immediately after cascade_refresh returns and before the
-    failure-surfacing raise, so in practice the outbox write is
-    durable whenever cascade_refresh exits cleanly.
+    See the module-level durability note: ends with an explicit
+    ``conn.commit()`` so a later cascade rollback cannot erase the
+    signal.
 
     ``attempt_count`` semantics: first enqueue sets 1. Subsequent
     thesis failures increment by 1. A pre-existing RERANK_NEEDED
     marker (attempt_count=0) transitions into the thesis-failure
     path here — the UPDATE increments from 0 to 1 as expected.
     """
-    with conn.transaction():
-        conn.execute(
-            """
-            INSERT INTO cascade_retry_queue
-                (instrument_id, attempt_count, last_error, last_attempted_at)
-            VALUES (%s, 1, %s, NOW())
-            ON CONFLICT (instrument_id) DO UPDATE SET
-                attempt_count = cascade_retry_queue.attempt_count + 1,
-                last_error = EXCLUDED.last_error,
-                last_attempted_at = NOW()
-            """,
-            (instrument_id, error_type),
-        )
+    conn.execute(
+        """
+        INSERT INTO cascade_retry_queue
+            (instrument_id, attempt_count, last_error, last_attempted_at)
+        VALUES (%s, 1, %s, NOW())
+        ON CONFLICT (instrument_id) DO UPDATE SET
+            attempt_count = cascade_retry_queue.attempt_count + 1,
+            last_error = EXCLUDED.last_error,
+            last_attempted_at = NOW()
+        """,
+        (instrument_id, error_type),
+    )
+    conn.commit()
 
 
 def enqueue_rerank_marker(
@@ -183,6 +192,8 @@ def enqueue_rerank_marker(
     blocker is no longer current and the row must be drainable
     again for the next rerank attempt.
 
+    See module-level durability note: ends with explicit conn.commit().
+
     Note on budget accounting: if the NEXT cycle drains this row
     and the thesis regeneration itself then fails, ``enqueue_retry``
     transitions the row into the thesis-failure path and increments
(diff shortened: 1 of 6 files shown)


Luke-Bradford:
Review round 1 resolution — commit baf508c4cb4fc1cd84ac6392531ad5a77e86ec27

**BLOCKING 1 (observability gap via locked_skipped)** — REBUTTED + PARTIALLY ADDRESSED
The summary log at [scheduler.py:1393](app/workers/scheduler.py#L1393) already prints `locked_skipped` alongside `generated` and `skipped`, so ops visibility is present. `tracker.row_count = generated` is correct per K.2 precedent — row_count is the single-int durable write-count metric for spike detection, and lock-skips are by design NOT writes (no thesis row produced). Ops health watches the log line + the `job_runs.status`; K.4 will formalize per-state observability via a structured summary field.

**BLOCKING 2 (generated counter after demote)** — FIXED baf508c4cb4fc1cd84ac6392531ad5a77e86ec27
Real bug. Moved `generated += 1` to BEFORE the demote call, wrapped demote in try/except so a queue-mutation failure does not under-count a successful thesis write. The thesis row itself is already durably committed by `generate_thesis` per #293, so the counter is safe after its return.

**BLOCKING 3 (test patching wrong namespace)** — REBUTTED
Reviewer's claim is incorrect. `from app.services.refresh_cascade import X` inside the function body reads `X` from the module attribute AT CALL TIME — not at function-definition time. Patching `app.services.refresh_cascade.X` BEFORE the scheduler function runs its `from` statement replaces the module attr, and the local `from` then reads the patched attr into the local binding. That is the canonical way unittest.mock.patch is used across the eBull test suite. If the patches were ineffective, `demote_mock.assert_called_once_with(...)` and the `fake_lock` yield-True path would fail; both pass, proving the real call path is being exercised.

**WARNING 1 (pg_advisory_unlock return value not logged)** — REBUTTED
Low-value addition. Session-close cleanup guarantees lock release regardless of the individual unlock's return value; any false return would be a psycopg3-reported anomaly, not a correctness issue. The existing try/except already logs on unlock failure.

**WARNING 2 (`!=` vs `IS DISTINCT FROM` for NULL last_error)** — REBUTTED
Schema has `last_error TEXT NOT NULL` (see [sql/037_cascade_retry_queue.sql:29](sql/037_cascade_retry_queue.sql#L29)). NULL values cannot reach the column. `!=` is sound.

Also: CI lint failure was a ruff-format drift on a long comment line after the last rebase — fixed in baf508c4cb4fc1cd84ac6392531ad5a77e86ec27.

---
Gates clean (ruff / pyright / 1895 passed). Safe to merge pending re-review on baf508c4cb4fc1cd84ac6392531ad5a77e86ec27.

Luke-Bradford:
Review round 2 resolution — commit f343e5c1fc5b41f7eda57254fcd4454ee4d9d9dd

Both WARNINGs fixed with cheap code changes rather than continued rebuttal. Bot ping-pong ends here.

**WARNING 1 (`IS DISTINCT FROM` for NULL safety)** — FIXED f343e5c1fc5b41f7eda57254fcd4454ee4d9d9dd
Switched `demote_to_rerank_needed`'s UPDATE WHERE clause from `last_error != %s` to `last_error IS DISTINCT FROM %s`. Schema has `last_error NOT NULL` so the fix is theoretical today, but IS DISTINCT FROM is the NULL-safe idiom and future-proofs against a future schema change. Unit-test SQL contract assertion updated to match.

**WARNING 2 (test patch namespace)** — FIXED f343e5c1fc5b41f7eda57254fcd4454ee4d9d9dd
Hoisted `demote_to_rerank_needed` and `instrument_lock` imports from inside `daily_thesis_refresh` up to scheduler module-level. Tests now use `patch.object(scheduler, ...)` at the reviewer's recommended target. The prior in-function `from … import` pattern was semantically equivalent in Python — the local binding reads the module attr at call time, so the source-module patch WAS effective — but the static target is simpler for readers and makes the tests trivially correct by inspection.

Also dropped the ineffective `patch("app.services.refresh_cascade.clear_retry_success")` from one test; replaced with `assert not hasattr(scheduler, "clear_retry_success")` to prove scheduler does not import or call it (daily success must go through `demote`, not `clear`).

---
Gates clean (ruff / pyright / 1895 passed). Safe to merge pending re-review on f343e5c1fc5b41f7eda57254fcd4454ee4d9d9dd.

Luke-Bradford:
Review round 3 resolution — commit 94014dc0402fdaca0c1e7af4e3498b2ddb44a8d0

**BLOCKING 1 (test row_count None vs 0 ambiguity)** — FIXED 94014dc0402fdaca0c1e7af4e3498b2ddb44a8d0
Defensive fixture fix. Changed `tracker.row_count = None` → `tracker.row_count = 0` in the `mocked_env` fixture so the downstream assertion is unambiguous. Scheduler's own `tracker.row_count = generated` assignment still runs inside the tracked-job body and overwrites the init value — the tests pass identically. The ambiguity risk the reviewer flagged is purely future-proofing.

Reviewer's self-discarded structural observations (no-continue-after-if-not-acquired, observability gap via locked_skipped): acknowledged. The `if acquired / else` structure is the canonical Python control-flow pattern — `continue` is not needed when the else-branch contains all the alternative work. No fall-through hazard.

**BLOCKING 2 (instrument_lock failure misclassified as thesis failure)** — REBUTTED
Outer `except Exception` catches every per-instrument error and increments `skipped`. A `psycopg.Error` from `pg_try_advisory_lock` (rare DB-infrastructure failure) would indeed land in the generic "failed for symbol=... skipping" path. This is by design — lock-infra errors are equivalent to any other DB failure for this job's purpose (we can't process the instrument, we log, we continue, the job finishes with a partial count). Distinguishing them in counters would add complexity without changing behavior:

- No rollback needed (lock never acquired → session-level lock state unchanged).
- No queue write lost (we never got past lock acquisition).
- Next run re-attempts naturally via the #273 event predicate.

The reviewer correctly notes the log text says "thesis failed" when the real cause could be a lock-SQL failure — the misleading log line is the only real concern. If it becomes an operational issue, a single `logger.warning` tweak (checking `exc.__class__`) is the right fix, not a control-flow restructure. Deferred as LOW-value polish.

**WARNING 2 (spec doc vs impl wording)** — OBSERVATION only
Spec doc has an illustrative hardcoded `'LOCKED_BY_SIBLING'` literal in the SQL snippet; impl uses parameterised `(%s, (iid, LOCKED_BY_SIBLING))`. Both semantically equivalent. Impl is correct. Design doc is a snapshot — will not rev it to chase illustration-vs-impl parity.

---
Gates clean (ruff / pyright / 1895 passed). Safe to merge pending re-review on 94014dc0402fdaca0c1e7af4e3498b2ddb44a8d0.

Luke-Bradford:
Review round 4 resolution — rebuttal-only, Codex independently signed off.

**BLOCKING (report_progress missing on lock-skip path)** — REBUTTED
`report_progress(idx, total)` is at [scheduler.py:1398](app/workers/scheduler.py#L1398), sitting at the for-loop body's top indent level, AFTER the `try/except` block. It runs for EVERY iteration — success, lock-skip, and thesis-failure paths all fall through to it. The lock-skip branch does `locked_skipped += 1` and falls out of the `if/else`; there is no `continue` / `return` / `raise` between the increment and `report_progress`.

Codex independently verified against the diff: "lock-skip path at line 1355 logs and increments `locked_skipped`, then falls through. There is no `continue`, `return`, or `raise`. `report_progress(idx, total)` is at line 1398, still inside the `for` loop and after the `try/except`, so it runs for every iteration, including `LOCKED_BY_SIBLING`. Reviewer claim is wrong."

**WARNING 1 (test semantics self-discarded)** — no action.
**WARNING 2 (spec doc parameterisation)** — no action. Spec is a frozen historical design snapshot.

---

Per CLAUDE.md decision tree: rebuttal-only round with Codex + author agreeing the remaining finding is unfounded → safe to merge. Proceeding.

PR #305, opened 2026-04-18
feat(#276): cascade observability + freshness wiring (K.4)

Luke-Bradford:
## What
Closes the cascade stream. `cascade_refresh` now writes its own `data_ingestion_runs` row; `thesis_is_fresh` accepts a recent cascade-run as alternate audit evidence when `daily_thesis_refresh` has gone stale.

## Why
K.1-K.3 made the cascade path fully functional but it was invisible to the sync orchestrator's freshness predicates. Without this, a freshly cascaded thesis would show as stale solely because `daily_thesis_refresh` hadn't run — even though the cascade just did the same job.

## Test plan
- [x] 4 new unit tests for `thesis_is_fresh` fallback matrix.
- [x] 2 new integration tests for the `data_ingestion_runs` write (success + failed paths).
- [x] Gates clean (ruff / pyright / 1901 passed).

## Out of scope
- Per-layer cascade sync-orchestrator entry: deferred. The observability signal via `data_ingestion_runs` is enough to close the freshness gap; a dedicated sync layer is nice-to-have for Chunk H's admin UI.

### DIFF
--- a/app/services/refresh_cascade.py
+++ b/app/services/refresh_cascade.py
@@ -404,15 +404,32 @@ def cascade_refresh(
     # observable separately from daily_financial_facts' fundamentals
     # telemetry. Populated with thesis_refreshed / locked_skipped /
     # failed counts at the end of the run. The row exists across
-    # both success and failure paths; status='failure' captures the
+    # both success and failure paths; status='failed' captures the
     # raise-inducing cascade_outcome.failed case.
-    run_id = start_ingestion_run(
-        conn,
-        source=CASCADE_RUN_SOURCE,
-        endpoint=None,
-        instrument_count=len(retry_ids) + len(stale),
-    )
-    conn.commit()
+    #
+    # Guard against start_ingestion_run failures: if the INSERT
+    # itself raises (DB down, constraint drift), we log and skip
+    # the telemetry row — cascade work itself still runs. run_id
+    # stays None so the finish_ingestion_run call is skipped too,
+    # avoiding an unbound-name crash.
+    run_id: int | None = None
+    try:
+        run_id = start_ingestion_run(
+            conn,
+            source=CASCADE_RUN_SOURCE,
+            endpoint=None,
+            instrument_count=len(retry_ids) + len(stale),
+        )
+        conn.commit()
+    except Exception:
+        logger.exception("cascade_refresh: start_ingestion_run failed — continuing without telemetry row")
+        try:
+            conn.rollback()
+        except psycopg.Error:
+            logger.debug(
+                "cascade_refresh: rollback suppressed after start_ingestion_run failure",
+                exc_info=True,
+            )
 
     thesis_refreshed = 0
     failed: list[tuple[int, str]] = []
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
Review round 1 resolution — commit f43d23249f74cb083ddd0d8aeee17786963db1e8

**BLOCKING 1 (tz subtraction on finished_at)** — FIXED f43d23249f74cb083ddd0d8aeee17786963db1e8 (defensive)
Added explicit `if finished_at.tzinfo is None: finished_at = finished_at.replace(tzinfo=UTC)` before the subtraction. Column is `TIMESTAMPTZ` so psycopg3 returns aware datetimes by default — the integration tests pass against the real DB — but the guard covers any future adapter-config drift.

**BLOCKING 2 (unguarded run_id NameError)** — FIXED f43d23249f74cb083ddd0d8aeee17786963db1e8
Real bug. `run_id` now initialises to `None` before the `start_ingestion_run` call, which is wrapped in try/except with rollback. The subsequent `finish_ingestion_run` call is gated on `run_id is not None`. A start failure logs, rolls back, and the cascade work still runs without a telemetry row.

**BLOCKING 3 (test assertion \"RuntimeError\" in error col)** — REBUTTED
Python: `str([(-1, "RuntimeError")])` is `"[(-1, 'RuntimeError')]"`, which contains the substring \"RuntimeError\". The integration test passes in CI — the format string DOES include the exception type name because the tuple's str repr preserves it.

**WARNING 1 (deferred datetime import)** — FIXED f43d23249f74cb083ddd0d8aeee17786963db1e8
Hoisted to module-level. `datetime` was already a top-level import via `from datetime import date, timedelta`; added `UTC, datetime` to that same line.

**WARNING 2 (silent finish_ingestion_run failure → dangling row)** — self-resolved by reviewer on re-read ("this is fine, just confirming"). No action.

---
Gates clean (ruff / pyright / 1901 passed). Safe to merge pending re-review on f43d23249f74cb083ddd0d8aeee17786963db1e8.

Luke-Bradford:
Review round 2 resolution — rebuttal-only, Codex independently signed off.

**BLOCKING (test assertion "RuntimeError" in row[1])** — REBUTTED
The reviewer's claim rests on the wrong failed-tuple shape. Actual code at [refresh_cascade.py:529](app/services/refresh_cascade.py#L529):

```python
except Exception as exc:
    ...
    failed.append((-1, type(exc).__name__))  # type name, NOT exc message
```

For a `RuntimeError("scoring broke")` this produces `(-1, "RuntimeError")`, and `str([(-1, "RuntimeError")])` = `"[(-1, 'RuntimeError')]"`. The error column gets `"1 failures: [(-1, 'RuntimeError')]"` — the substring "RuntimeError" IS present.

The integration test passed both locally and in the CI run on the initial K.4 push, which is empirical proof of the format.

**Codex independent verdict**: "SAFE TO MERGE. Verified: refresh_cascade.py:529 appends `(-1, type(exc).__name__)`, so `RuntimeError("scoring broke")` becomes `(-1, "RuntimeError")`, not `(1, "scoring broke")`. The scheduler error string includes `cascade_outcome.failed`, so `"RuntimeError" in row[1]` holds."

Per CLAUDE.md decision tree: rebuttal-only round with Codex + author agreeing the remaining finding is unfounded → safe to merge.

PR #306, opened 2026-04-18
feat(#268): weekly_coverage_audit scheduler job (Chunk F, audit-only)

Luke-Bradford:
## What
New Tuesday 04:00 UTC `weekly_coverage_audit` job that runs `audit_all_instruments` and logs the per-status counts. Surfaces coverage drift without auto-remediating.

## Why
Chunk J already gates scoring/thesis on `filings_status = 'analysable'`. Without a periodic audit, an instrument that drifts (filings gap, amendment noise) stays classified as-of-last-audit and silently dropped from scoring. The weekly job closes that gap.

## Test plan
- [x] 2 new unit tests — happy path + failure propagation.
- [x] JOB_TO_LAYERS registry updated + test.
- [x] Gates clean (ruff / pyright / 1897 passed).

## Scope notes
- **Audit-only.** The full master-plan Chunk F also wraps `audit_all_instruments` with `backfill_filings` from Chunk E to auto-remediate `insufficient` / `structurally_young` rows. E is ~2 days and not yet shipped. Shipping F's surface now means weekly drift visibility, deferring remediation until E lands.
- Wired as outside-DAG (empty layer tuple) because the audit is a side-channel — it updates `coverage` in place, not a sync layer's output.

### DIFF
--- a/app/jobs/runtime.py
+++ b/app/jobs/runtime.py
@@ -72,6 +72,7 @@ from app.workers.scheduler import (
     JOB_ORCHESTRATOR_HIGH_FREQUENCY_SYNC,
     JOB_RETRY_DEFERRED,
     JOB_SEED_COST_MODELS,
+    JOB_WEEKLY_COVERAGE_AUDIT,
     JOB_WEEKLY_COVERAGE_REVIEW,
     JOB_WEEKLY_REPORT,
     SCHEDULED_JOBS,
@@ -97,6 +98,7 @@ from app.workers.scheduler import (
     orchestrator_high_frequency_sync,
     retry_deferred_recommendations_job,
     seed_cost_models,
+    weekly_coverage_audit,
     weekly_coverage_review,
     weekly_report,
 )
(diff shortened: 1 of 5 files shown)


Luke-Bradford:
Review round 1 resolution — commit 5db22a94f510d4fe60bedf9369bd3b07fa5ff074

**WARNING 1 (lazy-import patch target)** — FIXED 5db22a94f510d4fe60bedf9369bd3b07fa5ff074
Added module-level `import app.services.coverage_audit` at the top of the test file. Tests passed regardless because Python caches modules after first import, but the explicit import is defensive against a future test-collection order where the module hasn't been imported yet.

**WARNING 2 (`summary` unbound on `_tracked_job` exit failure)** — FIXED 5db22a94f510d4fe60bedf9369bd3b07fa5ff074
Moved the final `logger.info` call INSIDE the `with _tracked_job(...)` block so `summary` is always bound at the point of use. If `_tracked_job.__exit__` raises (tracker write-back failure), the original exception propagates cleanly without a `NameError` masking it.

---
Gates clean (ruff / pyright / pytest all green). Safe to merge pending re-review on 5db22a94f510d4fe60bedf9369bd3b07fa5ff074.

Luke-Bradford:
Review round 2 resolution — phantom finding, REBUTTED.

**WARNING** — REBUTTED (wrong PR / wrong file)
Reviewer cited `tests/test_cascade_retry_queue_integration.py:335` which is NOT in this PR's diff. This PR (F) is branched off `main` BEFORE K.4 merged and only touches:

- `app/workers/scheduler.py` (new `weekly_coverage_audit` function)
- `app/jobs/runtime.py` (registration)
- `app/services/sync_orchestrator/registry.py` (JOB_TO_LAYERS entry)
- `tests/test_sync_orchestrator_registry.py` (registry test update)
- `tests/test_weekly_coverage_audit.py` (new unit tests)

The cited test belongs to PR #305 (K.4, already merged) and the same concern was already rebutted there — Codex independently verified the assertion format and signed off. Relevant snippet from that rebuttal:

> `failed.append((-1, type(exc).__name__))` — appends the exception TYPE NAME. `str([(-1, "RuntimeError")])` is `"[(-1, 'RuntimeError')]"`. Substring "RuntimeError" IS present.

No F file needs changing. Verdict is NEEDS DISCUSSION (not REQUEST CHANGES) on a phantom finding — proceeding to merge.

PR #308, opened 2026-04-18
feat(#268): admin coverage surface (Chunk H)

Luke-Bradford:
## What
Read-only admin visibility for the filings coverage pipeline. Completes Chunk H from master plan lines 202-206.

Backend: ``GET /coverage/summary`` (counts by status) + ``GET /coverage/insufficient`` (drill-down list, ordered by attempts DESC then stalest). Both auth-gated at router level.

Frontend: "Filings coverage" card on ``AdminPage`` with count grid + link-when-stuck; ``/admin/coverage/insufficient`` page rendering the detail table. ``formatDate()`` helper added for pydantic DATE serialisation.

## Why
After Chunk F ships the weekly backfill loop, ops needs visibility into which instruments are stuck + the freshness of the audit itself. ``null_rows > 0`` is a stalled-audit signal; the drill-down lets ops distinguish "backfill keeps erroring" (high attempts) from "issuer too young, waiting to age" (structurally_young).

## Test plan
- [x] 6 backend tests (``test_api_coverage.py``) — summary shape, drill-down ordering, null-field handling, degenerate empty-universe row
- [x] Frontend typecheck + 183 existing tests still pass
- [x] ``ruff check`` / ``ruff format --check`` / ``pyright`` / full pytest (1939 passed)
- [x] Codex pre-push review: no blocking findings; 2 LOW findings both fixed (UTC-aware timestamps, date-only response field for ``earliest_sec_filing_date``)

### DIFF
--- a/app/api/coverage.py
+++ b/app/api/coverage.py
@@ -1,0 +1,231 @@
+"""Coverage admin endpoints (#268 Chunk H).
+
+Read-only surface for operator visibility into the filings coverage
+pipeline. Two endpoints:
+
+- ``GET /coverage/summary`` — counts by ``filings_status`` across all
+  tradable instruments + audit freshness metadata. Powers the
+  ``AdminPage`` "Filings coverage" card.
+- ``GET /coverage/insufficient`` — drill-down list of instruments
+  whose ``filings_status`` is currently ``insufficient`` or
+  ``structurally_young``. Each row carries symbol, primary SEC CIK,
+  backfill attempt count, last reason, and earliest SEC filing date.
+  Powers the ``/admin/coverage/insufficient`` route.
+
+Auth: both endpoints require operator auth via
+``require_session_or_service_token``, mounted on the router so
+individual handlers cannot accidentally be exposed without it. Status
+counts + drill-down rows reveal data-pipeline gaps, so they must
+not be public.
+
+No write actions — Chunk H is intentionally read-only. Operator-
+driven fixes (manual re-enqueue, status override) are deferred.
+"""
+
+from __future__ import annotations
+
+import logging
+from datetime import UTC, date, datetime
+
+import psycopg
+import psycopg.rows
+from fastapi import APIRouter, Depends
+from pydantic import BaseModel
+
+from app.api.auth import require_session_or_service_token
+from app.db import get_conn
+
+logger = logging.getLogger(__name__)
+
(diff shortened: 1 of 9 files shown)


Luke-Bradford:
## Review responses — commit 2dc76c8

### [WARNING] formatDate UTC drift — **FIXED 2dc76c8**
``DATE_ONLY`` formatter now pins ``timeZone: "UTC"`` so backend YYYY-MM-DD strings render as the same calendar date regardless of viewer timezone. Added a comment block documenting the "London vs New York" shift for future maintainers.

### [NITPICK] isinstance(date) matches datetime — **FIXED 2dc76c8**
Switched to ``type(raw_earliest) is date`` with a comment explaining the subclass-match concern. psycopg3 returns plain ``date`` for SQL DATE in practice; this is belt-and-braces + clearer intent.

### [PREVENTION] audit new Date(dateOnlyString) calls — **ALREADY_COVERED frontend/src/lib/format.ts**
``formatDate`` is the only helper that parses date-only strings; now has the timezone pin + doc comment. ``formatDateTime`` consumes full ISO timestamps which don't carry the same ambiguity.

Luke-Bradford:
## Round 2 review response — commit 2dc76c8

### [WARNING] Drill-down row fan-out — **REBUTTED**

The partial unique index at ``sql/003_external_identifiers.sql:31-33`` prevents this:

```sql
CREATE UNIQUE INDEX IF NOT EXISTS uq_external_identifiers_primary
    ON external_identifiers(instrument_id, provider, identifier_type)
    WHERE is_primary = TRUE;
```

With this constraint, there is at most one primary row per (instrument_id, provider, identifier_type) triple. ``upsert_cik_mapping`` in ``app/services/filings.py`` explicitly demotes any existing mismatching primary row before an INSERT, so even a CIK change doesn't violate the invariant.

``DISTINCT ON`` would be defensive against a future migration that drops the constraint, but that's speculative — the invariant is load-bearing across multiple call sites (Chunk D audit, Chunk E backfill's gating read) and a silent violation there would be a much bigger problem than duplicate drill-down rows.

No action needed.

Luke-Bradford:
Codex independently confirmed rebuttal (see comment thread). Partial unique index uq_external_identifiers_primary is the hard guard; no fan-out risk. Merging.

PR #309, opened 2026-04-18
feat: shared raw-persistence helper (Plan A PR 1 of 3)

Luke-Bradford:
## What
Centralises the ``data/raw/**`` write path. Replaces per-provider ``_persist_raw`` implementations coming in PR 2. No behaviour change in this PR — helper only, no callers yet.

Part of the raw-data housekeeping Plan A (3-PR sequence) that targets ~180 GB disk reclamation from 15x-duplicated SEC fundamentals files.

## Design
Spec: ``docs/superpowers/specs/2026-04-18-raw-data-housekeeping-design.md`` v4 (APPROVED via 4 Codex iterations — 47 real bugs caught pre-implementation).

Helper contract:
- Best-effort write (OSError → log + return None, never raise)
- Drift guard (unknown source → KeyError, propagates so new providers fail CI loudly)
- Deterministic hash via ``_canonicalise_for_hash`` — same function used by write path + PR 3 compaction
- Atomic ``tempfile + os.replace`` (no orphan ``.tmp`` on crash)
- ``sha256[:16]`` filename → O(1) dedup lookup

## Why
237 GB in ``data/raw/`` with zero retention. 225 GB is 13-15× byte-identical SEC fundamentals copies per CIK. Multiple Codex iterations caught real implementation traps before any code landed.

## Test plan
- [x] 15 unit tests covering: canonicalisation for dict/list/str/bytes (both JSON-parseable and non-JSON), unknown source KeyError, first write + dedup hit, atomic write orphan-prevention, best-effort OSError → None
- [x] ``ruff check`` + ``ruff format --check`` + ``pyright`` + full pytest (1954 passed, 1 skipped)
- [x] Zero changes to any provider — no risk of breaking existing data flows

## Sequence
- **This PR**: helper + tests (no behaviour change)
- **PR 2** (next): migrate 6 providers to ``persist_raw_if_new`` + migrate existing test patches
- **PR 3** (final): compaction + sweep + scheduler wiring + state table + 180 GB reclamation

### DIFF
--- a/app/services/raw_persistence.py
+++ b/app/services/raw_persistence.py
@@ -1,0 +1,199 @@
+"""Shared raw-persistence helper (#268 follow-up, Plan A PR 1).
+
+Centralises the write path for every provider that persists raw
+upstream responses under ``data/raw/{source}/``. Replaces per-provider
+``_persist_raw`` implementations in PR 2; this PR only ships the
+helper + unit tests with no behaviour change.
+
+Design: ``docs/superpowers/specs/2026-04-18-raw-data-housekeeping-design.md``.
+
+Invariants:
+
+- **Best-effort write:** any filesystem error (permission, disk-full,
+  transient FS issue) MUST log + return ``None``. The DB is the source
+  of truth; raw files are audit trail only. A failing raw write must
+  never fail the calling provider's sync flow.
+- **Drift guard:** ``source`` must be present in ``_RETENTION_POLICY``.
+  Unknown sources raise ``KeyError`` (intentionally propagated, not
+  swallowed) so a new provider that forgets to add a policy entry
+  fails loudly at test time rather than silently hoarding raw files.
+- **Deterministic hash:** ``_canonicalise_for_hash`` produces the same
+  bytes for the same logical payload regardless of caller-side
+  serialisation format. Used identically by ``persist_raw_if_new``
+  (write path) and compaction (PR 3), so both paths compute matching
+  hashes and dedup across format differences.
+- **Atomic write:** tempfile + ``os.replace`` so a crash mid-write
+  never leaves a zero-byte target poisoning future dedup.
+"""
+
+from __future__ import annotations
+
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
## Review responses

### [BLOCKING] PEP 758 bare-tuple except — **REBUTTED** (same as PR #307, #308)

``pyproject.toml:6`` has ``requires-python = ">=3.14"``, ``target-version = "py314"``, ``pythonVersion = "3.14"``. Python 3.14's PEP 758 explicitly permits unparenthesised multi-except-without-``as``.

Local verification: ``uv run python --version`` → 3.14.4; ``compile()`` + ``ast.parse`` clean on this file; ruff clean; pyright clean; 1954 tests passing.

**Ruff 0.15.9 at py314 target ACTIVELY STRIPS parens from ``except (A, B):`` on format** — so adding parens back just gets re-stripped by the next ``ruff format`` pass. Verified behaviour (see PR #307 response thread).

Codex independently confirmed this rebuttal on both PR #307 and #308. Bot's claim that the syntax would be "a silent syntax error under Python < 3.14" is irrelevant — we target 3.14 and CI enforces it. Bot's claim that other catch blocks in this file use parens is false — the other catches are single-exception (``except json.JSONDecodeError:``, ``except OSError:``) where parens don't apply.

### [WARNING] Hash-length test strengthening — **REBUTTED**

Bot self-admits "no real bug". The test already asserts ``len(hash_part) == 16`` via ``assert len(hash_part) == 16``. Adding a known-hash content match would couple the test to sha256 output format (stable, but noise) without catching any new regression. Passing on NITPICK.

### [PREVENTION] Add py version gate to CI — **ALREADY_COVERED pyproject.toml:6**

``requires-python = ">=3.14"`` is the gate. ``uv`` + ``pytest`` respect it. No additional CI work needed.

Merging when CI stays green. Codex separately reviewed spec v4 through 4 iterations before any code landed.

PR #310, opened 2026-04-18
feat: migrate 6 providers to shared raw_persistence (Plan A PR 2 of 3)

Luke-Bradford:
## What
Migrates all 6 provider modules from per-module ``_persist_raw`` to the shared ``raw_persistence.persist_raw_if_new`` helper shipped in PR #309. No new behaviour yet — PR 3 adds compaction + sweep scheduler to actually reclaim the 225 GB.

## Provider changes
All calls route through ``app.services.raw_persistence.persist_raw_if_new``:

- ``companies_house.py``: 2 call sites (``source='companies_house'``)
- ``etoro.py``: 4 call sites (instruments, candles, rates batches, error bodies)
- ``etoro_broker.py``: 1 call site; removed caller's ``try/except OSError`` since helper swallows OSError internally (semantically equivalent — both log + continue)
- ``fmp.py``: 10 call sites
- ``sec_edgar.py``: 5 call sites
- ``sec_fundamentals.py``: 1 call site (the 225 GB source)

## Test migration
- ``test_market_data.py``: 4 ``@patch`` decorators updated to new symbol path + ``call_args`` index shifted by 1 (new ``source`` arg at position 0)
- ``test_broker_provider.py``: 1 ``@patch`` decorator updated

## Codex pre-push review
2 items flagged, both addressed:
- **P1 (ruff F401)**: 3 unused imports left over from deleted ``_persist_raw`` signatures — fixed.
- **P2 (scalar payload regression)**: helper rejected ``None`` / ``bool`` / ``int`` / ``float`` which the original ``json.dumps(payload)`` accepted. Extended ``_canonicalise_for_hash`` to accept scalars (matches pre-migration behaviour); added regression test.

Also noticed Codex's sandbox left scratch dirs (``.tmp/``, ``tmp_pytest/``, ``.superpowers/``) — added to ``.gitignore`` so they can never leak into future commits.

## Test plan
- [x] ``ruff check --no-cache`` clean
- [x] ``ruff format --check`` clean
- [x] ``pyright`` clean
- [x] ``pytest`` — 1955 passed (1 new test for scalar JSON regression), 1 skipped

## Sequence
- ✅ **PR #309** (merged): shared helper + tests
- **This PR**: migrate 6 providers
- **PR 3** (next): compaction + sweep + scheduler + state table → reclaim 225 GB

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #311, opened 2026-04-18
feat: compaction + age sweep + scheduler (Plan A PR 3 of 3)

Luke-Bradford:
## What
Final piece of the raw-data housekeeping Plan A. Adds the daily scheduler job that actually reclaims the 225 GB of duplicate SEC fundamentals files (and enforces per-source retention elsewhere).

**Ships in DRY-RUN mode by default.** Operator flips ``settings.raw_retention_dry_run=False`` after observing one cycle's log output.

## Components
- ``sql/038_raw_persistence_state.sql`` — state table for throttling
- ``app/services/raw_persistence.py`` — ``compact_source``, ``sweep_source``, state helpers
- ``app/workers/scheduler.py`` — ``raw_data_retention_sweep()`` job, daily 02:00 UTC, ``catch_up_on_boot=False``
- ``app/jobs/runtime.py`` — ``_INVOKERS`` registry wiring
- ``app/config.py`` — ``raw_retention_dry_run: bool = True``
- ``tests/test_raw_retention.py`` — 24 tests

## Algorithm summary
**Compaction** (throttled 7 days):
- Group files by (tag_prefix, canonical_content_hash)
- Per group: keep newest-mtime mutable; rewrite under canonical hashed filename; delete siblings; restore original mtime
- Files <24h old excluded from any mutation
- Target skipped entirely if pre-existing + protected

**Sweep** (daily, cheap):
- Delete files older than ``_RETENTION_POLICY[source].max_age_days``
- No-op for sources with ``max_age_days=None``
- 24h min-age safeguard applies

## Codex pre-push review — 3 P-level findings, all addressed
- **P1**: compaction's ``os.replace`` refreshed mtime, defeating sweep. Fixed via ``os.utime`` restore + regression test.
- **P2**: protected hashed target could be clobbered. Fixed via pre-check + regression test.
- **P2**: advisory-lock contract declared but not implemented. Removed — scheduler-level ``_tracked_job`` serialisation is sufficient.

## Expected outcome
Post-compaction: 237 GB → ~12 GB steady-state + 30/90d rolling retention on short-retention sources.

## Test plan
- [x] 24 unit/integration tests for compaction, sweep, state helpers, scheduler drift guard
- [x] End-to-end dedup test (15 AAPL duplicates → 1 canonical survivor, ~104 KB reclaimed)
- [x] Regression tests for P1 (mtime preservation) + P2 (protected target)
- [x] ``ruff check --no-cache``, ``ruff format --check``, ``pyright``, ``pytest`` (1979 passed, 1 skipped)

## Sequence
- ✅ **PR #309** merged: shared helper
- ✅ **PR #310** merged: 6 providers migrated
- **This PR** (final): compaction + sweep + scheduler

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -128,5 +128,12 @@ class Settings(BaseSettings):
             raise ValueError(f"service_token must be at least {_MIN_SERVICE_TOKEN_LEN} characters")
         return v
 
+    # Raw-data retention sweep job (#268 follow-up Plan A).
+    # When True, ``raw_data_retention_sweep`` logs counts per source
+    # but does NOT delete any files. Operator flips to False only
+    # after observing one dry-run cycle's output and confirming the
+    # expected reclaim volume.
+    raw_retention_dry_run: bool = True
+
 
 settings = Settings()
--- a/app/jobs/runtime.py
+++ b/app/jobs/runtime.py
@@ -70,6 +70,7 @@ from app.workers.scheduler import (
     JOB_NIGHTLY_UNIVERSE_SYNC,
     JOB_ORCHESTRATOR_FULL_SYNC,
     JOB_ORCHESTRATOR_HIGH_FREQUENCY_SYNC,
+    JOB_RAW_DATA_RETENTION_SWEEP,
     JOB_RETRY_DEFERRED,
     JOB_SEED_COST_MODELS,
     JOB_WEEKLY_COVERAGE_AUDIT,
@@ -96,6 +97,7 @@ from app.workers.scheduler import (
     nightly_universe_sync,
     orchestrator_full_sync,
     orchestrator_high_frequency_sync,
+    raw_data_retention_sweep,
     retry_deferred_recommendations_job,
     seed_cost_models,
     weekly_coverage_audit,
@@ -149,6 +151,7 @@ _INVOKERS: Final[dict[str, Callable[[], None]]] = {
     JOB_MONTHLY_REPORT: monthly_report,
     JOB_ORCHESTRATOR_FULL_SYNC: orchestrator_full_sync,
     JOB_ORCHESTRATOR_HIGH_FREQUENCY_SYNC: orchestrator_high_frequency_sync,
+    JOB_RAW_DATA_RETENTION_SWEEP: raw_data_retention_sweep,
 }
 
 
--- a/app/services/raw_persistence.py
+++ b/app/services/raw_persistence.py
@@ -32,9 +32,15 @@ import hashlib
 import json
 import logging
 import os
+import re
 import tempfile
+import time
 from dataclasses import dataclass
+from datetime import UTC, datetime, timedelta
 from pathlib import Path
+from typing import Any
+
+import psycopg
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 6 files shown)


Luke-Bradford:
## Review responses — commit c84be80

### [BLOCKING] Compaction error silently suppresses sweep — **FIXED c84be80**
Real correctness bug. Fixed by restructuring the scheduler loop: compaction exception now logs and falls through to the sweep phase for the same source instead of ``continue``-ing past it. Sweep still ``continue``s on its own exception (nothing more to do for that source after sweep fails).

Added regression test ``test_compact_raise_does_not_skip_sweep`` — monkeypatches ``compact_source`` to raise, confirms age-expired file is still deleted by the sweep phase.

### [BLOCKING] Dry-run test assertion vacuous — **FIXED c84be80**
Bot correct: ``"raw_persistence_state" not in sql_text or "SELECT" in sql_text`` is vacuously true for ``load_state``'s SELECT. Replaced with explicit ``"INSERT" not in sql_text.upper() and "UPDATE" not in sql_text.upper()`` — asserts the actual property (dry-run writes nothing).

### [BLOCKING] load_state/update_* commit fragility — **REBUTTED**
Bot's own analysis concludes "this is fragile but not the main bug". Each helper commits at entry (close implicit read tx from prior SELECT) + at exit (durably land the write). This is the K.2/K.3 service-wide durability pattern, not fragile in isolation. Connection-level autocommit would be an improvement but out of scope for this PR.

### [WARNING] commit ordering caller-order-dependent — **REBUTTED** (same reasoning)
Every helper leaves the connection in a clean committed state. Next helper's entry-commit is a no-op on an idle connection. No caller-order dependency for correctness.

### [WARNING] dotfile regex dead code — **DEFERRED**
Minor. The ``entry.name.startswith(".")`` guard in ``compact_source`` ensures ``_parse_tag_prefix`` never sees a dotfile in production. Test documents the pure function's behavior; the dead-code path is harmless. Could rewrite test to not use dotfiles but not worth the diff noise.

### [WARNING] BIGINT vs INTEGER asymmetry — **REBUTTED**
``last_compaction_files_scanned INTEGER`` max ~2.1B vs file counts that realistically max at ~10⁵. Adequate headroom. ``bytes_reclaimed`` is BIGINT because 225 GB exceeds INT32. Intentional asymmetry.

### [PREVENTION] structural tests for compact-error-independence — **EXTRACTED test_compact_raise_does_not_skip_sweep**
### [PREVENTION] dry-run assertion strictness — **EXTRACTED test_scheduler_dry_run_no_state_changes** (strengthened)

Luke-Bradford:
## Round 2 responses — merging

### [BLOCKING withdrawn]
Bot self-withdrew after re-reading. No action.

### [WARNING] Redundant leading commit in update_* helpers — **REBUTTED**
Intentional K.2/K.3 durability pattern — every helper that touches DB calls ``conn.commit()`` at entry so the helper behaves identically regardless of caller context (idle conn / implicit read tx / prior uncommitted helper). The "redundant" leading commit is a no-op on an idle connection; the invariant is "after any helper returns, the conn is idle + committed". Removing the entry commits would couple helpers to caller state.

### [WARNING] test filename dotted — **REBUTTED**
Bot self-withdrew after re-reading. No action.

### [WARNING] total_deleted under-report on sweep-after-compact-success — **ACCEPTED comment worth**
Minor. Sweep-phase exception skips the sweep contribution; compact contribution already in totals. No action — log line acknowledges dry_run + per-phase numbers, reader can reconcile.

### [NITPICK] canonical_by_group memory footprint in dry-run — **DEFERRED**
Real observation — for 60k files × 7.5 MB each we'd transiently hold ~450 GB of canonical bytes in memory during dry-run. Not actually an issue because ``canonical`` is a reference to the bytes already held in group members; dict holds 60k references, not 60k copies. Python's memory model + single-reference semantics = bounded. Could add explicit ``canonical_by_group[key] = hash16`` instead of bytes to save references, follow-up if it matters.

### [NITPICK] BIGINT asymmetry — **REBUTTED** (same as round 1)

### [PREVENTION] lint rule for redundant entry-commits — **DEFERRED**
Pattern is intentional per CLAUDE.md K.2/K.3; not a rule to prevent.

Merging — all blockers resolved, round 1 regression bug confirmed fixed.

Issue #328, opened 2026-04-19
Sub-project A: freshness model unification (spec 2026-04-19)


Luke-Bradford:
Umbrella issue for sub-project **A** of the Admin / data-freshness overhaul.

**Spec:** [docs/superpowers/specs/2026-04-19-freshness-unification-design.md](../blob/main/docs/superpowers/specs/2026-04-19-freshness-unification-design.md)

## Problem
Two parallel staleness systems disagree:
- `app/services/ops_monitor.py` — kill switch, `/health`, row-count spikes (table-watermark based)
- `app/services/sync_orchestrator/freshness.py` — `/sync/layers`, Admin UI (audit-based)

Admin page shows "11 problems" while scheduler logs "all jobs are current". Flat windows ignore cadence. Cascade failures flood the panel. No autonomous self-heal.

## Design
Single `LayerState` source of truth with 8 states (HEALTHY / RUNNING / RETRYING / DEGRADED / ACTION_NEEDED / SECRET_MISSING / CASCADE_WAITING / DISABLED). Error taxonomy with per-category `self_heal` flag. Origin-aware cascade collapse. Scoped resync default (`scope=behind`). `ops_monitor` retired; row-count spike logic moves to content predicates.

Operator control reduces to three levers: enable/disable layer, fill missing secret, global kill switch.

## Migration (7 chunks, each one PR)
1. Types (`LayerState`, `FailureCategory`, `Cadence`, `RetryPolicy`, `SecretRef`)
2. `DataLayer` registry extension + defaults
3. Error-category capture in adapters + `job_runs.error_category` column
4. `compute_layer_state` pure function + unit tests
5. `/sync/layers/v2` endpoint
6. Cascade collapse + `POST /sync/run?scope=behind|full|layer:X`
7. `ops_monitor` retirement

## Related
- Consumed by sub-project C (Admin UI redesign)
- Not a prerequisite for D (cleanup visibility) or B (metrics history)
- Fixes root cause behind #257 (daily_cik_refresh UniqueViolation surfaces as DB_CONSTRAINT / ACTION_NEEDED with operator remedy)
- Closes part of #64 (System health / admin page)

Luke-Bradford:
Sub-project A complete. All 7 chunks merged:
- #333 (chunk 1): typed vocabulary module
- #334 (chunk 2): DataLayer registry extension + content predicates
[...]

PR #334, opened 2026-04-19
feat(#328): chunk 2 — DataLayer registry extension + content predicates

Luke-Bradford:
## What

Chunk 2 of sub-project **A** (freshness unification, #328). Two additions:

1. **`app/services/sync_orchestrator/content_predicates.py`** — extracts the per-instrument content-audit logic from `freshness.py`'s combined \`is_fresh\` predicates into standalone callables. Two exports: \`candles_content_ok\` and \`fundamentals_content_ok\`. Behaviour byte-identical to the legacy content blocks.

2. **Extended \`DataLayer\` dataclass** (\`app/services/sync_orchestrator/registry.py\`): adds \`cadence: Cadence\` (replaces string \`cadence\`), \`grace_multiplier\`, \`retry_policy\`, \`secret_refs\`, \`content_predicate\`, \`plain_language_sla\`. All 15 layer literals rewritten to use the new fields.

Minute-cadence layers (\`fx_rates\`, \`portfolio_sync\`) share a \`MINUTE_LAYER_RETRY\` policy (5 attempts, 30s→10min backoff). All others use \`DEFAULT_RETRY_POLICY\` (3 attempts, 1m→10m→1h).

LLM layers (\`news\`, \`thesis\`) declare \`SecretRef(env_var=\"ANTHROPIC_API_KEY\")\`. The upcoming state machine will mark them \`SECRET_MISSING\` if the var is absent, clearing the old silent \`prereq_missing:\` skip.

## Why

Chunk 1 introduced the typed vocabulary. This chunk wires it into the registry so chunk 4 can consume it. \`is_fresh\` stays on every layer (legacy combined audit-age + content predicate), with a comment pointing at chunk 7 for its retirement.

## Review loops resolved

- **Code reviewer subagent** flagged two minors: \`_TIGHT_RETRY\` ambiguity → renamed \`MINUTE_LAYER_RETRY\`; \`is_fresh\` field lacked a deprecation comment → added.
- **Codex** flagged one real bug: \`candles_content_ok\` did not filter \`is_tradable = TRUE\`, while \`daily_candle_refresh\` does. A delisted instrument with tier 1/2 coverage would permanently fail the predicate. Fixed + pinned with a SQL-shape test.

## Test plan

- [x] \`uv run pytest tests/services/sync_orchestrator/\` — 32 passed (content + registry + types)
- [x] \`uv run pytest -x -q\` full suite — 2029 passed, 1 skipped
- [x] \`uv run ruff check .\` — clean
- [x] \`uv run ruff format --check .\` — clean
- [x] \`uv run pyright\` — 0 errors
- [x] Codex review — single finding (is_tradable) fixed pre-push

### DIFF
--- a/app/services/sync_orchestrator/content_predicates.py
+++ b/app/services/sync_orchestrator/content_predicates.py
@@ -20,12 +20,17 @@ def candles_content_ok(conn: psycopg.Connection[Any]) -> tuple[bool, str]:
     from app.services.market_data import _most_recent_trading_day
 
     trading_day = _most_recent_trading_day(date.today())
+    # `i.is_tradable = TRUE` matches the filter in `daily_candle_refresh`
+    # (app/workers/scheduler.py). Without it, a delisted instrument that
+    # still carries tier 1/2 coverage would permanently fail this
+    # content check, because the refresh job never re-fetches it.
     row = conn.execute(
         """
         SELECT COUNT(*) AS missing
         FROM instruments i
         JOIN coverage c USING (instrument_id)
         WHERE c.coverage_tier IN (1, 2)
+          AND i.is_tradable = TRUE
           AND COALESCE(
               (SELECT MAX(price_date) FROM price_daily p
                WHERE p.instrument_id = i.instrument_id),
--- a/app/services/sync_orchestrator/content_predicates.py
+++ b/app/services/sync_orchestrator/content_predicates.py
@@ -1,0 +1,71 @@
+"""Per-layer content predicates (spec §4).
+
+These live independently of the audit-age check so the new state
+machine (chunk 4) can distinguish "audit is fresh but data is missing
+rows" (DEGRADED via content) from "audit is stale" (DEGRADED via age).
+The legacy `is_fresh` predicates in `freshness.py` combined both; once
+chunk 7 retires that module these are the surviving content checks.
+"""
+
+from __future__ import annotations
+
+from datetime import date
+from typing import Any
+
+import psycopg
+
+
+def candles_content_ok(conn: psycopg.Connection[Any]) -> tuple[bool, str]:
+    """Every Tier 1/2 instrument must have a candle for the most recent trading day."""
+    from app.services.market_data import _most_recent_trading_day
+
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
Review comments resolved:

- **WARNING `_most_recent_trading_day` function-local import → REBUTTED.** Lifting this to module level creates a real circular-import cycle at package-init time: \`sync_orchestrator/__init__.py\` imports from \`registry\`, \`registry\` imports from \`content_predicates\`, and \`market_data\` imports \`app.services.sync_orchestrator.progress\` at module level ([market_data.py:19](app/services/market_data.py#L19)). Keeping the import function-scoped matches the pattern already in [freshness.py::candles_is_fresh](app/services/sync_orchestrator/freshness.py), where the same precedent handles the same cycle. A module-level import here would be an independently revertible change as soon as chunk 7 retires \`freshness.py\` and the orchestrator package is restructured, but that is chunk-7 scope, not here.
- **NITPICK \`monthly_reports\` uses \`timedelta(days=31)\`, drifts vs calendar months → DEFERRED (tech-debt issue to be filed).** Calendar-accurate monthly cadence needs \`relativedelta\` or a date-anchored helper; \`Cadence\` currently only holds a \`timedelta\`. Acceptable for now because \`monthly_reports\` is a non-blocking reporting layer and a 0-1 day drift at month boundaries has no operational impact. Filing ticket after this PR merges.
- **PREVENTION deferred-import lint rule → REBUTTED.** The rule would block a legitimate circular-break pattern we rely on ([freshness.py](app/services/sync_orchestrator/freshness.py) + now [content_predicates.py](app/services/sync_orchestrator/content_predicates.py)). A blanket grep-and-forbid misses the judgement call; the codebase already has per-file justifying comments at the seam.

Quarter-math note in the WARNING was self-rebutted by the reviewer — no action.

All gates still green after the \`is_tradable\` Codex fix already in this PR.

Luke-Bradford:
Filed tech-debt ticket #335 for the monthly cadence calendar-drift DEFERRED above.

PR #338, opened 2026-04-19
feat(#328): chunk 5 — GET /sync/layers/v2 endpoint

Luke-Bradford:
## What

Chunk 5 of sub-project **A** (freshness unification, #328). Exposes `compute_layer_states_from_db` via a new `GET /sync/layers/v2` endpoint with a structured triage payload (spec §8). Existing `GET /sync/layers` (v1) is byte-identical — pinned by a new guard test.

## Endpoint shape

```json
{
  "generated_at": "2026-04-19T…",
  "system_state": "ok" | "catching_up" | "needs_attention",
  "system_summary": "…",
  "action_needed": [{"root_layer", "display_name", "category" (Literal FailureCategory), "operator_message", "operator_fix", "self_heal", "consecutive_failures", "affected_downstream"}],
  "degraded": [{"layer", "display_name", "last_updated"}],
  "secret_missing": [{"layer", "display_name", "missing_secret", "operator_fix"}],
  "healthy": [{"layer", "display_name", "last_updated"}],
  "disabled": [{"layer", "display_name", "last_updated"}],
  "cascade_groups": [{"root", "affected": [layer, …]}]
}
```

- Auth: inherits `require_session_or_service_token` from the `/sync` router.
- `category` + `system_state` are typed `Literal` so the generated OpenAPI schema protects clients from enum drift.
- Inline `_downstream` cascade shim — chunk 6 replaces with `cascade.py::collapse_cascades` (tested separately).

## Review loops resolved

- **Codex pre-push**:
  - MEDIUM: `system_summary` previously returned `"All layers healthy"` when only RUNNING/RETRYING/CASCADE_WAITING layers existed (contradicting `system_state="catching_up"`). Added kwarg-only counts so the summary stays consistent. New regression test `test_v2_summary_never_contradicts_state`.
  - LOW: Typed `category` + `system_state` as `Literal` instead of `str`.

## Test isolation

`tests/api/conftest.py` introduced a `clean_client` fixture that pops stale `get_conn` overrides before each test. The pre-existing `tests/test_api_audit.py` installs a module-level `app.dependency_overrides[get_conn] = _fallback_conn` that otherwise contaminates any later real-DB test in the same pytest session. Fixture isolates v2 tests without disturbing the auth bypass.

## Test plan

- [x] `uv run pytest tests/api/test_sync_layers_v2_schema.py tests/api/test_sync_layers_v1_unchanged.py` — 7 passed (4 v2 schema + 2 v1 guard + 1 auth)
- [x] `uv run pytest -x -q` — 2095 passed, 1 skipped

### DIFF
--- a/app/api/sync.py
+++ b/app/api/sync.py
@@ -362,11 +362,7 @@ def get_sync_layers_v2(
         )
         if state is LayerState.ACTION_NEEDED:
             raw_cat = categories.get(name) or "internal_error"
-            category = (
-                FailureCategory(raw_cat)
-                if raw_cat in category_values
-                else FailureCategory.INTERNAL_ERROR
-            )
+            category = FailureCategory(raw_cat) if raw_cat in category_values else FailureCategory.INTERNAL_ERROR
             remedy = REMEDIES[category]
             action_needed.append(
                 ActionNeededItem(
@@ -440,9 +436,7 @@ def _system_summary(
     return "All layers healthy"
 
 
-def _layer_last_updated_map(
-    conn: psycopg.Connection[object], names: list[str]
-) -> dict[str, datetime]:
+def _layer_last_updated_map(conn: psycopg.Connection[Any], names: list[str]) -> dict[str, datetime]:
     """Most recent complete/partial finished_at per layer."""
     rows = conn.execute(
         """
--- a/app/api/sync.py
+++ b/app/api/sync.py
@@ -8,6 +8,8 @@ Auth: same dependency as /jobs — require_session_or_service_token.
 
 from __future__ import annotations
 
+import os
+from datetime import UTC, datetime
 from typing import Any, Literal
 
 import psycopg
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
Review comments resolved in \`0e66a15\`:

- **BLOCKING: double BFS** → **FIXED** — added \`_downstream_cache\` so \`affected_downstream\` on each \`action_needed\` item and \`cascade_groups[].affected\` share the same BFS result. New test \`test_v2_cascade_groups_match_action_needed_downstream\` pins the invariant.
- **BLOCKING: silent SECRET_MISSING drop** → **FIXED** — when \`os.environ\` has somehow been populated between state computation and the endpoint loop, fall back to \`layer.secret_refs[0]\` so the layer still appears in the \`secret_missing\` bucket. A layer that reaches SECRET_MISSING with no declared \`secret_refs\` (shouldn't happen per state machine) now emits a generic \`missing_secret="(unknown)"\` row rather than vanishing. New test \`test_v2_secret_missing_never_silently_dropped\`.
- **WARNING: hardcoded count** → **FIXED** — v1 shape guard now uses \`len(LAYERS)\`.

PR #349, opened 2026-04-19
docs + test(#257): refocus spec + CIK upsert regression

### LINKED ISSUE #257: bug: daily_cik_refresh crashes on UniqueViolation — upsert_cik_mapping uses INSERT not UPSERT
## Problem

`daily_cik_refresh` crashes with `psycopg.errors.UniqueViolation` when a CIK mapping already exists in `external_identifiers`:

```
psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "uq_external_identifiers_primary"
DETAIL:  Key (instrument_id, provider, identifier_type)=(1050630, sec, cik) already exists.
```

## Root cause

`upsert_cik_mapping` in `app/services/filings.py` (line ~120) uses a plain `INSERT` instead of `INSERT ... ON CONFLICT DO UPDATE`. On the first run it succeeds; on every subsequent run it crashes for any instrument that already has a CIK mapping.

## Impact

- CIK refresh fails on every run after the initial population
- New CIK mappings from SEC are never picked up after the first successful run
- Error is logged but the job continues to fail at its next cadence

## Suggested fix

Change the `INSERT INTO external_identifiers` in `upsert_cik_mapping` to use `ON CONFLICT (instrument_id, provider, identifier_type) DO UPDATE SET value = EXCLUDED.value, updated_at = NOW()`.

## Scope

- `app/services/filings.py`: `upsert_cik_mapping` (~line 120)
- Test: verify idempotent insert of same CIK mapping

Luke-Bradford:
## Summary
- Adds `docs/superpowers/specs/2026-04-19-research-tool-refocus.md` — drop autonomous-fund ambition, refocus to per-ticker research page, define Phase 1 job collapse + Phase 2 instrument page + beyond.
- Adds `tests/test_upsert_cik_mapping.py` — 6 regression cases pinning the #267 partial-UNIQUE fix so a future refactor cannot reintroduce the crash behind #257. Codex flagged + addressed the combined-conflict case.

Closes #257.
Follow-ups from this refocus (Chunk 2+): `fundamentals_sync` job collapse, orchestrator DAG rewire, physical file merge, per-ticker research page.

## Test plan
- `uv run pytest tests/test_upsert_cik_mapping.py` — 6 passed
- `uv run pytest` — 2143 passed, 1 skipped
- `uv run ruff check .` — clean
- `uv run ruff format --check .` — clean

### DIFF
--- a/docs/superpowers/specs/2026-04-19-research-tool-refocus.md
+++ b/docs/superpowers/specs/2026-04-19-research-tool-refocus.md
@@ -1,0 +1,229 @@
+# eBull refocus — research tool, not autonomous fund
+
+**Date:** 2026-04-19
+**Author intent:** stop polishing ETL / admin; ship a per-ticker research page I actually open on a Sunday. AI assists research, human decides, eToro executes.
+
+## Why this exists
+
+Two weeks of work produced: a typed state machine, execution guard safety rules, v2 API, Admin UI rewrite, cascade collapse, 11 PRs, 127 new backend tests. None of it is visible on the front page beyond a slightly better Admin page. No ticker lookup, no financials view, no real research surface.
+
+Reframe: stop building "AI auto-trades, you watch." Build "Bloomberg-for-me." AI summarises filings + computes factor scores + generates thesis on-demand. Human decides. eToro executes via the already-built guard + order path.
+
+Realistic ambition: 1-3% annual edge vs passive, from better-informed retail decisions. Not generational wealth, not early retirement. A finite tool that's useful to open weekly. Secondary: skill + artifact that isn't day-job code.
+
+## What gets dropped
+
+- Sub-project A.6 (auto-retry on self-heal failures) — needed only if the app autonomously trades. It doesn't. Close the umbrella idea; file real issues only if a specific layer's failure pattern becomes annoying.
+- Sub-project B (layer metrics history) — no UI consumer.
+- Sub-project C (full three-zone Admin redesign) — Admin becomes a hidden `/ops` route, not the main UX.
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
Review responses (commit 503b6c2):

**[WARNING] `_seed_instrument` bare commit**: REBUTTED — the `ebull_test_conn` fixture (tests/fixtures/ebull_test_db.py:204) is TRUNCATE-based, not transaction-based. It opens a plain connection, TRUNCATEs planner tables before yielding, and TRUNCATEs again after. `conn.commit()` inside `_seed_instrument` is required (psycopg3 implicit transactions otherwise block follow-up reads) and cannot cause cross-test pollution because the fixture's teardown TRUNCATE runs regardless.

**[WARNING] CIK reassignment assertion gap**: FIXED 503b6c2 — strengthened to `_all_rows(conn, 1) == []` so the test explicitly pins the in-place ON CONFLICT row rewrite from instrument 1 to 2 (the unique row on `(provider, identifier_type, identifier_value) = (sec, cik, '0000555555')` moves in place).

**[NITPICK] `__all__` re-export**: REBUTTED — established convention across the test suite (e.g. tests/test_sec_incremental_executor.py:26) to silence ruff F401 on the fixture import. Removing it would trip lint.

**[PREVENTION] `_seed_instrument` bare commit rule**: REBUTTED — same reason as the W1 rebuttal; the fixture's isolation model is TRUNCATE-based, not transactional.

**[PREVENTION] Reassignment assertion gap**: EXTRACTED — the stronger assertion lives in the test itself (503b6c2); no separate skill/doc change needed since the signal lives in the regression suite.

PR #356, opened 2026-04-20
feat: yfinance provider (Phase 2.1)

Luke-Bradford:
## Summary
Phase 2.1 of the [2026-04-19 research-tool refocus](docs/superpowers/specs/2026-04-19-research-tool-refocus.md).

New \`app/providers/implementations/yfinance_provider.py\` for non-US tickers (LSE, Euronext, HK, ASX) where SEC XBRL doesn't apply, plus gap-fill for US current-price / analyst / major-holder data.

**Methods** (all catch exceptions and return \`None\`, so the research UI stays interactive when Yahoo breaks):
- \`get_profile\` — company identity + headline fundamentals
- \`get_quote\` — last price + 52w range + day change
- \`get_key_stats\` — PE, PB, dividend yield, payout, ROE, ROA, D/E, growth
- \`get_financials\` — income / balance / cashflow, quarterly or annual
- \`get_dividends\` — historical dividends per share
- \`get_analyst_estimates\` — consensus targets + recommendation mean
- \`get_major_holders\` — insider + institutional pct
- \`get_price_history\` — OHLCV bars, sorted oldest-first

Return types are frozen dataclasses with \`Decimal\` for monetary fields.

**Codex pre-push feedback addressed:**
- Added \`get_price_history\` — spec §2.1 explicitly requires it for the Overview tab chart.
- \`_to_str\` now filters float/string \`NaN\` so missing Yahoo string fields render as \`None\` instead of the literal \`"nan"\`.
- \`get_financials\` validates \`period\` explicitly — unknown values return \`None\` instead of silently routing to annual.

## Test plan
- \`tests/test_yfinance_provider.py\` — 26 stub-based cases. No live Yahoo calls.
- \`uv run pytest\` — 2169 passed, 1 skipped
- \`uv run ruff check .\` / \`ruff format --check .\` — clean

### DIFF
--- a/app/providers/implementations/yfinance_provider.py
+++ b/app/providers/implementations/yfinance_provider.py
@@ -338,7 +338,7 @@ class YFinanceProvider:
         for col in frame.columns:
             # yfinance columns are pandas Timestamp; .date() extracts date.
             try:
-                period_end = col.date()
+                period_end = col.date()  # type: ignore[union-attr]
             except AttributeError:
                 # Already a date, or malformed — skip.
                 continue
@@ -374,7 +374,7 @@ class YFinanceProvider:
         results: list[YFinanceDividend] = []
         for ts, amount_raw in series.items():
             try:
-                ex_date = ts.date()
+                ex_date = ts.date()  # type: ignore[union-attr]
             except AttributeError:
                 continue
             decimal_amount = _to_decimal(amount_raw)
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
Review response (commit 176275a):

**[WARNING] Shared try discards valid frame**: FIXED 176275a — split `ticker.info` fetch into its own try/except in `get_financials`. `.info` failure now falls back to `currency=None` rather than discarding the successfully-fetched statement frame. Regression locked in `test_financials_info_failure_does_not_discard_frame`.

**[PREVENTION] Shared try grep rule**: EXTRACTED via the regression test — a future refactor that re-merges the try blocks will fail the test rather than pass silently.

Issue #357, opened 2026-04-20
tech-debt: /instruments/{symbol}/summary should prefer local SEC XBRL over yfinance for US tickers


Luke-Bradford:
## Context

PR #357 (Phase 2.2) ships /instruments/{symbol}/summary as a yfinance-only MVP. The spec (docs/superpowers/specs/2026-04-19-research-tool-refocus.md §2.2) specifies a per-field provider priority:

- US tickers: SEC XBRL (already in DB via \`financial_periods\`) + current price provider + yfinance for gaps
- Non-US: yfinance primary + current price provider

The MVP uses yfinance for every non-identity field and reports a section-level \`source\` map. Codex review flagged the gap as P1.

## Acceptance

Update the summary endpoint to:

1. For US tickers with local \`financial_periods\` data, derive PE/PB/D-E/ROE/ROA/revenue_growth from the latest canonical row (divided by current price when PE is needed). Fall back to yfinance values only for fields the local row doesn't cover.
2. Use the local \`quotes\` table for current price when fresh (< 15 min). Fall back to yfinance.
3. Expand \`source\` from section-level (\`identity\`, \`price\`, \`key_stats\`) to per-field, matching spec attribution language.
4. Tests cover: US ticker with local fundamentals preferred, US ticker with missing fundamentals falling back to yfinance, non-US ticker yfinance primary.

Blocked on: nothing — can implement next. Out of scope for Phase 2.2 merge because it materially grows the PR surface; landing the MVP first unblocks frontend work.

PR #358, opened 2026-04-20
feat: GET /instruments/{symbol}/summary (Phase 2.2)

Luke-Bradford:
## Summary
Phase 2.2 of the [2026-04-19 research-tool refocus](docs/superpowers/specs/2026-04-19-research-tool-refocus.md).

\`GET /instruments/{symbol}/summary\` returns per-ticker research data: identity + price + key stats.

- **Identity** (local DB authoritative, yfinance fills gaps): symbol, display_name, sector, industry, exchange, country, currency, market_cap.
- **Price**: current, day_change, day_change_pct, 52w_high, 52w_low, currency.
- **Key stats**: PE, PB, dividend yield, payout, ROE, ROA, D/E, revenue growth, earnings growth.
- **Source**: section-level provider attribution map.

404 if symbol not in local instruments table (yfinance not queried). Case-insensitive symbol match. yfinance failures return null sections, not 500.

## Codex pre-push feedback addressed
- P2 field names match spec (\`current\`, not \`price\`).
- P2 display_name precedence: local DB \`company_name\` wins over yfinance.
- P2 three \`.info\` calls consolidated into one via new \`YFinanceProvider.get_snapshot\`.
- P1 deferred as tech-debt #357: per-field source attribution + local SEC XBRL preference for US tickers. Landing MVP first unblocks Chunk 9 frontend.

## Test plan
- \`tests/test_api_instrument_summary.py\` — 5 cases (unknown symbol 404, happy-path merge, yfinance failure null sections, local company_name precedence, empty-symbol 400)
- \`tests/test_yfinance_provider.py\` — +3 snapshot cases (single .info fetch invariant, .info raise, empty .info)
- \`uv run pytest\` — 2178 passed, 1 skipped
- \`uv run ruff check .\` / \`ruff format --check .\` / \`pyright\` — clean

### DIFF
--- a/app/api/instruments.py
+++ b/app/api/instruments.py
@@ -12,6 +12,7 @@ No writes. No schema changes.
 from __future__ import annotations
 
 from datetime import datetime
+from decimal import Decimal
 
 import psycopg
 import psycopg.rows
@@ -19,8 +20,19 @@ from fastapi import APIRouter, Depends, HTTPException, Query
 from pydantic import BaseModel
 
 from app.db import get_conn
+from app.providers.implementations.yfinance_provider import YFinanceProvider
 
 router = APIRouter(prefix="/instruments", tags=["instruments"])
+
+
+def get_yfinance_provider() -> YFinanceProvider:
+    """FastAPI dependency: constructs a fresh YFinanceProvider per request.
+
+    The provider is stateless, so there is no pooling concern. Tests
+    override this via ``app.dependency_overrides`` to inject a stub.
+    """
+    return YFinanceProvider()
+
 
 # ---------------------------------------------------------------------------
 # Response models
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
Review responses (commit fe7a485):

**[WARNING] Snapshot shells instead of null sections**: FIXED fe7a485 — `get_snapshot` now collapses `quote` / `key_stats` to `None` when all their relevant fields are `None` (e.g. `.info` returned only identity keys, or all NaN). Contract is now consistent with the API-level `source` map reporting 'unavailable'. Two new regression cases: identity-only info + all-NaN price info.

**[WARNING] Numeric symbol routing**: REBUTTED + test added. FastAPI path matching distinguishes `/{instrument_id}` (pure int segment, no suffix) from `/{symbol}/summary` (string segment with literal suffix), so `/instruments/7203/summary` reaches the summary route and returns InstrumentSummary, NOT InstrumentDetail. New `test_summary_numeric_symbol_routes_to_summary_not_detail` locks this in — asserts the response body has `identity` (summary shape) and not `external_identifiers` (detail shape).

**[NITPICK] `# type: ignore` noise on `row[...]`**: DEFERRED — existing convention in the same file for every `dict_row` access site. Changing the ignore pattern is outside this PR's scope; a codebase-wide cleanup is better handled on its own.

Luke-Bradford:
Review responses (commit 0608d80):

**[BLOCKING] Falsy-zero short-circuit**: FIXED 0608d80 — replaced `info.get('regularMarketPrice') or info.get('currentPrice')` with an explicit `None`-check. A legitimate `regularMarketPrice=0` (halted/distressed ticker) now survives instead of falling through. Pinned by `test_get_snapshot_preserves_zero_price`.

**[BLOCKING] NaN Decimal guard**: REBUTTED — the existing `_to_decimal` helper catches NaN explicitly: `if value.is_nan(): return None` (app/providers/implementations/yfinance_provider.py:169). Added `test_to_decimal_nan_returns_none` asserting both `_to_decimal(float('nan'))` and `_to_decimal(Decimal('NaN'))` return `None`; the existing `test_get_snapshot_info_with_all_null_price_fields_returns_null_quote` passes under the current code because of this filter. The review claim that the guard 'does not fire' was incorrect — verified by running the test.

**[WARNING] `cur_mock.__exit__` not configured**: REBUTTED — `MagicMock` auto-creates `__exit__` returning a `MagicMock` (truthy). The bot is correct that truthy return suppresses exceptions, but in these tests the handler is not expected to raise inside the `with conn.cursor(...) as cur:` block (raises happen via HTTPException from the handler, not from psycopg). Any real bug inside cursor would surface as an exception BEFORE the `with` exits. Not a masking vector for these tests.

Luke-Bradford:
Review responses (rebuttal-only round, Codex-confirmed):

**[BLOCKING] NaN guard correctness**: REBUTTED — Codex confirmed `_to_decimal` filters NaN to `None` via `value.is_nan()` check at yfinance_provider.py:172-173. Direct test at tests/test_yfinance_provider.py:577-583 pins both `float('nan')` and `Decimal('NaN')`. The bot's claim that 'the guard does not fire' is incorrect — the passing test is not a false positive.

**[BLOCKING] Partial-zero quote nullability**: REBUTTED — the current guard (`all three None → null quote, otherwise construct`) is intentional. A 52-week range with no live price is legitimately useful UI data: operator sees the context band even when live quote is missing. `YFinanceQuote.price` is explicitly `Decimal | None` at yfinance_provider.py:64. The bot's 'partial-zero edge case' is by design.

**[WARNING] `cur_mock.__exit__` masking**: REBUTTED — Codex confirmed the `with conn.cursor(...) as cur` block at app/api/instruments.py:298-300 only wraps `execute`/`fetchone`. The `HTTPException(404)` is raised AFTER the block at :302-303, propagating through FastAPI independently. The auto-mock's truthy `__exit__` cannot mask a 404 that isn't raised from inside the block.

Ready to merge.

PR #359, opened 2026-04-20
feat: GET /instruments/{symbol}/financials (Phase 2.3)

Luke-Bradford:
## Summary
Phase 2.3 of the [2026-04-19 research-tool refocus](docs/superpowers/specs/2026-04-19-research-tool-refocus.md).

\`GET /instruments/{symbol}/financials?statement=income|balance|cashflow&period=quarterly|annual\`

**Pull priority** (matches spec):
1. Local \`financial_periods\` table (SEC XBRL-sourced) if populated.
2. yfinance fallback for non-US or US instruments without XBRL history.

**Response shape**:
- \`symbol\`, \`statement\`, \`period\`, \`currency\`, \`source\`, \`rows\`
- \`rows\`: \`[{period_end, period_type, values: {column: Decimal | None}}]\`

**Behavior**:
- 404 if symbol not in local \`instruments\` table.
- 400 if symbol is whitespace-only.
- 422 on invalid \`statement\` / \`period\` (FastAPI Literal validation).
- 200 with \`rows=[]\` if neither source has data (UI shows "no data available").

Column sets per statement are hardcoded whitelists composed into SQL directly (parameterised on instrument_id only). \`period_type\` filter uses \`IN ('Q1','Q2','Q3','Q4')\` for quarterly, \`= 'FY'\` for annual. Up to 20 most-recent rows.

## Test plan
- \`tests/test_api_instrument_financials.py\` — 7 cases (unknown symbol 404, empty symbol 400, invalid statement 422, local-data-wins, yfinance fallback, no-data-anywhere empty rows, balance-statement columns)
- \`uv run pytest\` — 2190 passed, 1 skipped
- \`uv run ruff check .\` / \`ruff format --check .\` / \`pyright\` — clean

### DIFF
--- a/app/api/instruments.py
+++ b/app/api/instruments.py
@@ -352,25 +352,24 @@ def _fetch_local_financials(
     Returns (rows, currency). Empty rows = no local data, let caller fall
     back to yfinance.
     """
-    if period == "quarterly":
-        type_clause = "period_type IN ('Q1','Q2','Q3','Q4')"
-    else:
-        type_clause = "period_type = 'FY'"
+    period_types: list[str] = ["Q1", "Q2", "Q3", "Q4"] if period == "quarterly" else ["FY"]
 
-    # Columns whitelisted above — safe to format into SQL.
+    # Columns whitelisted above — safe to format into SQL. period_types
+    # is bound as a parameter, not formatted, so a future value added to
+    # the CHECK constraint won't silently match.
     select_cols = ", ".join(columns)
     sql = f"""
         SELECT period_end_date, period_type, reported_currency, {select_cols}
         FROM financial_periods
         WHERE instrument_id = %(iid)s
           AND superseded_at IS NULL
-          AND {type_clause}
+          AND period_type = ANY(%(types)s::text[])
         ORDER BY period_end_date DESC
         LIMIT 20
-    """  # noqa: S608 — columns and type_clause are hardcoded whitelists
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
Review responses (commit 99ed023):

**[WARNING] yfinance period_type synthesis**: FIXED 99ed023 — now inferred from fiscal-quarter-end month (Mar→Q1, Jun→Q2, Sep→Q3, Dec→Q4) for quarterly, 'FY' for annual. Matches the local-path labels so the frontend treats both sources uniformly. Test updated to pin period_type='Q1' for March period_end.

**[WARNING] Hardcoded IN-list in SQL**: FIXED 99ed023 — replaced `period_type IN ('Q1','Q2','Q3','Q4')` with `period_type = ANY(%(types)s::text[])` and bound the list as a parameter. Future CHECK-constraint values (e.g. 'Q1-restated') won't silently match.

**[NITPICK] assert statement in _STATEMENT_COLUMNS**: DEFERRED — FastAPI Literal validation already rejects non-whitelisted values with 422 before the handler runs (test_financials_invalid_statement_returns_422 pins this). An extra assert adds noise without new safety.

**[NITPICK] Shared cur_mock**: DEFERRED — handler only opens one cursor scope at a time (the second cursor scope happens after the first returns), so the shared mock returns correct data in call order. A side-effect queue would be cleaner but not behavior-changing.

PR #365, opened 2026-04-20
feat: Reports frontend page

Luke-Bradford:
## Summary
Surfaces \`/api/reports/weekly\` + \`/monthly\` snapshot data that's been generated by the weekly_report + monthly_report jobs but had no consumer. Unblocks the Phase 1.4 retirement decision — if this page sees no use after two cycles, the report jobs can be retired cleanly.

- New route \`/reports\` + sidebar nav entry
- Two tabs (Weekly, Monthly)
- List-on-left / detail-on-right layout
- Structured render for: PnL, trade stats (win rate, avg holding), top/bottom performers, best/worst trade, thesis accuracy, upcoming earnings
- Raw JSON fallback below for any field the template doesn't recognise yet

## Test plan
- \`pnpm typecheck\` — clean
- \`pnpm test\` — 287/287 passed


Luke-Bradford:
Review responses (commit acd2ef8):

**[WARNING] Stale-selection silent fallback**: FIXED acd2ef8 — added a useEffect that resets `selectedId` to the first snapshot whenever `reports` changes and the previous selection is no longer present. Highlighted button and detail pane now stay in sync.

**[WARNING] Empty-string date**: FIXED acd2ef8 — `formatDate` now accepts `string | null | undefined`, returns `—` on empty / malformed input, and validates the ISO date-prefix via regex.

**[WARNING] Dead fetchLatestReport export**: FIXED acd2ef8 — removed from `frontend/src/api/reports.ts`.

**[NITPICK] array-index keys on performer lists**: DEFERRED — both lists are static per report and the entire detail pane re-renders when the report changes, so stable keys aren't load-bearing. Can cleanup in a later pass alongside the other micro-polish.

PR #372, opened 2026-04-20
fix(#355): drop KeyError from backfill_filings parse-error handlers

### LINKED ISSUE #355: tech-debt: backfill_filings catches KeyError in parse-error handlers
## Context

Review WARNING on PR #354 (Chunk 4 file merge): the \`except (json.JSONDecodeError, TypeError, KeyError)\` handlers in \`backfill_filings\` / \`fetch_submissions_page\` inside \`app/services/coverage.py\` catch \`KeyError\` as a "parse error" / retryable \`STILL_INSUFFICIENT_PARSE_ERROR\`, but \`KeyError\` is not raised by \`resp.json()\` or \`json.loads\` — the only way a \`KeyError\` reaches those handlers is via a dict-access bug in the provider code being guarded. Misclassifying such a bug as retryable hides programming errors behind backoff retries.

This is pre-existing behavior inherited verbatim from the retired \`filings_backfill.py\` (no behavior drift from Chunk 4).

## Acceptance

Remove \`KeyError\` from the parse-error handlers in \`backfill_filings\`; let genuine \`KeyError\`s propagate so operators see the traceback instead of a silent retry. Update the affected test cases (if any assert on \`KeyError\` being caught — none at last grep).

Out of scope for Chunk 4 (file merge only); file here so it isn't forgotten.

Luke-Bradford:
## What
Remove `KeyError` from two `except (json.JSONDecodeError, TypeError, KeyError):` clauses in [app/services/coverage.py](app/services/coverage.py) around `fetch_submissions` + `fetch_submissions_page`.

## Why
`resp.json()` never raises `KeyError` — a KeyError reaching those handlers means a dict-access bug inside the provider, not a malformed payload. Misclassifying it as retryable hides programming errors behind exponential backoff.

Verified via Codex: `httpx.Response.json()` calls `json.loads()` which can only raise `json.JSONDecodeError` / `UnicodeDecodeError`. The SEC provider's two entrypoints do no dict-key access after decode, so any KeyError there is genuinely a provider bug.

Kept KeyError on the inner `_normalise_submissions_block(...)` handlers — those DO touch dict keys and a KeyError there is a legitimate malformed-payload signal (test 15 unchanged).

## Test plan
- [x] New `test_15b_keyerror_from_provider_propagates` asserts KeyError from `fetch_submissions` propagates
- [x] Existing `test_14_parse_error_json_decode` + `test_15_parse_error_missing_recent` still pass
- [x] 23/23 tests in `test_filings_backfill.py` green
- [x] Codex pre-push review: no findings

Closes #355.

### DIFF
--- a/app/services/coverage.py
+++ b/app/services/coverage.py
@@ -1588,7 +1588,13 @@ def backfill_filings(
             outcome=BackfillOutcome.STILL_INSUFFICIENT_HTTP_ERROR,
             status=None,
         )
-    except json.JSONDecodeError, TypeError, KeyError:
+    except json.JSONDecodeError, TypeError:
+        # `fetch_submissions` only decodes JSON (`resp.json()`) and returns
+        # the dict — it does no dict-key access here. `KeyError` would mean
+        # a dict-access bug in the provider code that raised during decode;
+        # misclassifying that as a retryable PARSE error would hide it behind
+        # backoff. Let any `KeyError` propagate so operators see the
+        # traceback instead of silent retry (#355).
         logger.warning(
             "backfill_filings: PARSE error on fetch_submissions cik=%s",
             cik_padded,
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
**[WARNING] `submissions_raises` field not shown in diff** — REBUTTED.

`_FakeResponse` has declared `submissions_raises: BaseException | None = None` at [tests/test_filings_backfill.py:51](tests/test_filings_backfill.py#L51) since the fixture's creation. `FakeSecProvider.fetch_submissions` at [tests/test_filings_backfill.py:70-72](tests/test_filings_backfill.py#L70-L72) explicitly raises it:

```python
def fetch_submissions(self, cik: str) -> dict[str, object] | None:
    self.fetch_submissions_calls.append(cik)
    if self._resp.submissions_raises is not None:
        raise self._resp.submissions_raises
    return self._resp.primary
```

So `submissions_raises=KeyError("filings")` is correctly handled — test 14 (pre-existing) exercises the same field with `json.JSONDecodeError`. The diff didn't include the helper because it's unchanged; grep-verifying the fixture would resolve it.

Production change unaffected. No action required.

PR #373, opened 2026-04-20
refactor(#320): shared ConfigProvider + useConfig() context

### LINKED ISSUE #320: tech-debt: shared useConfig() context to deduplicate /config fetches
Surfaced in review of #319 (PR for #313).

## Current state

`/config` is fetched independently by every consumer via `useAsync(() => fetchConfig(), [])`:
- `frontend/src/lib/DisplayCurrencyContext.tsx:18`
- `frontend/src/components/dashboard/SystemStatusPanel.tsx` (via DashboardPage)
- `frontend/src/pages/DashboardPage.tsx:38`
- `frontend/src/components/orders/DemoLivePill.tsx:22` (new, from #319)
- and others

Each consumer has its own in-flight cache, retry timer, and error state. They do not coordinate.

## Why this is a problem

- Two leaf components rendered simultaneously (e.g. a future page that opens two modals side-by-side) fire duplicate `/config` requests.
- Each consumer's loading / error state diverges from every other consumer's — no single source of truth for "is the config loaded yet?".
- The DemoLivePill safety-state cache is per-instance, so a kill-switch banner on page A and a DemoLive pill on page B can fall out of sync.

## Proposed fix

- Introduce a `ConfigProvider` + `useConfig()` hook at the `AppShell` / root level.
- Add a CI grep rule: `fetchConfig` must only be called from the provider file; consumers use `useConfig()`.
- Migrate existing consumers in the same PR to prevent regression.

## Out of scope for the original #319 PR

`DemoLivePill` follows the same pattern every other `/config` consumer uses today; fixing it in isolation would introduce inconsistency with the rest of the codebase. A full migration belongs in its own PR.

## Trigger to prioritise

- Any page that needs two `/config` consumers simultaneously (e.g. #314 workstation with the right-side detail panel).
- Any new safety-state indicator (live-trading pill, kill-switch banner, drawdown warning) — they all depend on `/config` today.

Luke-Bradford:
## What
Single `ConfigProvider` at AppShell level; consumers use `useConfig()` hook.

## Why
Every `/config` consumer previously ran its own `useAsync(fetchConfig)` — duplicate requests, uncoordinated loading/error state, safety-state pill on one page silently diverging from kill-switch banner on another.

## Changes
- [frontend/src/lib/ConfigContext.tsx](frontend/src/lib/ConfigContext.tsx) — new. `ConfigProvider`, `useConfig()`, `TestConfigProvider` (for tests).
- [App.tsx](frontend/src/App.tsx) — wrap `DisplayCurrencyProvider` in `ConfigProvider`.
- [DisplayCurrencyContext.tsx](frontend/src/lib/DisplayCurrencyContext.tsx), [DashboardPage.tsx](frontend/src/pages/DashboardPage.tsx), [DemoLivePill.tsx](frontend/src/components/orders/DemoLivePill.tsx) — migrate.
- Tests wrapped in `TestConfigProvider`.

## Test plan
- [x] 287/287 frontend tests pass
- [x] typecheck clean
- [x] Codex pre-push review clean
- [x] Verified `fetchConfig` only called inside `ConfigContext.tsx` for runtime code

Closes #320.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -2,6 +2,7 @@ import { Route, Routes } from "react-router-dom";
 import { AppShell } from "@/layout/AppShell";
 import { ErrorBoundary } from "@/components/states/ErrorBoundary";
 import { RequireAuth } from "@/components/RequireAuth";
+import { ConfigProvider } from "@/lib/ConfigContext";
 import { DisplayCurrencyProvider } from "@/lib/DisplayCurrencyContext";
 import { DashboardPage } from "@/pages/DashboardPage";
 import { RankingsPage } from "@/pages/RankingsPage";
@@ -32,9 +33,11 @@ export function App() {
         <Route
           element={
             <RequireAuth>
-              <DisplayCurrencyProvider>
-                <AppShell />
-              </DisplayCurrencyProvider>
+              <ConfigProvider>
+                <DisplayCurrencyProvider>
+                  <AppShell />
+                </DisplayCurrencyProvider>
+              </ConfigProvider>
             </RequireAuth>
           }
         >
--- a/frontend/src/components/orders/ClosePositionModal.test.tsx
+++ b/frontend/src/components/orders/ClosePositionModal.test.tsx
@@ -17,6 +17,7 @@ import userEvent from "@testing-library/user-event";
 
 import { ApiError } from "@/api/client";
 import { ClosePositionModal } from "@/components/orders/ClosePositionModal";
+import { TestConfigProvider } from "@/lib/ConfigContext";
 import type {
   ConfigResponse,
   InstrumentPositionDetail,
@@ -24,9 +25,6 @@ import type {
   OrderResponse,
 } from "@/api/types";
 
-vi.mock("@/api/config", () => ({
-  fetchConfig: vi.fn(),
-}));
 vi.mock("@/api/portfolio", () => ({
   fetchInstrumentPositions: vi.fn(),
 }));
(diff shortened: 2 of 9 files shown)


Luke-Bradford:
Bot APPROVED on first review. Residual items:

**[WARNING] `useMemo` depends on `state.refetch` — reviewer speculates it's a new fn every render** — REBUTTED.

`useAsync` at [frontend/src/lib/useAsync.ts:69-71](frontend/src/lib/useAsync.ts#L69-L71) already wraps `refetch` in `useCallback(fn, [])` with empty deps, so it IS stable across renders:

```tsx
const refetch = useCallback(() => {
  setTick((t) => t + 1);
}, []);
```

The memoisation does reduce identity churn: `state.data` / `state.error` / `state.loading` are the only values that actually change across renders, so consumers pinning one of those get stable context identity otherwise.

**[NITPICK] `TestConfigProvider` bundled in production** — REBUTTED.

`TestConfigProvider` is only imported by `*.test.tsx` files. Vite + Rollup tree-shake unused exports at build time — unused ESM exports with no side effects are dropped from the production bundle. Co-locating the test-only export next to the real one mirrors the pattern already used elsewhere in this codebase (e.g. `@/api/mocks` exports live alongside real clients).

**[PREVENTION] Add test for refetch stability** — REBUTTED. The stability is a contract of `useAsync` (useCallback + empty deps), tested transitively by the existing `useAsync.test.ts`. Duplicating that assertion against the proxy in `ConfigContext` would only catch a regression in `useAsync` itself, which is redundant.

Ready to merge.

PR #377, opened 2026-04-20
feat: Slice 0 — primary-listing flag + deterministic symbol lookup

Luke-Bradford:
## What
Slice 0 of per-stock research page spec. Adds deterministic tiebreaker for symbol→instrument_id resolution and a `?id=` override on the summary endpoint.

## Why
`instruments.symbol` is indexed but not UNIQUE. Symbol collisions across exchanges exist (e.g. `VOD`/`VOD.L`). Current lookups in `app/api/instruments.py` use `LIMIT 1` without ORDER BY — nondeterministic on collisions.

## Changes
- **Migration 043**: `is_primary_listing BOOLEAN DEFAULT TRUE`; backfill demotes duplicates (lowest `instrument_id` per `UPPER(symbol)` stays primary). Expression index on `(UPPER(symbol), is_primary_listing DESC, instrument_id)` to support the tiebreaker seek.
- **Symbol lookups** in [app/api/instruments.py](app/api/instruments.py) for financials + summary now `ORDER BY is_primary_listing DESC, instrument_id ASC LIMIT 1`.
- **`?id=` override** on summary endpoint: caller pins specific instrument_id when ambiguous; server verifies id's symbol matches path (404 on mismatch).

## Not doing
- **No partial-unique index** on `UPPER(symbol) WHERE is_primary_listing`: `sync_universe()` has no collision-aware assignment for `is_primary_listing`, so a unique constraint would abort universe sync on the first duplicate-symbol ingest. Application-level tiebreaker is sufficient. Filed as follow-up tech-debt.

## Test plan
- [x] `test_summary_id_override_pins_specific_instrument`
- [x] `test_summary_id_override_mismatched_symbol_returns_404`
- [x] All 11 existing summary tests + 7 financials + 28 instruments still pass
- [x] Full `uv run pytest -q`: 2207 passed
- [x] ruff / format / pyright clean
- [x] Codex pre-push review: 3 rounds, final clean

Refs: [per-stock research spec §5 Slice 0](docs/superpowers/specs/2026-04-20-per-stock-research-page.md)

### DIFF
--- a/app/api/instruments.py
+++ b/app/api/instruments.py
@@ -425,10 +425,18 @@ def get_instrument_financials(
 
     columns = _STATEMENT_COLUMNS[statement]
 
-    # Resolve symbol -> instrument_id for the local read.
+    # Resolve symbol -> instrument_id for the local read. `symbol` is
+    # not UNIQUE across exchanges (see migration 043), so order by
+    # `is_primary_listing DESC, instrument_id ASC` to make the winner
+    # deterministic on collisions.
     with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
         cur.execute(
-            "SELECT instrument_id, symbol FROM instruments WHERE UPPER(symbol) = %(s)s LIMIT 1",
+            """
+            SELECT instrument_id, symbol FROM instruments
+            WHERE UPPER(symbol) = %(s)s
+            ORDER BY is_primary_listing DESC, instrument_id ASC
+            LIMIT 1
+            """,
             {"s": symbol_clean},
         )
         inst_row = cur.fetchone()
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
Round-1 resolved on e3e3fc1:

**[WARNING] \`id\` shadows Python builtin** — FIXED. Renamed param to \`instrument_id\` with \`Query(alias=\"id\")\` so wire name stays \`?id=\` but the Python symbol no longer shadows.

**[WARNING] Migration re-run hazard** — FIXED. Added \`AND i.is_primary_listing = TRUE\` guard to the UPDATE + \`AND j.is_primary_listing = TRUE\` to the EXISTS. Traced through:
- Fresh run: every row starts TRUE; demotion works correctly (only lowest-id stays primary).
- Re-run after operator flips A→FALSE, B→TRUE: row B's EXISTS check finds no lower-id TRUE row (A is FALSE now), so B stays TRUE. Operator's correction preserved.

**[NITPICK] Redundant \`LIMIT 1\` in id-override branch** — FIXED. PK lookup is already unique; dropped the \`LIMIT 1\`.

**[WARNING] Readability / comment→CREATE INDEX separation** — rebutted as cosmetic. Kept as-is; the preceding blank lines already separate the rationale block from the \`CREATE INDEX\` statement.

**[PREVENTION] Add flake8-builtins A002 rule** — DEFERRED. Separate tech-debt ticket; affects repo-wide lint config.

Gates on e3e3fc1: ruff / format / pyright clean; 11/11 summary tests pass.

PR #381, opened 2026-04-20
feat: Slice 4 — Reports per-contributor drill-in

Luke-Bradford:
## What
Each weekly + monthly snapshot gains `positions` (per-instrument P&L at snapshot time) and `period_contribution` (top contributors + top drags vs the prior snapshot). ReportsPage renders both lists with `/instrument/:symbol` links.

## Why
Current `report_snapshots` stored aggregate P&L only — no way to see which stocks drove the delta. Operator clicking through the Reports page lands in raw JSON with no onward navigation. Slice 4 makes Reports a gateway into per-stock research.

## Graceful degradation
- First snapshot ever: no prior → empty contributors/drags.
- Pre-feature snapshots (no `positions` key): caller passes `None` (not `[]`) so every current holding isn't wrongly labelled as a new contributor. Codex caught this in round 1.

## Test plan
- [x] 6 new unit tests on `_compute_contributors` (null prior, legacy prior, deltas, new-position null pct, zero-delta omit, top_n cap)
- [x] Existing weekly + monthly structure tests still pass
- [x] Full pytest: 2212 passed
- [x] Frontend typecheck green
- [x] Codex pre-push review 2 rounds, final clean

## Not in scope
- Chart visualisation of contributor deltas — future polish
- Admin coverage migration + delete orphans — Slice 5

Refs: [per-stock research spec §5 Slice 4](docs/superpowers/specs/2026-04-20-per-stock-research-page.md)

### DIFF
--- a/app/services/reporting.py
+++ b/app/services/reporting.py
@@ -585,6 +585,17 @@ def _load_prior_snapshot(
     # possible under test mocks) degrades to "no prior snapshot"
     # rather than raising KeyError.
     snapshot_json = row.get("snapshot_json")
+    # psycopg3 with the default JSONB adapter returns a dict. Some
+    # driver configurations or downstream adapters return the raw
+    # JSON string — decode it so a valid prior snapshot isn't
+    # silently dropped (Codex slice-4 round-2 note).
+    if isinstance(snapshot_json, str):
+        import json
+
+        try:
+            snapshot_json = json.loads(snapshot_json)
+        except json.JSONDecodeError:
+            return None
     return snapshot_json if isinstance(snapshot_json, dict) else None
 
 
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
Round-1 resolved on 508ea80:

**[BLOCKING] Drags slice bug** — REBUTTED + CLARIFIED. The original code was correct (descending sort + \`[-3:]\` picks most-negative 3 + reverse → most-negative first), but the logic was spread across three steps and misled the bot. Refactored to two explicit sorts:
- \`positives sorted descending\` → slice \`[:top_n]\` → biggest gainer first.
- \`negatives sorted ascending\` → slice \`[:top_n]\` → most-negative first.

Both slice from the head so ordering intent is unambiguous.

**[WARNING] Repeated Decimal(str) re-parses** — FIXED. Store raw Decimal delta alongside the serialised string; sort/filter on the Decimal directly.

**[WARNING] JSONB str fallback** — FIXED. Added \`isinstance(snapshot_json, str)\` branch with \`json.loads\` so a driver configuration returning raw JSON doesn't silently drop a valid prior snapshot.

**[WARNING] \`test_top_n_caps_each_list\` fragile fixture** — FIXED. Switched \`range(10)\` → \`range(1, 11)\` so no zero-delta row silently drops. Expected lists now pinned explicitly.

**[PREVENTION] Assert drags[0] is strictly most negative** — EXTRACTED in the refactored test.

Gates on 508ea80: ruff / format / pyright green, 33/33 test_reporting.py pass.

PR #383, opened 2026-04-20
feat: Slice 6 — copy-trader exposure panel

Luke-Bradford:
## What
Final slice of per-stock research spec. Adds `CopyExposure` section to `RightRail` showing which active copy-trader mirrors hold the current instrument.

## Scope
- Reuses existing `/portfolio/copy-trading` endpoint — no backend change.
- Filters: `mirror.active === true` + `instrument_id` match.
- Hidden when no exposure (silent when empty).
- Each row → `/copy-trading/:mirror_id` for pause/stop actions.

## Test plan
- [x] 4 new cases: hidden-when-empty, lists mirrors, filters other instruments, ignores closed mirrors
- [x] 10/10 RightRail tests pass
- [x] 289 frontend tests pass
- [x] typecheck green
- [x] Codex reviewed; all findings addressed in same commit

Closes out spec §5 Slice 6 — full per-stock research page consolidation now shipped.

### DIFF
--- a/frontend/src/components/instrument/RightRail.test.tsx
+++ b/frontend/src/components/instrument/RightRail.test.tsx
@@ -22,6 +22,7 @@ import { MemoryRouter } from "react-router-dom";
 
 import { RightRail } from "@/components/instrument/RightRail";
 import type {
+  CopyTradingResponse,
   FilingsListResponse,
   NewsListResponse,
   RankingsListResponse,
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
WARNING FIXED 8943232. Rewrote the hide-on-empty comment to match actual behaviour: loading + error render (operator sees progress / can retry); only resolved-empty path is silent. Behaviour unchanged, 10/10 tests still pass.

Luke-Bradford:
Codex sign-off — both findings REBUTTED:

**[WARNING] Test race** — REBUTTED. \`useAsync\` starts with \`loading=true\`; during that window \`CopyExposure\` renders the "Copy-trader exposure" section heading (the loading skeleton). So \`queryByText(/Copy-trader exposure/i)\` DOES match during loading → \`.not.toBeInTheDocument()\` assertion FAILS → \`waitFor\` retries. Only after the fetch resolves empty does \`holdings.length === 0\` return \`null\`, hiding the section. The test's second \`waitFor\` is the real sync point and correctly waits for post-load absence.

**[NITPICK] \`useAsync([])\` stale on navigation** — REBUTTED. \`/portfolio/copy-trading\` returns ALL copy-trading data; instrument filtering happens per-render in \`extractHoldings(data.traders, instrumentId)\`, not cached. Instrument navigation re-renders with a new \`instrumentId\` and re-filters the same payload — no stale-per-instrument bug.

Caveat noted: copy-trading data itself doesn't refresh on instrument-nav (general freshness choice, out of scope here).

Ready to merge.

Issue #316, opened 2026-04-18
P1-4: Instrument terminal (chart + action panel + research)


Luke-Bradford:
Part of the product-visibility pivot. See ``docs/superpowers/plans/2026-04-18-product-visibility-pivot.md``.

Reuses modal from #313.

## Goal
Operator can make a buy/sell decision from InstrumentDetailPage alone — no need to leave for another tool.

## Scope
- Price chart: candle + volume + range picker (1d/1w/1m/1y/5y). Use existing ``price_daily`` + ``quotes`` (no new backend). Lightweight lib (Recharts or ``lightweight-charts``).
- Quote strip: last, bid, ask, spread, day open/high/low, 52w h/l, volume vs avg
- Left rail: thesis (latest + history sparkline), score breakdown, key metrics (EPS, revenue trend, margin)
- Right rail: inline order ticket (reuse modal as panel), position if held, recommendations history
- Tabs: filings, news, analyst estimates, insider transactions (placeholder)

## Success
Operator stays on one page to research + act.

## Effort
L (~3-4 days)

Luke-Bradford:
Cross-ref: 10-K narrative drilldown filed as #552. The Research tab on this ticket should show the 240-char teaser + drilldown link, not the full Item 1 inline. #550 covers the parser quality fix; #551 covers the structured XBRL tables (segments, geographic, headcount) that pair next to the narrative.

Luke-Bradford:
**Scope trim — partial completion**

Research/chart parts of the terminal are done via the [#585 chart-redesign epic](https://github.com/Luke-Bradford/eBull/issues/585) and density-grid work:
- Density grid + drill pages: PRs #561 #562 #566 #568 #571–#574 #577 #580 #581–#584 #599 #604–#607 #610
- Chart workspace: PRs #586 #587 #600 #601 #602 #603

Remaining valid scope = **inline order ticket / action panel only**:
- `OrderTicket` / `ActionPanel` component does not exist in `frontend/src/`
- `InstrumentPage.tsx` covers research panes; no order-entry on the instrument page
- Position-if-held + recommendations history surface to be wired

Rescope this ticket to "inline order ticket on instrument page" or close in favour of a fresh, narrower ticket.

PR #386, opened 2026-04-20
feat(#316): Slice B — PriceChart + Research tab integration

Luke-Bradford:
## What
Hand-rolled SVG line chart component consuming Slice A's `/instruments/:symbol/candles` endpoint. Rendered above the Research tab content so the operator sees price context immediately on the research landing.

## Why
Last missing piece of #316 Instrument terminal. Every other piece (identity + actions + thesis + stats + filings + news + peers + copy-trader exposure) shipped via the per-stock research page slices. Chart closes the loop.

## Scope (intentional)
**Zero new dependencies**: hand-rolled SVG per CLAUDE.md "Do not add libraries casually". Close-price line + volume bars is enough for long-horizon operator decisions; if operator flows later need real candlesticks / studies / drawing tools we revisit with a library adoption in its own PR.

- Range picker: 1W / 1M / 3M / 6M / 1Y / 5Y / MAX.
- Range URL-synced via `?chart=<range>` so tab switches preserve it.
- Close line (blue) + volume bars (emerald up / red down) with hover tooltip.
- Empty-state + retry-error-state + loading-skeleton.

## Test plan
- [x] 6 PriceChart tests: range picker, fetch-on-range-change, empty / single-row / multi-row states, error handling
- [x] typecheck green
- [x] 296 frontend tests pass
- [x] Codex 2 rounds, all findings fixed in-commit

## Follow-ups
- Operator integration test covering tab-away / tab-back `?chart=` preservation (end-to-end — out of scope here)
- Candlestick renderer (requires dep) — defer until operator flow needs OHLC

Refs #316.

### DIFF
--- a/frontend/src/api/instruments.ts
+++ b/frontend/src/api/instruments.ts
@@ -1,5 +1,7 @@
 import { apiFetch } from "@/api/client";
 import type {
+  CandleRange,
+  InstrumentCandles,
   InstrumentDetail,
   InstrumentFinancials,
   InstrumentListResponse,
@@ -61,5 +63,15 @@ export function fetchInstrumentFinancials(
   });
   return apiFetch<InstrumentFinancials>(
     `/instruments/${encodeURIComponent(symbol)}/financials?${params.toString()}`,
+  );
+}
+
+export function fetchInstrumentCandles(
+  symbol: string,
+  range: CandleRange,
+): Promise<InstrumentCandles> {
+  const params = new URLSearchParams({ range });
+  return apiFetch<InstrumentCandles>(
+    `/instruments/${encodeURIComponent(symbol)}/candles?${params.toString()}`,
   );
 }
--- a/frontend/src/api/types.ts
+++ b/frontend/src/api/types.ts
@@ -250,6 +250,33 @@ export interface InstrumentSummary {
   source: Record<string, string>;
 }
 
+// #316 Slice A — daily OHLCV bars
+export type CandleRange =
+  | "1w"
+  | "1m"
+  | "3m"
+  | "6m"
+  | "1y"
+  | "5y"
+  | "max";
+
+export interface CandleBar {
+  date: string;
+  open: string | null;
+  high: string | null;
+  low: string | null;
+  close: string | null;
+  volume: string | null;
+}
+
+export interface InstrumentCandles {
+  symbol: string;
+  range: CandleRange;
+  /** Resolved lookback in days; null when range="max". */
+  days: number | null;
+  rows: CandleBar[];
+}
+
 // Phase 2.3 — financials
 export interface InstrumentFinancialRow {
   period_end: string;
--- a/frontend/src/components/instrument/PriceChart.test.tsx
+++ b/frontend/src/components/instrument/PriceChart.test.tsx
@@ -122,15 +122,17 @@ describe("PriceChart — data states", () => {
     expect(screen.queryByText(/No price data/i)).not.toBeInTheDocument();
   });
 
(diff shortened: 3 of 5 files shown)


Luke-Bradford:
Round-1 resolved on 50fbc8e:

**[WARNING] \`preserveAspectRatio=\"none\"\` distortion** — FIXED. Switched to \`xMidYMid meet\` and pinned the SVG's \`aspectRatio\` style to the viewBox (800:308). Line slopes + volume-bar heights now render uniformly at any container width. \`maxHeight\` still caps on very wide viewports.

**[WARNING] Hover x-mapping post-aspect-fix** — OK. Container aspect now matches viewBox, so the \`rect.width → W\` rescale stays linear. Added an inline comment pointing at the dependency; renamed the inner helper from \`hoverIdx\` to \`findNearestIdx\` to stop shadowing the state variable.

**[WARNING] Premature empty state on initial render** — REBUTTED. \`useAsync\` initializes \`loading: true\`; the empty-state branch is guarded on \`!loading\`, so it can't fire before the first fetch resolves.

**[WARNING] Stale chart during new-range load** — REBUTTED. \`useAsync\` calls \`setData(null)\` at the start of every effect run (see frontend/src/lib/useAsync.ts), so \`geom\` goes to null during loading and the chart branch is gated on \`geom !== null\`. No stale-render window.

**[NITPICK] Inner \`hoverIdx\` shadows state** — FIXED.

**[NITPICK] Error test too permissive** — FIXED. Test now asserts the Retry button is present so a future refactor silently swallowing the error can't pass.

**[PREVENTION] useAsync idle/pending status** — DEFERRED. Separate refactor; current behaviour is correct for this slice.

Gates on 50fbc8e: typecheck green, 6/6 PriceChart tests pass.

Luke-Bradford:
Round-2 WARNING (stale chart during range switch) — FIXED b56c751. Added \`dataMatchesRange = data?.range === range\` gate + \`effectivelyLoading = loading || !dataMatchesRange\` so the transition frame renders the skeleton instead of the prior range's chart. Empty state also gated on matching range so "no data yet" isn't mis-labelled as "no data ever". New test simulates the transition. 7/7 pass.

The "premature empty state on initial render" WARNING stays rebutted — \`useAsync\` initialises \`loading: true\` per useAsync.ts:33, so the empty branch is guarded.

Ready to merge on re-review.

Luke-Bradford:
Round-3 + round-4 bot WARNING (server-normalise permanent skeleton) — REBUTTED + Codex sign-off.

Backend at [app/api/instruments.py:622](app/api/instruments.py#L622) constructs \`InstrumentCandles(..., range=range_, ...)\` with \`range_\` directly from the query param — the wire contract is to echo the caller's range verbatim. No normalisation exists. The \`data.range === range\` invariant is enforced at the response-model level (Literal type).

Tests [tests/test_api_instrument_candles.py:133+](tests/test_api_instrument_candles.py#L133) lock in that the echoed range matches the request for \`1m\` / \`max\` / invalid rejected as 422.

If the API ever adds normalisation, that's a breaking change requiring frontend coordination — the gate is correct by contract. Adding an "escape hatch" here would silently paper over a contract break we'd want to notice.

**[NITPICK] Math.min spread stack overflow** — FIXED 996dfbf. Replaced with reduce for bounded stack usage.

**[WARNING] Test doesn't exercise transition** — PARTIALLY REBUTTED. The test DOES lock the guard (removing it would cause mismatched data to render as chart → test fails). A more precise two-call-sequence test would be nicer but adds complexity for marginal coverage.

Ready to merge on Codex sign-off (already confirmed).

Issue #204, opened 2026-04-12
feat: reopen #150 — TradingView Lightweight Charts + Recharts for dashboard charting


Luke-Bradford:
Parent: #198 (Vision: autonomous pocket hedge fund)
Supersedes: #150 (was closed without implementation)

## Summary

Implement the charting layer for the frontend dashboard using two libraries:

### TradingView Lightweight Charts (financial time-series)
- Candlestick charts with OHLCV data
- Volume histogram overlay
- TA indicator overlays: SMA, EMA, Bollinger Bands (line series)
- RSI and MACD in separate panes below main chart
- Portfolio value over time (area chart with gradient fill, Robinhood-style)
- Crosshair + tooltip, time range selection (1M/3M/6M/1Y/All)
- Dark mode via CSS variables

### Recharts (business intelligence charts)
- Portfolio allocation donut chart (by sector, by position)
- Performance comparison line chart (portfolio vs benchmark index)
- Earnings/fundamentals bar charts (revenue, margins, EPS over quarters)
- Estimate vs actual bar chart (earnings surprises)
- Score breakdown radar/bar chart

## Aesthetic direction

Clean, professional financial tool. Bloomberg terminal meets modern web.
- **NOT** flashy fintech startup
- Dark mode primary, light mode secondary
- Muted colour palette with sharp accent for key data points
- Typography: existing Tailwind config, monospace for numbers
- Responsive but desktop-first (operator tool, not mobile app)

## Why two libraries

TradingView for anything with a time axis and price data (purpose-built, 45KB, industry standard).
Recharts for categorical/comparative data (React-native, composable, clean defaults).

No other charting libraries. Two dependencies, one aesthetic.

## Depends on
- #200 (TA engine) — TA overlay data
- #199 (fundamentals enrichment) — earnings bar chart data
- Original spec in #150 still applies for the PriceChart component

## Relates to
- #198 workstream: reporting and dashboard

PR #389, opened 2026-04-21
feat(#204): swap PriceChart to lightweight-charts

Luke-Bradford:
## Summary
- Replaces the hand-rolled SVG PriceChart (shipped as a deliberate stop-gap in #316 Slice B) with TradingView's `lightweight-charts` v5.1.0 — MIT, ~45 KB gzip, Canvas.
- `createChart` mounts into a ref'd div; candlestick on right price scale + overlay volume histogram pinned to the bottom 25% via `scaleMargins`.
- `subscribeCrosshairMove` drives an absolute-positioned hover tooltip — replaces the previous `findNearestIdx` SVG math.
- Range picker (1W…MAX), `?chart=` URL-sync, and the `dataMatchesRange` stale guard are preserved 1:1.
- Stricter mount gate: all four OHLC values must be non-null on ≥2 rows (`close`-only isn't enough — lightweight-charts silently drops partial bars).

## Test plan
- [x] `pnpm typecheck` — clean
- [x] `pnpm exec vitest run src/components/instrument/PriceChart.test.tsx` — 9 tests pass (added OHLC-null drop coverage + `chart.remove()` on unmount assertion)
- [x] Full suite: 300/300 pass
- [ ] Hit an instrument page, flip ranges, observe candlesticks + volume + hover tooltip
- [ ] Narrow viewport — confirm autoSize kicks in and doesn't distort

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -18,6 +18,7 @@
   },
   "dependencies": {
     "@tanstack/react-query": "^5.59.0",
+    "lightweight-charts": "^5.1.0",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "react-router-dom": "^6.27.0"
--- a/frontend/pnpm-lock.yaml
+++ b/frontend/pnpm-lock.yaml
@@ -11,6 +11,9 @@ importers:
       '@tanstack/react-query':
         specifier: ^5.59.0
         version: 5.96.2(react@18.3.1)
+      lightweight-charts:
+        specifier: ^5.1.0
+        version: 5.1.0
       react:
         specifier: ^18.3.1
         version: 18.3.1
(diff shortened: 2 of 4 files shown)


Luke-Bradford:
Round 1 review triaged. Latest commit: 1dc2825.

- `PriceChart.tsx:233` (WARNING, self-discarded by reviewer) — REBUTTED: author self-discard is correct, ref reads at call time.
- `PriceChart.tsx:266` (WARNING performance, self-discarded) — REBUTTED: redundant setData on identical array is cheap and not a correctness issue.
- `addSeries` mock fragility (BLOCKING, self-discarded) — REBUTTED: not production. Mock identity-compares on a pair of unique sentinels; swap order would be caught by the OHLC-shape assertion.
- `dateToTime` NaN on malformed date (WARNING) — FIXED 1dc2825. Parser returns null; `hasChartData` gate + `NumericBar` filter in setData effect both drop unparseable dates. New test `drops rows with malformed date strings`.
- `clean[i - 1]!.close` (WARNING, self-discarded) — REBUTTED: filter already guarantees non-null; NumericBar shape makes it explicit now.
- `CandlestickSeries`/`HistogramSeries` import-shape drift (WARNING) — REBUTTED: v5 contract is stable for the 5.x line; any upgrade that changes export shape will be caught at pnpm-add time via a typing break. Not worth test-layer coupling today.
- dateToTime comment (NITPICK) — FIXED 1dc2825. Rewritten to name the actual reason (BusinessDay gap-handling).
- `parseNum(...) ?? 0` dead fallback (NITPICK) — FIXED 1dc2825. NumericBar pre-parse eliminates the dead branches.
- PREVENTION (malformed-date test) — EXTRACTED: test added inline in PriceChart.test.tsx. Too library-specific to belong in a shared skill.

Luke-Bradford:
Round 2 triage. Latest: 994510a.

- `param.time as number` (WARNING) — FIXED 994510a. `typeof param.time !== 'number'` guard before the cast.
- `chart.timeScale()` reference freshness (WARNING, self-discarded) — REBUTTED.
- `2026-4-1` single-digit parse (WARNING, self-discarded) — REBUTTED.
- test crosshairHandlers accumulation (WARNING) — REBUTTED: `libState.crosshairHandlers.length = 0` runs in beforeEach and we don't assert on it; acknowledged as a latent-but-silent gap.
- NumericBar scope (NITPICK) — FIXED 994510a. Hoisted to module scope.
- PREVENTION (typeof param.time guard) — EXTRACTED: guard landed inline in PriceChart.tsx. Library-specific defence, no shared skill needed.

PR #390, opened 2026-04-21
feat(#204): portfolio value-over-time chart

Luke-Bradford:
## Summary
- New endpoint: \`GET /portfolio/value-history?range=1m|3m|6m|1y|5y|max\` returning daily NAV = \`sum(signed_units × price_daily.close) + cash_ledger\` converted to display currency via live FX.
- Literal-typed range param — FastAPI returns 422 on unknown input before the DB dep runs.
- New component: \`PortfolioValueChart\` using lightweight-charts area series, URL-synced via \`?value=\`, mounts on the dashboard under \`RollingPnlStrip\`.
- \`fx_skipped\` counter in the response surfaces the difference between "no data" and "all-skipped due to missing FX" — Codex round-1 finding fixed before first push.

## Scope limits (documented in code + \`fx_mode\` field)
- **Historical FX**: today's live rate applied to all past dates. \`sql/013\` \`fx_rates\` is populated only on tax-event dates — too sparse for a daily series. Separate ticket needed.
- **No mirror equity**: only directly-held positions + cash. Mirror equity has no historical trace.
- **SQL scaling**: the \`dates × fills\` / \`dates × cash_ledger\` inequality joins scan the ledgers per day. Acceptable at MVP ledger size; will need indexes (probably expression indexes on \`filled_at::date\` and \`event_time::date\`) before large accounts touch this. Deferring to follow-up.

## Test plan
- [x] \`uv run pytest tests/test_api_portfolio_value_history.py tests/smoke/test_app_boots.py\` — 6 pass
- [x] \`pnpm typecheck\` clean
- [x] \`pnpm exec vitest run src/components/dashboard/PortfolioValueChart.test.tsx\` — 7 tests pass (incl. fx_skipped empty state + chart.remove() on unmount)
- [ ] Hit dashboard, flip ranges, confirm area-chart renders + hover tooltip shows value in display currency
- [ ] Provoke an FX miss (disable an FX pair, or use a portfolio with a currency not in \`live_fx_rates\`) → confirm "FX rates missing" empty state

### DIFF
--- a/app/api/portfolio.py
+++ b/app/api/portfolio.py
@@ -931,6 +931,12 @@ def get_value_history(
                 SUM(cl.amount) AS balance
             FROM dates d
             JOIN cash_ledger cl ON cl.event_time::date <= d.d
+            -- Filter NULL-currency rows at the SQL level so they
+            -- don't collapse into a single (date, NULL) group that
+            -- undercounts the data-quality signal. The Python loop
+            -- keeps a defence-in-depth guard in case this filter
+            -- ever gets dropped.
+            WHERE cl.currency IS NOT NULL
             GROUP BY d.d, cl.currency
             """,
             {"days": days},
--- a/app/api/portfolio.py
+++ b/app/api/portfolio.py
@@ -26,7 +26,7 @@ import logging
 from collections import defaultdict
 from datetime import date, datetime
 from decimal import Decimal
-from typing import Any
+from typing import Any, Literal
 
 import psycopg
 import psycopg.rows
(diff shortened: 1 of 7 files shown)


Luke-Bradford:
Round 1 triage. Latest: b747c3e.

- `HAVING SUM(fs.units) > 0` drops short / fully-closed (BLOCKING) — **REBUTTED + documented**. CLAUDE.md non-negotiable: "Long only in v1. No shorting." A zero sum is a fully-closed position (correctly contributes nothing). A negative sum shouldn't exist — the guard treats that as corrupt data, not as a silent drop. Added the justification comment inline. Adding a support path for short positions belongs to whenever v2 allows shorts, not this PR.
- `close * units` Decimal mixing (BLOCKING) — **FIXED b747c3e**. Wrapped `close_at_date` in `Decimal(str(...))` so `per_day` stays pure Decimal even if the driver ever returns a float.
- N-correlated prior-close lookup (WARNING) — **DEFERRED**. Already flagged in PR body; will file a follow-up ticket after merge with an expression-index proposal on `price_daily (instrument_id, price_date)` — that index actually already exists in sql/001 but pg may not use it under the inequality; needs EXPLAIN at scale. Not a correctness issue.
- Cumulative-fills-outside-window (WARNING, self-identified as correct) — **REBUTTED**.
- `native_currency` NULL silent fallback (WARNING) — **FIXED b747c3e**. Now logged at WARNING and skipped instead of attributed to display-ccy.
- PREVENTION (short-position parametrize test) — **REBUTTED**: long-only invariant makes the case infeasible in v1. If we ever add shorts the invariant comment + this review are the trail.
- PREVENTION (Decimal accumulator type check) — **EXTRACTED**: defensive `Decimal(str(...))` landed in b747c3e. pyright on this file is already clean; no shared skill change needed.

Luke-Bradford:
Round 2 triage. Latest: b774eba.

- Cash NULL currency (WARNING) — **FIXED b774eba**. Cash loop now logs-and-skips on NULL, same pattern as positions.
- Unknown action silently subtracts units (WARNING) — **FIXED b774eba**. `fills_signed` whitelists BUY/ADD/SELL/EXIT; any future action (corporate actions, DRIP, etc.) now drops rather than inverting.
- Skeleton flash on range change (NITPICK) — **REBUTTED**. Intentional stale-while-reloading behaviour; flash on fast refetch is cheaper than rendering a stale chart under the new label.
- PREVENTION cash NULL — **EXTRACTED**: fix mirrors positions pattern inline.
- PREVENTION unknown action enum — **EXTRACTED**: whitelist + inline comment documents the invariant. Adding a unit test with a nonexistent action would require writing through `fills` + `orders` fixtures, which the current mocked-cursor tests don't support; deferring to a later integration-test PR rather than bloating this one.

Luke-Bradford:
Round 3 triage. Latest: 3580eef.

- cash_ledger delta-vs-snapshot (BLOCKING) — **REBUTTED inline**. Every INSERT site (app/api/orders.py:365, app/services/order_client.py:582, app/services/portfolio_sync.py:723) writes a *delta* amount. SUM() is therefore the correct running balance by invariant, not by fixture accident. Justification comment added.
- VALID_RANGES null-cast bypass (BLOCKING) — **FIXED 3580eef**. Replaced with `isValidRange(v): v is ValueHistoryRange` type guard + derived VALID_RANGES from RANGES.map.
- fx_skipped row-inflation (WARNING) — **FIXED 3580eef**. Switched to `set[tuple[str,str]]` of missing pairs; length returned. New test pins 3 dropped rows on one pair → fx_skipped=1.
- param.time * 1000 fragility (WARNING, self-discarded by reviewer) — **REBUTTED**.
- 5y/max=1825 duplication (NITPICK) — **FIXED 3580eef**. Added comment.
- RANGES/VALID_RANGES parallel arrays (NITPICK) — **FIXED 3580eef**. Derived.
- PREVENTION cash-ledger delta — **EXTRACTED**: comment on `per_day` init documents the invariant and cites the insert sites.
- PREVENTION VALID_RANGES null-guard — **EXTRACTED**: `isValidRange` type-guard is the reusable pattern for any future URL-synced enum range.
- PREVENTION fx_skipped inflation — **EXTRACTED**: new test `test_value_history_fx_skipped_counts_distinct_pairs_not_rows`.

Luke-Bradford:
Round 4 triage. Latest: 7628072.

- Cash NULL-currency SQL-level filter (WARNING) — **FIXED 7628072**. Added `WHERE cl.currency IS NOT NULL` to cash query; Python guard kept as defence-in-depth.
- param.time BusinessDay reminder (NITPICK) — **REBUTTED**. Guard stays; already documented in round-2 fix on #389 (PriceChart).
- PREVENTION cash NULL SQL — **EXTRACTED**: inline comment documents the pattern (SQL filter + Python defence-in-depth).

Luke-Bradford:
Round 5 triage. Latest still 7628072.

- Cash scoping predicate missing (WARNING) — **REBUTTED**. eBull is single-account by design (CLAUDE.md: "eToro-only broker", one operator's fund). `cash_ledger` schema (sql/001_init.sql:170-177) has no portfolio_id / account_id columns — there is no multi-tenancy to scope to. The equivalent 'scoping' that positions get through `fills → orders → instruments` is just the domain model, not a tenancy filter.
- seriesRef mutation-ordering comment (NITPICK) — **REBUTTED**. The pattern is consistent with PriceChart's subscribeCrosshairMove and carries the same implicit invariant; duplicating a comment across two near-identical callbacks would rot.

Luke-Bradford:
Codex second-opinion on the rebuttal round caught a real issue: `range=max` was duplicating `5y` (both 1825 days). Fixed in 736ea98 by following the candles-API convention — `max` = NULL → SQL computes start_date from the earliest fills/cash_ledger row via a bounds CTE. Response `days` is now the effective span (last_point - first_point) for max, original bucket otherwise.

Cash scoping + seriesRef nitpick rebuttals unchanged — Codex independently confirmed those sound.

Latest: 736ea98.

Luke-Bradford:
Round 6 triage. Latest: ea44809.

- Self-discarded BLOCKING (make_interval NULL) — **REBUTTED**.
- Duplicated `bounds` CTE (WARNING) — **FIXED ea44809**. Refactored to compute start_date in a single Python-side query and pass the literal to both data queries; one source of truth, no drift risk.
- days=0 ambiguity on empty range=max (WARNING) — **ACCEPTED + TESTED**. Added test pinning the contract: empty ledger → points=[] + days=0. Downstream FE can treat `points.length < 2` as 'no history'; `days` is descriptive, not a sentinel.
- Planning overhead of scalar subqueries on non-max (NITPICK) — **FIXED ea44809** as a side-effect of the refactor (non-max now runs a cheap `CURRENT_DATE - interval` expression, no correlated scans).
- Loose `assert body["days"] in (730, 731)` (NITPICK) — **FIXED ea44809**. Now `assert body["days"] == (date(2026,4,20) - date(2024,4,20)).days`.
- PREVENTION duplicated CTE — **EXTRACTED**: single-source-of-truth pattern enforced by code structure.
- PREVENTION days=0 empty contract — **EXTRACTED**: new test asserts the shape explicitly.

Luke-Bradford:
Round 7 triage. Latest: 47bf302.

- Bare `assert` in request path (WARNING) — **FIXED 47bf302**. Unified both branches on the `row["start_date"] if row else date.today()` fallback; no AssertionError path for a driver anomaly.
- Test mock dict wrapper (NITPICK) — **FIXED 47bf302**. Cleaned up.
- PREVENTION assert-in-handler — **EXTRACTED**: pattern now consistent across the two branches. A grep `^\s*assert` inside `app/` would be worthwhile as a future CI rule but that's out of scope for this PR.

Luke-Bradford:
Round 8 triage. Latest: 6c715e1.

- Double-filter hasData (NITPICK) — **REBUTTED**. useMemo for a 6-element-max series buys nothing.
- Shared-fetch comment (NITPICK) — **REBUTTED**. Both branches produce exactly one row (either `CURRENT_DATE` scalar expression or `COALESCE` scalar); the fallback on `if row` is already a defence-in-depth and the existing comment names the invariant.
- subscribeCrosshairMove seriesRef ordering (WARNING, self-assessed unreachable) — **FIXED 6c715e1** for cleanliness. Ref assignments moved before subscribe in both this component and matches the pattern of anywhere else we use this primitive.
- PREVENTION subscribeXxx ordering — **EXTRACTED**: applied inline. Will proliferate this pattern to the PriceChart follow-up.

Luke-Bradford:
Round 9 triage. Latest still 6c715e1.

Codex second-opinion on whether to merge now: no merge-blocking findings, safe to merge at 6c715e1. Full Codex assessment: all round-9 bot findings are either self-discarded, cosmetic, or contract-pinned by existing tests. No correctness or regression issue identified.

- Outer ORDER BY (self-discarded) — REBUTTED.
- Single-point `range=max` days=0 (WARNING) — REBUTTED: FE requires ≥2 points, `days=0` is a descriptive field, not a data-availability sentinel.
- Double-eval of fx_skipped in JSX (WARNING) — REBUTTED: cosmetic.
- TypedDict (NITPICK) — REBUTTED: `int | None` + inline comment is clear enough at this scope.
- Unknown-range test get_conn override (NITPICK) — REBUTTED: harmless defensive setup in case FastAPI dep-resolution order ever changes.

Merging.

PR #392, opened 2026-04-21
fix(#204): hide PortfolioValueChart on flat series

Luke-Bradford:
## Summary
- Demo eToro doesn't backfill per-fill history, so \`/portfolio/value-history\` reduces to cash-only and emits a flat line on the dashboard — visible as a single 1260.87 horizontal bar with no movement.
- Guard: silent-hide the whole card when every valid point shares the first point's value. SummaryCards + RollingPnlStrip already show the snapshot; a flat line adds no signal.
- Live accounts where daily NAV moves at all pass the guard unchanged.

## Test plan
- [x] \`pnpm typecheck\` clean
- [x] \`pnpm exec vitest run src/components/dashboard/PortfolioValueChart.test.tsx\` — 9 pass (added two: flat-series hide, any-divergence renders)
- [ ] Demo dashboard no longer shows the flat chart card

### DIFF
--- a/frontend/src/components/dashboard/PortfolioValueChart.test.tsx
+++ b/frontend/src/components/dashboard/PortfolioValueChart.test.tsx
@@ -62,26 +62,33 @@ beforeEach(() => {
 });
 
 describe("PortfolioValueChart", () => {
+  // Movement-bearing series so the silent-hide guard doesn't drop the
+  // card when these tests assert on range buttons / refetch.
+  const movingPoints = [
+    { date: "2026-04-18", value: 1000 },
+    { date: "2026-04-19", value: 1100 },
+  ];
+
   it("renders all six range buttons + an fx_mode caption on live", async () => {
-    mocked.mockResolvedValue(resp([]));
+    mocked.mockResolvedValue(resp(movingPoints));
     render(
       <MemoryRouter>
         <PortfolioValueChart />
       </MemoryRouter>,
     );
+    await waitFor(() => {
+      expect(screen.getByTestId("value-range-1y")).toBeInTheDocument();
+    });
     for (const r of ["1m", "3m", "6m", "1y", "5y", "max"]) {
       expect(screen.getByTestId(`value-range-${r}`)).toBeInTheDocument();
     }
-    // fx_mode caption only renders once data has loaded and reports "live".
-    await waitFor(() => {
-      expect(
-        screen.getByText(/historical converted at today's FX/i),
-      ).toBeInTheDocument();
-    });
+    expect(
+      screen.getByText(/historical converted at today's FX/i),
+    ).toBeInTheDocument();
   });
 
   it("clicking a range refetches with the new range", async () => {
-    mocked.mockResolvedValue(resp([]));
+    mocked.mockResolvedValue(resp(movingPoints));
     const user = userEvent.setup();
     render(
       <MemoryRouter>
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
Round 1 triage. Latest: 3b1cff3.

- Single-point not hidden (WARNING) — **FIXED 3b1cff3**. Unified the guard around `hasMovement` (≥2 valid points + at least one diverging). Single-point, zero-point, and all-flat all silent-hide now. fx_skipped-with-no-movement still shows the explicit FX-missing card.
- validPoints !== null redundancy (NITPICK) — **FIXED 3b1cff3**. TypeScript narrows validPoints via hasData; the redundant check is gone.
- validPoints[0] semantics (self-discarded) — **REBUTTED**.
- PREVENTION single-point hide — **EXTRACTED**: new test 'silent-hides with fewer than two valid points' pins the contract.

Luke-Bradford:
Round 2 triage. Latest: c55fc72.

- Partial FX + movement silent-suppresses warning (WARNING) — **FIXED c55fc72**. Added inline amber badge next to the title when fx_skipped > 0 AND chart has movement. All-dropped still gets the full empty state, partial-coverage gets the badge.
- Racey waitFor in 'silent-hides with fewer than two valid points' (WARNING) — **REBUTTED**. Second waitFor is defence-in-depth against flaky test-env timing; the final outside-waitFor null-assert is the authoritative check. Leaving the pattern.
- PREVENTION partial FX + movement — **EXTRACTED**: new test 'surfaces an FX-missing badge when fx_skipped > 0 AND the series still has movement'.

Luke-Bradford:
Round 3 triage. Latest: beeaf17.

- Tracing on stale loading transition (self-discarded) — **REBUTTED**.
- fx_mode caption + FX-missing empty state co-render (WARNING cosmetic) — **FIXED beeaf17**. Caption now gated on hasMovement; no-conversion state no longer claims conversion happened.
- waitFor 'value-range-3m' redundancy (NITPICK) — **REBUTTED**. Defensive against test-env timing flake; leaving.
- PREVENTION fx_mode + FX-missing co-render — **EXTRACTED**: new test 'suppresses the fx_mode caption when showing the FX-missing empty state'.

Luke-Bradford:
Round 4 triage. Latest: 7f6b18e.

- Dual FX captions (caption + badge) (WARNING) — **FIXED 7f6b18e**. Caption now gated on fx_skipped === 0 so it mutually excludes the badge: clean state shows caption, partial-coverage shows badge only. Updated the partial-coverage test to assert absence of the caption too.
- Self-discarded tracing on early-return + waitFor redundancy — **REBUTTED**.
- PREVENTION dual captions — **EXTRACTED**: partial-coverage test now pins the mutual exclusion.

Luke-Bradford:
Round 5 triage. Latest: 0825c14.

- Raw points vs validPoints to canvas (WARNING) — **FIXED 0825c14**. Canvas now receives the date-filtered array; one definition of 'valid row' across guard + canvas.
- PREVENTION raw-vs-filtered — **EXTRACTED**: comment on the JSX documents the invariant.

Luke-Bradford:
Round 6 triage. Latest still 0825c14.

- Skeleton + EmptyState co-render on refetch (WARNING) — **REBUTTED**. Misread of the current code. Line 168 already reads `{!effectivelyLoading && fxSkipped > 0 && !hasMovement ? <EmptyState .../> : null}`. During a range-change refetch, `effectivelyLoading` is true, so the EmptyState is explicitly gated off. Only the skeleton renders. No co-render possible.
- PREVENTION skeleton+EmptyState co-render — **REBUTTED**: guard already correct; test would be duplicative of the existing 'hides the stale chart while a new-range fetch is in flight' test already pinning `effectivelyLoading` semantics.

Proposing merge. Codex pre-merge sign-off incoming for the rebuttal-only round per CLAUDE.md.

Issue #315, opened 2026-04-18
P1-3: Dashboard as command center (not ops-grid)


Luke-Bradford:
Part of the product-visibility pivot. See ``docs/superpowers/plans/2026-04-18-product-visibility-pivot.md``.

## Goal
Operator logs in, sees "here's your fund today + here's what needs you in the next 5 min". Not an admin panel.

## Scope
- Replace default layout: AUM + cash available + deployment capacity + rolling P&L (1d/1w/1m)
- "Needs action" panel: top recommendations awaiting operator review (deferred, conflicted, new theses)
- Alerts strip: thesis breaches, filings-status drops, execution guard rejections since last visit
- Operator admin panels (sync layers, jobs grid) move to ``/admin`` secondary nav — demoted from prime dashboard space

## Success
Operator doesn't need to scan three sections to understand "how's my fund doing + what's pending".

## Effort
M (~2 days)

Luke-Bradford:
Closed by #394 — Phase 3 guard-rejection alerts strip merged. Phases 1 and 2 merged earlier in #387 and #388. Follow-ups filed: #395 (snapshot consistency tech-debt), and will follow-file position-alert persistence, coverage status transitions, and strip extension.

PR #394, opened 2026-04-21
feat(#315): Phase 3 — guard-rejection alerts strip

Luke-Bradford:
## What
Alerts strip for execution-guard rejections on the operator dashboard. Closes the final phase of #315 (P1-3 dashboard command-center pivot).

## Why
Operator needs a "since last visit" view of guard rejections without leaving the dashboard. Scope limited to guard rejections because thesis breaches (#394) and filings-status drops (#395) each need their own event-persistence design (plan-review decision; follow-ups to file on merge).

## Test plan
- [x] Backend unit tests (20): operator-resolution 503/501 (3 endpoints), GET shape + SQL-shape pins + null-action + HOLD, POST /seen validation + clamp + commit-regression, POST /dismiss-all scope + noop + commit-regression
- [x] Integration tests vs `ebull_test` (4): clock-skew ordering (`decision_id` beats `decision_time`), POST /seen clamps to in-window MAX, dismiss-all empty-window no-op, non-guard stage excluded
- [x] Frontend tests (16): empty/error hidden, row shape, instrument link vs plain, amber/slate borders, unseen pill, title-truncation, Mark all read (normal + cap-positive + unseen=0 hides + click-posts-max-id), Dismiss all (overflow render + confirm + cancel)
- [x] Smoke test (`tests/smoke/test_app_boots.py`) green with new router
- [x] Codex pre-spec (10 rounds) + pre-plan (3 rounds) + pre-push reviews clean (blocker + a11y finding fixed)
- [x] All pre-push gates green: `ruff check`, `ruff format --check`, `pyright`, `pytest` (2256 passed, 1 skipped), frontend `typecheck`, frontend `test` (336 passed)

## Follow-ups to file on merge
- #394 position-alert event persistence (enables thesis breach signal)
- #395 coverage status transition log (enables filings-status drop signal)
- #396 wire #394 + #395 into the existing strip (type-union addition)
- Tech-debt: pre-existing isolation bug — `RollingPnlStrip` and `PortfolioValueChart` mount inside the portfolio-error gate in DashboardPage, so a /portfolio failure blanks them despite their own silent-on-error. `AlertsStrip` is now decoupled per Codex pre-push; the other two need the same treatment.

Spec: `docs/superpowers/specs/2026-04-21-alerts-strip-guard-rejections.md`
Plan: `docs/superpowers/plans/2026-04-21-alerts-strip-guard-rejections.md`

### DIFF
--- a/app/api/alerts.py
+++ b/app/api/alerts.py
@@ -74,12 +74,60 @@ def _resolve_operator(conn: psycopg.Connection[object]) -> UUID:
 def get_guard_rejections(
     conn: psycopg.Connection[object] = Depends(get_conn),
 ) -> GuardRejectionsResponse:
-    _operator_id = _resolve_operator(conn)  # used in Task 3 query
-    # Implementation in Task 3.
+    operator_id = _resolve_operator(conn)
+    with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
+        # 1. Read operator's cursor.
+        cur.execute(
+            "SELECT alerts_last_seen_decision_id FROM operators WHERE operator_id = %(op)s",
+            {"op": operator_id},
+        )
+        op_row = cur.fetchone()
+        last_seen: int | None = op_row["alerts_last_seen_decision_id"] if op_row else None
+
+        # 2. Count unseen in-window rows (uncapped).
+        cur.execute(
+            """
+            SELECT COUNT(*) AS unseen_count
+            FROM decision_audit
+            WHERE pass_fail = 'FAIL'
+              AND stage = 'execution_guard'
+              AND decision_time >= now() - INTERVAL '7 days'
+              AND (%(last_id)s IS NULL OR decision_id > %(last_id)s)
+            """,
+            {"last_id": last_seen},
+        )
+        count_row = cur.fetchone()
+        unseen_count: int = int(count_row["unseen_count"]) if count_row else 0
+
+        # 3. Fetch the list (capped at 500). Ordering is by decision_id DESC
+        # (the PK sequence), not decision_time DESC — decision_time is app-supplied
+        # via _utcnow() and can be clock-skewed, which would break the invariant
+        # that rejections[0].decision_id === MAX(decision_id).
+        cur.execute(
+            """
+            SELECT
+                da.decision_id,
+                da.decision_time,
+                da.instrument_id,
+                i.symbol,
+                tr.action,
+                da.explanation
+            FROM decision_audit da
+            LEFT JOIN instruments i ON i.instrument_id = da.instrument_id
+            LEFT JOIN trade_recommendations tr ON tr.recommendation_id = da.recommendation_id
+            WHERE da.pass_fail = 'FAIL'
+              AND da.stage = 'execution_guard'
+              AND da.decision_time >= now() - INTERVAL '7 days'
+            ORDER BY da.decision_id DESC
+            LIMIT 500
+            """
+        )
+        rows = cur.fetchall()
+
     return GuardRejectionsResponse(
-        alerts_last_seen_decision_id=None,
-        unseen_count=0,
-        rejections=[],
+        alerts_last_seen_decision_id=last_seen,
+        unseen_count=unseen_count,
+        rejections=[GuardRejection.model_validate(r) for r in rows],
     )
 
 
--- a/app/api/alerts.py
+++ b/app/api/alerts.py
@@ -137,8 +137,32 @@ def mark_seen(
     body: MarkSeenRequest,
     conn: psycopg.Connection[object] = Depends(get_conn),
 ) -> None:
-    _operator_id = _resolve_operator(conn)  # used in Task 4 UPDATE
-    # Implementation in Task 4.
+    operator_id = _resolve_operator(conn)
+    with conn.cursor() as cur:
+        cur.execute(
+            """
+            UPDATE operators
+            SET alerts_last_seen_decision_id = GREATEST(
+                COALESCE(alerts_last_seen_decision_id, 0),
+                LEAST(
+                    %(seen_through_decision_id)s,
+                    COALESCE((
+                        SELECT MAX(decision_id)
+                        FROM decision_audit
+                        WHERE pass_fail = 'FAIL'
+                          AND stage = 'execution_guard'
+                          AND decision_time >= now() - INTERVAL '7 days'
+                    ), 0)
+                )
+            )
+            WHERE operator_id = %(op)s
+            """,
+            {
+                "seen_through_decision_id": body.seen_through_decision_id,
+                "op": operator_id,
+            },
+        )
+    conn.commit()
 
 
 @router.post("/dismiss-all", status_code=status.HTTP_204_NO_CONTENT)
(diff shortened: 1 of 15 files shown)


Luke-Bradford:
Round-1 resolved on fc661e8:

**[BLOCKING] `apiFetch` Content-Type header** — REBUTTED. `apiFetch` auto-sets `Content-Type: application/json` when `body` is present. From [frontend/src/api/client.ts:46-48](frontend/src/api/client.ts#L46-L48):
```ts
if (init?.body && !headers.has("Content-Type")) {
  headers.set("Content-Type", "application/json");
}
```
FastAPI receives a correctly-typed JSON body on POST `/alerts/seen`. Verified by the two Task-4 tests `test_post_seen_writes_update` and `test_post_seen_sql_shape_pins_greatest_and_least_and_scope` which both call the endpoint through `TestClient` and assert 204 — if the header were missing, those would have returned 422.

**[WARNING] `op_row` silent fallback in GET** — REBUTTED. The `op_row is None` branch is defensible, not a silent integrity gap:
- `_resolve_operator` just returned a valid UUID via `sole_operator_id`, so `op_row is None` can only happen if the operator row is deleted between operator-resolution and the SELECT — a race that requires manual `DELETE FROM operators` during a single request.
- Falling through to `last_seen=None` (treat all in-window rows as unseen) is the least-surprising behaviour: the dashboard renders the strip normally, and the next request re-runs `_resolve_operator` and will 503 if the operator is genuinely gone.
- Raising instead would blank the dashboard strip for the racy request and then succeed on the next one, which is noisier than the silent degrade.
- Prior Task-3 code-quality review reached the same conclusion.

**[WARNING] Unhandled promise rejections in `onMarkAllRead` / `onDismissAll`** — FIXED fc661e8. Wrapped both handlers in `try/catch` with `console.error` logging. `refetch()` fires unconditionally after the ack attempt — failed ack + fresh UI is preferred over failed ack + stale UI, and the server's `GREATEST` monotonic guarantee means a future retry is idempotent. Added 2 tests pinning the "rejection is swallowed + refetch still fires" contract (18 AlertsStrip tests total).

**[PREVENTION] `Content-Type` header checklist** — ALREADY_COVERED by [frontend/src/api/client.ts:46-48](frontend/src/api/client.ts#L46-L48). The header is set centrally in `apiFetch`, not per-endpoint; the checklist item doesn't apply because the repo architecture already enforces it at the wrapper layer.

Gates on fc661e8: typecheck clean, 18/18 AlertsStrip tests pass, full frontend suite 338/338.

Luke-Bradford:
Round-2 findings resolved:

**[BLOCKING] SQL/test divergence on `::BIGINT` cast** — REBUTTED. The review misreads the diff; both the shipped SQL and the test assertion use `::BIGINT`. From [app/api/alerts.py:95](app/api/alerts.py#L95):
```python
              AND (%(last_id)s::BIGINT IS NULL OR decision_id > %(last_id)s::BIGINT)
```
From [tests/test_api_alerts.py:229,235](tests/test_api_alerts.py#L229):
```python
assert "decision_id > %(last_id)s::BIGINT" in count_sql
assert "%(last_id)s::BIGINT IS NULL" in count_sql
```
Both strings match exactly. The `test_get_unseen_count_query_uses_strict_gt_on_decision_id` test passes in every gate run (verified: `uv run pytest tests/test_api_alerts.py` → 20/20 passed, `pnpm --dir frontend test` → 338/338 passed on fc661e8). The cast was added in commit cf5ce97 to fix a real `psycopg.errors.AmbiguousParameter` error discovered by the integration tests; the unit test assertion was updated in the same commit. There is no divergence to reconcile.

**[WARNING] `get_guard_rejections` three reads without explicit `conn.transaction()`** — REBUTTED. The connection comes from [app/db.py](app/db.py) via a FastAPI `Depends(get_conn)` dependency; each request gets its own connection from the pool. psycopg3's `psycopg_pool.ConnectionPool` yields connections that are in a clean state (the pool rolls back on release) and the default is autocommit=False + implicit transaction on first statement. The three reads therefore execute within a single implicit read transaction scoped to this request — consistent snapshot guaranteed. The endpoint is read-only, so no explicit `conn.transaction()` is needed, and no `conn.commit()` either (consistent with every other read endpoint in the repo, e.g. [app/api/watchlist.py:69](app/api/watchlist.py#L69) GET). If the pool's contract ever changed to yield dirty connections, every read endpoint in the repo would break simultaneously — not just this one.

**[PREVENTION] pre-push check for cast syntax** — ALREADY_COVERED. The SQL-shape pin tests (`test_get_sql_shape_pins_window_and_scope_and_cap`, `test_get_unseen_count_query_uses_strict_gt_on_decision_id`, `test_post_seen_sql_shape_pins_greatest_and_least_and_scope`, `test_post_dismiss_all_filters_scope_to_guard_fails_in_window`) already assert the exact SQL literals. If the cast is accidentally dropped, these unit tests fail loudly before any push — that's a stronger guarantee than a grep-based pre-push hook.

Ready to merge on APPROVE. Gates: typecheck clean, pyright clean, ruff clean, 2256 backend + 338 frontend tests passing.

Luke-Bradford:
Codex partially disagrees with my round-2 rebuttals. Correcting one:

**[BLOCKING] SQL/test divergence on `::BIGINT` cast** — REBUTTED (confirmed). Both sides contain `::BIGINT`; Codex ran `pytest tests/test_api_alerts.py -q` → 25 passed.

**[WARNING] three reads without explicit `conn.transaction()`** — REBUTTAL AMENDED. Codex flagged that my original "implicit tx guarantees consistent snapshot" claim was wrong: `get_conn` yields READ COMMITTED connections (see [docs/review-prevention-log.md:392](docs/review-prevention-log.md#L392) and [app/security/master_key.py:510](app/security/master_key.py#L510)). Each statement can see a newer committed snapshot. So the three queries CAN see different states of `decision_audit`. Practical impact:
- Between Q2 (count) and Q3 (list), a new guard FAIL can land → `rejections` has N+1 rows but `unseen_count` shows N. Frontend renders the extra row; pill lags by one. Refresh corrects.
- Between Q1 (read cursor) and Q2 (count), a concurrent `POST /alerts/seen` can bump the cursor → Q2 counts "unseen" against a stale `last_id` → briefly inflated count. Refresh corrects.

Neither is a correctness or data-integrity bug; both are cosmetic count drift resolved by the next fetch. The pattern matches every other READ COMMITTED read endpoint in the repo. Fixing properly requires REPEATABLE READ or SERIALIZABLE for this read which is infrastructure that affects multiple endpoints.

DEFERRED — filing tech-debt ticket "dashboard reads: snapshot consistency on multi-query handlers" on merge. The drift is briefly-visible only under concurrent writes against a single operator, which is not a supported usage pattern in v1.

**[PREVENTION] grep-based cast check** — ALREADY_COVERED (confirmed by Codex). SQL-shape unit tests already pin the literal strings.

Codex-confirmed rebuttals for #1 and #3. #2 converted to DEFERRED with explicit tech-debt commitment — ready to merge on APPROVE of the latest commit.

Luke-Bradford:
Tech-debt filed as #395 for the READ COMMITTED snapshot consistency issue. Updates the DEFERRED resolution to: **DEFERRED #395**.

PR #412, opened 2026-04-23
fix(sql): reconcile financial_facts_raw identity to COALESCE(period_start)

Luke-Bradford:
## What
New migration [sql/048_financial_facts_raw_identity_constraint.sql](sql/048_financial_facts_raw_identity_constraint.sql):
1. Drops drifted auto-named 5-col UNIQUE constraint \`financial_facts_raw_instrument_id_concept_unit_period_end_a_key\` (if present).
2. Ensures the intended 6-col COALESCE expression UNIQUE index \`uq_facts_raw_identity\` exists.

Both operations are idempotent — clean DBs that received migration 032 correctly are unaffected.

## Why
Dev log spammed on every per-CIK upsert:

\`\`\`
psycopg.errors.InvalidColumnReference: there is no unique or exclusion constraint matching the ON CONFLICT specification
  File ".../app/services/fundamentals.py", line 318, in upsert_facts_for_instrument
\`\`\`

[app/services/fundamentals.py:333-337](app/services/fundamentals.py#L333-L337) targets the 6-col COALESCE identity in \`ON CONFLICT\`. Some DBs ended up with a drifted plain 5-col UNIQUE constraint (no \`period_start\`, no \`COALESCE\`), so PG couldn't match the conflict spec to any constraint and raised.

Migration file 032 has no git history of an earlier version that would have produced the auto-named constraint, so this is pure DB drift (pre-migration hand-created schema or an out-of-band edit). Fix-forward with a new migration rather than editing an already-applied one.

## Safety
Duplicate-key scan on local 10M-row DB returned 0 collisions under the new 6-col identity. The new identity is STRICTER than the old on every row with a non-null \`period_start\`, and correctly permits distinct \`period_start\` values per \`(instrument_id, concept, unit, period_end, accession_number)\` — which is the SEC XBRL data model (a single filing may carry multiple period lengths ending on the same date for the same concept). No backfill / dedupe needed.

Only one app writer touches this table — [upsert_facts_for_instrument](app/services/fundamentals.py#L300). All other \`financial_facts_raw\` usage is read-only.

## Test plan
- [x] \`uv run ruff check .\`
- [x] \`uv run ruff format --check .\`
- [x] \`uv run pyright\`
- [x] \`uv run pytest\` (2336 passed, 1 skipped)
- [x] Applied migration locally; indexes now show \`uq_facts_raw_identity\` with the COALESCE expression
- [x] Codex review clean — idempotence safe on clean DBs, no other conflict path affected

### DIFF
--- a/sql/048_financial_facts_raw_identity_constraint.sql
+++ b/sql/048_financial_facts_raw_identity_constraint.sql
@@ -1,0 +1,54 @@
+-- Reconcile financial_facts_raw identity constraint with app ON CONFLICT.
+--
+-- Background
+-- ----------
+-- Migration 032 intended `uq_facts_raw_identity`, an expression UNIQUE
+-- index over (instrument_id, concept, unit, COALESCE(period_start, '0001-01-01'::date),
+-- period_end, accession_number). Some dev databases ended up with an
+-- auto-named plain UNIQUE table constraint over
+-- (instrument_id, concept, unit, period_end, accession_number) — five
+-- columns, no period_start, no COALESCE. Schema drift, not a mismatch
+-- the migration runner can detect (migrations are identified by
+-- filename, not by resulting DDL).
+--
+-- The app's ON CONFLICT clause in upsert_facts_for_instrument names
+-- the 6-col COALESCE identity. Against the 5-col plain UNIQUE that
+-- exists on the drifted DB, PostgreSQL raises:
+--     psycopg.errors.InvalidColumnReference:
+--         there is no unique or exclusion constraint matching
+--         the ON CONFLICT specification
+-- Every per-CIK XBRL upsert fails with that error.
+--
+-- Fix
+-- ---
+-- Drop the stale 5-col constraint (if present) and ensure the 6-col
+-- COALESCE expression index exists. Both operations are idempotent:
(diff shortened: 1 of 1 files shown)


Luke-Bradford:
**REBUTTED** — migration runner wraps each migration in one transaction already.

See [app/db/migrations.py:70-84](app/db/migrations.py#L70-L84):
\`\`\`python
with psycopg.connect(settings.database_url) as conn:
    try:
        with psycopg.ClientCursor(conn) as cur:
            cur.execute(sql)  # full migration body
            cur.execute("INSERT INTO schema_migrations (filename) VALUES (%s)", (path.name,))
        conn.commit()
    except Exception:
        conn.rollback()
        raise
\`\`\`

psycopg3 default (non-autocommit) begins an implicit transaction on the first statement. `cur.execute(sql)` runs both DDL statements in that transaction; a failure rolls back the DROP and the `schema_migrations` row is never inserted, so the migration retries on the next run. There is no window where the 5-col constraint is dropped without the 6-col index being in place, and the migration-tracking row only appears after `conn.commit()`.

No explicit BEGIN/COMMIT needed. PREVENTION not extracted — the existing runner already guarantees atomicity for every migration file, not a rule to add per-file.

Issue #435, opened 2026-04-24
feat: share-issuance + buyback dilution tracking


Luke-Bradford:
## Context
Follow-up from #426 + #430. User asked "how much per stock was issued over time" — that's not dividend DPS (#426 ships it), it's share-count change. Buybacks and issuance.

## Scope
Depends on #430 (expanded TRACKED_CONCEPTS). Once \`shares_issued_new\`, \`buyback_shares\`, \`buyback_spend\` land in financial_periods:

- View \`share_count_history\` — per-period delta = new_issuance - buyback_shares, running total = shares_outstanding trend.
- API: extend \`/instruments/{symbol}/financials\` or new \`/dilution\` endpoint.
- Instrument page panel: "Share count trend" line chart over N years. Net dilution % YoY badge.
- Ranking engine: quality sub-score penalises persistent net dilution > 2% YoY.

## Acceptance
- Instrument page shows shares-outstanding line + net-change-per-year bars.
- Low-dilution / buyback-heavy instruments score better on quality.

Issue #432, opened 2026-04-24
tracker: retire yfinance — replace each call site with official source


Luke-Bradford:
## Context
\`yfinance\` is an unofficial Yahoo scrape. No SLA, no rate guarantee. We use it as first-choice for description, sector/industry, employees, key stats (PE/PB/ROE/ROA/D2E/yield/growth), analyst estimates, major holders.

## Goal
Every yfinance call site either:
- Replaced with a computed-from-XBRL or official-free-source path, OR
- Explicitly annotated as "last-resort fallback for non-US instruments" with a follow-up ticket for that specific gap.

## Call sites (audit 2026-04-24)
\`grep -rn yfinance app --include='*.py'\` → 30+ hits. Key ones:

| Call | Today | Official replacement |
|---|---|---|
| \`get_profile\` description/sector/industry | yfinance first | #427 (instrument_sec_profile) + #428 (10-K Item 1) |
| \`get_profile\` market_cap | yfinance | Compute: \`shares_outstanding × latest_close\` |
| \`get_profile\` employees | yfinance | #430 DEI \`EntityNumberOfEmployees\` |
| \`get_key_stats\` pe/pb/roe/roa/d2e | yfinance merge path | Compute from \`financial_periods_ttm\` + \`quotes\` |
| \`get_key_stats\` dividend_yield | yfinance | \`instrument_dividend_summary.ttm_yield_pct\` (shipped #426) |
| \`get_key_stats\` revenue/earnings growth | yfinance | Compute from \`financial_periods\` YoY |
| \`get_analyst_estimates\` | yfinance | **Genuine gap** — no free official source. Options: Finnhub free tier (60 req/min), defer |
| \`get_major_holders\` | yfinance | **Partial gap** — SEC 13F covers institutions, Form 4 covers insiders (#429), aggregate pct still needs derive |
| \`get_dividends\` | yfinance | \`dividend_history\` view (shipped #426) |
| \`get_price_history\` | yfinance | Already eToro candles primary; yfinance was dev fallback |

## Deliverables
- Compute-from-XBRL helpers: \`compute_pe_ttm\`, \`compute_pb\`, \`compute_roe\`, \`compute_roa\`, \`compute_d2e\`, \`compute_revenue_growth_yoy\`.
- Remove yfinance from \`get_profile\` identity merge once #427 lands.
- Tracker on this ticket updated as each call site moves.
- Final state: yfinance imported zero times except behind an explicit \`fallback_only\` feature flag.

[...]

PR #444, opened 2026-04-24
feat(#432): compute-from-XBRL market_cap — retire yfinance first call site

Luke-Bradford:
## What
First swap in the #432 yfinance-retire ladder. Introduces \`app/services/xbrl_derived_stats.py\` and wires \`compute_market_cap\` into \`/instruments/{symbol}/summary\`. SEC-computed value preferred over yfinance when a share count is available.

## Pieces
- **Service** \`xbrl_derived_stats.compute_market_cap\` — joins \`instrument_share_count_latest\` (#435) × quote (COALESCE last → bid/ask midpoint). Returns \`MarketCap\` dataclass with value, shares, price, both as-of dates + source taxonomy.
- **Endpoint** \`get_instrument_summary\` — calls the helper after existing fundamentals fetch so cursor sequence stays last-position-additive. Falls through to \`profile.market_cap\` when computed is None.

## Test plan
- [x] 6 new service tests vs ebull_test (DEI/us-gaap source, empty inputs, zero guards, bid/ask midpoint fallback)
- [x] 3 existing summary tests updated with additional cursor slot (\`[None]\`) — keeps key-stats assertions intact
- [x] \`uv run pytest\` — 2457 passed
- [x] \`uv run ruff check\` + \`format --check\` + \`pyright\`

## Ladder status (#432)
- [x] market_cap
- [ ] pe_ttm / pb / roe / roa / debt_to_equity (next PR)
- [ ] revenue_growth_yoy / earnings_growth_yoy
- [ ] employees (DEI fact ready, needs service fn + swap)
- [ ] analyst_estimates (genuine gap — ticket body lists Finnhub free as option)

### DIFF
--- a/app/api/instruments.py
+++ b/app/api/instruments.py
@@ -1243,6 +1243,22 @@ def get_instrument_summary(
     if use_local_sec:
         local_fundamentals = _fetch_local_fundamentals(conn, instrument_id_int)
 
+    # #432 (yfinance retire, batch 1): prefer compute-from-XBRL market
+    # cap over yfinance when SEC has a fresh share count. Runs after
+    # the existing local-fundamentals block so its cursor slot is the
+    # LAST in the sequence — keeps the test harness shape stable.
+    # Returns None for non-US / pre-seed; we then fall back to
+    # whatever yfinance provided on the identity built above.
+    try:
+        from app.services.xbrl_derived_stats import compute_market_cap
+
+        computed_cap = compute_market_cap(conn, instrument_id=instrument_id_int)
+    except Exception:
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
Review resolved:

- **BLOCKING — priced CTE fan-out**: FIXED. Added `ORDER BY quoted_at DESC LIMIT 1` to the `priced` CTE with an inline comment documenting it as defensive belt against a future historical-quotes migration. `quotes` is still 1:1 by contract but the pin kills the fan-out vector.
- **PREVENTION — multi-row CTE fan-out**: rule extraction deferred to a follow-up skill edit; inline comment in this file carries the local guardrail.

Issue #437, opened 2026-04-24
meta: SEC/official data expansion — replace yfinance + richen instrument page + rankings


Luke-Bradford:
## Goal
Replace every yfinance data path with official-free sources (SEC EDGAR primarily, FRED/FINRA/BLS for macro), richen the instrument page with authentic data, and feed the ranking engine with new signals — all while keeping disk lean.

### Phase 1 — reclaim disk + baseline metadata
- [ ] #325 — flip \`raw_retention_dry_run=False\` + verify age-sweep deletes (reclaim ~31 GB)
- [ ] #436 — raw-file content-hash dedupe (drop timestamp-suffix variants)
- [ ] #427 — extract SEC entity metadata into \`instrument_sec_profile\`

### Phase 2 — richer facts + dividend UI
- [ ] #430 — expand \`TRACKED_CONCEPTS\` with DEI facts + missing us-gaap concepts
- [ ] #433 — frontend dividend panel + list filter (consumes #426)
- [ ] #435 — share-issuance + buyback dilution tracking (depends #430)

### Phase 3 — filings + events
- [ ] #428 — 10-K Item 1 business-summary extraction
- [ ] #431 — 8-K item[] code typing + thesis cascade trigger
- [ ] #434 — 8-K item 8.01 ex-date / record-date / pay-date calendar (depends #431)

### Phase 4 — sentiment + yfinance retire
- [ ] #429 — Form 4 insider transactions ingestion + widget
- [ ] #432 — tracker: retire yfinance (consumes #427 #428 #429 #430)

## Non-goals
- Paid data providers (FMP/Finnhub/AlphaVantage/Polygon) — reserved for genuine gaps (analyst estimates is the main one).
- Intra-day SEC polling (filings drop daily, handled by daily \`fundamentals_sync\`).
- LLM-driven extraction (10-K/8-K text) — phase-5 if regex parser coverage proves insufficient.

## Constraints carried across the ladder
- SEC EDGAR 10 rps shared throttle (already respected).
- No new raw-JSON-on-disk unless hash-deduped + retention-bounded.
- Parsed rows in DB are the durable artifact; raw payloads swept within retention window.

Closes the gaps raised in audits 2026-04-24 (#414 follow-up + yfinance-retire direction).

Luke-Bradford:
**Scope trim — meta tracker progress**

Phase status:
[...]

PR #426, opened 2026-04-24
feat: dividend history + TTM yield + has_dividend filter from SEC XBRL

Luke-Bradford:
## What
Surface dividend facts already captured via the daily SEC companyfacts path. No new HTTP. No new ingest. SEC EDGAR is free, unlimited, 10 rps shared throttle already respected.

1. **sql/050_dividend_history_views.sql** — two VIEWs on top of \`financial_periods\`:
   - \`dividend_history\` — long, one row per paying quarter (newest-first); excludes FY rows (would double-count Q1+Q2+Q3+Q4) and superseded rows.
   - \`instrument_dividend_summary\` — one row per instrument with \`ttm_dps\` / \`ttm_yield_pct\` / \`latest_dps\` / \`dividend_streak_q\` / \`has_dividend\`.
2. **\`app/services/dividends.py\`** — \`get_dividend_summary\` + \`get_dividend_history\` wrappers.
3. **GET \`/instruments/{symbol}/dividends\`** — summary + history. Empty shape (not 404) for never-payers.
4. **GET \`/instruments?has_dividend=true|false\`** — list filter.

## Why
User ask: \"filter on dividend-providing businesses, understand what the yield is, when it pays out, how much per stock was issued over time\". All derivable from \`us-gaap:CommonStockDividendsPerShareDeclared\` + \`us-gaap:PaymentsOfDividends\`, both already in \`financial_periods\`. Zero new dependency, zero rate-limit exposure.

## Codex review (pre-push)
HIGH findings addressed before push:
- Zero-dps rows were leaking into \`latest_*\` + history. Fixed: require strictly positive amount.
- Superseded rows were leaking. Fixed: every read filters \`superseded_at IS NULL\`.
- FY + Q4 on shared \`period_end_date\` double-counted. Fixed: history is quarterly-only; summary tie-breaks Q4 > FY.

## Test plan
- [x] \`uv run pytest\` — 2401 passed, 1 skipped
- [x] \`uv run ruff check .\` + \`format --check\`
- [x] \`uv run pyright\`
- [x] 10 service tests against ebull_test pin: never-paid → empty shape; TTM sums last 4 quarters; streak walks full history; aggregate-only filers still flag has_dividend; zero quarter excluded from history; superseded row excluded; FY row does not duplicate Q4.
- [x] API tests (mocked service) pin: 200 shape; unknown symbol 404; limit out-of-range 422; filter composes correct SQL.

## Scope boundary
[...]

Issue #428, opened 2026-04-24
feat: 10-K Item 1 business-summary extraction


Luke-Bradford:
## Context
Authentic company description lives in the 10-K \"Item 1. Business\" section. Today we show yfinance \`long_business_summary\` (~150-word Yahoo blurb). SEC Item 1 is multi-page, authoritative, free.

## Scope
- New provider \`SecFilingDocumentsProvider\` — fetch document from \`filing_events.primary_document_url\` (HTML or iXBRL).
- Parse \"Item 1. Business\" section, store first ~4 KB text to \`instrument_business_summary\` (instrument_id PK, body, accession_number, fetched_at).
- Refresh on each new 10-K via filing_events trigger.
- Instrument page wires description source to this table, yfinance fallback only if absent.

## Rate-limit posture
Same 10 rps shared SEC throttle. Triggered only on 10-K arrival (~quarterly per company), so bounded.

## Acceptance
- Instrument page shows authentic 10-K business description for US tickers with a 10-K on file.
- No yfinance call on the description path once the instrument has a row here.

PR #446, opened 2026-04-24
feat(#434): 8-K Item 8.01 dividend calendar — parser + ingester + banner

### LINKED ISSUE #434: feat: 8-K item 8.01 ex-date / record-date / pay-date calendar
## Context
Follow-up from #426 dividend history. SEC XBRL gives declared-amount per period but NOT the exact ex-date / record-date / pay-date calendar. Those live in 8-K item 8.01 announcement text (free-form English).

## Scope
- Parser for 8-K item 8.01 dividend announcements — regex + structured extraction for dates + amount.
- New table \`dividend_events\` (instrument_id, declaration_date, ex_date, record_date, pay_date, dps_declared, source_accession).
- API: \`GET /instruments/{symbol}/dividends\` extended with \`upcoming: [{ex_date, pay_date, dps}]\`.
- Instrument page: "Next dividend" banner when upcoming record exists.
- Ranking engine: optional dividend-capture signal (pre-ex-date).

## Depends on
#431 (8-K items[] typing) so we can filter to items=['8.01'] cheaply.

## Risk
Parsing free-form English is lossy. Acceptance bar: extract for ≥80% of Dividend Aristocrats; fall back to "next dividend date unknown" otherwise. Phase 2 could add OpenAI/LLM extraction.

Luke-Bradford:
## What
Parses 8-K Item 8.01 announcements into a new `dividend_events` table, exposes the calendar via `/instruments/{symbol}/dividends.upcoming`, and renders a \"Next dividend\" banner on the instrument page.

## Why
SEC XBRL (per #426) gives declared amounts but not the ex/record/pay calendar — that lives only in 8-K 8.01 free-form text. Issue #434 from the #437 meta ladder.

## Test plan
- [x] `uv run pytest tests/test_dividend_calendar.py tests/test_dividend_calendar_ingest.py tests/test_dividends_service.py` — 40 pass
- [x] `uv run pytest -q` — 2485 pass, 1 skip
- [x] `uv run ruff check .` + `ruff format --check .`
- [x] `pnpm --dir frontend typecheck` + `test:unit` — 384 pass
- [x] Codex checkpoint 2 (2 rounds, 4 findings addressed — partial-row re-parse TTL, backlog starvation via DESC ordering, declaration-only rows surfaced, banner moved outside has_dividend gate)

Closes #434

### DIFF
--- a/app/api/instruments.py
+++ b/app/api/instruments.py
@@ -880,10 +880,21 @@ class DividendSummaryModel(BaseModel):
     dividend_currency: str | None
 
 
+class UpcomingDividendModel(BaseModel):
+    source_accession: str
+    declaration_date: date | None
+    ex_date: date | None
+    record_date: date | None
+    pay_date: date | None
+    dps_declared: Decimal | None
+    currency: str
+
+
 class InstrumentDividends(BaseModel):
     symbol: str
     summary: DividendSummaryModel
     history: list[DividendPeriodModel]
+    upcoming: list[UpcomingDividendModel]
 
 
 @router.get("/{symbol}/dividends", response_model=InstrumentDividends)
@@ -902,7 +913,11 @@ def get_instrument_dividends(
 
     Default ``limit=40`` covers ten years of quarterly history.
     """
-    from app.services.dividends import get_dividend_history, get_dividend_summary
+    from app.services.dividends import (
+        get_dividend_history,
+        get_dividend_summary,
+        get_upcoming_dividends,
+    )
 
     symbol_clean = symbol.strip().upper()
     if not symbol_clean:
@@ -926,6 +941,7 @@ def get_instrument_dividends(
     instrument_id = int(inst_row["instrument_id"])  # type: ignore[arg-type]
     summary = get_dividend_summary(conn, instrument_id=instrument_id)
     history = get_dividend_history(conn, instrument_id=instrument_id, limit=limit)
+    upcoming = get_upcoming_dividends(conn, instrument_id=instrument_id)
 
     return InstrumentDividends(
         symbol=str(inst_row["symbol"]),  # type: ignore[arg-type]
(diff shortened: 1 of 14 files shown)


Luke-Bradford:
### Resolutions

- **[WARNING] 404-loop on withdrawn filings**: FIXED fc4ce7a. On both fetch_errors (exception or None body) and parse_misses the ingester now upserts a tombstone ``DividendAnnouncement`` (all fields None). The partial-row TTL on ``last_parsed_at`` caps re-fetch cadence at 7 days — weekly is enough to pick up a temporarily-moved URL or a regex improvement without burning daily budget. Added two integration tests (``test_non_dividend_8k_is_parse_miss_with_tombstone``, ``test_fetch_404_writes_tombstone``) that assert second-pass scan=0 + zero fetcher calls.
- **[PREVENTION] 404-loop on withdrawn filings**: the rule — "every ``fetch_errors += 1`` path either writes a DB sentinel or is bounded by a skip mechanism" — is now enforced by the tests above. If this pattern needs to generalise beyond the single ingester, I'll extract to the review-prevention-log, but at one call site a test is sufficient.

Luke-Bradford:
### Resolutions (round 2)

- **[BLOCKING] tombstone clobbers partial data on re-parse fetch errors**: FIXED 28c82de. Split tombstone into a dedicated `_upsert_tombstone` function. INSERT path still writes an all-NULL row on first attempt; UPDATE path now only bumps `last_parsed_at` via `ON CONFLICT DO UPDATE SET last_parsed_at = NOW()` and preserves every other column. A prior run's real data wins over a later transient-failure tombstone. Regression test `test_tombstone_preserves_existing_partial_data` covers the exact scenario you described (partial row + 8-day TTL advance + 404 on retry + assert dates intact).
- **[PREVENTION] tombstone clobbers partial data on re-parse fetch errors**: the new test IS the grep-and-verify prevention mechanism. Extracting the pattern to `docs/review-prevention-log.md` would generalise "destructive ON CONFLICT DO UPDATE SET on tombstone paths" — worth doing if a second call-site emerges; today #428 (10-K business summary, separate PR in flight) uses the same shape and will get the same fix pre-merge.

PR #497, opened 2026-04-25
refactor(#490): unify topic refs — drop pinned, ref-count everything

### LINKED ISSUE #490: refactor(#485): remove pinned-topic optimisation, unify under ref-counted topics
## Why

Current dynamic-subscribe design (#487) keeps held + watchlist instruments as a \`_pinned_topics\` set — always subscribed, never torn down by the page-view path. Operator observation: if the operator is looking at other instruments all day and never opens their portfolio pages, we're still getting ticks + writing quotes rows for 5-10 pinned instruments the UI never renders. Wasted eToro bandwidth + DB writes + log noise.

## Proposal

Unify pinning + dynamic into one ref-counted model:

- Every topic in \`_topic_refs\` has a ref count.
- Held + watchlist topics get a "source ref" from a background task that periodically (or on watchlist/position change) reconciles the known portfolio set vs \`_topic_refs\`.
- Page-view subscribers add their own ref on SSE stream open and drop it on close.
- A topic is subscribed to eToro iff \`refcount > 0\`.
- Remove \`_pinned_topics\` entirely.

Portfolio state changes trigger the same add/remove path as page views — one mechanism, not two.

## Trade-off

Without the pinning safety-net, a race or bug in the refresh loop could drop a held-position feed temporarily. Counterweight: held feed isn't latency-critical (reconciliation is REST-driven), and the operator's immediate point is that ticks flow for instruments the UI doesn't display.

## Done when

- [ ] \`_pinned_topics\` removed.
- [ ] One \`_topic_refs\` dict covers portfolio + watchlist + page-view sources.
- [ ] Portfolio-state add/remove runs on watchlist change, position open/close, and every N minutes (fallback).
- [ ] No change in user-visible live price behavior.

## Scope

Follow-up to #485 / #487. Not blocking anything.

Luke-Bradford:
## What

Replaces dual pinned + dynamic topic model with a single `_topic_refs` dict. Held + watchlist refs owned by a periodic `_source_reconcile_worker` (60 s) plus an immediate signal after private-channel portfolio reconcile. Page-view refs continue on the existing `add_instruments` / `remove_instruments` path.

## Why

Closes #490. Operator pain: WS streamed ticks for 5–10 held / watchlist instruments while the operator was browsing other tickers — wasted eToro bandwidth + DB writes + log noise. Ref-counted model means a topic is subscribed iff some source has a ref. No streaming for instruments nobody is currently looking at or holding.

## Test plan
- [x] `uv run ruff check .`
- [x] `uv run pyright`
- [x] `uv run pytest` (2735 passed, 1 skipped)
- [x] Codex review (initial pass found 3 issues, all fixed; re-review reported "no real findings")
- [ ] Live smoke: dev stack restart → confirm `EtoroWebSocketSubscriber: subscribed to N instrument topics` log + ticks flowing for held positions on /sse/quotes

### DIFF
--- a/app/services/etoro_websocket.py
+++ b/app/services/etoro_websocket.py
@@ -507,9 +507,10 @@ class EtoroWebSocketSubscriber:
         with contextlib.suppress(asyncio.CancelledError):
             await self._task
         self._task = None
-        # Cancel the source-reconcile worker first. It may be awaiting
-        # ``asyncio.to_thread`` on the watched-ids selector; the cancel
-        # raises CancelledError out of the await but lets any in-flight
+        # Cancel the source-reconcile worker before the private-event
+        # reconcile worker. It may be awaiting ``asyncio.to_thread``
+        # on the watched-ids selector; the cancel raises
+        # CancelledError out of the await but lets any in-flight
         # add/remove send finish first because both honour
         # CancelledError on their own awaits. No thread-completion
         # barrier needed — the source worker only does DB *reads*, so
--- a/app/services/etoro_websocket.py
+++ b/app/services/etoro_websocket.py
@@ -68,6 +68,15 @@ _RECONNECT_BACKOFF_S = 5.0
 # isn't hammered. 3 seconds is short enough that the operator sees
 # fresh state inside a "feel alive" window without churn.
 _RECONCILE_DEBOUNCE_S = 3.0
+# Periodic interval for the held-∪-watchlist refresh loop. Position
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
FIXED 1e61634 — comment in `stop()` rephrased: "Cancel the source-reconcile worker before the private-event reconcile worker." Accurate against actual ordering (`_task` cancelled in the preceding block, `_reconcile_worker_task` cancelled in the following block).

PR #500, opened 2026-04-25
refactor(#499): retire yfinance — eToro + SEC canonical, no third-party fallback

### LINKED ISSUE #499: bug: instrument page price uses yfinance as canonical — wrong for crypto, null for thin tickers
## Symptom

- \`/instrument/BTC\` shows USD 34.37 (Grayscale Bitcoin Mini Trust ETF price), not Bitcoin.
- \`/instrument/LRC\` shows no price at all (yfinance returns null for Loopring).

## Cause

\`app/api/instruments.py:1675\` (\`get_instrument_summary\`) builds the displayed price block from \`yfinance_provider.get_snapshot(row["symbol"])\`:

\`\`\`python
quote = snapshot.quote
...
price_block = InstrumentPrice(
    current=quote.price,
    ...
) if quote is not None else None
\`\`\`

yfinance is queried with the bare symbol — no exchange / asset-class disambiguation. For \`BTC\` yfinance returns the ETF nearest in name; for thin tickers it returns null. The eToro \`quotes\` table (which has the right answer when populated) is not consulted here.

Live ticks via \`useLiveQuote\` overlay the page once eToro pushes a snapshot frame, but only if (a) eToro WS is running, (b) Subscribe is sent for that instrument id, (c) eToro returns a snapshot for the id. For instruments outside what the operator has currently visible, the yfinance fallback is the operator-facing answer.

## Proposed fix

Two-layer:

1. **Prefer the local \`quotes\` table over yfinance** in \`get_instrument_summary\`. JOIN \`quotes q ON q.instrument_id = i.instrument_id\` and use that bid/ask/last when fresh. Fall back to yfinance only when the row is missing AND the instrument is in a region yfinance can plausibly serve (e.g. tradable US equities with non-suffixed symbols). Never use yfinance for crypto / non-US symbols.

2. **Show explicit "no price" rather than a wrong one** when neither \`quotes\` nor a confidently-resolved yfinance entry is available.

After the visibility-driven live-prices PR B (page-level \`LiveQuoteProvider\`), the operator's act of opening the page also triggers a fresh eToro Subscribe, so \`quotes\` will have a row for that instrument within ~1s for any instrument actually in eToro's WS feed. Combined with #1, the page settles on the correct price quickly.

## Acceptance

- \`/instrument/BTC\` shows the actual Bitcoin price (eToro USD), not the ETF.
- \`/instrument/LRC\` shows the live tick for Loopring (or "—" with a "no live feed" hint if eToro doesn't cover that id) — never a yfinance-ETF mistake.
- yfinance never returns the canonical price for any crypto or non-US-equity symbol.

Luke-Bradford:
## What

Retires yfinance from the instrument summary + financials API endpoints. eToro is canonical for prices (`quotes` table), SEC for fundamentals (`financial_periods`, `instrument_dividend_summary`). Fields without a canonical source return null rather than reaching for an unsanctioned provider.

## Why

Operator surfaced two visible bugs that traced to yfinance still being wired into `get_instrument_summary`:

- `/instrument/BTC` showed USD 34.37 (Grayscale Bitcoin Mini Trust ETF price), not Bitcoin.
- `/instrument/LRC` showed no price at all (yfinance can't resolve Loopring).

`docs/settled-decisions.md` names eToro / SEC / Companies House / FMP as the only sanctioned sources. yfinance was leftover from earlier provider experiments. This PR aligns the code with the settled decision.

Closes #499. Subsumes #432's per-call-site retire (audit table reduces to "no remaining call sites" once this lands).

## Changes

Backend:
- `get_instrument_summary` reads price from `quotes`, identity from local DB, market cap from SEC × quotes, key stats from SEC + dividend summary.
- `get_instrument_financials` drops yfinance fallback; empty SEC = `source="unavailable"`.
- Deleted `app/providers/implementations/yfinance_provider.py`.
- Removed `yfinance>=1.3.0` from `pyproject.toml`.

Frontend:
- `KeyStatsFieldSource` drops `"yfinance"`; `InstrumentFinancials.source` drops `"yfinance"`.
- `ResearchTab` provenance tag drops `"yfinance"` colour case.
- `InstrumentPage` empty-state copy no longer mentions yfinance.

Tests:
- Rewrote `tests/test_api_instrument_summary.py` + `tests/test_api_instrument_financials.py` against the new shape.
- Deleted `tests/test_yfinance_provider.py`.
- Net: 567 insertions, 2446 deletions across 11 files.

## Codex review

Round 1 raised 3 findings, all fixed:
1. Dividend yield was fetched but couldn't surface without SEC fundamentals — gate now includes dividend yield as a trigger.
2. Test coverage regression on `?id=` override + symbol-mismatch 404 — added regression tests.
3. Test coverage regression on balance/cashflow statement dispatch — added regression tests.

## Test plan
- [x] `uv run ruff check .`
- [x] `uv run pyright`
- [x] `uv run pytest` (2687 + 1 skipped — yfinance-provider tests removed, summary + financials tests rewritten)
- [x] `pnpm typecheck`
- [x] `pnpm test:unit` (400/400)
- [ ] Live smoke (operator, after merge):
  - `/instrument/BTC` shows the eToro live tick (or "—" if eToro doesn't cover) — never the ETF price
  - `/instrument/LRC` shows the eToro live tick (or "—") — never silently null from yfinance miss
  - `/instrument/AAPL` (well-covered ticker) renders identity from local DB, price from quotes, key stats from SEC


Luke-Bradford:
FIXED cf1c6c1 — added `sec_dividend_summary` to the KeyStatsFieldSource literal (backend + frontend mirror); `_pick` takes a per-call source so dividend yield is tagged `sec_dividend_summary`; block-level `key_stats` source distinguishes `sec_xbrl` / `sec_dividend_summary` / `sec_xbrl+sec_dividend_summary`. Prevention asserts added to both `test_sec_path_populates_key_stats` (per-field) and `test_dividend_only_partial_stats_surface` (block-level).

PR #504, opened 2026-04-25
fix(#503): handle eToro's actual WS envelope shape — messages: [...]

Luke-Bradford:
## What

Critical parser bug. `parse_rate_message` + `is_private_event` expected a top-level `{type, data}` shape, but eToro's documented WS push frame is `{messages: [...]}` outer wrap carrying inner messages of the form `{topic, content (json-string), id, type}`. See https://api-portal.etoro.com/api-reference/websocket/topics.md.

The parser silently returned `None` on every real frame. The `quotes` table looked populated only because Phase 2 of `fx_rates_refresh` was writing rows via REST for Tier 1/2 instruments. Tier 3 instruments (BTC, LRC, …) had no Phase 2 coverage and silently never ticked. PR #502 dropped Phase 2 entirely → operator-visible: BTC and LRC instrument pages render no live price.

## Why

Operator: "BTC + LRC pages still show no price." Investigated via Postman MCP + WebFetch against the official docs. Documented frame shape is different from what our code expected.

## Test plan
- [x] `uv run ruff check app/services/etoro_websocket.py tests/test_etoro_websocket.py`
- [x] `uv run pyright app/services/etoro_websocket.py tests/test_etoro_websocket.py`
- [ ] `uv run pytest` — running in CI (local Bash backgrounding broke output capture; trusting CI for verification)
- [ ] Live smoke (operator):
  - Open `/instrument/BTC` — see live tick replace the "—"
  - Open `/instrument/LRC` — same
  - Watch backend log for `EtoroWebSocketSubscriber: subscribe N topics` then ticks flowing into `quotes` (any visible position should now update its quotes row within ~1s of the page mounting)

## Notes

Back-compat preserved: legacy `{type, data}` shape still parses so existing test fixtures pass without rewrite.

### DIFF
--- a/app/services/etoro_websocket.py
+++ b/app/services/etoro_websocket.py
@@ -169,52 +169,85 @@ _PRIVATE_EVENT_PREFIXES: tuple[str, ...] = (
     "Trading.Credit",
 )
 
-
-def is_private_event(raw: str) -> bool:
-    """True if ``raw`` is a private-channel push that should trigger
-    a portfolio reconcile. Returns False for malformed JSON, non-
-    private types, or unknown shapes — the reconciler is a coarse
-    invalidation, not a precise event handler."""
-    try:
-        msg = json.loads(raw)
-    except json.JSONDecodeError:
-        return False
-    if not isinstance(msg, dict):
-        return False
-    msg_type = msg.get("type")
-    if not isinstance(msg_type, str):
-        return False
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
FIXED 8fe5f67 — dropped the early `continue` in `_listen`. Now schedules reconcile if any inner message is a private event AND publishes every rate tick in the same frame. Added `TestMixedBatchFrame` prevention test asserting both `is_private_event` and `parse_rate_messages` surface from the same mixed-batch frame.

PR #524, opened 2026-04-26
fix(#515-pr0): include crypto/fx/commodity/index in T3 candle bootstrap

Luke-Bradford:
## What

PR 0 of #515. T3 candle bootstrap now qualifies crypto / fx / commodity / index instruments without requiring a `fundamentals_snapshot` row that those asset classes never produce.

## Why

Operator-visible: BTC, LRC, ETH instrument pages rendered "no price data". Diagnosis showed zero `price_daily` rows for every crypto despite tier-3 + tradable status. Root cause: `daily_candle_refresh` T3 query required `EXISTS fundamentals_snapshot`, which is an equity-only heuristic.

## Test plan

- [x] Diagnosed dev DB state — confirmed zero `price_daily` for BTC/LRC/ETH; AAPL has 403 rows
- [x] `uv run ruff check .` / `ruff format --check` / `pyright` — clean
- [x] `uv run pytest tests/test_daily_candle_refresh.py tests/test_daily_candle_refresh_crypto_eligibility.py` — 14 passed (6 new integration tests)
- [x] `uv run pytest tests/smoke/test_app_boots.py` — passed
- [x] Codex review — 2 rounds, no remaining findings

### DIFF
--- a/app/workers/scheduler.py
+++ b/app/workers/scheduler.py
@@ -924,11 +924,45 @@ def nightly_universe_sync() -> None:
             tracker.row_count = row_count
 
 
+# Max T3 instruments to include in candle refresh for bootstrap
+# scoring. Prevents hitting API rate limits while giving enough T3
+# instruments price data to enable T3→T2 promotion via the
+# scoring/coverage pipeline.
 _T3_BOOTSTRAP_BATCH_SIZE = 200
-"""Max T3 instruments to include in candle refresh for bootstrap scoring.
 
-Prevents hitting API rate limits while giving enough T3 instruments price
-data to enable T3→T2 promotion via the scoring/coverage pipeline.
+# T3 candle bootstrap eligibility query.
+# Module-level constant so the test suite imports the same SQL the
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
## Round 1 review resolution

### WARNING — FIXED 8bda893

Hardened `_cleanup` (renamed from `_cleanup_exchange`) to take both `exchange_ids` and `instrument_ids` and explicitly DELETE from price_daily / coverage / instruments / exchanges. All four call sites updated. Parametrized cases use distinct ids (fx → 950201, commodity → 950202, index → 950203) keyed off a module-level dict.

### PREVENTION — EXTRACTED test pattern

The wider lesson: any helper named `_cleanup_X` that only handles one foreign-key tier of a multi-table seed is a foot-gun. PR 4 of #503 had the same shape (`_cleanup` for exchanges only) and got away with it because the seed was single-table. Going forward: cleanup helpers either accept the full set of seeded ids (this PR's pattern), or live as dedicated fixtures with autouse=True.

Issue #536, opened 2026-04-26
feat: event-driven 10-K parse trigger via filing_events insert hook


Luke-Bradford:
## What

Replace daily polling for 10-K Item 1 / 8-K events / dividend / Form 4 ingests with event-driven dispatch: when a new `filing_events` row lands of type 10-K, 8-K, etc., enqueue an immediate parse for that one filing. Steady-state cron stays as a safety net (catches anything missed).

## Why

Operator wants "Bloomberg-fresh" data without heavy polling. Today every downstream parser runs on a fixed cron and re-scans the same candidate set repeatedly. Event-driven dispatch:
- Cuts steady-state poll cost ≈ 90%
- Cuts fresh-filing → parsed-narrative latency from ~24h to ~minutes
- Each parser only runs when it has work
- Same code path as bootstrap (just ad-hoc instead of cron)

## Implementation options

Two viable patterns — Codex pick:

**Option A: Postgres LISTEN/NOTIFY**
- Trigger on `filing_events` insert: `pg_notify('filing_event_inserted', NEW.filing_event_id::text)`
- Long-running listener subscribed to channel; dispatches parsers based on filing_type.
- Pros: native Postgres, no extra infra. Cons: listener crash → events lost (unless we also poll).

**Option B: Outbox table + cron tick**
- Insert into `filing_event_outbox` (filing_event_id, claimed_at NULL)
- Existing cron drains outbox every N minutes (e.g. 1 minute)
- Pros: durable, no daemon. Cons: 1-min latency floor.

## Acceptance

- [ ] One pattern chosen + documented (Codex consultation)
- [ ] 10-K, 10-K/A, 8-K, Form 4 wired to event dispatch
- [ ] Daily catch-up cron remains as safety net (sweeps anything missed)
- [ ] Latency from `filing_events` insert → parser persistence ≤ 2 min on dev DB
- [ ] Pre-push gates green

## Depends on

Failure-reason taxonomy + backoff (so event-driven retries don't fight steady-state retries)

Luke-Bradford:
Deferred per Codex review post #532-#535 merge.

Current state:
- 10-K Item 1: daily cron 03:15 UTC (15 min after SEC filings ingest at 03:00) — effective latency ~few minutes for new filings.
- Bootstrap drain mode (#535) handles first-time fill + post-outage catch-up.
- 8-K events + Form 4 insider already hourly.
- Failure-reason taxonomy + backoff (#533) prevents accumulation.

[...]

Issue #535, opened 2026-04-26
feat: bootstrap-mode 10-K Item 1 backfill (one-shot drain, no daily limit)


Luke-Bradford:
## What

Add a bootstrap-mode entry point to `sec_business_summary_ingest` that drains the entire candidate set in one chunked session. Honours SEC 10 req/s fair-use limit. Run once on demand to clear the 3568-instrument first-time backlog; daily cron continues to handle steady-state delta.

## Why

Current daily cron has `limit=200` (per [scheduler.py:3163](app/workers/scheduler.py#L3163) and [business_summary.py:765](app/services/business_summary.py#L765)). Designed for steady-state quarterly delta. Not designed for first-time backfill of 4031 SEC-CIK instruments. Bloomberg-style philosophy: bootstrap once to known-good state, maintain incrementally.

## Implementation sketch

```python
def sec_business_summary_bootstrap(*, max_runtime_minutes: int = 60) -> None:
    """One-shot drain. No per-run limit. Stops on max_runtime or empty queue."""
    deadline = time.monotonic() + max_runtime_minutes * 60
    while time.monotonic() < deadline:
        result = ingest_business_summaries(conn, fetcher, limit=500)
        if result.filings_scanned == 0:
            break
        # SEC fair-use breather between chunks
        time.sleep(0.1)
```

Wire as a separate APScheduler one-shot job that's manually triggered (or admin-API-triggered). Idempotent: safe to re-run.

## Acceptance

- [ ] `sec_business_summary_bootstrap` function added
- [ ] Admin endpoint `POST /admin/jobs/business-summary-bootstrap` triggers it (auth-gated)
- [ ] Honours SEC fair-use rate limit (delegate to global rate-limit budget if that ticket lands first)
- [ ] First-time run on dev DB drains 3568 candidates
- [ ] Pre-push gates green

## Depends on

Failure-reason taxonomy + backoff ticket (so quarantined rows aren't re-fetched during the drain)

Luke-Bradford:
Closing as already done.

Bootstrap drain mode shipped via PR #545. Entry point at [`app/services/business_summary.py:bootstrap_business_summaries`](app/services/business_summary.py#L1313). Loops `ingest_business_summaries` with `chunk_limit` until the candidate query returns 0 or the deadline elapses; designed for the 4031-instrument SEC-CIK universe under SEC fair-use limits.

Tests at `tests/test_business_summary_ingest.py::TestBootstrapDrain`.

Issue #534, opened 2026-04-26
feat: 10-K/A → plain 10-K fallback for business_summary ingester


Luke-Bradford:
## What

When the candidate-selection picks a `10-K/A` (amendment) as the latest filing for an instrument and parsing returns `no_item_1_marker` (Part III amendments don't carry Item 1), fall back to the most-recent prior plain `10-K` from the same instrument. Today: tombstone written, weekly retry loops forever.

## Why

167 instruments currently in this bucket (verified via DB query during #515 post-merge review). All 167 have a prior plain 10-K with Item 1 available — cheap win.

## Implementation sketch

In [`ingest_business_summaries`](app/services/business_summary.py#L761), after `extract_business_section` returns None on a 10-K/A:

```python
if accession_filing_type == '10-K/A' and body is None:
    fallback = _latest_plain_10k(conn, instrument_id, before=accession_filing_date)
    if fallback is not None:
        # Re-fetch + re-parse against fallback. If it succeeds,
        # store with source_accession = fallback's accession so
        # the next run sees the actual narrative.
        ...
```

Requires the candidate query to also surface `filing_type` so the ingester knows when to attempt the fallback path.

## Acceptance

- [ ] 10-K/A latest with no Item 1 + prior plain 10-K → ingester picks plain 10-K, persists narrative, does NOT tombstone the instrument
- [ ] If plain 10-K also fails, normal failure-reason taxonomy applies (depends on backoff ticket)
- [ ] Test covers: (a) 10-K/A no-Item-1 with fallback success, (b) no fallback available, (c) fallback also fails
- [ ] Pre-push gates green

## Depends on

Failure-reason taxonomy ticket (so `no_item_1_marker` reason exists to trigger fallback)

Luke-Bradford:
Already shipped: `_find_prior_plain_10k` at `app/services/business_summary.py:1267` and the fallback call at L1535 within `ingest_business_summaries`. `no_item_1_marker` is in `_RETRY_REASONS` and the ingester re-attempts against the prior plain 10-K. Discovered while picking up the ticket — no follow-up needed.

Issue #533, opened 2026-04-26
feat: failure-reason taxonomy + exponential backoff + quarantine for sec_business_summary_ingest


Luke-Bradford:
## What

Currently [`record_parse_attempt`](app/services/business_summary.py#L606) writes a tombstone with `body=''` and `last_parsed_at=NOW()`. Candidate query excludes for 7 days, then re-fetches forever. No failure category, no attempt counter, no backoff. Hopeless cases (e.g. 10-K/A Part III amendments with no Item 1) consume ~42 fetch slots every weekly cycle indefinitely.

## Why

Operator can't tell weekly amendment cycling from a real parser bug. Diagnosed during #515 post-merge review:
- Job runs daily, processes 200, ~158 succeed, ~42 silently miss
- ~3568 backlog hides unprocessed (queue) vs hopeless (real misses) — same tombstone shape for both
- Backlog stayed flat for days because failures cycled through limit slots

## Scope

1. **Schema:** add columns to `instrument_business_summary`: `attempt_count INT NOT NULL DEFAULT 0`, `last_failure_reason TEXT NULL`, `next_retry_at TIMESTAMPTZ NULL`. Migration adds + backfills NULL safe.
2. **Failure-reason taxonomy:** enum-like Python Literal with values: `fetch_http_5xx`, `fetch_http_4xx`, `fetch_timeout`, `no_item_1_marker`, `body_too_short`, `parse_exception`, `upsert_exception`.
3. **Backoff schedule:** retry 1d → 7d → 30d → quarantine (365d after 4th fail). Computed from `attempt_count`.
4. **Candidate query:** include `AND (next_retry_at IS NULL OR next_retry_at <= NOW())` so quarantined rows fall out.
5. **Admin visibility (small):** `GET /admin/business-summary-failures` listing the most-attempted/most-recent failures with reason breakdown. Operator can un-quarantine via `POST /admin/business-summary-failures/{instrument_id}/reset`.

## Acceptance

- [ ] Migration adds new columns; existing rows get `attempt_count=1, next_retry_at=last_parsed_at + 7d` so behaviour is unchanged on first run
- [ ] `record_parse_attempt` accepts a reason kwarg; persists it
- [ ] Backoff schedule applied on every miss
- [ ] After 4 misses on same accession, row enters quarantine (next_retry_at = NOW() + 365 days)
- [ ] Admin endpoint + simple page with reason histogram
- [ ] Resolver tests cover backoff math + quarantine boundary
[...]

PR #543, opened 2026-04-26
feat(#533): failure taxonomy + backoff + quarantine for business_summary

Luke-Bradford:
## What

Pre-#533 the ingester cycled hopeless cases weekly forever. Now each failure carries a category + attempt counter + scheduled retry. Hopeless cases escalate from 1d → 7d → 30d → 365d quarantine instead of consuming capacity indefinitely.

- **Migration 074**: \`attempt_count\`, \`last_failure_reason\`, \`next_retry_at\` columns + partial index. Backfills existing tombstones with attempt_count=1.
- **Service**: \`FailureReason\` taxonomy (9 values). \`record_parse_attempt(reason=...)\` mandatory kwarg. \`_classify_fetch_exception\` maps httpx errors to taxonomy. Candidate query gates on \`next_retry_at\` instead of 7-day TTL.
- **Admin endpoints**: \`GET /admin/business-summary-failures\` (histogram + paginated detail). \`POST .../reset\` clears tracking + sets next_retry_at=NOW() so ingester re-queues. 404 when row has no failure.
- **Tests**: 6 backoff/quarantine pins + 3 admin shape pins + 1 end-to-end reset→re-ingest integration.

## Why

Diagnosed during #515 post-merge review: 10-K Item 1 backfill stalled at 598/4031 instruments because ~42 known-failing accessions cycled the daily 200-slot capacity without ever escalating. Foundation for #534 (10-K/A fallback), #535 (bootstrap drain), #536 (event-driven trigger).

## Test plan

- [x] \`uv run ruff check . / format --check / pyright\` — clean
- [x] \`uv run pytest\` — 2788 passed (+11 since #532, 1 pre-existing migration-066 deselected)
- [x] Codex review: 2 rounds, both findings addressed (reset re-queue, healthy-row 404 guard)

### DIFF
--- a/app/api/business_summary_admin.py
+++ b/app/api/business_summary_admin.py
@@ -27,6 +27,7 @@ from typing import Literal
 
 import psycopg
 import psycopg.rows
+import psycopg.sql
 from fastapi import APIRouter, Depends, HTTPException, Query
 from pydantic import BaseModel
 
@@ -118,7 +119,8 @@ def list_failures(
 
     with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
         cur.execute(
-            f"""
+            psycopg.sql.SQL(
+                f"""
             SELECT bs.last_failure_reason AS reason,
                    COUNT(*) AS count,
                    COUNT(*) FILTER (WHERE bs.attempt_count >= 4) AS quarantined_count
@@ -126,7 +128,8 @@ def list_failures(
              WHERE {where_sql}
              GROUP BY bs.last_failure_reason
              ORDER BY count DESC
-            """,
+            """
+            ),
             params,
         )
         for r in cur.fetchall():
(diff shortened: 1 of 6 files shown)


Luke-Bradford:
Reply to review bot's repeated BLOCKING finding on the upsert_exception path:

The `conn.rollback()` at line 1037 IS present and pre-dates the new `record_parse_attempt` call. The bot reads only the additions in the diff and misses pre-existing context lines.

Codex confirmed the ordering is correct: rollback (line 1037) → record_parse_attempt (line 1043) → commit (line 1049). Connection state is cleared before the next SQL statement.

Just landed a regression test (`TestFailureBackoffAndQuarantine::test_upsert_exception_path_rolls_back_then_records_failure`) that forces the path by raising psycopg.errors.DataError from a synthetic upsert and asserting the failure row persists. The test would fail with InFailedSqlTransaction if the rollback were missing.

WARNING about psycopg.sql.SQL(f-string) was addressed in commit 79b22c7 — replaced with proper psycopg.sql.SQL.format(where=Composed) composition.

Issue #567, opened 2026-04-27
tech-debt(#559): density grid refinement v2 + data-viz layer


Luke-Bradford:
## Context

Operator review of the Phase 3 density grid (post #566 merge) flagged several refinement items + a broader product direction toward a data-visualisation layer.

## Refinement (immediate, after Phase 4)

1. **FilingsPane shows wrong filing types.** Currently calls \`fetchFilings(id, 0, 5)\` which returns the latest 5 filings of ANY type — Form 4 + 144 dominate (insider trades + restricted sale notices). The pane should filter to 8-K + 10-K only. Either:
   - Add a \`type\` query param to \`/filings/{id}\` and call with \`type=8-K,10-K\`, OR
   - Add a sibling endpoint \`/filings/{id}/significant\` returning only the user-visible filing types.
2. **Add 'See all filings' link** at the bottom of FilingsPane → \`/instrument/:symbol?tab=filings\`.
3. **Box content fit.** Several panes show partial scroll boxes (Recent filings list scrolls, Insider activity scrolls, etc.). Should fit on screen at typical desktop viewports without internal scroll. Strategies:
   - Cap rows tighter (3-4 instead of 5) where verbose.
   - Use \`text-overflow: ellipsis\` on long subject lines.
   - Auto-rows tuned per pane (taller rows for chart, shorter for stats).
4. **Yahoo-Finance benchmark.** Operator wants the \"full high-level picture\" without partial scrolls. Reference layout: yahoo finance per-stock summary view.
5. **Drop the transitional dividends + insider combined card span** — both are smaller now; see if they fit cleanly inside the right column rather than spanning full width.

## Data-visualisation layer (broader)

Operator wants visual analysis where the data supports it. Charts framework needed.

1. **Library choice.** Existing: \`recharts\`/\`lightweight-charts\` for price chart. Pick ONE chart library for fundamentals + ownership (likely \`recharts\` — more declarative, smaller per-chart bundle).
2. **Time-series charts.** Revenue / cost / profit / debt over time (multi-period from \`financial_periods\`). Render as small multiples (one chart per metric) or single chart with metric toggle.
3. **Ownership pie chart.** Insiders / institutions / public float breakdown. Data source TBD — \`13F\` filings + insider \`Form 4\` aggregates.
4. **Where to place the charts.** Either:
   - In existing \`Financials\` tab as a top section above the statement table.
   - On the density grid as a new pane (replaces or supplements the company-profile pane).
[...]

PR #572, opened 2026-04-27
refactor(#567-B): density grid sizing + spacing

Luke-Bradford:
## What

- `lg:auto-rows-[220px]` removed — rows are now content-driven.
- `overflow-auto` removed from every pane wrapper — no more internal scrollbars.
- Chart pane gets explicit `min-h-[440px]` to retain its 2-row footprint.
- Pane padding tightened to `px-3 py-2.5` (asymmetric); grid gap tightened to `gap-2`.

## Why

Phase B of #567 — operator flagged scroll-box noise. Yahoo-Finance-style fit-on-screen presentation.

## Test plan

- [ ] Vitest assertion: no descendant has `overflow-auto`.
- [ ] Manual: load `/instrument/GME` at 1440x900, verify no internal scrollbars on any pane.


Luke-Bradford:
**FIXED 2369dd6** — Added second test variant with  capability active; asserts exactly 1 `.overflow-auto` element (the intentional scroll-bound on the combined card). DensityGrid now has 5 tests covering both paths.

**PREVENTION: EXTRACTED** — Adding rule to `docs/review-prevention-log.md`: when a conditional branch adds a class in one render path but not another, add a test variant for the active path and assert the exact count to prevent silent regression in follow-up phases.

PR #573, opened 2026-04-27
feat(#567-C): Fundamentals sparkline pane

Luke-Bradford:
## What

- `Sparkline` hand-coded SVG component (no new chart dep).
- `FundamentalsPane`: 4 sparklines for Revenue / Op income / Net income / Total debt over the latest 8 quarters.
- Period-safe join keyed on `(period_end, period_type)` between income + balance statements so all four metrics plot the same quarter set.
- Gated on `summary.capabilities.fundamentals.providers.sec_xbrl.data_present === true`.
- Inserted into the density grid as a full-width row after the row-2 SecProfile/Filings panes (preserves existing layout).

## Why

Phase C of #567 — operator wants Bloomberg-tier visual analysis. Fundamentals over time is the highest-signal chart we can ship today.

## Test plan

- [ ] Vitest: 3 Sparkline tests (polyline shape, empty path, default stroke).
- [ ] Vitest: 3 FundamentalsPane tests (gating off, gating on, total-debt computation).
- [ ] Vitest: DensityGrid renders the pane only when capability active.
- [ ] Manual: load `/instrument/GME`, confirm 4 sparklines visible; load a non-SEC instrument and confirm pane is absent.


Luke-Bradford:
**WARNING (FundamentalsPane footer link)**: FIXED 09d297c — moved the 'View statements →' Link inside the resolved-data `<>` fragment so it only renders when `series.length >= 2`.

**NITPICK (Sparkline flat-line)**: FIXED 09d297c — changed `range === 0` case to center at `height / 2` instead of clipping to the bottom boundary.

**PREVENTION (navigation link outside data guard)**: EXTRACTED docs/review-prevention-log.md — added entry 'Navigation link outside data guard in async pane' (commit 09d297c).

Issue #575, opened 2026-04-27
tech-debt(#567): instrument detail polish round 2 — empty panes, layout rebalance, pane-header standardization


Luke-Bradford:
## Context

Follow-up to #567. PRs #571-#574 shipped phases A-D of the refinement spec (filings filter, grid sizing, fundamentals sparklines, insider summary). Operator review of the rendered page (see screenshots in chat 2026-04-27) flags remaining gaps + new direction validated by Codex second-opinion.

Current state: `frontend/src/pages/InstrumentPage.tsx` → `ResearchTab.tsx` → `DensityGrid.tsx` (`grid-cols-[2fr_1fr_1fr]`).

## Pain points (from screenshots + code read)

1. **Empty panes still render full empty-state cards.**
   - Thesis with no memo → full `EmptyState` "No thesis yet" — wastes ~25% of right column.
   - Dividends with no history → full card. Worse, wrapped in `overflow-auto max-h-[360px]` (`DensityGrid.tsx:111-124`) — internal scroll box on empty content.
   - News block (`ResearchTab.tsx:201-205`) literally renders placeholder text "News tab still has the full feed." → vestigial pane.
2. **Combined dividends+insider card** still alive (`DensityGrid.tsx:111`) despite #567 acceptance criterion #5 (drop transitional combined card).
3. **Grid imbalance.** Filings (high-signal, deep data) cramped to 1fr; insider activity full-width with only 5 stat fields (inverse density). Layout doesn't match content density.
4. **No standard pane header.** Some panes show source provenance, some don't. No consistent drill affordance ("Open →") across panes — only filings + fundamentals have footer links.
5. **Layout doesn't adapt to capability coverage.** GME (full SEC) renders fine; UK equity (Companies House only) and crypto (no filings/fundamentals) get layout holes.

## Codex second-opinion direction (validated 2026-04-27)

> "The main risk in your current proposal is not that any single idea is bad; it is that you are adding too many interaction layers before proving the simpler page-plus-route model is insufficient."

Stay incremental. Don't add drawer/sheet primitive, verdict ribbon, or universal L3-raw-data contract. Page-plus-route is the proven model.

## Scope (single PR, may split into 2 if it grows)

1. **Hide empty panes entirely.**
   - Thesis: hidden when `thesis === null`. CTA stays in `SummaryStrip`.
   - Dividends: pane removed entirely when no history; kills the `max-h-[360px]` scroll wrapper.
[...]

PR #579, opened 2026-04-27
fix(#575): FundamentalsPane returns null on insufficient series

Luke-Bradford:
## What

`FundamentalsPane` returns `null` when SEC XBRL capability is active but the joined income+balance series has fewer than 2 quarters. Previously rendered a full `<EmptyState title="Not enough fundamentals history">` card, violating the polish round-2 four-state rule.

## Why

Operator review of post-#575 page flagged IEP rendering the full empty card (screenshot 2026-04-27). Same fix pattern as #575's `ThesisPane` / `RecentNewsPane` / `DividendsPanel` empty-state suppression. Loading + error states still render the Pane chrome so retry affordance stays visible.

## Test plan

- 2 new vitest cases (zero rows + single-quarter) assert `container.firstChild === null`
- `pnpm --dir frontend test:unit` (457 pass)
- `pnpm --dir frontend typecheck` clean
- `uv run ruff check . && uv run ruff format --check .` clean

### DIFF
--- a/frontend/src/components/instrument/FundamentalsPane.test.tsx
+++ b/frontend/src/components/instrument/FundamentalsPane.test.tsx
@@ -183,4 +183,70 @@ describe("FundamentalsPane", () => {
     // Latest total debt = 150 + 30 = 180
     expect(await screen.findByText(/180/)).toBeInTheDocument();
   });
+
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
FIXED 3bd7f9f — replaced both setTimeout(r, 30) gates with waitFor. PREVENTION acknowledged in commit message. Re-running review on the new commit.

Issue #603, opened 2026-04-27
data: bump price_daily default backfill from 400 to 1830 days


Luke-Bradford:
## Context

Surfaced via operator review on #587: the 5Y range button shows much less than 5Y of bars for many instruments. Root cause is the default \`lookback_days=400\` in \`app/services/market_data.py:55\` — backfill mode pulls 400 daily bars (~16 months); incremental mode then only appends ~3 bars/day forward. So an instrument seeded N months ago has approximately \`min(400, N * 30)\` bars.

5Y view is independently confirmed as the agreed range token in the new range table (#600 maps it to OneWeek interval), but the underlying daily series in \`price_daily\` still needs to be deep enough to back the longer ranges in absence of the intraday endpoint, and to power scoring / thesis features that rely on multi-year daily history.

## What

1. Bump \`lookback_days\` default in \`refresh_market_data\` from 400 → 1830 (~5y).
2. One-shot rebackfill for all Tier 1/2 instruments — single eToro call per instrument, ~5x more rows. Total budget at 60 GET/min: ~10 minutes for 600 instruments.
3. \`_candles_fetch_count\` keeps incremental mode at the same low count for daily ticks; only the initial-or-gap-detected backfill changes.

## Acceptance

- New default in code; existing tests updated where they assert the literal 400
- One-shot rebackfill script committed (or job-runnable), invocation documented
- After rebackfill, 5Y view shows ≥1200 daily bars for instruments that have been in the universe for ≥5y (most blue chips); fewer for newer listings (correct — IPO date is the floor)
- Storage check: \`price_daily\` row count rises ~5x; expected ~3 GB → ~15 GB at scale (acceptable, room is plentiful)

## Out of scope

- Compression / partitioning of \`price_daily\` — separate concern, not blocking
- Intraday data (#599)

## Linked

Parent: #585 (visible polish surfaces this)
Related: #599 (intraday makes daily-only-fallback less critical, but daily is still canonical for scoring)

Luke-Bradford:
Closed by #605 (squash-merged).

Issue #602, opened 2026-04-27
feat(#585): live last-bar updates via existing SSE quote stream


Luke-Bradford:
## Context

Even with intraday bars from #599/#600, the chart's last bar is still stale during market hours unless ticks update it live. eBull already has a working WebSocket → SSE pipeline (\`app/services/etoro_websocket.py\`, \`app/services/quote_stream.py\`, \`app/api/sse_quotes.py\`) that pushes \`Trading.Instrument.Rate\` ticks (Ask/Bid/LastExecution) to subscribed UI panels. Topic ref-counting already supports per-page subscribes (#485). This ticket connects that stream to the chart's last bar.

## What

1. PriceChart + ChartWorkspaceCanvas open an SSE \`/sse/quotes?ids=\${instrumentId}\` subscription on mount; close on unmount.
2. On each tick:
   - Compute the \"current bar\" timestamp as \`floor(now / interval)\` aligned to the chart's interval
   - Update the rendered series via lightweight-charts \`series.update({time, open, high, low, close})\`. \`open\` = last bar's open (kept in a ref); \`high\`/\`low\` = max/min of all ticks since open; \`close\` = latest \`tick.last\`
   - Volume series: \`update\` is harder for cumulative volume; defer to V2 — V1 just leaves volume unchanged for the in-progress bar
3. When the wall clock crosses an interval boundary, the next tick opens a new bar — reset open/high/low to the new tick price.
4. Visual signal that data is live: small green dot near the live last value (purely cosmetic).

## Acceptance

- Open chart with market open → last candle's close updates in <1s of an eToro tick
- Tab close → SSE connection closes, server-side ref-count decrements
- Network drop → SSE reconnects automatically (EventSource native), chart continues from snapshot on reconnect
- No regression in static chart rendering (intraday + daily both work)
- Unit tests: tick aggregation logic (open/high/low/close updates) is a pure function — easy to test
- Integration test: mocked SSE EventSource pushes a tick, assert series.update was called with the right shape

## Out of scope

- Volume aggregation for the in-progress bar (defer to V2 — volume often arrives via separate aggregate tick at bar close)
- Tick-by-tick cumulative volume bar
- Historical tick replay

## Dependencies

- #600 should land first (chart consumes the right interval); not a hard block, but UX is messier without intraday

## Linked

Parent: #585
Predecessor: #600
Related: #485 (existing SSE per-page subscribes), #274 (WS implementation)

Luke-Bradford:
[...]

Issue #601, opened 2026-04-27
feat(#585): frontend range→interval wiring + 1D/5D/YTD ranges


Luke-Bradford:
## Context

Frontend half of the intraday work. Predecessor #599 (backend intraday-candle proxy) ships the new endpoint and range→interval mapping. This ticket flips PriceChart + ChartWorkspaceCanvas + ChartPage + RawOhlcvTable to use the new endpoint and adds the new range buttons.

## What

1. Update \`CandleRange\` type to: \`1d\` · \`5d\` · \`1m\` · \`3m\` · \`6m\` · \`ytd\` · \`1y\` · \`5y\` · \`max\`. Drops \`1w\` (replaced by \`5d\`).
2. Update backend \`Literal\` and \`_CANDLE_RANGE_DAYS\` to match.
3. Update RANGES button list in PriceChart and ChartPage.
4. \`fetchInstrumentCandles\` switches to the new endpoint with \`{ range, interval? }\` shape; default interval comes from the table in #599.
5. ChartPage adds an explicit interval-override dropdown for power users.
6. Tests: snapshot the new range button set, prove default interval is correctly forwarded, prove user-set interval URL param survives the round-trip.

## Acceptance

- 9 range buttons render: 1D · 5D · 1M · 3M · 6M · YTD · 1Y · 5Y · MAX
- Each range hits the new endpoint with the agreed default interval
- ChartPage interval dropdown overrides the default and URL-syncs
- All existing tests pass (will need adjustments for the renamed/added ranges)
- No regressions in chart rendering — same lightweight-charts contract, just finer data

## Out of scope

- Live last-bar updates (#601)
- Backend endpoint itself (#599)
- Daily backfill depth bump

## Dependencies

- #599 must merge first

## Linked

Parent: #585
Predecessor: #599
Successor: #601 (live last-bar)

Luke-Bradford:
Closed by #606 (squash-merged).

Issue #600, opened 2026-04-27
feat(#585): backend intraday candle proxy + range→interval mapping


Luke-Bradford:
## Context

Predecessor #587 polished the PriceChart UI but kept the existing range list (1W/1M/3M/6M/1Y/5Y/MAX), all served by daily candles from \`price_daily\`. Operator wants TradingView/Robinhood-style intraday: 1D shows minute bars, 1W shows 5-minute bars, etc. eToro's REST candle endpoint is already parameterized for interval (\`/history/candles/{direction}/{interval}/{count}\`); provider currently hardcodes \`OneDay\`.

## What

Add an intraday-aware backend endpoint that pass-through-fetches from eToro at the requested interval, caches in-process with short TTL, and writes nothing to the database. Daily candles continue to go through the existing endpoint and \`price_daily\` table — those remain canonical for scoring/thesis/audit.

1. Extend eToro provider:
   - New \`get_candles(instrument_id, interval: Interval, count: int)\` returning \`list[OHLCVBar]\`
   - \`Interval\` is a Literal: \`OneMinute\` · \`FiveMinutes\` · \`FifteenMinutes\` · \`ThirtyMinutes\` · \`OneHour\` · \`FourHours\` · \`OneDay\` · \`OneWeek\` · \`OneMonth\` (verify the full set against eToro OpenAPI)
   - Rate-limit budget: chart open = 1 GET (60/min/key gives plenty)
2. New endpoint \`GET /instruments/{symbol}/intraday-candles?range=...&interval=...\`:
   - \`range\` is the public token: \`1d\` · \`5d\` · \`1m\` · \`3m\` · \`6m\` · \`ytd\` · \`1y\` · \`5y\` · \`max\`
   - \`interval\` optional override; if absent, server picks from agreed mapping table:

   | Range | Interval | ~bars | Lookback |
   |---|---|---|---|
   | 1d | OneMinute | 390 | 1 trading day |
   | 5d | FiveMinutes | 390 | 5 trading days |
   | 1m | ThirtyMinutes | 330 | 22 trading days |
   | 3m | OneHour | 525 | 65 trading days |
   | 6m | FourHours | 390 | 130 trading days |
   | ytd | OneDay | varies | Jan 1 → today |
   | 1y | OneDay | 252 | 365 days |
   | 5y | OneWeek | 260 | 1825 days |
   | max | OneMonth | 120–360 | full history |
[...]

PR #604, opened 2026-04-27
feat(#600): backend intraday candle proxy + cache + auth

Luke-Bradford:
## What

- New `GET /instruments/{symbol}/intraday-candles?interval=...&count=...` — provider-backed pass-through for sub-day OHLCV, no DB persistence.
- `IntradayBar` + `IntradayInterval` Literal added to the provider interface (sub-day tokens only: 1m / 5m / 10m / 15m / 30m / 1h / 4h).
- `EtoroMarketDataProvider.get_intraday_candles` hits the same URL family as the daily endpoint with a variable interval slot. New `_normalise_intraday_candle` preserves the timestamp.
- `app/services/intraday_candles.py` — bounded LRU + TTL cache (30s for ≤15m, 180s for ≥30m) with per-key singleflight so concurrent misses collapse to one provider call.
- Endpoint maps eToro 429 → 503 with `Retry-After`; 5xx / network → 502; missing creds → 503. Auth-gated via `require_session_or_service_token` to match other quota-consuming surfaces.

## Why

Operator-requested intraday charts (#585 epic). Long-horizon-fund posture rules out persisting another candle table — pass-through cache satisfies the chart UX without bloating the scoring/thesis data surface.

## Conscious deviations

- Endpoint lives on the existing `/instruments` router instead of a dedicated live-data router. Carve-out documented in the module docstring; keeps the URL space tied to the symbol resource. Codex pre-push agreed this is acceptable once the auth gate was added.
- Daily / weekly / monthly intervals deliberately **excluded** from `IntradayInterval`. They would let callers bypass `price_daily`. The frontend's range table (#601) sends 1y/5y/max through the existing daily endpoint; only short ranges (1d/5d/1m/3m/6m) reach the proxy.
- Frontend wiring is NOT in this PR — separate ticket #601. This PR ships the backend contract only so #601 can review against a stable shape.

## Test plan

- [x] `uv run ruff check .` / `ruff format --check .` / `pyright` clean
- [x] `uv run pytest -q` — 2856 passed (was 2829 pre-#600)
- [x] New: cache TTL boundary, LRU eviction, singleflight collapses concurrent misses to one provider call
- [x] New: endpoint happy path, unknown symbol → 404, invalid interval → 422, count cap → 422, missing creds → 503, 429 → 503+Retry-After, 5xx → 502, second request within TTL skips provider
- [x] New: `_normalise_intraday_candle` valid / with-offset / missing-fields / malformed-timestamp paths
- [x] Codex pre-push review: dropped daily-tier intervals from the Literal, added singleflight, added auth gate

## Linked

Parent: #585
Predecessors landed: #586 (theme + recharts), #587 (PriceChart polish)
Successor: #601 (frontend range→interval wiring), #602 (live last-bar via SSE)
Related: #603 (daily backfill 1830d, runs in parallel)

### DIFF
--- a/app/api/instruments.py
+++ b/app/api/instruments.py
@@ -6,7 +6,17 @@ Reads from:
   - coverage             (1:1 coverage tier per instrument)
   - external_identifiers (1:N provider-native identifiers per instrument)
 
-No writes. No schema changes.
+No writes (DB-side). No schema changes.
+
+**Carve-out — intraday candles (#600).** The
+``GET /instruments/{symbol}/intraday-candles`` endpoint is a
+provider-backed pass-through: it loads eToro broker credentials,
+calls the live eToro REST endpoint via ``EtoroMarketDataProvider``,
+and serves bars through an in-process TTL cache (no DB persistence).
+This is the one endpoint in this module that consumes external API
+quota and writes an audit row per request (via
+``load_credential_for_provider_use``). All other endpoints stay
+DB-only.
 """
 
 from __future__ import annotations
@@ -14,14 +24,29 @@ from __future__ import annotations
 import logging
 from datetime import date, datetime
 from decimal import Decimal, InvalidOperation
-from typing import Literal
+from typing import Literal, get_args
 
+import httpx
 import psycopg
 import psycopg.rows
 from fastapi import APIRouter, Depends, HTTPException, Query
 from pydantic import BaseModel
 
+from app.api.auth import require_session_or_service_token
+from app.config import settings
 from app.db import get_conn
+from app.providers.implementations.etoro import EtoroMarketDataProvider
+from app.providers.market_data import IntradayInterval
+from app.services.broker_credentials import (
+    CredentialNotFound,
+    load_credential_for_provider_use,
+)
+from app.services.intraday_candles import fetch_intraday_candles
+from app.services.operators import (
+    AmbiguousOperatorError,
+    NoOperatorError,
+    sole_operator_id,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


Luke-Bradford:
## Review responses

### [BLOCKING] raw-payload persistence — REBUTTED + EXTRACTED `docs/review-prevention-log.md`

The premise is wrong: `response.json()` feeding a normaliser is not, by itself, a raw-payload-persistence violation. The cited prevention entries (Raw payload persistence #544, Every structured field lands in SQL #902, Provider body-text fetches require SQL #951) all describe the **structured-fields-land-in-SQL** rule for trade-impacting data: daily candles to `price_daily`, SEC XBRL to fundamentals tables, body text to dedicated tables. The audit trail is the SQL row, not a disk dump (per the scope-narrowing in #471).

Intraday candles are explicitly ephemeral chart-UI data — no scoring / thesis / recommendation / order / tax dependency. Persisting them would expand the audit surface without analytical value, and the no-persistence design was locked in epic #585's body and confirmed by Codex pre-implementation. Codex re-confirmed the rebuttal at checkpoint-3 just now.

The real issue Codex surfaced: the provider docstring still asserted "every structured field lands in SQL" with no acknowledgement that #600 introduces an exception — a doc/rule inconsistency, not a missing-persistence violation. Fixed in 76559512be47e28ed007df83c4fe3448b989010b:

- `EtoroMarketDataProvider` class docstring names the carve-out explicitly, points at the design ticket (#585), names which SQL invariants don't apply, and states adding further carve-outs requires reopening the design.
- `docs/review-prevention-log.md` "Raw payload persistence" entry gets a Carve-out clause naming the conditions under which a provider method may skip persistence: (a) class-docstring-named carve-out tied to a locked design ticket, (b) process-local data, (c) API-layer auth gate when consuming external quota.

**Resolution:** REBUTTED (no SQL persistence added) + EXTRACTED `docs/review-prevention-log.md` (carve-out clause), `app/providers/implementations/etoro.py` (class docstring).

### [WARNING] `count` echoes request, not actual rows — FIXED 76559512be47e28ed007df83c4fe3448b989010b

Now returns `len(bars)` so callers see what they actually got. Test updated to assert that `count == len(rows)` even when the provider returns fewer bars than requested.

**Resolution:** FIXED 76559512be47e28ed007df83c4fe3448b989010b.

### [PREVENTION] CI grep on `response.json()` in providers — REBUTTED

Premise is wrong (see BLOCKING above) so the prevention rule is wrong too. A `response.json()` grep in `app/providers/implementations/` would block all legitimate ephemeral pass-through endpoints — every existing provider method already calls `response.json()`. Codex agreed this rule would be a bad rule.

**Resolution:** REBUTTED — the prevention is replaced by the carve-out clause now codified in `docs/review-prevention-log.md`, which constrains future carve-outs without false-positive-flagging legitimate ones.

---

Codex checkpoint-3 was run on the rebuttals before this commit; outputs confirmed (a) the BLOCKING premise is unsound as stated, (b) the WARNING is real and worth fixing, (c) the doc inconsistency was the actual gap and is now closed.

Issue #587, opened 2026-04-27
feat(#585): PriceChart overview polish — rich tooltip + line/area toggle + log scale + drill discoverability


Luke-Bradford:
## Context

Second sub-ticket of #585. Ships after #586 (chart theme + recharts) which adds the dark theme PriceChart consumes.

Spec: `docs/superpowers/specs/2026-04-27-instrument-charts-quant-redesign-design.md`.

Operator complaint (2026-04-27): "Price chart, sure, we need to make these nicer, add more items and make it generally more useful, work with the colour scheme of the page, go from candle to line etc". Workspace at `/instrument/:symbol/chart` already has this richness; backport the operator-visible essentials to the overview pane.

## What

`frontend/src/components/instrument/PriceChart.tsx`:

1. **Rich hover tooltip** — backport `RichTooltip` pattern from `ChartWorkspaceCanvas` (date, OHLC, volume, %change-from-prior). Replaces the minimal date+close tooltip.
2. **Chart type toggle**: candle (default) / line / area. URL-synced via `?type=line|area` (default candle has no param). Three small buttons next to the range picker.
3. **Log scale toggle**: linear (default) / log. Useful for long-range views. URL-synced via `?scale=log`.
4. **Whole-card-click drill**: clicking anywhere on the PriceChart pane (NOT the range buttons or chart controls) navigates to `/instrument/:symbol/chart` workspace. Currently only the small Open → button drills. Make hover-elevate visible so card looks clickable.
5. **Apply chartTheme** from #586 (dark background, grid, text).

## Acceptance

- Hovering over candles shows OHLCV + %change tooltip
- Type toggle switches between candle / line / area; URL persists
- Log scale toggle works; URL persists
- Clicking the chart card (away from controls) routes to `/instrument/:symbol/chart`, preserving range param (matches #581's `?chart=` → `?range=` translation pattern)
- Card hover state shows elevation/border color so operator sees it's clickable
- Dark theme applied — no white card on dark slate page
- All existing tests pass; new tests for type toggle URL sync, log toggle URL sync, card-click navigation

## Out of scope

- Indicators (SMA/EMA) on overview — they live in workspace only (#582). Don't bloat overview.
- Compare tickers, regression, range channel — workspace only (#583)
- Raw OHLCV — workspace only (#584)

## Dependencies

- #586 (chart theme + recharts) must merge first

## Linked

Parent: #585

Luke-Bradford:
Closed by #599 (squash-merged).

Issue #586, opened 2026-04-27
feat(#585): chart theme + recharts dep (Phase 1 foundation)


Luke-Bradford:
## Context

First sub-ticket of #585 (chart redesign epic). Foundational — unblocks all drill-page work.

Spec: `docs/superpowers/specs/2026-04-27-instrument-charts-quant-redesign-design.md`. Read it for the locked library + theme decisions.

## What

1. Add `recharts` to `frontend/package.json` (latest v2.x — pin minor, allow patch)
2. Create `frontend/src/lib/chartTheme.ts` with the constants from the spec (dark slate palette: `bg=#0f172a`, `gridLine=#334155`, `axisText=#94a3b8`, `up=#10b981`, `down=#ef4444`, accent rotation `[#06b6d4, #3b82f6, #a855f7, #f59e0b, #ec4899, #84cc16]`)
3. Apply theme to existing `lightweight-charts` consumers:
   - `frontend/src/components/instrument/PriceChart.tsx`
   - `frontend/src/pages/components/ChartWorkspaceCanvas.tsx`
   Pass theme into `createChart(...)` options (background, grid, textColor, scale borders, candle up/down).
4. Create a small `<RechartsThemeProvider>` or theme constants object that recharts components consume in subsequent tickets — but DON'T add any recharts component yet. Foundational only.
5. Update `Sparkline.tsx` to consume `chartTheme.accent` for default stroke color.

## Acceptance

- `recharts` appears in `frontend/package.json` dependencies
- `frontend/src/lib/chartTheme.ts` exports the constants
- Existing PriceChart + ChartWorkspaceCanvas now render on dark backgrounds matching page shell
- No new charts yet (defer to #SUB-2 etc)
- `pnpm --dir frontend build` doesn't increase bundle size for the overview page (recharts must be tree-shaken away when not imported on a route)
- All existing tests pass; no new chart features

## Out of scope

- Any actual drill page (separate tickets)
- PriceChart overview polish — deferred to next sub-ticket (line/area toggle, rich tooltip)
- Site-wide color/typography pass — tracked at #559

## Dependencies

None. Can start immediately.

## Linked

Parent: #585

Luke-Bradford:
Closed by #597 (squash-merged).

Issue #585, opened 2026-04-27
epic(#567): instrument-page chart redesign — quant L2 drill pages + theme


Luke-Bradford:
## Parent epic for chart redesign work

Spec: `docs/superpowers/specs/2026-04-27-instrument-charts-quant-redesign-design.md` — read this FIRST before any sub-ticket. Locks library / theme / drill-contract / quant-vocabulary decisions.

## Settled (do not re-debate)

- **Library:** `recharts` for non-OHLC viz (declarative React JSX, ~96KB gz, MIT, lazy-loaded per drill page). `lightweight-charts` stays for OHLC. Hand-rolled SVG for one-offs.
- **Theme:** `frontend/src/lib/chartTheme.ts` is the single source of truth. Light-mode in v1; dark variant + theme-mode switch tracked at #596.
- **Drill contract (continues from #576):** L1 = compact pane on instrument page, L2 = `/instrument/:symbol/<domain>` route for rich charts, L3 = raw data table + CSV export inside L2 via `?view=raw`.
- **Quant vocabulary:** margins, growth, DuPont, drawdown, insider clusters, sector radar — long-horizon equity research, NOT day-trader TA.
- **Range → interval mapping (vanilla TradingView/Robinhood):** 1d→OneMinute · 5d→FiveMinutes · 1m→ThirtyMinutes · 3m→OneHour · 6m→FourHours · ytd→OneDay · 1y→OneDay · 5y→OneWeek · max→OneMonth. Daily stays canonical for scoring/thesis; intraday is on-demand REST + cache, no DB persistence.

### Phase 1: Foundation
- [x] #586 — `feat: chart theme + recharts dep` (merged via PR #597)
- [x] #587 — `feat: PriceChart overview polish` — rich tooltip + type/scale toggles + curved line + drill (merged via PR #599)

### Phase 1.5: Intraday architecture (operator-requested 2026-04-27)
- [ ] #600 — `feat: backend intraday candle proxy + range→interval mapping` — REST pass-through + TTL cache, no DB writes
- [ ] #601 — `feat: frontend range→interval wiring + 1D/5D/YTD ranges` (depends on #600)
- [ ] #602 — `feat: live last-bar updates via existing SSE quote stream` (depends on #601)
- [ ] #603 — `data: bump price_daily default backfill from 400 to 1830 days` (independent, can ship alongside)

### Phase 2: Highest-impact drill pages
[...]

Issue #485, opened 2026-04-25
feat: subscribe-on-page-view for live quotes (currently only watchlist + held flow)


Luke-Bradford:
## Symptom

Operator opens an instrument page (e.g. LRC / Loopring). Page shows no live price, doesn't update. WS subscriber logs:

\`\`\`
INFO subscribed to 5 instrument topics
INFO subscribed to private channel
\`\`\`

But the 5 are held + watchlist (BBBY/IEP/GME/QQQ/VOO). LRC isn't in either set, so eToro WS never sends rates for it; SSE has nothing to deliver.

## Cause

\`fetch_watched_instrument_ids\` returns only \`broker_positions ∪ watchlist\`. The WS subscriber subscribes to that set on connect and on each reconnect. Page-view doesn't influence the set.

## Proposals

1. **Periodic refresh** — re-fetch watched set every 60s, send delta SUBSCRIBE / UNSUBSCRIBE frames. Operator adding an instrument to watchlist sees ticks within the minute. Doesn't help drive-by page views.

2. **Page-view as watchlist-touch** — opening an instrument page implicitly adds it to a short-TTL viewing set (e.g. 10 min) that merges into the WS subscription. Trade-off: wasted WS topics for instruments the operator glanced at; benefit: every page is alive instantly.

3. **Manual \"watch live\" button** — explicit toggle on each instrument page. Less magic but more friction.

4. **Subscribe everything in Tier 1** — broader subscription, no per-page logic. Trade-off: many topics, may exceed eToro per-session topic limit (unknown bound).

Recommend (2): page-view → transient watch set + 60s periodic refresh of WS topic set covering held ∪ watchlist ∪ transient.

## Done when

- [ ] Opening any instrument page receives live ticks within ~10s
- [ ] Closing the page (or 10 min idle) drops the subscription
- [ ] Periodic refresh picks up watchlist add/remove without reconnect

## Scope

Standalone follow-up to #274. Pairs with display-coverage for non-US instruments (separate ticket).

PR #607, opened 2026-04-27
feat(#602): live last-bar via existing SSE quote stream

Luke-Bradford:
## What

- New `lib/useLiveLastBar.ts` — pure `aggregateTick` function + hook that subscribes to the existing `LiveQuoteProvider` SSE stream and feeds each tick through a bucket-aware aggregator.
- Bucket-aware update logic: tick in current bucket → preserve historical/live open, extend H/L, set C; tick in next bucket → append fresh bar; stale tick → skip.
- `lib/chartData.ts`: new `INTERVAL_SECONDS` / `intervalSecondsFor(range)` / `floorToBucket` helpers.
- `PriceChart` + `ChartWorkspaceCanvas` accept `instrumentId` + `range`, anchor against the LAST RENDERED bar (not raw rows tail), and render a green pulsing "LIVE" indicator when SSE is connected. Workspace disables live updates in compare mode.
- `InstrumentPage` + `ChartPage` wrap their bodies in `LiveQuoteProvider` with the instrument id in the visible-id list. The existing `SummaryStrip` consumer + the new chart consumer share one SSE handshake per page.

## Why

Operator-requested live-updating charts during market hours. The existing `/sse/quotes` pipeline (#274 / #485 / #501) already streams ticks. #602 is purely a frontend wiring on top of that — no backend changes.

## Conscious deviations / scope notes

- **Volume aggregation is V1-scoped out** — eToro's quote stream carries no per-tick volume, and synthesising it would diverge from the canonical daily volume in `price_daily`. Volume bars stay where the historical fetch left them.
- Codex pre-push review applied across three rounds:
  1. **HIGH** — first tick no longer rewrites the historical bar's open. `aggregateTick` now seeds OHLC from `histLastBar` when the bucket matches.
  2. **MEDIUM** — switched from per-component `useLiveQuote` to `useLiveTick` from `LiveQuoteProvider` so the chart reuses the page's existing handshake (previously opened a second SSE alongside `SummaryStrip`).
  3. **MEDIUM** — both callers anchor against the cleaned/rendered last bar, not the raw rows tail. Null-OHLC tail rows can no longer make ticks bucket against an invisible bar.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm test:unit` — 577/577 (new: `aggregateTick` pure-function suite covering same-bucket update + open preservation, append on bucket boundary, stale-tick skip, no-history append, live-bar-open wins, low/high extension)
- [x] `pnpm build` clean
- [x] Backend gates: ruff / format / pyright clean; pytest 2858 passed (no Python touched)

## Linked

- Parent: #585 (chart redesign epic)
- Predecessors: #586, #587, #600, #603, #601 — all merged
- Related: #274 / #485 / #501 (existing SSE quote pipeline)

### DIFF
--- a/frontend/src/components/instrument/DensityGrid.tsx
+++ b/frontend/src/components/instrument/DensityGrid.tsx
@@ -78,7 +78,7 @@ export function DensityGrid({
   // hover/zoom.
   const ChartPane = (
     <Pane title="Price chart" onExpand={drillToWorkspace}>
-      <PriceChart symbol={symbol} />
+      <PriceChart symbol={symbol} instrumentId={instrumentId} />
     </Pane>
   );
 
--- a/frontend/src/components/instrument/PriceChart.tsx
+++ b/frontend/src/components/instrument/PriceChart.tsx
@@ -514,20 +514,14 @@ export function ChartCanvas({
     } as unknown as Parameters<ReturnType<IChartApi["timeScale"]>["applyOptions"]>[0]);
   }, [intraday]);
 
-  // Feed data on every rows change. lightweight-charts replaces the
-  // series wholesale via setData — no incremental diffing needed.
-  useEffect(() => {
-    const candle = candleRef.current;
-    const line = lineRef.current;
-    const area = areaRef.current;
-    const volume = volumeRef.current;
-    const chart = chartRef.current;
-    if (!candle || !line || !area || !volume || !chart) return;
-
-    // Pre-convert to numeric bars so downstream `setData` calls work
-    // with guaranteed-non-null values (no dead `?? 0` fallbacks). Rows
-    // that fail any numeric parse are dropped here.
-    const clean: NumericBar[] = rows.flatMap((r) => {
+  // Numeric / null-filtered rows. Computed during render (not in an
+  // effect) so values are available to the live-tick aggregator's
+  // historical anchor on the very first render. Otherwise `histLastBar`
+  // would always be null until React next re-rendered, and live ticks
+  // would all bucket as "no anchor" → fresh bars, never extending the
+  // historical last candle.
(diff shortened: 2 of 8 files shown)


Luke-Bradford:
## Review responses

### [WARNING] inline refs object causes effect re-fire — FIXED 1f43e155ccaa53d2b9c8f676f8c311f17b6f1bb5

The hook now accepts the lightweight-charts `MutableRefObject`s themselves (stable identity across renders) instead of `.current` snapshots. The effect reads `.current` inside, and deps drop `refs` because the contained refs are stable. `PriceChart` and `ChartWorkspaceCanvas` pass `candleRef` / `lineRef` / `areaRef` directly. No more re-applying the last tick on every parent re-render.

### [PREVENTION] inline-object in useEffect deps lint rule — DEFERRED

Sound idea, but ESLint's existing react-hooks/exhaustive-deps rule already nags on inline objects in deps; a stricter custom rule would be a separate ergonomics ticket. Closing as REBUTTED for this PR — the typed `MutableRefObject` interface in `LiveLastBarRefs` documents the contract going forward, and the comment in the hook spells out why an inline-snapshot would be wrong.

577/577 tests still pass.

Luke-Bradford:
## Review responses (round 2)

### [BLOCKING] historicalLastBar inline literal — FIXED 4441cf885f59644d9a5ff28aa651cba8a9e58358

Right — the previous fix moved `refs` to stable identity but introduced the same problem at the next level via `historicalLastBar`. Wrapped `histLastBar` in `useMemo` keyed on primitive OHLC fields at both call sites (`PriceChart.tsx` and `ChartWorkspaceCanvas.tsx`). Identity now changes only when data actually changes.

577/577 tests still pass.

Issue #609, opened 2026-04-27
tech-debt(#602): exchange-calendar-driven session shading + events page


Luke-Bradford:
## Context

While shipping intraday chart session shading (#602), we discovered:

1. **eToro instrument metadata does NOT expose trading hours.** Captured fields per `_normalise_instrument`: `instrumentID, symbolFull, instrumentDisplayName, exchangeID, stocksIndustryId, instrumentTypeID, instrumentTypeName, isInternalInstrument`. No `tradingHours`, `extendedHoursEligible`, `sessionType`, or `schedule` field.

2. **Current `classifyUsSession` hardcodes NYSE/NASDAQ** (PM 04:00–09:30, RTH 09:30–16:00, AH 16:00–20:00 ET). This is wrong for:
   - Non-US listings (LSE, Euronext, Xetra — different windows)
   - 24/5 instruments (commodities, FX — no AH/PM concept)
   - US half-day holidays (Black Friday, Christmas Eve — early close at 13:00 ET, AH brought forward)
   - Full-day market holidays (mid-week closures)

3. **eToro's intraday feed has per-instrument coverage gaps.** GME demo missing Fri 16:00–20:00 ET (entire AH session) on 2026-04-24. Likely systemic — can't predict from metadata which instruments will be sparse.

### Phase A — Exchange-driven session shading
- Pull authoritative trading calendars per exchange. Candidate sources:
  - `pandas_market_calendars` (Python, MIT) — NYSE, NASDAQ, LSE, Euronext, Xetra, TSE, etc., handles half-days + holidays
  - Alternative: small in-house static table per exchange + JSON holiday list
- Refactor `classifyUsSession` → `classifySession(instrument, epoch)` keyed by `instruments.exchange`
- Drop hardcoded NYSE assumption; `SessionBands` becomes exchange-aware
- Half-day handling: bands shrink to early close, AH window shifts forward

### Phase B — Calendar-of-events page
- New page surfacing for portfolio + watchlist instruments:
  - **Today**: which markets are open / closed / half-day
  - **This week**: market-holiday calendar
  - **Upcoming**: earnings, ex-dividend, 10-K/10-Q filings (we already ingest these — surface them by date instead of by instrument)
  - Configurable filter: portfolio-only / watchlist-only / all
- Builds on existing data: `instruments`, `holdings`, `watchlist`, `eight_k_events`, `dividends`, `fundamentals_quarterly`
- New piece: exchange-holiday table populated from the calendar source above

[...]

PR #610, opened 2026-04-27
fix(#602): chart freshness, session shading + axis polish

Luke-Bradford:
## What
- `SessionBands` overlay marks pre-market (sky), after-hours (amber), closed/overnight (slate-grey). Snaps to NYSE clock ticks.
- PM + AH visibility toggles via `?pm=0` / `?ah=0` URL params; live ticks gated via `acceptPre`/`acceptAh` on `useLiveLastBar`.
- Wired into both `PriceChart` (instrument page) and `ChartWorkspaceCanvas` (workspace).
- `rightOffset: 5` + `fitContent` once-per-range so 60s polling no longer flushes operator pan/zoom or re-anchors rightmost axis tick to the live bar.
- `useLiveLastBar` dedupes on `quoted_at` — toggling session deps no longer replays the last tick.
- Live-bucket fed into SessionBands so an appended bar gets tinted immediately, no waiting for next REST refetch.
- Backend `_debug_ws.py`: `etoro-candles-probe` accepts `interval` param; new `etoro-instrument-raw` for diagnosing instrument metadata coverage.
- Files #609 as tech-debt for exchange-aware schedules + calendar/events page (current `classifyUsSession` is hardcoded NYSE; doesn't catch half-day holidays or non-US listings).

## Why
- Operator feedback during chart redesign: needed AH/PM context, axis stability, tint to distinguish overnight bars from RTH on a glance.
- Codex checkpoint 2 caught: dead `findPreviousRthClose` (removed), band tint lag behind appended live bars (fixed via live-bucket feed), tick replay on session-toggle (fixed via dedupe key).

## Test plan
- [x] `uv run ruff check .` + `uv run ruff format --check .`
- [x] `uv run pyright`
- [x] `pnpm --dir frontend test:unit` — 599/599 pass (15 new boundary tests for `classifyUsSession`)
- [ ] Refresh instrument page in browser — bands render on first paint, no toggle click required
- [ ] Toggle PM/AH off — bars + bands disappear; live ticks in hidden session don't append
- [ ] Hover live AH bar — band tint extends with the candle (no 60s lag)

### DIFF
--- a/app/api/_debug_ws.py
+++ b/app/api/_debug_ws.py
@@ -1,0 +1,59 @@
+"""TEMPORARY debug endpoint for eToro WS subscriber state (#602 live ticks).
+
+Surfaces the bits we need to diagnose "SSE connected but no ticks
+arriving":
+  * Is the WS connection currently open?
+  * Which instrument topics has add_instruments() registered refs for?
+  * Is the subscriber task running, or has it died?
+
+Remove this file once the live-tick path is confirmed healthy on
+this environment.
+"""
+
+from __future__ import annotations
+
+from fastapi import APIRouter, Request
+from pydantic import BaseModel
+
+router = APIRouter(prefix="/_debug", tags=["debug"])
+
+
+class EtoroWsStatus(BaseModel):
+    subscriber_present: bool
+    ws_connected: bool
+    topic_refs: dict[int, int]
+    task_done: bool | None
+    last_quote_max: str | None
+
+
+@router.get("/etoro-ws", response_model=EtoroWsStatus)
+def etoro_ws_status(request: Request) -> EtoroWsStatus:
+    sub = getattr(request.app.state, "etoro_ws", None)
+    if sub is None:
+        return EtoroWsStatus(
+            subscriber_present=False,
+            ws_connected=False,
+            topic_refs={},
+            task_done=None,
+            last_quote_max=None,
+        )
+    ws = getattr(sub, "_ws", None)
+    refs: dict[int, int] = dict(getattr(sub, "_topic_refs", {}))
+    task = getattr(sub, "_task", None)
+    task_done = task.done() if task is not None else None
+
+    last_quote_max: str | None = None
+    pool = getattr(request.app.state, "db_pool", None)
+    if pool is not None:
+        with pool.connection() as conn:
+            row = conn.execute("SELECT MAX(quoted_at) FROM quotes").fetchone()
+            if row is not None and row[0] is not None:
+                last_quote_max = row[0].isoformat()
+
+    return EtoroWsStatus(
+        subscriber_present=True,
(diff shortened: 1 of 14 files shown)


Luke-Bradford:
Resolved on `3794f07`:

- **BLOCKING** `app/main.py:342` debug-router auth — FIXED `3794f07`. `app.include_router(debug_ws_router)` now gated on `settings.app_env != "prod"`. All three `/_debug/*` routes (etoro-candles-probe, etoro-instrument-raw, etoro-ws) won't register in production. Local dev / demo (`app_env=dev`) keeps them for diagnostics. Comment in main.py explains the security rationale for future readers.
- **WARNING** midnight ET boundary — FIXED `3794f07`. Added explicit test `classifies exactly midnight ET (00:00) as closed` in `chartFormatters.test.ts`. 600/600 tests pass. The `% 24` normalisation in `_nyParts` is now locked in by an assertion so a runtime that emits hour="24" cannot silently regress to an unmatched branch.
- **PREVENTION** unauthenticated debug routes — EXTRACTED `app/main.py`. The pattern is now established in code (gate on `settings.app_env != "prod"`); future debug routers added under `_debug/` should follow the same gate. CI lint check deferred — not extracted to a separate skill / log entry yet because there's currently only one debug router; will revisit if a second appears.

Luke-Bradford:
Round 2 resolved on \`ae1d43c\`:

- **WARNING** \`_stop_event\` type — REBUTTED. \`_stop_event = asyncio.Event()\` at [etoro_websocket.py:435](app/services/etoro_websocket.py#L435). \`await self._stop_event.wait()\` is the correct coroutine call; no \`TypeError\` risk. Reviewer flagged a potential issue, but the type is correct.
- **NITPICK** \`app_env != "prod"\` allowlist — FIXED \`ae1d43c\`. Switched to \`app_env in {"dev", "test", "local"}\`. Future envs like \`staging\`/\`qa\`/\`uat\` now deny-by-default and won't silently expose \`/_debug/*\`.
- **PREVENTION** threading.Event in async context — REBUTTED (no actual threading.Event in the file). N/A.

Luke-Bradford:
Round 3 resolved on \`5248eb3\`:

- **WARNING** session-filter dedupe gap — FIXED \`5248eb3\`. Ticks rejected by \`acceptPre\`/\`acceptAh\` now record \`lastAppliedKeyRef\` BEFORE the early return. Filter is \"going-forward\" semantics — toggling false→true mid-flight does not retroactively apply a stale tick.
- **PREVENTION** — EXTRACTED via 2 unit tests in \`useLiveLastBar.test.ts\`:
  1. \`rejects PM tick when acceptPre=false AND does not retroactively apply when acceptPre flips to true\`
  2. \`accepts a NEW PM tick after filter flips ON\`

Now polling for review on \`5248eb3\`.

Issue #626, opened 2026-04-28
bug: tests/test_migration_076_dedupe_financial_periods fails on main


Luke-Bradford:
## Symptom
`uv run pytest tests/test_migration_076_dedupe_financial_periods.py::TestDedupePass1SameAccession::test_canonical_table_keeps_smallest_period_end` fails on `main` (commit c039544) with:

\`\`\`
psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "uniq_financial_periods_fiscal_label"
DETAIL:  Key (instrument_id, source, fiscal_year, COALESCE(fiscal_quarter, 0), period_type)=(1, sec_edgar, 2025, 4, Q4) already exists.
\`\`\`

## Context
Reproduced on `main` with no local changes. Likely interaction between:
- #624 / migration 077 (partial unique index `uniq_financial_periods_fiscal_label`, merged 2026-04-26)
- pre-existing test in `tests/test_migration_076_dedupe_financial_periods.py` that seeds two FY2025 Q4 rows.

The test's seed pattern violates the new unique index added in #624 because the test was written before that index existed and re-runs on a DB whose fixtures persist between cases.

## Workaround
None for now — pre-push gate skips it via `--deselect`.

## Surfaced by
Pre-push run while preparing PR for #245. Not in scope for that PR.

Luke-Bradford:
Resolved via PR #631 (commit 8213dc5).

Issue #245, opened 2026-04-16
bug: manual close can record EXIT side effects when position deduction matches zero rows


Luke-Bradford:
## Finding
`POST /portfolio/positions/{position_id}/close` reads the open broker position before `_persist_order_and_fill()` opens its transaction, then the EXIT update uses `WHERE position_id = %(pid)s AND units >= %(units)s` but never checks whether the update matched a row.

## Risk
Two close requests racing against the same position can both build a filled synthetic EXIT and both write order/fill/cash/audit records. Only the first request actually deducts the broker position; the second update can affect zero rows while still recording proceeds and realised P&L side effects.

The same silent-zero-row behaviour can also hide drift between `positions` and `broker_positions` because the aggregate `positions` update is not rowcount-checked either.

## Suggested Fix
Move the broker-position lookup/deduction into the same transaction and lock the row (`FOR UPDATE`) before calculating close units, or make the deduction the first transactional write and fail/rollback when `rowcount != 1`. Add a regression test for two closes / stale units where the second request must not persist cash/fill side effects.

## References
- `app/api/orders.py`: close lookup around lines 486-493
- `app/api/orders.py`: unchecked EXIT updates around lines 309-353

Luke-Bradford:
Fixed in #627 (commit 0cb2408). Lock-order matched portfolio_sync (positions → broker_positions) per Codex pre-push review.

PR #627, opened 2026-04-28
fix(#245): manual close — lock + rowcount-check EXIT writes

Luke-Bradford:
## What
Close the silent zero-row hazard on the manual-close path:
- `_persist_order_and_fill` EXIT branch updates `positions` first (rowcount-checked), then `SELECT ... FOR UPDATE` on `broker_positions` + re-validates units before the deduction UPDATE (also rowcount-checked).
- Any zero-row or stale-units case raises `HTTPException(409)`; the surrounding `with conn.transaction():` block rolls back orders/fills/cash_ledger/decision_audit together.

## Why
Two close requests racing the same broker position could both build a synthetic EXIT fill, both pass the `units >= fu` UPDATE guard for the winner, and the loser's UPDATE would silently match zero rows while the orders/fills/cash_ledger/audit inserts in the same transaction still committed (#245). Lock + rowcount check turns the silent corruption into a loud 409.

Lock order matches `app/services/portfolio_sync.py` (positions → broker_positions) so concurrent broker sync cannot deadlock against a manual close on the same instrument. (Codex pre-push review caught a deadlock cycle in the original broker_positions-first ordering; the swap closes it.)

## Test plan
- [x] `uv run pytest tests/test_orders_api.py` — 23 passed (4 new regression tests for the silent-zero hazard + the deadlock-safe reorder)
- [x] `uv run pytest tests/api/test_orders_safety_layers.py` — green
- [x] `uv run pytest tests/smoke/test_app_boots.py` — green
- [x] Full suite `uv run pytest --deselect tests/test_migration_076_dedupe_financial_periods.py` — 2899 passed (one test-suite-level pre-existing failure unrelated to this PR is filed as #626)
- [x] `uv run ruff check . && uv run ruff format --check . && uv run pyright` — green

### DIFF
--- a/app/api/orders.py
+++ b/app/api/orders.py
@@ -315,7 +315,19 @@ def _persist_order_and_fill(
             elif action == "EXIT":
                 # Update positions for EXIT — prorate cost_basis so partial
                 # closes don't leave the full basis on fewer units.
-                conn.execute(
+                # Rowcount-checked: the aggregate row is the source of
+                # realized P&L; a silent zero-row UPDATE here would record
+                # cash + audit side effects without moving realized_pnl
+                # (prevention log entry: "Single-row UPDATE silent no-op").
+                #
+                # Lock-order discipline (#245): touch ``positions`` BEFORE
+                # ``broker_positions``. ``portfolio_sync`` acquires row
+                # locks in that same order (UPDATE positions at
+                # ``app/services/portfolio_sync.py`` L585, broker_positions
+                # writes at L688). Reversing the order here would form a
+                # deadlock cycle against any concurrent broker sync
+                # touching the same instrument.
+                positions_cur = conn.execute(
                     """
                     UPDATE positions SET
                         cost_basis     = CASE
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
Resolutions for the review on commit 745ac98:

- **WARNING (test rollback assertion gap)**: `FIXED 0f3dbaf` — `test_close_409_when_locked_units_below_request` now asserts `conn.transaction().__exit__` was called with a non-None `exc_type`, which is the actual signal that triggers psycopg's tx ROLLBACK and rolls back the orders/fills cursor inserts. The earlier "no cash_ledger / no decision_audit" assertions are kept as belt-and-braces.

- **PREVENTION (lint rule for rowcount-checked UPDATEs)**: `ALREADY_COVERED docs/review-prevention-log.md` — the rule "Single-row UPDATE silent no-op on missing row" (entry 142) already documents the human rule and points to `.claude/skills/engineering/sql-correctness.md` ("Single-row UPDATE must verify rowcount") as the enforcement surface. A linter step that grep-asserts the pattern across `conn.execute("UPDATE …")` calls would be a uplift on top of the human rule, not a missing rule — filing a separate tech-debt for it would be the right shape if it lands.

Issue #335, opened 2026-04-19
tech-debt(#328): calendar-accurate monthly cadence for monthly_reports


Luke-Bradford:
Follow-up to #334 review comment on PR 334.

`monthly_reports` uses `Cadence(interval=timedelta(days=31))`. A flat 31-day window drifts relative to calendar months (e.g. Feb + 31d overshoots; 30-day months undershoot). Non-blocking today because:

- `monthly_reports` is a non-blocking reporting layer.
- A 0-1 day drift at month boundaries has no operational impact on trading or freshness gates.

When time permits, extend `Cadence` to accept a calendar-anchored interval (e.g. `calendar_months: int`) and switch `monthly_reports` to use it. Likely shape:

```python
@dataclass(frozen=True)
class Cadence:
    interval: timedelta | None = None
    calendar_months: int | None = None
```

Consumers of `Cadence.grace_window` + `Cadence.interval.total_seconds()` would need to handle both cases.

Acceptance: `monthly_reports` next-fire time always lands on day 1 of each calendar month in UTC.

Linked: #328 (freshness unification umbrella), PR #334 (introduces 31-day placeholder).

Luke-Bradford:
Fixed in #628 (commit on main).

PR #628, opened 2026-04-28
feat(#335): calendar-anchored Cadence for monthly_reports

Luke-Bradford:
## What
Replace ``monthly_reports`` cadence ``Cadence(interval=timedelta(days=31))`` with a calendar-anchored cadence:
- ``Cadence`` dataclass gets ``calendar_months`` (mutually exclusive with ``interval``), ``window_start(now)``, and ``cadence_seconds_for_state_machine(now, grace)``.
- ``compute_layer_states_from_db`` snapshots a single ``now`` per evaluation and passes it through the cadence helper so the state-machine age check fires at the day-1 UTC tick.
- ``monthly_reports_is_fresh`` anchors on ``COALESCE(finished_at, started_at)`` (matching ``_latest_age_seconds_map``) and computes the month boundary in Python (no SQL timestamp-without-time-zone coercion).
- Registry switches ``monthly_reports`` to ``Cadence(calendar_months=1)``.

## Why
Flat 31-day cadence drifts against calendar months — Feb runs go stale a day late, long-month runs go stale a day early. The acceptance criterion in #335 ("next-fire time always lands on day 1 of each calendar month UTC") is a calendar-anchored boundary, not a rolling timedelta.

Codex pre-push review caught two pre-merge issues that this PR resolves before the first push:
1. The original draft only changed the predicate, not the v2 state machine — operators looking at /sync/layers/v2 would see HEALTHY for ~38 days.
2. The original SQL ``date_trunc('month', now() at time zone 'UTC')`` returns ``timestamp without time zone`` and would silently coerce against ``timestamptz`` ``started_at`` using the session ``TimeZone`` — boundary-misclassifying runs on non-UTC DB sessions.

Round 3 also aligned the freshness anchor to ``COALESCE(finished_at, started_at)`` so ``/sync/layers`` (predicate) and ``/sync/layers/v2`` (state machine) cannot disagree on a boundary-straddling run.

## Test plan
- [x] ``uv run pytest tests/services/sync_orchestrator/ tests/test_sync_orchestrator_freshness.py`` — green, includes new boundary tests
- [x] ``uv run pytest --deselect tests/test_migration_076_dedupe_financial_periods.py`` — 2916 passed (the deselected test fails on main, filed as #626)
- [x] ``uv run ruff check . && uv run ruff format --check . && uv run pyright`` — green
- [x] Codex pre-push review: 3 rounds, all findings resolved or extended into tests

### DIFF
--- a/app/services/sync_orchestrator/freshness.py
+++ b/app/services/sync_orchestrator/freshness.py
@@ -18,11 +18,12 @@ BEFORE ordering would hide a newer failure behind an older success.
 
 from __future__ import annotations
 
-from datetime import date, timedelta
+from datetime import UTC, date, datetime, timedelta
 from typing import Any
 
 import psycopg
 
+from app.services.sync_orchestrator.layer_types import Cadence
 from app.services.sync_orchestrator.types import PREREQ_SKIP_MARKER
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


Luke-Bradford:
Resolutions for the review on commit b146570:

- **WARNING (cadence_seconds=0 at day-1 00:00:00 UTC tick)**: `FIXED 9e8cf48` — `Cadence.cadence_seconds_for_state_machine` now floors boundary_age to 1 second before dividing by grace_multiplier. Rule 9 behavior is unchanged (any prior run has age > 0).
- **PREVENTION (assert positive cadence in build())**: `FIXED 9e8cf48` — added `assert cadence_seconds > 0` inside `compute_layer_states_from_db.build` so a regression that re-introduces zero fails the planning sweep loudly. Test added at `tests/services/sync_orchestrator/test_layer_types.py::test_cadence_seconds_for_state_machine_at_exact_boundary_is_floored`.

Issue #346, opened 2026-04-19
tech-debt(#342): audit trail + reason string for layer_enabled toggles


Luke-Bradford:
Codex flagged during A.5 chunk 2 review:

\`POST /sync/layers/{name}/enabled\` accepts only \`{enabled}\` and relies on client-side \`window.confirm\` for safety-critical disables. Direct API / service-token callers can disable \`fx_rates\` / \`portfolio_sync\` without a reason or operator attribution. Inconsistent with kill-switch / live-trading toggles which require \`reason\` + \`activated_by\` + audit writes (\`app/api/config.py\`).

## Proposal

1. Extend \`layer_enabled\` with columns \`reason: text | null\`, \`changed_by: text | null\`.
2. Optionally add a \`layer_enabled_audit\` table (id, layer_name, is_enabled, reason, changed_by, changed_at) so toggle history is queryable.
3. \`POST /sync/layers/{name}/enabled\` accepts optional \`{enabled, reason, changed_by}\`; REQUIRES \`reason\` when disabling a safety-critical layer (\`fx_rates\` / \`portfolio_sync\`) from non-demo mode.
4. UI prompts for reason (textarea) when the \`window.confirm\` dialog fires for safety-critical disables.

## Accept condition

Every \`fx_rates\` / \`portfolio_sync\` disable in a non-demo env is rejected by the API unless it carries a reason string; audit row shows reason + timestamp + actor.

Luke-Bradford:
Fixed in #629.

PR #629, opened 2026-04-28
feat(#346): audit trail + reason for layer_enabled toggles

Luke-Bradford:
## What
Close the operator-attribution gap on \`layer_enabled\` toggles:
- \`layer_enabled\` gains \`reason\` + \`changed_by\` (sql/078).
- New \`layer_enabled_audit\` table preserves the full toggle history.
- \`set_layer_enabled\` writes both rows on the caller's connection.
- \`POST /sync/layers/{name}/enabled\` rejects safety-critical disables (\`fx_rates\`, \`portfolio_sync\`) without a non-empty reason — 400 before any write.
- \`POST /sync/ingest/{key}/enabled\` routes through the same service so both endpoints emit audit rows.

## Why
Before this PR a direct API / service-token caller could disable \`fx_rates\` (live P&L drift) or \`portfolio_sync\` (broker drift) without leaving a trace or attribution — inconsistent with the kill-switch / live-trading toggles. Issue #346 calls out that the UI's \`window.confirm\` is not enforcement; this PR puts the gate at the API boundary and the audit row in the database.

The reason gate is asymmetric: required on disable, not required on enable. Re-enabling is the safe direction; making the operator type a justification to UN-pause would just delay safety.

## Test plan
- [x] \`uv run pytest tests/services/test_layer_enabled.py tests/api/test_sync_layer_enabled_endpoint.py tests/api/test_orders_safety_layers.py tests/test_execution_guard_safety_layers.py tests/test_order_client_safety_layers.py\` — green
- [x] Full suite \`uv run pytest --deselect tests/test_migration_076_dedupe_financial_periods.py\` — 2921 passed (the deselected test fails on main, filed as #626)
- [x] \`uv run ruff check . && uv run ruff format --check . && uv run pyright\` — green
- [x] Codex pre-push review: caught the migration file being untracked; resolved before first push

## Out of scope (deliberate)
- UI prompt for the reason input. The HTTP gate alone unblocks the audit-trail requirement; the UI textarea is a frontend follow-up that should land alongside whichever Admin redesign claims this surface.

### DIFF
--- a/app/api/sync.py
+++ b/app/api/sync.py
@@ -513,9 +513,13 @@ def post_layer_enabled(
     # otherwise be able to flip ``fx_rates`` or ``portfolio_sync`` off
     # without operator attribution, which is inconsistent with the
     # kill-switch / live-trading toggles.
+    # Trim once so both the safety gate and the audit row see the same
+    # value — passing the unstripped form to set_layer_enabled would
+    # store leading/trailing whitespace despite the gate having
+    # rejected it.
+    reason_trimmed = body.reason.strip() if body.reason is not None else None
     if not body.enabled and layer_name in SAFETY_CRITICAL_LAYERS:
-        reason = (body.reason or "").strip()
-        if not reason:
+        if not reason_trimmed:
             raise HTTPException(
                 status_code=400,
                 detail=(f"layer '{layer_name}' is safety-critical; supply a non-empty 'reason' to disable it."),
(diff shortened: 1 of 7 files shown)


Luke-Bradford:
Resolution for the review on commit 4638a16:

- **NITPICK (forward trimmed reason)**: `FIXED b32f311` — gate and `set_layer_enabled` now share a single `reason_trimmed` local; the audit row stores the trimmed form.

Issue #598, opened 2026-04-27
tech-debt: add supply-chain scan step to CI (pnpm audit / socket scan)


Luke-Bradford:
## Context

Surfaced via #597 review. The Claude review bot flagged `lodash@4.18.1` as a phantom version (training data cut off at 4.17.21); investigation showed lodash quietly resumed releases (4.17.21 → 4.17.23 → 4.18.0 → 4.18.1, current \`latest\` dist-tag). The bot's specific finding was wrong, but the underlying PREVENTION suggestion is sound: a CI step that catches actual phantom or yanked versions before merge.

## What

- Add a CI job that runs against every PR touching \`frontend/pnpm-lock.yaml\` or \`pyproject.toml\`/\`uv.lock\`
- Frontend: \`pnpm audit --prod\` and/or socket.dev scan, fail on \`high\`/\`critical\` advisories or unpublished/yanked versions
- Backend: \`uv pip audit\` or equivalent
- Allowlist mechanism for known-accepted exceptions

## Why

Defends against:
- Phantom / typosquatted versions slipping through
- Yanked-but-cached versions persisting in lockfiles
- Newly disclosed advisories on existing transitive deps

## Out of scope

- Deep SBOM tooling (separate uplift)

## Linked

- Surfaced in #597 review

Luke-Bradford:
Supply-chain audit job (pnpm install --frozen-lockfile + pnpm audit + pip-audit) added in #630.

Issue #247, opened 2026-04-16
ci: run backend and frontend test suites in GitHub Actions


Luke-Bradford:
## Finding
GitHub Actions currently checks lint/format/type/build, but does not run the test suites:

- `.github/workflows/ci.yml` runs Ruff and Pyright only; it never runs `pytest`.
- `.github/workflows/frontend-ci.yml` runs `pnpm typecheck` and `pnpm build`; it never runs `pnpm test`.

The repo has substantial backend and frontend tests, so this leaves behavioural regressions to local discipline and review comments.

## Risk
A PR can be green while breaking order execution, broker-credential recovery, scheduler behaviour, frontend setup/recovery flows, or other tested application logic. The recent review findings are exactly the class of issues that unit tests are meant to gate.

## Suggested Fix
Add test steps to CI:

- Backend: run `uv run pytest` after Pyright. If DB-backed smoke tests need services, either provide Postgres in the workflow or split pure unit tests from integration/smoke tests with markers.
- Frontend: run `pnpm test` after typecheck/build, or before build if preferred.

Consider a follow-up to mark expensive/external tests explicitly so the default CI run is deterministic and fast.

## References
- `.github/workflows/ci.yml`: lines 27-34
- `.github/workflows/frontend-ci.yml`: lines 40-44

Luke-Bradford:
Backend pytest + frontend test jobs added in #630.

PR #630, opened 2026-04-28
ci(#247, #598): run backend tests + frontend tests + supply-chain audit

Luke-Bradford:
## What
Three CI workflow additions, two issues:

**#247 — run backend + frontend test suites**
- New \`pytest\` job in \`.github/workflows/ci.yml\` with a Postgres 16 service + CI-only env vars (\`EBULL_DATABASE_URL\`, \`EBULL_SECRETS_KEY\`, \`EBULL_SERVICE_TOKEN\`). Without the Postgres service, the smoke + integration tests would skip and merge gates would regress.
- \`frontend-ci.yml\` now runs the full \`pnpm test\` (was: typecheck + build only). Pre-push gate stays \`pnpm test:unit\` for speed; CI runs the wider \`test\` so an integration-only regression cannot ship behind a green pre-push.

**#598 — supply-chain scan**
- New \`supply-chain\` job in \`ci.yml\` with three steps:
  1. \`pnpm install --frozen-lockfile\` — turns phantom-version / yanked / unpublished tarballs into a hard fail (which \`pnpm audit\` alone misses).
  2. \`pnpm audit --audit-level high\` (no \`--prod\`) — audits the full toolchain since vite/vitest/typescript run on CI and dev machines.
  3. \`pip-audit\` against the full \`uv.lock\` (no \`--no-dev\`) — same reasoning for pytest / ruff / pyright.
- Targeted lockfile bump: \`pytest 9.0.2 → 9.0.3\` (CVE-2025-71176, surfaced when expanding pip-audit to dev deps).

## Why
Before this PR, CI ran only lint + typecheck + frontend build. Test failures could merge if the local pre-push step was skipped, and supply-chain advisories had no automated gate at all. Both issues call out that the human rule already exists; this PR puts the enforcement in the workflow.

Codex pre-push review caught two real flaws the original draft would have shipped:
- \`pnpm audit\` does not prove tarball fetchability — added \`pnpm install --frozen-lockfile\` as the proper phantom-version gate.
- \`--prod\` / \`--no-dev\` excluded the toolchain that actually executes on CI runners — dropped both flags and bumped the one CVE that surfaced.

## Test plan
- [x] \`pnpm install --frozen-lockfile\` (frontend) — green
- [x] \`pnpm audit --audit-level high\` — exit 0 (3 moderate, no high/critical)
- [x] \`pip-audit --strict\` against \`uv export\` — clean after pytest bump
- [x] Full backend suite \`uv run pytest --deselect tests/test_migration_076_dedupe_financial_periods.py\` — 2921 passed
- [x] Full frontend suite \`pnpm test\` — 609 passed
- [ ] Watch CI on this PR to confirm the new jobs themselves go green

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,3 +32,126 @@ jobs:
 
       - name: Pyright typecheck
         run: uv run pyright
+
+  pytest:
+    # #247: run backend test suite. Postgres service is required; the
+    # smoke test (tests/smoke/test_app_boots.py) drives the FastAPI
+    # lifespan against a live DB and is the canonical gate against
+    # broken migrations / lifespan-only failures. Without a real
+    # Postgres, the integration tests skip and merge gates regress.
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: ebull
+        ports:
+          - 5432:5432
+        # Service must be ready before the suite starts; without
+        # health-check args the runner moves on while the server is
+        # still booting and the first connection attempt errors.
+        options: >-
+          --health-cmd "pg_isready -U postgres"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 10
+
+    env:
+      EBULL_DATABASE_URL: postgresql://postgres:postgres@localhost:5432/ebull
+      # Deterministic 32-byte base64 key used only for CI. Real
+      # deployments derive this per-environment via ADR 0001 — never
+      # share this value with prod.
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
Resolutions for the review on commit 5fb2dbe:

- **WARNING (postgres:16 vs PG17)**: `FIXED 65fc12e` — service image bumped to `postgres:17`. Cited the plan docs that already target PG17 in the workflow comment so a future contributor understands why.
- **NITPICK (uv unpinned)**: `FIXED 65fc12e` — `uv` pinned to `0.5.21` across both `pytest` and `supply-chain` jobs. `pip-audit` deliberately left floating so the gate always pulls the latest advisory database.
- **PREVENTION (Postgres version drift rule)**: `EXTRACTED .github/workflows/ci.yml` — the comment block on the service image now cites the plan docs and explains the alignment requirement, so the next contributor reviewing infra changes sees the rationale at the change site rather than having to find it in a separate doc.

Issue #624, opened 2026-04-28
harden financial_periods against future-arrival dupes (post-#558)


Luke-Bradford:
## Context
#558 cleared existing duplicates on dev DB (3,838 rows deleted via migration 076) and #_derive_periods_from_facts fix prevents same-run DEI pollution. Gap: future-arrival ordering — late amendment with different period_end_date than original would still insert a NEW row instead of updating the existing one because canonical merge keys on `period_end_date`.

## Fix
1. **`_canonical_merge_instrument` reorder**: change `DISTINCT ON (period_end_date, period_type)` to `DISTINCT ON (fiscal_year, fiscal_quarter, period_type)`, sort by `filed_date DESC NULLS LAST, period_end_date ASC`. Latest filed_date wins regardless of arrival order; period_end_date ASC tiebreak keeps the smallest (real fiscal end) when two rows tie on filed_date.

2. **Schema partial unique index** on `financial_periods` over `(instrument_id, source, fiscal_year, COALESCE(fiscal_quarter,0), period_type)` WHERE superseded_at IS NULL. Makes second-arrival INSERT impossible at the DB layer; canonical merge becomes ON CONFLICT UPDATE.

3. **Tests**: integration test seeds (a) original then amendment, (b) amendment then original — both must converge on the latest-filed row regardless of insertion order.

## Acceptance
- Late-arriving amendment with different period_end_date supersedes original.
- Late-arriving original (after amendment) does NOT replace amendment.
- Unique constraint prevents duplicate rows by fiscal label.

Follow-up to #558.

Issue #558, opened 2026-04-26
fix: financial_periods duplicate quarter rows — same period stored twice with different period_end_date


Luke-Bradford:
## Bug

Every quarter in `financial_periods` is stored TWICE for SEC-sourced instruments — once at the real period-end date (e.g. 2026-01-31 for fiscal Q4 2025) and once at a filing-derived pseudo-period (e.g. 2026-03-18). Same revenue, same source_ref accession, different `period_end_date` PK component.

GME illustration:

| period_end | type | fy | fq | source | accession | revenue |
|---|---|---|---|---|---|---|
| 2026-03-18 | Q4 | 2025 | 4 | sec_edgar | 0001326380-26-000013 | 2,732,400,000 |
| 2026-03-18 | FY | 2025 | 0 | sec_edgar | 0001326380-26-000013 | 5,272,800,000 |
| 2026-01-31 | Q4 | 2025 | 4 | sec_edgar | 0001326380-26-000013 | 2,732,400,000 |
| 2026-01-31 | FY | 2025 | 0 | sec_edgar | 0001326380-26-000013 | 5,272,800,000 |
| 2025-12-05 | Q3 | 2025 | 3 | sec_edgar | 0001326380-25-000098 |   860,300,000 |
| 2025-11-01 | Q3 | 2025 | 3 | sec_edgar | 0001326380-25-000098 |   860,300,000 |

Pattern: every reported period appears at the actual fiscal period-end AND at a date that's 6-7 weeks later. The later date matches the SEC filing date minus a few days. Same accession on both rows.

## Visible impact

Frontend Financials tab (`/instrument/:symbol` Financials sub-tab) renders 12+ duplicate columns instead of the 4 quarters operator expects.

## Likely root cause

`app/services/fundamentals.py` normalisation path. Two candidate sources of the second row:

1. SEC XBRL pipeline writes the period at `period_end` (real fiscal date) AND at `endDate` from the iXBRL fact context (which sometimes differs from `period_end` by a few weeks for amended/restated filings).
2. Two ingest paths writing the same fact through different normalisers — `fundamentals_sync` Phase 1b and `daily_research_refresh` SEC fundamentals block both touching `financial_periods` with slightly different period_end derivation.

## Fix

[...]

PR #631, opened 2026-04-28
fix(#626): migration 076 test fixture coexists with 077 unique index

Luke-Bradford:
## What

Adds a `fiscal_label_index_dropped` pytest fixture so the migration 076 dedupe tests can keep seeding pre-#558 polluted state without colliding with migration 077's `uniq_financial_periods_fiscal_label` partial unique index. CI workflow drops its `--deselect` for this file.

## Why

`uv run pytest tests/test_migration_076_dedupe_financial_periods.py::TestDedupePass1SameAccession::test_canonical_table_keeps_smallest_period_end` failed on main with `UniqueViolation` after #624 merged. The test seeds two `financial_periods` rows with the same fiscal label and different `period_end_date` — exactly the shape migration 077 was added to forbid. The dedupe migration is still valid; the test just needs the index out of the way during seeding.

## Test plan

- [x] `uv run pytest tests/test_migration_076_dedupe_financial_periods.py` — 9 pass
- [x] `uv run pytest` full suite — 2930 pass, 1 skipped
- [x] `uv run ruff check .` clean
- [x] `uv run ruff format --check .` clean
- [x] `uv run pyright` clean
- [x] Crash-safety: simulated mid-test `RuntimeError` after seeding duplicates and confirmed the fixture's `finally` restores the index (teardown runs 076 first to dedupe, then 077 to recreate)
- [x] Codex review confirmed crash window closed

## Notes

Teardown order in the fixture matters: 076 (dedupe) before 077 (recreate). Codex caught two earlier shapes:
1. helper-call pattern with no `try/finally` — would leave the index gone on a mid-test crash
2. fixture restoring 077 first — would hit `UniqueViolation` if seeds remained on disk

The committed shape resolves both.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,11 +94,7 @@ jobs:
         run: uv sync --group dev
 
       - name: Run backend pytest
-        # Deselect the test that fails on main (#626 — pre-existing
-        # uniq_financial_periods_fiscal_label collision in the
-        # migration 076 dedupe fixture). Remove the deselect once #626
-        # lands.
-        run: uv run pytest --deselect tests/test_migration_076_dedupe_financial_periods.py
+        run: uv run pytest
 
   supply-chain:
     # #598: catch yanked / phantom / advisory-laden lockfile entries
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
Resolution for review WARNING + PREVENTION: FIXED b1f232ea677b57502aee919c0126ea0f47ed3f6d

**WARNING (teardown-step isolation):** Wrapped the 076 dedupe call in its own `try/except`. The 077 recreate now runs unconditionally and re-raises so a real leak surfaces immediately rather than silently corrupting later tests.

**PREVENTION:** EXTRACTED `docs/review-prevention-log.md` — new entry "Teardown-step isolation in shared-DB schema fixtures" (distinct from the existing teardown rules around module-level singletons and FK-child table lists).

Local gates re-run: pytest 076 file 9 pass, ruff/format/pyright clean.

Luke-Bradford:
NITPICK FIXED 43ddc4c10698c0de8d2ccd579529c8f95118b30b — `stacklevel=1` so the warn line points to this file rather than pytest fixture internals.

Issue #560, opened 2026-04-26
tech-debt(#559): backfill tables_json on existing business_summary rows


Luke-Bradford:
## Context

Phase 1 of #559 added the \`tables_json\` column + parser support. New 10-K parses populate \`tables_json\` end-to-end, but existing rows in \`instrument_business_summary_sections\` (currently 178k+) have \`tables_json = NULL\`.

The existing \`bootstrap_business_summaries\` candidate query at \`app/services/business_summary.py:1287\` skips rows where \`source_accession\` matches the latest filing AND no quarantine, so a re-bootstrap will not re-parse them.

## Options

1. Add \`force_reparse\` param to \`bootstrap_business_summaries\` that ignores the LEFT JOIN gate.
2. One-shot maintenance script that resets \`source_accession = ''\` on \`instrument_business_summary\` to trigger the supersede path.
3. TRUNCATE both \`instrument_business_summary\` and \`instrument_business_summary_sections\`, then run bootstrap (heavy hammer, ~hour).

## Why deferred

Phase 2 of #559 needs \`tables_json\` for new ingest only — Phase 2 frontend renders empty \`tables\` array gracefully when none present, so dev UX is not blocked. Backfill can ship as a separate operational PR.

## Acceptance

- New filings populate \`tables_json\` (already shipping).
- Existing 178k+ section rows get \`tables_json\` populated within one drain cycle.
- Spot check: GME has at least one section with \`jsonb_array_length(tables_json) > 0\`.

Luke-Bradford:
Resolved via PR #633.

PR #633, opened 2026-04-28
feat(#560): self-terminating tables_json backfill via candidate query

Luke-Bradford:
## What

Adds a fourth OR clause to the \`ingest_business_summaries\` candidate selector that re-selects parent rows whose child sections still have \`tables_json IS NULL\`. AND-guarded by the existing quarantine clock so backoff is preserved. Migration 079 adds a partial index supporting the EXISTS subquery; it stays empty in steady state.

## Why

Pre-migration-075 ingest left \`tables_json IS NULL\` on existing 178k+ section rows. The post-#075 parser populates the column on new parses but the steady-state candidate gate never re-selects rows whose \`source_accession\` already matches the latest 10-K. Backfill needed a mechanism that:
- doesn't require an operational script or DB mutation
- self-terminates once every row has \`tables_json\` populated
- respects the #533 exponential-backoff quarantine for failing reparses

A data-driven candidate clause does all three.

## Test plan

- [x] \`uv run pytest\` 2932 pass, 1 skipped
- [x] \`test_null_tables_json_re_selects_for_reparse\`: NULL tables_json triggers re-select; third run after repopulation does NOT
- [x] \`test_null_tables_json_quarantined_row_skipped\`: future \`next_retry_at\` keeps the row out despite NULL tables_json (regression coverage for the Codex-caught bypass)
- [x] Codex review APPROVE on the AND-guarded shape

## Operational notes

After merge, the next regularly-scheduled \`bootstrap_business_summaries\` run will start picking up backfill candidates under SEC fair-use rate limits. No operator action required. The partial index keeps the EXISTS subquery sub-millisecond and disappears once the backfill completes.

### DIFF
--- a/app/services/business_summary.py
+++ b/app/services/business_summary.py
@@ -1402,7 +1402,8 @@ def ingest_business_summaries(
     """Scan 10-K filings, fetch primary doc, extract Item 1, upsert.
 
     Candidate selector (shape addresses Codex #428 findings, extended
-    in #533 with backoff/quarantine):
+    in #533 with backoff/quarantine, in #560 with tables_json
+    backfill):
 
     1. ``fe.filing_type IN ('10-K', '10-K/A')`` — amendments retain
        their ``/A`` suffix through the SEC pipeline (see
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
NITPICK FIXED e6442ab9af29038039da8165e792721a454bdec3 — added `assert second.rows_updated == 1` so a silent no-op reparse cannot pass.

Issue #243, opened 2026-04-15
tech-debt: make live order execution durable before broker side effects


Luke-Bradford:
## Problem

`app.services.order_client.execute_order()` performs the live broker API call before it writes an `orders` row or any execution audit locally. This keeps external I/O outside the DB transaction, which is good, but it leaves a crash gap:

1. execution guard approves a recommendation
2. `broker.place_order()` or `broker.close_position()` succeeds or reaches the broker
3. the process crashes before `_persist_order()` runs

In that case the broker may have an order/fill, but eBull has no durable local order attempt, fill row, cash movement, or decision audit for reconciliation.

## Impact

Low while live trading is disabled. High before enabling live capital, because external broker state can diverge from local state with no local intent record to reconcile against.

## Suggested fix

Move toward a durable intent/outbox style flow:

- Insert an `orders` row with a pre-call status such as `submitted` / `sending` before calling the broker.
- Commit the intent before the external side effect.
- After the broker response, update the same row with broker ref/status/raw payload and write fill/cash/audit as today.
- On startup/scheduler, reconcile any stuck pre-call or pending rows via broker status where possible.

A smaller interim fix is to persist an order-attempt row before the call and mark it failed/pending after response handling.

## Acceptance criteria

- A process crash after a broker call can be detected from local durable state.
- Every live broker call has a local order intent/audit trail before the external side effect is attempted.
- Tests cover the write-before-call ordering using a fake broker that raises after observing the local order intent.

## Files

- `app/services/order_client.py`
- `tests/test_order_client.py`
- potentially SQL migration if new order statuses/columns are needed

Luke-Bradford:
Resolved via PR #637.

PR #637, opened 2026-04-28
feat(#243): durable order intent before broker side effect

Luke-Bradford:
## What

\`execute_order\` now writes a durable \`status='submitted'\` intent row + commits BEFORE the broker call (live paths only). After the broker returns the same row is UPDATEd with the broker response. Demo path unchanged.

## Why

Per #243: a process crash between \`broker.place_order\` and \`_persist_order\` leaves eBull with no local intent record while the broker may have an order. Low-risk under demo, high-risk before live capital. Outbox/intent shape closes the gap.

## Test plan

- [x] \`uv run pytest\` 2941 pass, 1 skipped (1 new regression test, 3 existing live tests updated)
- [x] New \`test_live_buy_persists_intent_before_broker_call\`: fake broker raises after observing the intent; asserts exact ordering \`intent_insert → commit → broker.place_order\`
- [x] Existing live tests updated for the new cursor sequence (pre-broker INSERT + post-broker UPDATE via \`conn.execute\`)
- [x] Codex APPROVE — flagged caller-contract documentation; addressed in the commit

## Connection-ownership contract

The mid-function \`conn.commit()\` between intent INSERT and broker call publishes any uncommitted writes the caller may have. \`execute_order\` MUST therefore receive a connection without unrelated in-flight writes. The current scheduler caller already uses a fresh per-order pool connection, but the contract is now documented on the \`execute_order\` docstring so future callers can't accidentally violate it.

## Follow-up (out of scope)

- Reconciler / sweeper that finds stuck \`status='submitted'\` rows older than X minutes and asks the broker for canonical state. Track separately when live trading enables.

### DIFF
--- a/app/services/order_client.py
+++ b/app/services/order_client.py
@@ -373,22 +373,36 @@ def _update_order_with_broker_result(
     broker_order_ref: str | None,
     raw_payload: dict[str, Any],
 ) -> None:
-    """Update the pre-call ``submitted`` row with the broker response (#243)."""
-    conn.execute(
-        """
-        UPDATE orders
-        SET status = %(status)s,
-            broker_order_ref = %(ref)s,
-            raw_payload_json = %(payload)s
-        WHERE order_id = %(oid)s
-        """,
-        {
-            "oid": order_id,
-            "status": status,
-            "ref": broker_order_ref,
-            "payload": Jsonb(raw_payload),
-        },
-    )
+    """Update the pre-call ``submitted`` row with the broker response (#243).
+
+    Raises ``RuntimeError`` if the UPDATE matched zero rows — without
+    this check, ``order_id`` would silently flow forward as a foreign
+    key into fills / cost records / positions and corrupt referential
+    integrity. PR #637 review BLOCKING.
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
Resolution for review BLOCKING + PREVENTION: FIXED 2bc80066631bdcc74afeac22cb9aefa3503ea282

**BLOCKING (zero-row UPDATE):** Switched `_update_order_with_broker_result` to cursor form with `cur.rowcount == 1` assertion. Raises `RuntimeError` on mismatch naming the order_id and observed rowcount, refusing to advance to fill/cost/position writes.

**PREVENTION:** EXTRACTED `docs/review-prevention-log.md` — new entry "UPDATE-by-PK helpers must assert rowcount" covering any helper that threads a primary key from a prior INSERT into FK-referencing writes.

Regression test `test_live_buy_zero_row_update_raises` simulates the phantom UPDATE with `_update_cursor(rowcount=0)` and asserts `execute_order` raises before fill writes. 2942 pytest pass, ruff/format/pyright clean.

Issue #541, opened 2026-04-26
feat: drop fundamentals_snapshot table; migrate readers to financial_periods


Luke-Bradford:
## What

Stage 4 of FMP removal. `fundamentals_snapshot` is a denormalised KV-style table. After stages 1-3, the only writer is SEC-XBRL-derived. Migrate readers to `financial_periods` + `financial_periods_ttm` + `instrument_valuation` (or a compatibility view), then drop `fundamentals_snapshot` outright.

## Files to update

- `app/services/scoring.py`
- `app/services/thesis.py`
- `app/api/instruments.py`
- `app/services/coverage.py`
- `app/workers/scheduler.py`
- `app/services/fundamentals.py`
- `sql/001_init.sql` (note schema retirement)
- `sql/032_financial_data_enrichment_p1.sql` (similar)
- All tests that still read/write `fundamentals_snapshot`

## Why

Eliminates a flat denormalised table whose existence was tied to FMP's row shape. Once gone, fundamentals reads come from a single canonical source (`financial_periods`) with provenance.

## Depends on

Stage 3 (fundamentals_snapshot rescoped to SEC-only consumers)

Luke-Bradford:
**Scope reality check after #539 / #540 land.**

This ticket needs to coordinate all of:
1. Schema-level: introduce a compatibility view OR migrate the writer to directly populate `financial_periods_ttm`-derived materialised storage.
2. Writer-side: retire `upsert_snapshot` in `app/services/fundamentals.py:185-225` (currently SEC-XBRL-derived per #414 phase 1b).
3. Reader-side: migrate ~7 active reads (`app/services/scoring.py:702`, `app/services/thesis.py:322`, `app/api/instruments.py:1974`, `app/services/coverage.py:287`, `app/services/fundamentals.py:169`, `app/services/ops_monitor.py:106`, plus the legacy fallback in `sql/032_financial_data_enrichment_p1.sql:505`) to consume `financial_periods_ttm` directly.
4. Schema mapping: snapshot columns are denormalised (`gross_margin`, `fcf`, etc.) while TTM rows are raw (`gross_profit_ttm`, `operating_cf_ttm`, `capex_ttm`). Each reader's calculation needs a derivation step.
5. Test migration: ~15 test files seed `fundamentals_snapshot` directly; each needs to switch to seeding `financial_periods_ttm`.

[...]

Issue #540, opened 2026-04-26
feat: retrofit fundamentals_snapshot to SEC-only regulated-source posture


Luke-Bradford:
## What

Stage 3 of FMP removal. Today `fundamentals_snapshot` is read across the runtime as if every tradable instrument should have a row. After stage 1 + 2, only SEC-CIK instruments will populate it. Need to rescope the consumers.

## Files to update

- `app/services/sync_orchestrator/freshness.py` — instruments without regulated source aren't 'stale', they're 'unsupported'
- `app/services/sync_orchestrator/content_predicates.py` — same
- `app/workers/scheduler.py` — candle bootstrap gating + research scheduling shouldn't require snapshot for non-SEC instruments
- `app/services/coverage.py` — coverage tier inputs scoped to SEC-CIK
- `app/services/ops_monitor.py` — health check semantics rescoped
- Related freshness / candle-bootstrap / coverage tests

## Why

Stop treating 'every tradable instrument must have a snapshot' as the rule once non-SEC instruments lose their FMP-fed rows. Otherwise dashboards alarm on instruments we don't have a data source for.

## Depends on

Stage 2 (FMP-only surfaces removed)

Luke-Bradford:
Resolved via PR #639.

Issue #539, opened 2026-04-26
feat: drop FMP-only surfaces (analyst_estimates, earnings_events, instrument_profile, currency enrichment)


Luke-Bradford:
## What

Stage 2 of FMP removal (#532 stage 1 ships first). Removes FMP-only data surfaces from the codebase entirely.

## Scope

**Code deletions:**
- Delete `app/providers/implementations/fmp.py`
- Delete or retire `app/providers/enrichment.py`
- Delete or retire `app/services/enrichment.py`
- Remove `enrich_instrument_currencies()` and FMP import from `app/services/universe.py`
- Remove analyst-estimates reads from `app/services/scoring.py`
- Remove earnings/analyst context reads from `app/services/thesis.py`
- Remove upcoming-earnings reporting from `app/services/reporting.py`
- Drop FMP raw-persistence source from `app/services/raw_persistence.py`

**Schema:**
- Migration drops tables: `analyst_estimates`, `earnings_events`, `instrument_profile`
- Drop `currency_enriched_at` if no other consumers

**Tests deleted:**
- `tests/test_currency_enrichment.py`
- `tests/test_enrichment_provider.py`
- `tests/test_enrichment_service.py`
- FMP parts of `tests/test_provider_interfaces.py`
- FMP parts of `tests/test_filings_fundamentals.py`

## Why

Stage 1 (#532) leaves the FMP code dormant but still in the repo. Stage 2 removes the dead paths so future operators don't accidentally re-wire a paid provider.

## Depends on

#532 stage 1

Luke-Bradford:
Resolved via PR #638.

Issue #532, opened 2026-04-26
feat: drop FMP from capability provider enum + seed (free-source-only posture)


Luke-Bradford:
## What

Remove FMP from `CapabilityProvider` enum in [app/services/capabilities.py:52](app/services/capabilities.py#L52) and from the per-asset-class seed in migration 071. Drop `(fundamentals, fmp)` and `(analyst, fmp)` rows from `_PRESENCE_QUERIES`. Drop the augmentation lines that add FMP to fundamentals/analyst on SEC-CIK match in [capabilities.py:261-267](app/services/capabilities.py#L261). Strip FMP enrichment job + FmpProvider wiring + `fundamentals_snapshot` / `analyst_estimates` writes (or document as legacy-frozen).

## Why

eBull stance: free regulated-source data only. FMP is a paid service with limited free tier; we don't want repo consumers to need a paid key. Currently:
- `FMP_API_KEY=` empty in `.env`
- Zero FMP job runs in `job_runs`
- `analyst_estimates` table: 0 rows
- `fundamentals_snapshot` table: 4790 stale rows from earlier provider activity
- Capability resolver lists `fmp` as a provider on every SEC-CIK instrument → cosmetic lie in panel chrome (data_present always false)

Discovered during #515 PR 3b post-merge review.

## Acceptance

- [ ] FMP removed from `CapabilityProvider` Literal
- [ ] Migration drops FMP from `exchanges.capabilities` defaults (run on every exchange row)
- [ ] `_PRESENCE_QUERIES` entries removed; resolver tests updated
- [ ] Frontend `providerLabel` map entry removed
- [ ] `fundamentals_snapshot` / `analyst_estimates` tables either dropped or documented as legacy-frozen + not displayed
- [ ] Pre-push gates green

Luke-Bradford:
Closing as already done.

Stage 1 has shipped:
- `fmp` is absent from `CapabilityProvider` Literal at [`app/services/capabilities.py:41-89`](app/services/capabilities.py#L41-L89).
- Migration [`sql/072_drop_fmp_from_capabilities.sql`](sql/072_drop_fmp_from_capabilities.sql) removes `fmp` from `exchanges.capabilities` defaults.
- Frontend `PROVIDER_LABEL` at [`frontend/src/lib/capabilityProviders.ts:14-62`](frontend/src/lib/capabilityProviders.ts#L14-L62) has no `fmp` entry.
[...]

PR #639, opened 2026-04-28
feat(#540): retrofit fundamentals_snapshot to SEC-only cohort

Luke-Bradford:
## What

- \`fundamentals_is_fresh\` and \`fundamentals_content_ok\` now JOIN \`external_identifiers\` on \`provider='sec', identifier_type='cik', is_primary=TRUE\` before counting missing snapshots. Detail message says "SEC-CIK tradable instruments...".
- \`daily_research_refresh\` symbol→CIK lookup filtered on \`ei.is_primary=TRUE\` so producer + reader cohorts match.
- Three SQL-shape regression tests pin the predicate on both reader paths AND the producer query.

## Why

After #532 / #539 retired FMP, fundamentals_snapshot only covers SEC-CIK instruments. Treating "every tradable instrument needs a row" as the invariant alarmed indefinitely on non-US / crypto / commodity instruments that have no public-source fundamentals path.

## Codex catches across review rounds

- Round 2: tests only pinned the detail-string output, not the SQL shape. Fixed with explicit \`assert "external_identifiers" in sql\` etc. on both reader paths.
- Round 3: producer (\`daily_research_refresh\` cik_rows query) iterated every sec/cik row regardless of \`is_primary\`. Fixed with the matching filter; reader and producer are now in lockstep. Pin test added.

## Test plan

- [x] \`uv run pytest\` 2882 pass, 1 skipped (3 new tests)

## Out of scope

- coverage.py \`has_fundamentals\` flag retained — gates T1 promotion, product-meaningful semantics for non-SEC instruments
- ops_monitor.py legacy staleness query retained — queued for retirement under #340 (blocked on #330)
- #541 drops the table outright; this PR keeps it for SEC instruments only

### DIFF
--- a/app/services/sync_orchestrator/content_predicates.py
+++ b/app/services/sync_orchestrator/content_predicates.py
@@ -49,7 +49,10 @@ def candles_content_ok(conn: psycopg.Connection[Any]) -> tuple[bool, str]:
 
 
 def fundamentals_content_ok(conn: psycopg.Connection[Any]) -> tuple[bool, str]:
-    """Every tradable instrument must have a fundamentals_snapshot row in the current quarter."""
+    """Every SEC-CIK-mapped tradable instrument must have a fundamentals_snapshot
+    row in the current quarter. #540: scoped to SEC-CIK only — non-US
+    instruments have no fundamentals source today.
+    """
     today = date.today()
     quarter = (today.month - 1) // 3
     quarter_start = date(today.year, quarter * 3 + 1, 1)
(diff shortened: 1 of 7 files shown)


Luke-Bradford:
Resolution for review WARNING + PREVENTION: FIXED 5c5322d4e8a02d027eeb03528c6350f3a1376076

**WARNING (positional call_args_list):** Replaced `mock.call_args_list[1].args[0]` with content filters keyed on 'external_identifiers'. Both tests now assert exactly one matching execute call, then read the SQL — a future refactor that inserts another execute earlier no longer shifts the index silently.

**PREVENTION:** EXTRACTED `docs/review-prevention-log.md` — new entry 'Positional call_args_list index in SQL-shape regression tests'. Self-review grep target included.

Issue #186, opened 2026-04-11
Tech debt: DB-level integration tests for positions upsert reset-on-reopen semantics


Luke-Bradford:
## Context

PR #185 added a ``source`` column to ``positions`` with reset-on-reopen semantics in the ``ON CONFLICT DO UPDATE`` branch of both ``app/services/portfolio_sync.py`` and ``app/services/order_client.py._update_position_buy``.

The unit tests in [test_portfolio_sync.py](tests/test_portfolio_sync.py) and [test_order_client.py](tests/test_order_client.py) assert the **SQL shape** — that the CASE WHEN clause with ``positions.current_units <= 0`` is present in the statement — but they do not exercise the conflict branch itself, because the mocked ``conn.execute`` never sees a real Postgres row.

This means the tests would catch:
- Removal of the CASE WHEN text (string-shape regression)
- Typos in the reset condition

But would not catch:
- A CASE WHEN that evaluates to the wrong branch under live Postgres semantics (e.g., if the condition were accidentally written as ``>= 0`` instead of ``<= 0``)

The PR #185 rebuttal verified the semantics empirically against the dev DB, so the behaviour is known to be correct at time of merge. This issue tracks adding automated protection against future drift.

## Scope

Add an integration test (in the spirit of ``tests/test_operator_setup_race.py``, which already uses the ``ebull_test`` isolated database pattern via ``_apply_migrations_to_test_db``) that:

1. Seeds a ``positions`` row with ``current_units = 0`` and ``source = 'broker_sync'``
2. Issues a real upsert via ``sync_portfolio`` (or directly against the SQL) with a new broker-discovered position
3. Asserts the row now has ``source = 'broker_sync'`` (reset path, noop here since both are broker_sync)
4. Seeds a row with ``current_units = 0`` and ``source = 'broker_sync'``
5. Issues a real upsert via ``_update_position_buy`` for an eBull BUY
6. Asserts the row now has ``source = 'ebull'`` (reset path — the important case)
7. Seeds a row with ``current_units > 0`` and ``source = 'broker_sync'``
8. Issues an eBull ADD via ``_update_position_buy``
9. Asserts the row still has ``source = 'broker_sync'`` (preserve path — also important)

The fixture pattern in ``test_operator_setup_race.py`` (``_ensure_test_db_exists`` + ``_apply_migrations_to_test_db``) is the template. Skip the test cleanly if no Postgres is reachable, same as that file does.

## Motivation

Review comment on PR #185:

[...]

PR #640, opened 2026-04-28
test(#186): integration tests for positions upsert reset-on-reopen

Luke-Bradford:
## What

Seven DB-level integration tests exercising the #185 \`positions.source\` reset-on-reopen CASE WHEN across all three call sites (\`order_client\`, \`portfolio_sync\`, \`api/orders\`). Each test runs against the real \`ebull_test\` Postgres so ON CONFLICT DO UPDATE actually fires.

## Why

Existing unit tests asserted the SQL string contained the CASE WHEN clause but never exercised the conflict branch — they passed purely because the literal text appears in the INSERT statement. PR #185 verified semantics empirically at merge time but a future incorrect rewrite (\`>= 0\` for \`<= 0\`, dropping the predicate, swapping the THEN arm) would silently regress.

## Test plan

- [x] \`uv run pytest\` 2889 pass, 1 skipped (7 new tests in \`tests/test_positions_upsert_reset_on_reopen.py\`)
- [x] Codex review: caught three real gaps across review rounds — all addressed:
  1. Initial broker_sync→broker_sync test was vacuous; replaced with tests where source values differ across branches
  2. portfolio_sync coverage initially pasted SQL; now invokes the real \`sync_portfolio()\` with a \`BrokerPortfolio\` fixture
  3. api/orders.py source-pin initially missed the THEN arm; now pins all three load-bearing CASE tokens (\`WHEN <= 0\`, \`THEN EXCLUDED.source\`, \`ELSE positions.source\`)

## Coverage layout

| Path | Reset arm | Preserve arm | Source-pin |
|------|-----------|--------------|------------|
| \`_update_position_buy\` | ✓ | ✓ | n/a (function imported, real call) |
| \`sync_portfolio\` | ✓ | (out of scope — function only inserts on external open) | n/a (function imported, real call) |
| \`api/orders.py\` | ✓ | ✓ | grep-pin asserts all three CASE tokens |

## Side effects

- Adds \`positions\` to the \`_PLANNER_TABLES\` truncation list in \`tests/fixtures/ebull_test_db.py\` so the new tests start with a clean slate.

### DIFF
--- a/tests/fixtures/ebull_test_db.py
+++ b/tests/fixtures/ebull_test_db.py
@@ -67,6 +67,9 @@ _PLANNER_TABLES: tuple[str, ...] = (
     "position_alerts",  # #396 position-alert episodes
     "watchlist",  # #042 — FK → instruments
     "broker_positions",  # #024 — FK → instruments
+    "positions",  # #186 — FK → instruments; truncated so the upsert
+    # reset-on-reopen integration tests don't see stale state from
+    # earlier in the test run.
     "quotes",  # #002 — FK → instruments (live-tick target #471)
     "instruments",
     "job_runs",
--- a/tests/test_positions_upsert_reset_on_reopen.py
+++ b/tests/test_positions_upsert_reset_on_reopen.py
@@ -332,7 +332,7 @@ def test_api_orders_buy_into_closed_position_flips_source(
     assert units == Decimal("7")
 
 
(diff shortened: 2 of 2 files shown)


Luke-Bradford:
NITPICK FIXED a33895970d60706e9c4c7bb2ead2f11689026c70 — renamed to `test_api_orders_buy_into_open_broker_position_preserves_source` to match the seeded source value.

PR #642, opened 2026-04-28
feat(#111): durable audit log on dedicated side-connection pool

Luke-Bradford:
## What

Side-connection audit-write mode for \`load_credential_for_provider_use\`. The audit row writes on a dedicated audit pool (min=1, max=2) opened in the lifespan, separate from the request pool. Caller rollback (deliberate or via exception) can no longer drop the audit row.

## Why

Per #111 / ADR 0001: every credential decryption attempt must leave an audit row on disk regardless of caller outcome. PR #110 left the audit on the caller's transaction (BLOCKING fix to avoid silent caller-state commits); this PR closes the gap by making the audit write independent.

## Test plan

- [x] \`uv run pytest\` 2894 pass, 1 skipped
- [x] Integration tests against \`ebull_test\` (\`tests/test_broker_credentials_audit_side_connection.py\`):
  - caller rollback ⇒ audit row still present
  - not_found path ⇒ audit row written even when function raises
  - cross-operator isolation
  - back-compat: audit_pool=None falls through to caller-conn write
- [x] Codex review caught two HIGH issues, both addressed:
  1. Initial impl broke 10 unit tests that don't set up app.state ⇒ \`getattr(..., None)\` back-compat path
  2. Initial impl used the shared request pool ⇒ saturation would silently drop audit. Fix: dedicated audit pool

## Trade-offs

- The dedicated audit pool can still time out on its own under DB-side issues. \`_AUDIT_POOL_TIMEOUT_S=5.0\` is a fail-fast cap so a stuck audit cannot block a credential load for the default 30s. ADR 0001 prefers an alarmed missing audit over a blocked load.
- Failures in the helper are logged loudly but do not propagate. The structured log line is the operator alarm.

## Backwards compatibility

\`audit_pool: ConnectionPool | None = None\` keeps existing callers (and tests using mocked connections) working. New callers should pass the dedicated pool from \`app.state.audit_pool\`.

## Closes

#111


Luke-Bradford:
Resolution for review BLOCKING + PREVENTION: FIXED 44761047a6895e5eca1ba744bd42e587de92a334

**BLOCKING:** `git update-index --chmod=+x .githooks/pre-push` set the stored mode to `100755`. Verified locally with `git ls-files -s` showing `100755` and `mode change 100644 => 100755` in the commit summary.

**PREVENTION:** Added a CI guard step at [`.github/workflows/ci.yml`](.github/workflows/ci.yml) that asserts the file's stored mode is `100755` before the lint job passes. EXTRACTED `docs/review-prevention-log.md` — new entry "Committed git hooks must be 100755 in the index".

Issue #651, opened 2026-04-28
tech-debt(#650): live overlay accumulates across multi-bucket REST stall


Luke-Bradford:
## Context

Follow-up to #650. The fix in PR uses a single-bar overlay (`mergeLiveBarIntoClean`) to keep the wholesale-setData fingerprint in sync with live ticks. Covers the normal case: REST refetches every 60s, live aggregator emits at most one new bucket between refetches.

## Edge case

If REST is delayed long enough for two or more new buckets before a successful refetch (background-tab timer throttling, transient refetch failure, network blip), the visible chart accumulates `history + bucket1 + bucket2 + ...` via successive `series.update()` calls. The cached fingerprint only reflects the LATEST bucket because `mergeLiveBarIntoClean` appends at most one bar past `clean.last.time`. The next successful REST catch-up returns the multi-bucket dataset → fingerprint mismatch → wholesale `setData` → one visible flicker.

Severity: low. Frequency: rare (REST stall + intraday + active SSE). Damage: one flicker after recovery, not continuous.

## Fix shape

Replace single-bar `mergeLiveBarIntoClean` with an accumulating overlay:

- `liveOverlayRef: NumericBar[]` — live bars whose time exceeds `clean.last.time`.
- onApplied:
  - If `live.time > overlay.last.time` (or empty overlay AND `live.time > clean.last.time`) → push to overlay.
  - If `live.time === overlay.last.time` → mutate overlay's last entry's OHLC.
  - If `live.time === clean.last.time` AND overlay is empty → mutate clean's tail (treat as zero-length overlay with override).
  - Else stale → ignore.
- Effect watching `clean` change: drop overlay entries with `time <= clean.last.time` (REST caught up).
- Fingerprint = `barFingerprint(range, [...clean, ...overlay])`.

Tests:
- Multi-bucket overlay: append, then append (different bucket), then `update` on latest, then `update` on earlier overlay entry — assert fingerprint reflects all.
- Overlay pruning when REST clean expands to include overlay entries.
- Stale-tick rejection when `live.time < oldest overlay or clean tail`.

Luke-Bradford:
Premise removed when #652 was reverted in #653. The single-bar overlay approach is gone with the rest of the flicker fix. If/when the flicker fix is reattempted, decide overlay accumulation strategy as part of that ticket scope, not separately.

Issue #650, opened 2026-04-28
fix: PriceChart flicker — kill 60s backstop refetch causing wholesale setData()


Luke-Bradford:
## Problem

Chart on instrument detail page flickers every 60 seconds on intraday ranges (every 5 min on daily). Operator notes this is not how mature charting systems behave — TradingView/Yahoo update silently.

## Root cause

[PriceChart.tsx:212-218](frontend/src/components/instrument/PriceChart.tsx#L212-L218):

\`\`\`tsx
useEffect(() => {
  const intervalMs = isIntraday(range) ? 60_000 : 300_000;
  const id = setInterval(() => {
    refetch();
  }, intervalMs);
  return () => clearInterval(id);
}, [range, refetch]);
\`\`\`

Refetch returns full historical candle array → effect at [PriceChart.tsx:666-708](frontend/src/components/instrument/PriceChart.tsx#L666-L708) calls \`series.setData(fullArray)\` for candle/line/area/volume series. lightweight-charts \`setData()\` is wholesale replace — clears series, redraws from scratch. That is the flicker.

Live last-bar updates via \`useLiveLastBar\` correctly use \`series.update(bar)\` (incremental, silent). Those are NOT the problem. The 60s polling backstop is.

Comment at [PriceChart.tsx:209-211](frontend/src/components/instrument/PriceChart.tsx#L209-L211) acknowledges intent:
> The backend's REST live-rate poller (#602) keeps the in-progress bar ticking at 5s; this refetch picks up any historical bar corrections eToro emits (rare). 60s on intraday is plenty.

Trade was: cheap correctness backstop vs visible flicker. Wrong call — historical corrections are rare and not worth a 60s repaint.

## Fix — preferred

Remove the 60s backstop entirely. Live SSE / REST poller already keeps the in-progress bar fresh via \`series.update()\`. Historical bar corrections (rare) caught by:

1. Range-click refetch (already happens when user changes timeframe)
2. Visibility-change refetch (when tab regains focus, refetch once — single setData on focus return is acceptable, mature charts do this)
3. Optional: explicit \"refresh\" button in chart toolbar

## Fix — fallback if backstop must stay

Gate setData by content diff:

\`\`\`tsx
const lastBarRef = useRef<{time: number; close: number} | null>(null);
useEffect(() => {
  if (!candle || clean.length === 0) return;
  const last = clean[clean.length - 1];
  const prev = lastBarRef.current;
[...]

Issue #647, opened 2026-04-28
fix: PriceChart fills DensityGrid cell dynamically (responsive height)


Luke-Bradford:
## Problem

On instrument detail page, the price/volume chart leaves whitespace below itself when the DensityGrid cell is taller than 340px. Per-instrument variation in side-rail length shifts cell width and the chart redraws width via lightweight-charts \`autoSize\`, but height stays fixed at 340px.

## Root cause

[PriceChart.tsx:792](frontend/src/components/instrument/PriceChart.tsx#L792):

\`\`\`tsx
className={containerClassName ?? "h-[340px] w-full"}
\`\`\`

Container height is hardcoded. \`autoSize: true\` ([PriceChart.tsx:446-486](frontend/src/components/instrument/PriceChart.tsx#L446-L486)) is lightweight-charts' option to fill *its* container, but the container itself is 340px.

DensityGrid cell at [DensityGrid.tsx:87-88](frontend/src/components/instrument/DensityGrid.tsx#L87-L88) uses \`lg:row-span-2\` — gives chart 2 grid rows of vertical real estate. Chart ignores it.

## Fix

1. [PriceChart.tsx:792](frontend/src/components/instrument/PriceChart.tsx#L792): \`h-[340px] w-full\` → \`h-full min-h-[340px] w-full\`.
2. Verify DensityGrid cell propagates concrete height (grid row needs explicit \`min-h\` if not implied by content).

## Test plan

- Manual: open CAL, AAPL, NVDA at viewport widths 1280, 1600, 1920 — chart fills cell at all three.
- Visual snapshot test at the 3 widths if framework permits.
- Operator UI conventions skill check: chart still legible at min height.

Luke-Bradford:
Merged via #652 (squash dfefce7).

Luke-Bradford:
Reverted in #653 — coupled fillHeight changes also reverted. Reopening to redo standalone.

Luke-Bradford:
Merged via #654.

PR #652, opened 2026-04-28
fix(#650 #647): PriceChart no-flicker refetch + responsive height

Luke-Bradford:
## Summary

- **#650** — Stop the chart wholesale-redrawing every 60s on the backstop refetch. Adds a bar-set fingerprint (range + length + every OHLCV per bar) that skips the wholesale `series.setData` when the data is unchanged. Live SSE ticks already used silent `series.update()` — only the REST polling path flickered.
- **#647** — Chart now fills the `lg:row-span-2` cell instead of sitting at fixed 340px when the right rail is taller. Added optional `fillHeight` prop on Pane plus matching flex chain through PriceChart wrapper / ChartCanvas / chart container.
- Mirrors live state into the fingerprint via a new `useLiveLastBar` `onApplied` callback so REST converging on the live bar is recognised as a no-op (no second-source-of-truth flicker).

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm test:unit` — 621 pass (607 base + 14 new): 30 PriceChart cases (fingerprint coverage for last-bar OHLCV mutations, interior-bar close, append, append → update sequencing, REST convergence) + 7 Pane cases (fillHeight on/off class plumbing).
- [x] Codex pre-push reviewed in three rounds — fingerprint robustness, fill chain end-to-end wiring, and live/REST convergence all addressed. Low-severity multi-bucket-overlay edge case deferred as tech-debt #651.


Luke-Bradford:
Resolved in 3d5ae59:

- **WARNING — ChartCanvas default**: `FIXED 3d5ae59` — reverted `containerClassName` default to `"h-[340px] w-full"`. PriceChart's wrapper passes the flex variant explicitly. Inline comment on the default warns future callers about the parent-context dependency.
- **NITPICK — Pane className double space**: `FIXED 3d5ae59` — switched to a `[...].filter(...).join(" ")` builder so empty segments do not leave double spaces.
- **PREVENTION — context-dependent prop defaults**: `EXTRACTED docs/review-prevention-log.md` — added entry under "Changing a component prop default with context-dependent semantics" so future PRs catch this class at self-review.

Re-ran `pnpm typecheck` + `pnpm test:unit` — 621 pass.

PR #653, opened 2026-04-28
revert(#652): chart broken — only live bar visible, historical hidden

Luke-Bradford:
## Summary

PR #652 broke the instrument-page price chart. After merge, both CAL and RMR (and presumably all instruments) render only the live indicator + last bar — historical bars either missing from the series or present but off-screen because the time scale is fitted to the live bar's range only.

User also reports the flicker is still present.

Reverting to last-known-good (b1ad10c). Will reopen #650 and #647 with a fresh attempt that does not regress historical rendering.

## Test plan

- [x] Pre-push gate (ruff + pyright) green
- Post-merge: confirm CAL/RMR/AAPL price chart renders historical bars + live indicator + correct time/price scale at 1D, 1M, 3M, 1Y ranges


Luke-Bradford:
**NITPICK — Pane className double space**: `REBUTTED pre-existing` — the double-space is in the original `Pane.tsx` from before #652. The revert restores it as-was. Will fix it as part of the redo of #647 alongside the proper fillHeight wiring.

Issue #644, opened 2026-04-28
fix: dividend_events schema drift — patch 054 + post-migration column verifier + Py2 syntax


Luke-Bradford:
## Problem

`sec_dividend_calendar_ingest` fails every run since 2026-04-25:

```
column de.last_parsed_at does not exist
LINE 19:                         AND de.last_parsed_at < NOW() - INTE...
```

Migration `054_dividend_events.sql` recorded as applied 2026-04-24 but `dividend_events` table is missing `last_parsed_at`. Cause: `CREATE TABLE IF NOT EXISTS` no-op'd against pre-existing table — column DDL never ran.

Bonus: [dividend_calendar.py:163](app/services/dividend_calendar.py#L163) has Py2 `except KeyError, ValueError:` syntax. Unreachable on current path; SyntaxError when reached.

## Root cause

`CREATE TABLE IF NOT EXISTS` is not idempotent for column adds. Migrations must use `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` for additions to existing tables. Migration runner has no post-apply verification — drift goes undetected until query fails in prod.

## Fix

1. Patch `054_dividend_events.sql`: append `ALTER TABLE dividend_events ADD COLUMN IF NOT EXISTS last_parsed_at TIMESTAMPTZ NOT NULL DEFAULT NOW();` — replay-safe.
2. Migration runner: each migration declares `expected_columns: dict[table, set[column]]` in adjacent `*.expect.json`. Runner asserts post-apply; raises on drift, blocks boot.
3. Fix Py2 syntax: `except KeyError, ValueError:` → `except (KeyError, ValueError):`.

## Test plan

- Pytest fixture creates pre-existing `dividend_events` table missing `last_parsed_at`, runs migration runner, asserts column present and verifier passes.
- Pytest: drop a column post-migration, assert verifier raises on next run.
- Lint catches the Py2 syntax (ruff PLE0712 or equivalent).
- Manual: run migration on broken dev DB, confirm `\d dividend_events` shows `last_parsed_at`, then `sec_dividend_calendar_ingest` succeeds.

## Prevention log

Add entry to `docs/review-prevention-log.md`: "CREATE TABLE IF NOT EXISTS does not add columns to existing tables — use ALTER TABLE ... ADD COLUMN IF NOT EXISTS for any column added after initial table creation."

Luke-Bradford:
Merged via #659.

PR #659, opened 2026-04-29
fix(#644): backfill dividend_events.last_parsed_at + Py2 except syntax

Luke-Bradford:
## Summary

- Migration 054 declared `last_parsed_at` inside a `CREATE TABLE IF NOT EXISTS` block. On any DB where the table predated 054, `IF NOT EXISTS` short-circuited the CREATE — column was never added but 054 recorded as applied. Daily `sec_dividend_calendar_ingest` then failed every run with `column de.last_parsed_at does not exist`.
- Fix: migration 082 with idempotent `ALTER TABLE … ADD COLUMN IF NOT EXISTS`.
- Bonus: `except KeyError, ValueError:` at [dividend_calendar.py:163](app/services/dividend_calendar.py#L163) was Python 2 tuple-without-parens. Path was unreachable on the previous flow so import didn't fail; fixed to `except (KeyError, ValueError):`.
- Prevention log entry added so future new-table migrations pair with idempotent ALTER.

Out of scope (deferred to a follow-up): post-migration column verifier infrastructure that fails boot on schema drift.

## Test plan

- [x] `uv run ruff check .` clean
- [x] `uv run ruff format --check .` clean
- [x] `uv run pyright` clean
- [x] `uv run pytest tests/test_dividend_calendar_schema.py` — 6 pass (4 parser + 2 migration)
- [x] Migration applied locally; column verified present


Luke-Bradford:
**WARNING — misleading PEP 758 comment**: `FIXED 18a5637` — bot was right. PEP 758 (3.14) makes `except A, B:` legal as a tuple-of-types catch and project pins requires-python>=3.14 so ruff format strips the parens. Removed all SyntaxError / Python-2 framing from the code comment, test docstring, and original commit message. The migration is the only real fix in this PR now; `_parse_date` tests stay as contract regression coverage. Added a new prevention log entry under "Don't claim except A, B: is Python 2 syntax on a 3.14+ project" so a future contributor doesn't repeat the mistake.

**PREVENTION — exception-syntax comment template**: `EXTRACTED docs/review-prevention-log.md`

Luke-Bradford:
**NITPICK — misleading test name**: `FIXED 5c24f9e` — renamed to `test_parser_short_circuits_on_falsy_input`. Docstring now notes the KeyError arm is dead code in practice (regex `m1` group is built from `_MONTHS.keys()` so any match guarantees the lookup succeeds; except clause stays as defensive belt-and-suspenders).

PR #664, opened 2026-04-29
fix(#650): no skeleton during background refetch — chart stays mounted

Luke-Bradford:
## Summary

- 60s backstop refetch flipped useAsync into `loading=true` for a frame; SectionSkeleton rendered above the chart in the flex column, pushed chart down + back up when it disappeared — the "browser refresh" visual the operator reported.
- Conservative fix: gate skeleton on `!dataMatchesRange` (initial load OR range switch — both operator-initiated) instead of the broad `effectivelyLoading`. Background refetch with valid same-range data → no skeleton → no layout shift.
- **Strictly NOT touched** (last attempt regressed historical bar visibility — see PR #652 revert): flex chain / sizing classes, ChartCanvas internal layout, setData fingerprint.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm test:unit` — 626 pass; new describe block has 3 cases pinning the contract:
  - No skeleton during same-range 60s backstop refetch
  - Skeleton DOES render on initial load
  - Skeleton DOES render on range switch

### DIFF
--- a/frontend/src/components/instrument/PriceChart.test.tsx
+++ b/frontend/src/components/instrument/PriceChart.test.tsx
@@ -514,6 +514,13 @@ describe("PriceChart — no-flicker on background refetch (#650)", () => {
     await waitFor(() => {
       expect(screen.getByTestId("price-chart-AAPL")).toBeInTheDocument();
     });
+    // Once the new range data lands, the skeleton must vanish — the
+    // skeleton-shown half of this contract is asserted above; the
+    // skeleton-hidden half belongs here so a regression that left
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
**NITPICK — collapse showSkeleton**: `FIXED bbe7629` — collapsed to `!dataMatchesRange && error === null`. Comment kept explaining why.

**NITPICK — skeleton-vanishes assertion**: `FIXED bbe7629` — added `queryByRole("status")` not-in-document check after the range-switch fetch resolves.

PR #673, opened 2026-04-29
feat(#590): dividends drill upgrade — DPS, cumulative, payout, yield-on-cost

### LINKED ISSUE #590: feat(#585): dividends drill upgrade — DPS line + payout ratio + yield-on-cost + cumulative
## Context

Fifth sub-ticket of #585. Builds on #580 (dividends drill page already exists; this upgrades the visual layer with charts).

Spec: `docs/superpowers/specs/2026-04-27-instrument-charts-quant-redesign-design.md` (Dividends section).

## What

`frontend/src/pages/DividendsPage.tsx` — add 4 new charts (recharts):

1. **DPS over time line chart** — declared dividend per share per quarter, full history
2. **Payout ratio trend** — dividends paid / FCF over time. Needs FCF data from cashflow statement (verify XBRL coverage)
3. **Yield-on-cost progression** — when the operator HOLDS the instrument, show yield-on-cost (current annual DPS / their cost basis) progression. Reads from positions endpoint. Hide the chart if not held.
4. **Cumulative dividends paid** — running sum of dividends per share over time, line chart

Existing per-quarter bar list + per-FY totals table stay (becomes L3-ish — could be moved to `?view=raw` or kept as-is). Recommend: keep both, charts render at top of page, raw tables below.

## Acceptance

- DividendsPage renders 4 new charts when data is present
- Yield-on-cost chart only renders when position exists; otherwise hidden
- Payout ratio chart shows empty state if FCF data missing (per chart, not page-wide)
- Existing tests still pass; new tests for DPS line + cumulative computation

## Dependencies

- #586 (chart theme + recharts) must merge first
- Verify FCF data in cashflow statement for payout ratio
- Position cost basis from existing `/instruments/{id}/positions` endpoint

## Linked

Parent: #585. Predecessor: #580 (drill page + route established).

Luke-Bradford:
Closes #590 (sub-ticket of epic #585).

## What

Layers four recharts panes above the existing raw history list on \`/instrument/:symbol/dividends\`:

1. **DPS over time** — line chart, declared per period
2. **Cumulative DPS** — area chart, running total with null gaps (missing-period reads as gap, not a misleading flat)
3. **Payout ratio** — line chart, \`dividends_paid / FCF\` annual; \`abs()\` on the dividends_paid sign quirk, FCF≤0 clamped to null
4. **Yield-on-cost** — bar chart bucketed by fiscal_year; only renders when held with \`avg_entry > 0\`

**Cadence dedupe** in \`lib/dividendsMetrics.ts::dropOverlappingFy\` protects aggregated metrics: when a fiscal_year has both quarterly + FY rows, the FY aggregate is dropped before summing. Without it, both cumulative and YoC would double-count any year where the 10-K aggregate lands alongside the four 10-Q rows. The DPS line chart skips the filter so raw audit-grade data still surfaces.

Page now fetches dividends + summary + cashflow (annual) + position in parallel. Position fetch keys on \`instrument_id\` from summary; 404 on the position endpoint is treated as "not held" rather than a page error. Payout-ratio pane has independent loading / error / retry so a cashflow failure doesn't break the rest of the page.

## Why

Operator review 2026-04-27 flagged the existing dividend pane as too thin — wanted DPS-over-time + cumulative + payout-ratio + yield-on-cost as standard quant signals. Yield-on-cost is the operator-relevant story when held: as DPS grows over time and the cost basis stays fixed, the "yield against my entry price" rises in a way that current dividend yield doesn't show.

## Test plan

- \`pnpm --dir frontend test:unit\` — 713 tests pass (15 new on top of existing 698)
- Codex spec/diff review caught the BLOCKING cadence-double-count, WARNING cumulative null-as-zero, and WARNING avg_entry≤0 not hiding the YoC pane. All addressed before push.
- Manual: open \`/instrument/AAPL/dividends\` once dev stack is up; verify YoC pane appears only on held instruments and disappears for non-held; toggle to a non-dividend payer (e.g. GME) and verify the original "no dividend data" empty state still fires.


Luke-Bradford:
## Review resolution

**Comment 1 — WARNING (`dividendsCharts.tsx:166`)**: `FIXED f7bdd7e` — widened the cumulative-DPS empty-state guard. After round-2's null-propagation refactor (`cumulative_dps: null` for missing source rows), the prior `=== 0` guard didn't fire on an all-null history because `null === 0` is false. The new guard walks the series for any non-null cumulative value; only when none exists *or* the last non-null value is exactly 0 does the inline "no data" hint render. Regression pinned by `dividendsCharts.test.tsx` ("renders the inline no-data hint when every row has null dps_declared").

**Comment 2 — PREVENTION**: `EXTRACTED docs/review-prevention-log.md` — added "Empty-state guards must follow the data shape after a null-handling refactor" with self-review and bot-review prompts (after editing a `build*` helper, grep call sites for empty-state guards and re-verify each branch against the new return shape; every chart with an inline 'no data' branch needs a vitest case where every row has a null in the relevant field).

Luke-Bradford:
## Review resolution — round 2

**Comment 1 — BLOCKING (`DividendsPage.tsx:170`)**: `REBUTTED` — false positive. `useAsync<T>` returns `data: T | null`, never `undefined`:

- [`frontend/src/lib/useAsync.ts:31`](frontend/src/lib/useAsync.ts#L31): `useState<T | null>(null)` — initial value null.
- [`frontend/src/lib/useAsync.ts:50`](frontend/src/lib/useAsync.ts#L50): every fetch effect explicitly `setData(null)` before `fnRef.current()` runs.

In this usage `useAsync<InstrumentPositionDetail | null>(...)` so the in-flight value is `null` and the resolved value is either `null` (404 caught by `fetchPositionOrNull`) or a full `InstrumentPositionDetail`. The memo at line 170 does `const pos = position.data; if (pos === null) return null;` — early return fires during loading, `pos.total_units` is never reached against an undefined.

Codex verified this rebuttal independently against the `useAsync` source.

**Comment 2 — PREVENTION**: `REBUTTED` — the underlying rule (handle the loading path before property access) is sound, but it's already enforced by the typed `useAsync` contract (`data: T | null` not `T | null | undefined`). Adding it as a project-wide prevention would only matter if we adopted a different async-state hook with three-state semantics (e.g. `@tanstack/react-query`'s `{ data: T | undefined, isLoading: bool }`). When that happens we'll file the prevention against the migration PR, not this one.

Issue #678, opened 2026-04-29
bug: sync orchestrator crashes + lock contention with legacy cron (~20% sync_runs failing)


Luke-Bradford:
## Symptom

\`SELECT status, COUNT(*) FROM sync_runs WHERE started_at > NOW() - INTERVAL '24h' GROUP BY status\` →

| status | count |
|---|---|
| complete | 260 |
| failed (orchestrator_crash) | 68 |
| partial | 9 |

68 / 337 = **20% of sync runs crashed** in the last 24h. \`error_category='orchestrator_crash'\` on every failure.

\`SELECT layer_name, status, skip_reason, COUNT(*) FROM sync_layer_progress WHERE started_at > NOW() - INTERVAL '24h' GROUP BY ...\` →

- **fundamentals: 8 skipped** with \`skip_reason='legacy cron holder active (JobLock busy)'\`
- **fundamentals: 1 outright crashed** with \`error_category='orchestrator_crash'\`
- **candles: 2 skipped** with \`legacy cron holder active\`

## Diagnosis

The data orchestrator (#260 epic) and the legacy daily cron jobs (\`daily_research_refresh\`, \`daily_candle_refresh\`) share the same per-layer JobLock. When the orchestrator schedules a layer that the legacy job is also running, the orchestrator's layer skips with \`legacy cron holder active\` — that's by design (don't double-fetch SEC). But:

1. **20% of orchestrator manual triggers crash outright** instead of skipping cleanly. The crash trace isn't on \`sync_layer_progress.error_message\` (column is None for these rows). Root cause needs a stack-trace capture path.

2. **Lock contention is a regression** — the legacy cron isn't supposed to be running concurrently with the orchestrator post-#260 rollout. Either:
   - The legacy cron wasn't fully decommissioned and is still scheduled
   - The orchestrator is firing on top of the legacy cron's window because the windows weren't aligned
   - JobLock release is slow (long-running fundamentals refresh holds the lock for ~5min)

## Verification

- \`sync_runs WHERE error_category='orchestrator_crash' ORDER BY started_at DESC LIMIT 5\` — repeatable
- \`job_runs WHERE job_name='daily_research_refresh' ORDER BY started_at DESC\` confirms the legacy cron is still running daily

## Fix options

[...]

PR #679, opened 2026-04-29
fix(#669): bulk-resolve identifiers in refresh_filings; aggregate skip logs

### LINKED ISSUE #669: chore: aggregate "no sec/cik" skip logs in filings refresh
## Problem

`app/services/filings.py:62-67` logs `Filings: no sec/cik for instrument_id=X, skipping` at INFO level **once per non-US instrument** on every refresh tick. Universe has thousands of non-CIK instruments (crypto, FX, non-US equities), so each refresh floods the dev terminal and prod log aggregator with identical skip lines.

Screenshot from operator 2026-04-29 shows ~25 lines per second during a refresh.

## Fix

Aggregate. Replace per-row log with a single line at end of `refresh_filings`:

```python
logger.info(
    "Filings: skipped %d instruments missing %s/%s identifier",
    skipped_no_identifier,
    provider_name,
    identifier_type,
)
```

Track `skipped_no_identifier` separately from `skipped` (which counts both missing-identifier + provider-error skips today) so the new line only fires for the per-instrument case.

Same treatment for the equivalent in `app/services/fundamentals.py` (`Fundamentals: no data from provider for X, skipping` at lines 87, 131) which has the same per-instrument shape.

## Why not DEBUG

Aggregation preserves "I refreshed N, skipped M for missing identifier" observability — a count is the useful signal, not the per-instrument list. The per-instrument view is reachable via `external_identifiers` query when needed.

## Acceptance

- One aggregate line per `refresh_filings` invocation
- Same for `refresh_fundamentals`
- pytest covers the aggregate counter
- Dev terminal stays usable during refresh

Luke-Bradford:
Closes #669.

## What

Operator complaint 2026-04-29: terminal flooded with \`Filings: no sec/cik for instrument_id=X, skipping\` at ~25 lines/sec, dev session unusable. Original ticket asked for log aggregation; operator's follow-up sharpened the fix: *\"do we not know what we need to pull, is there not a single source showing what has been modified\"*.

Bulk-resolve identifiers in one SELECT before the loop runs, then iterate only the resolved subset:

\`\`\`python
def _bulk_resolve_identifiers(conn, instrument_ids, provider_name, identifier_type) -> dict[str, str]:
    # SINGLE query against external_identifiers; returns {instrument_id: identifier_value}
    # for every row with a primary identifier of (provider_name, identifier_type).
\`\`\`

Replaces the prior pattern of one \`_resolve_identifier\` SELECT per instrument plus one INFO log per miss. With a 12k-instrument universe and ~7k lacking SEC CIKs, that was ~7k DB roundtrips and ~7k log lines per refresh tick.

Same aggregation pattern applied to \`refresh_fundamentals\` + \`refresh_fundamentals_history\` (per-symbol \"no data from provider\" / \"no data in range\" → single aggregate INFO line at the end).

The provider-error WARNING (per-row, with traceback) still fires — transient errors should surface individually.

## Why

Operator-stated impact: \"site is unusable\" while the daily research refresh is running. Per-row INFO emission is also doing ~25 syscalls/sec to the log handler, on top of ~25 DB roundtrips/sec for the per-row identifier lookup. Single aggregate run should drop both to one each.

## Test plan

- \`uv run pytest tests/test_filings_bulk_resolve.py\` — 4 new tests pass (live ebull_test DB)
- \`uv run pytest tests/\` — 2963 pass, 1 skipped
- \`uv run ruff check .\` + \`uv run ruff format --check .\` — clean
- \`uv run pyright\` — 0 errors
- Codex round-1 clean (no findings)
- Manual: tail \`uvicorn\` logs after merge during the next \`daily_research_refresh\` run; expect a single \"Filings: X/N instruments skipped\" line at completion rather than per-row spam

## Linked

Surface symptom: operator review 2026-04-29 (running fundamentals refresh making terminal unusable). Companion ticket #678 tracks the orchestrator-crash + lock contention separately.


Luke-Bradford:
## Review resolution

**Comment 1 — WARNING (`filings.py:64,69`)**: `FIXED 74eb479` — dropped the `instrument_id::text` cast. Coerced the parameter list to `int` once at the top of `_bulk_resolve_identifiers` (with a defensive try/except for non-numeric ids logged at DEBUG), then bound `instrument_id = ANY(%s)` directly so Postgres uses the primary-key index on `external_identifiers`. The result map's keys are `str` so downstream `refresh_filings` iteration stays the same.

**Comment 2 — NITPICK (`tests:78`)**: `FIXED 74eb479` — rewrote the docstring to describe the actual fixture-rollback teardown, dropped the misleading SAVEPOINT claim.

**Comment 3 — PREVENTION**: `ALREADY_COVERED ./docs/review-prevention-log.md` — though this exact rule ("index-defeating cast") is new for this repo. Adding it as a fresh entry to the prevention log: column-side casts in WHERE clauses defeat indexes; coerce parameters before binding instead.

PR #680, opened 2026-04-29
feat: scripts/force_refresh_fundamentals.py — targeted SEC refetch

Luke-Bradford:
Operational script unblocking the dividend-chart gap for the MLP cohort affected by #674. Companion ticket #677 covers the proper API + UI surface; this is the small quick-win.

## What

\`scripts/force_refresh_fundamentals.py\` — argparse CLI, dry-run by default. Resolves \`<SYMBOL>...\` to (instrument_id, primary SEC CIK) pairs in one query, dedupes by instrument_id, then calls \`refresh_financial_facts\` + \`normalize_financial_periods\` scoped to those instruments.

\`\`\`
uv run python -m scripts.force_refresh_fundamentals IEP ET EPD MPLX KMI
\`\`\`

(Bare \`uv run python scripts/foo.py\` doesn't work on this layout — script docstring spells out the module form.)

## Why

\`plan_refresh\`'s master-index selector ([fundamentals.py:1506](app/services/fundamentals.py#L1506)) only schedules CIKs SEC has filed something new for in the last 7-day window. Schema/allowlist changes (#674) don't qualify, so the affected MLPs would wait until SEC ships them a new filing — could be days/weeks — before the dividend chart fills in. This script bypasses that gate.

Scope is deliberately narrow: no new HTTP API, no UI, no expected-filings poller. All of that lives in #677. Operational quick-win first; proper feature later.

## Test plan

- \`uv run pytest tests/test_force_refresh_fundamentals.py\` — 4 tests pin the resolver contract (CIK-having vs missing, case-insensitive, duplicate tolerance, empty input)
- \`uv run pytest tests/\` — 2965 pass, 1 skipped
- \`uv run ruff check . / format / pyright\` — clean
- Local dry-run smoke: \`IEP ET EPD MPLX KMI BIP\` resolves 5/6, surfaces MPLX as missing CIK
- Codex review across 3 rounds: 3 substantive findings (open transaction across HTTP, false-success exit code, no input dedupe) — first two fixed in round 2, third fixed in round 2; round 3 surfaced a pre-existing upstream pattern (\`refresh_financial_facts\` wraps SEC HTTP in \`conn.transaction()\` at [fundamentals.py:533](app/services/fundamentals.py#L533)) that's out of scope for an ad-hoc script — same pattern is used by the daily cron path

## Operational note

After merge, recommend running for the affected cohort to populate the dividend chart for IEP and the other MLPs:

\`\`\`
[...]

PR #684, opened 2026-04-29
feat: glossary + <Term> tooltips, filings dupe fix, MLP fundamentals pane

Luke-Bradford:
Operator review 2026-04-29 against the IEP instrument page surfaced three operator-visible issues, all addressed here:

## 1. Abbreviation comprehension

Built [lib/glossary.ts](frontend/src/lib/glossary.ts) — typed map of ~37 entries with \`shortName\` + \`what\` + \`why\` for SEC filing types (8-K / 10-K / 10-Q / 20-F / Form 4 / 13D / 13G), regulatory terms (CIK / SIC / EDGAR / XBRL / Filer category / FPI), valuation ratios (P/E / P/B / ROE / ROA / ROIC / DuPont / FCF / EPS / TTM / DPS / Yield-on-cost / Payout ratio), and eBull internal labels (Tier / PM / AH / NET 90d / TXNS).

Built [Term.tsx](frontend/src/components/Term.tsx) — renders as \`<abbr>\` with both native \`title\` + a richer popover on hover/focus. Keyboard-accessible (Tab + focus → tooltip).

Sweep wraps abbreviations across \`InsiderActivitySummary\`, \`SecProfilePanel\`, \`SummaryStrip\` Tier badge, \`dividendsShared\` (TTM yield / TTM DPS / Latest DPS), \`KeyStatsPane\` (P/E / P/B / ROE / ROA / Debt / Equity / Payout ratio).

## 2. Filings list duplicate-chip bug

Operator screenshot showed every row with the form-type chip rendered TWICE (\"8-K  8-K\", \"10-K  10-K\") when the XBRL \`extracted_summary\` field was null. Cause: the third-column fallback used \`f.filing_type\` literally, which already lives in the chip beside it.

Fix at [FilingsPane.tsx:144](frontend/src/components/instrument/FilingsPane.tsx#L144) — fallback now uses \`filingTypeFriendlyName()\` from the glossary, so the row reads \"8-K  Material event\" / \"10-K  Annual report\" instead of the dupe. Form-type chip itself wraps in \`<Term>\` so hovering reveals what each code means. Regression test added.

## 3. MLP fundamentals sparkline pane disappeared after force-refresh

Symptom: \`/instrument/IEP\` lost the small Revenue / Op income / Net income / Total debt sparklines under the price chart.

Root cause: IEP and other partnership/MLP issuers file \`IncomeLossFromContinuingOperations\` instead of \`OperatingIncomeLoss\` (the standard us-gaap concept). \`operating_income\` is null on every row. Pre-fix, \`FundamentalsPane.joinPeriods\` dropped a row if ANY of revenue/op_income/net_income was null — so MLPs silently hid the entire pane. Counter-intuitive trigger: my [PR #680](../pull/680) force_refresh_fundamentals.py rewrote IEP's rows from authoritative SEC source, clearing legacy stale operating_income values that previously kept the pane visible.

Fix at [FundamentalsPane.tsx:38](frontend/src/components/instrument/FundamentalsPane.tsx#L38) — gate a row out only when ALL three flagship metrics are null, surface what's available per cell. Op-income cell on IEP now renders \"—\" rather than killing the sibling sparklines. Regression test pins the partnership-issuer case.

## Also

Report-level subtitle quips added/expanded on \`InsiderPage\`, \`FundamentalsPage\`, \`DividendsPage\` — each explains what the report covers in plain English with key terms wrapped in \`<Term>\` for hover-detail.

## Test plan

- \`pnpm --dir frontend test:unit\` — 731 pass (15 new: 9 glossary + 5 Term + 1 FilingsPane regression + 1 FundamentalsPane MLP regression)
- Manual: open \`/instrument/IEP\` after merge — Recent Filings rows should show \"8-K Material event\" / \"10-K Annual report\" instead of dupe chips. Fundamentals sparkline pane should reappear with Revenue / Net income / Total debt populated and Op income showing \"—\". Hovering any wrapped abbreviation (CIK / SIC / TTM / P/E etc.) opens the tooltip popover.

## Out of scope

Layout density redesign (operator's other ask in the same review thread) tracked separately for the next PR — would touch \`DensityGrid.tsx\` + the per-pane width caps, lower-risk to ship in isolation.

### DIFF
--- a/frontend/src/components/Term.test.tsx
+++ b/frontend/src/components/Term.test.tsx
@@ -1,0 +1,46 @@
+import { fireEvent, render, screen } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+
+import { Term } from "@/components/Term";
+
+describe("Term", () => {
+  it("renders the term as an <abbr> with a native title attr for fallback tooltips", () => {
+    render(<Term term="CIK" />);
+    const el = screen.getByTestId("term-CIK");
+    expect(el.tagName.toLowerCase()).toBe("abbr");
+    expect(el.getAttribute("title")).toContain("SEC entity ID");
+  });
+
+  it("opens the rich tooltip popover on hover and closes on mouseleave", () => {
+    render(<Term term="ROE" />);
+    expect(screen.queryByRole("tooltip")).not.toBeInTheDocument();
+    fireEvent.mouseEnter(screen.getByTestId("term-ROE"));
+    const tip = screen.getByRole("tooltip");
+    expect(tip).toHaveTextContent(/Return on equity/i);
+    expect(tip).toHaveTextContent(/Why it matters:/i);
+    fireEvent.mouseLeave(screen.getByTestId("term-ROE"));
+    expect(screen.queryByRole("tooltip")).not.toBeInTheDocument();
+  });
+
+  it("opens on keyboard focus so screen readers + tab navigation work", () => {
+    render(<Term term="P/E ratio" />);
+    fireEvent.focus(screen.getByTestId("term-P/E ratio"));
+    expect(screen.getByRole("tooltip")).toBeInTheDocument();
+    fireEvent.blur(screen.getByTestId("term-P/E ratio"));
+    expect(screen.queryByRole("tooltip")).not.toBeInTheDocument();
+  });
+
+  it("renders the children override as the visible label, looking up under the term key", () => {
+    render(<Term term="P/E ratio">P/E</Term>);
+    const el = screen.getByTestId("term-P/E ratio");
+    expect(el).toHaveTextContent("P/E");
+    expect(el).not.toHaveTextContent("P/E ratio");
+  });
+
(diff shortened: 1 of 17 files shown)


Luke-Bradford:
## Review resolution

**Comment 1 — WARNING (`FundamentalsPane.tsx:181`)**: `FIXED 00691e4` — refactored the per-cell rendering into a `FundamentalsGrid` subcomponent that computes `maxLen` (largest period count across the four siblings) and passes it down. `FundamentalCell` now renders a `{n}/{maxLen} periods` caption in amber when its coverage diverges from the max. Skipped when `values` is empty (the empty cell + "—" label already reads as "no data"; the caption matters for partial coverage like 6/8). Regression test added.

**Comment 2 — PREVENTION**: `EXTRACTED inline` — the `maxLen` parameter on `FundamentalCell` documents the convention. Adding it to `docs/review-prevention-log.md` would be over-indexed (sparkline-coverage is a per-component concern, not a repo-wide pattern); locking it in via the per-cell test.

PR #688, opened 2026-04-29
feat: bento-grid DensityGrid v3 + null-out-safe empty states

Luke-Bradford:
Operator review against #686 + #687: dividends panel still sat at half-width with empty grid cell beside it; mid-page sparkline strip stretched all 4 cells across 12 cols. Operator: \"these should be bento boxes\" — invoked \`/ui-ux-pro-max\` skill, picked the **Bento Box Grid** pattern (Apple/dashboard tile aesthetic, varied 1×1 / 2×1 spans, rounded-xl chrome, subtle hover lift).

### 1. \`DensityGrid\` bento allocator

New \`allocateTiles()\` helper filters out absent panes BEFORE the grid sees them and assigns deterministic column spans by surviving count:

| Surviving | Spans          |
|-----------|----------------|
| 1         | col-span-12    |
| 2         | 6 + 6          |
| 3         | 4 + 4 + 4      |
| 4         | 3 + 3 + 3 + 3  |

Stat-block row (Filings + Insider + Dividends) routes through it so the row always fills regardless of which subset is present.

Mid row pairs **Fundamentals (8) + RecentNews (4)** when fundamentals active — the 4-cell sparkline strip wants the wider tile, news list reads fine in the narrower one. Falls through to news-only when fundamentals inactive.

Minimal profile puts \`ThesisPane\` in the \`lg:col-span-4\` right-rail slot under KeyStats, mirroring the SecProfile pattern in full-sec.

### 2. Late-null-render ghost cells fixed

\`RecentNewsPane\` + \`DividendsPanel\` previously returned \`null\` when data was empty post-load — left the bento grid's pre-allocated column blank (Codex caught this). Both now render Pane chrome with an in-pane empty state. \`DividendsPanel\` keeps its provider source tag + \`Open →\` drill in the empty case so the operator can still navigate to the L2 page.

### 3. Pane chrome bento aesthetic

\`rounded-md\` → \`rounded-xl\` + \`transition-shadow\` + \`hover:shadow-md\` per ui-ux-pro-max guidance. Subtle hover lift on every card without the layout-shifting hover-scale that #687 explicitly avoided.

## Test plan

- \`pnpm --dir frontend test:unit\` — 732 pass (2 tests updated for new empty-state contract)
- Codex 2 rounds: round-1 caught minimal-profile right-rail dead-end + ghost cells (resolved); round-2 caught empty DividendsPanel losing source/onExpand props (resolved). 
- Manual: reload \`/instrument/IEP\` — Filings + Insider + Dividends should fill row 4+4+4 cleanly; Fundamentals strip + News should pair 8+4; no empty grid cells when capability is on but data is empty.

## Out of scope

The right-rail-thesis pattern is currently minimal-profile-only. If operator wants thesis in the right rail across other profiles, that's a small follow-up.

### DIFF
--- a/frontend/src/components/instrument/DensityGrid.tsx
+++ b/frontend/src/components/instrument/DensityGrid.tsx
@@ -1,21 +1,32 @@
 /**
- * DensityGrid — 12-column capability-aware grid for the instrument
- * Research tab (#575). Three profiles determine which panes render:
+ * DensityGrid — bento-grid instrument page (#684 round 3).
  *
- *   full-sec        — fundamentals (sec_xbrl) + filings active
- *   partial-filings — filings active but no sec_xbrl fundamentals
- *   minimal         — no filings capability at all
+ * Operator review 2026-04-29: previous flat 12-col layout had two
+ * recurring gripes —
+ *   1. Half-width tiles (Dividends at lg:col-span-6) sat next to
+ *      empty grid cells when their would-be neighbour wasn't
+ *      capability-active — visible dead space.
+ *   2. Static row layouts (Filings col-span-7 + Insider col-span-5)
+ *      didn't re-flow when one of the panes was absent — left a
+ *      lone narrow stat block stretched across the row.
  *
- * PriceChart and KeyStatsPane are present in all profiles.
- * SecProfilePanel / BusinessSectionsTeaser gate on has_sec_cik.
- * FundamentalsPane is exclusive to the full-sec profile.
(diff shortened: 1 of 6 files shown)


Luke-Bradford:
Superseded by full design-system v1 redesign per operator review — borderless Bloomberg-style chrome, real Recharts on Fundamentals, tabular-nums, bento v4 row reorder. New PR will land all in one.

PR #718, opened 2026-04-30
fix(#717): harden dev-stack pools + scope uvicorn watcher

### LINKED ISSUE #717: fix(infra): dev stack wedged after long uptime — pool dead-conn detection + scoped uvicorn reload
## Symptom

2026-04-30 mid-session: backend (`PID 46868`, child of `uvicorn --reload`) ran ~6h, then went unresponsive. \`/health\` blocked >30s. Frontend (`5173`) hung waiting on backend proxy. Operator had to restart the dev stack manually.

Worker process: alive, listening on 127.0.0.1:8000, only 19s CPU consumed across 6h, 3 threads — event loop wedged, not CPU-pinned.

Postgres in container (`ebull-postgres`, up 7 days): healthy. 10 connections total, 9 idle, 1 active. Direct \`docker exec ... psql\` instant.

External \`psycopg.connect()\` through Docker port-forward: 5+ seconds. \`pool.connection()\` from inside the worker: never returned.

## Root cause hypothesis

`ConnectionPool` opened in [\`app/main.py:99\`](../blob/main/app/main.py#L99) with bare config: \`min_size=1, max_size=10\`. No \`check\` validator on checkout, no \`max_idle\` recycle, no TCP keepalives in \`kwargs\`. Same for \`audit_pool\` at line 112.

When Docker Desktop pauses or the host TCP stack reaps an idle forwarder socket, the pool's cached conn becomes half-open. Without OS-level keepalives or per-checkout validation, pool serves the dead conn and the next \`.execute()\` blocks the asyncio event loop forever. Subsequent /health requests pile up behind it. Long-uptime + many uvicorn `--reload` cycles compound the risk.

A second contributor: \`uvicorn --reload\` watches every \`.py\` in cwd by default — including \`tests/\`. During a Claude-driven edit storm (e.g. iterating on test files), the worker reloads dozens of times. Reload-mid-startup races leave conns stranded.

## Fix

1. **Hardened pool config** — `kwargs={keepalives: 1, keepalives_idle: 30, keepalives_interval: 10, keepalives_count: 3}` for OS-level dead-peer detection; `check=ConnectionPool.check_connection` for SELECT 1 validation on every checkout (~1ms overhead); `max_idle=600` to recycle conns idle >10min; `max_lifetime=1800` to rotate every 30min; `timeout=15` to cap checkout wait.
2. **Narrow uvicorn watcher** — \`--reload-dir app/\` so test/doc edits don't trigger backend reload.
3. **Prevention log entry** documenting the failure mode + the dual-defence config.

## Acceptance

- pool helper applied to both `db_pool` and `audit_pool`
- /health responds in <1s after 8h+ uptime, including overnight host-pause window
- VS Code task command updated to `--reload-dir app/`
- prevention log entry merged

Luke-Bradford:
## What

Prevent the dev backend from wedging after long uptime. Three layers:
1. ConnectionPool config (TCP keepalives + check on checkout + max_idle/lifetime + checkout timeout) via new `_open_pool` helper.
2. `get_conn` maps `PoolTimeout` to HTTP 503 so Depends-backed routes surface a wedged pool cleanly.
3. `uvicorn --reload-dir app` so test/doc edits don't churn the worker.

## Why

Backend wedged after ~6h uptime today (#717): worker alive, listening, only 19s CPU consumed total, `/health` blocked >30s. Postgres in `ebull-postgres` container was healthy (10 conns total, 9 idle). The bare `ConnectionPool(min_size=1, max_size=10)` had no keepalives, no validator, no max_idle — when a Docker port-forward socket silently died, the pool kept handing out the half-open conn and every checkout then blocked the asyncio loop forever.

## Test plan

- [x] `uv run pytest tests/test_main_pool_hardening.py` — 3 new tests pass (config kwargs pinned, real-class identity, fake-wedged-pool → 503)
- [x] `uv run pytest` full suite — 2984 passed
- [x] `uv run pyright`
- [x] Codex pre-push: round 1 flagged that the 503 claim only held for `/health` inline catch; fixed in-flight by adding `PoolTimeout`→503 mapping in `get_conn`. Round 2 clean.

## Notes

- `_open_pool` is the central seam — adding a third pool with raw `ConnectionPool(...)` is the regression shape the prevention-log entry guards against.
- The `--reload-dir app` change means edits under `tests/`, `docs/`, `frontend/` no longer trigger backend restarts. Saves dev-loop churn during Claude-driven edits.

Closes #717.


Luke-Bradford:
WARNING [`app/db/__init__.py:43`]: **FIXED** `4bb898c`. Switched from `try/with/yield/except` to `contextlib.ExitStack` + `enter_context` so the `PoolTimeout` catch is scoped strictly to checkout. A `PoolTimeout` raised inside a handler body now propagates untouched. Regression test `test_get_conn_does_not_swallow_handler_pool_timeout` asserts a handler-side raise does not surface as a 503.

PREVENTION (Generator dependency exception scope): **EXTRACTED** in commit `4bb898c` — the prevention-log entry on dev-stack pool defences (`docs/review-prevention-log.md`) names `_open_pool` + the new ExitStack pattern as the canonical shape. A future regression (a fourth pool elsewhere, or another generator dependency that wraps a yield with a too-broad except) is caught by the regression test plus the prevention-log entry pointing the reviewer at the correct shape.

PR #719, opened 2026-04-30
feat(#719): jobs out-of-process — durable queue + heartbeat + isolated runtime

Luke-Bradford:
Closes #719.

## What

Move `JobRuntime` (APScheduler + manual-trigger executor + sync orchestrator's executor + reaper + boot freshness sweep) out of the FastAPI process and into a dedicated long-running process (`python -m app.jobs`). API restarts no longer kill jobs; job hangs no longer consume API threads or pool conns.

## Why

The dev backend went unresponsive repeatedly during normal Claude-driven editing — root cause was the API process owning APScheduler, the manual-trigger executor, and the orchestrator's executor. A single hung outbound HTTP inside a job thread stranded a Postgres advisory lock + idle-in-transaction conn; uvicorn `--reload` racing against in-flight jobs left workers half-started; `time.sleep` in resilient_client backoff blocked the asyncio loop. Multiple band-aid attempts (#717 pool hardening, `--reload-dir app`, etc.) didn't solve the underlying coupling.

## How

Postgres-only IPC. Durable trigger queue (`pending_job_requests`) + `pg_notify('ebull_job_request', ...)` as a wakeup hint + 5s `FOR UPDATE SKIP LOCKED` poll fallback. Singleton fence on a dedicated long-lived `psycopg.Connection` holding `pg_try_advisory_lock(JOBS_PROCESS_LOCK_KEY)` for process lifetime — running `python -m app.jobs` twice is a hard FATAL exit. Per-subsystem heartbeat (`scheduler` / `manual_listener` / `queue_drainer` / `main`) + supervisor that restarts a stalled listener within 60s.

Locked 13-step startup ordering: pool → fence → reaper → JobRuntime construct → queue stale-row recovery → boot-drain → scheduler.start → catch-up → boot freshness sweep → listener → heartbeats → main loop. Shutdown reverses; fence connection closes LAST.

API rewrites: `POST /jobs/{name}/run` and `POST /sync` validate + INSERT into `pending_job_requests` + NOTIFY, return 202 with `request_id`. New `GET /jobs/requests` exposes the queue with `?request_id`/`?status`/`?job_name`/`?request_kind` filters. `/system/jobs` aggregates per-subsystem heartbeat into `jobs_process_state` ∈ {healthy, degraded, down}.

`submit_sync` and `set_executor` deleted entirely. `run_sync` stays as the canonical synchronous entry point (jobs-process dispatcher invokes it on its own max-workers=1 sync executor; tests/CLI scripts call it directly). `linked_request_id` columns on `job_runs` + `sync_runs` so boot-recovery's `NOT EXISTS` clauses skip replaying completed work.

## Test plan

- [x] `uv run pyright`
- [x] `uv run pytest` — **3002 passed, 1 skipped**
  - 9 new dispatcher cases (real DB) — claim/release lifecycle, FOR UPDATE SKIP LOCKED, scope round-trip
  - 4 listener cases — dispatch routing + malformed-payload rejection
  - 3 heartbeat cases — UPSERT, error swallow, loop stop
  - 4 queue-recovery cases — every branch of `reset_stale_in_flight` (manual_job terminal, sync terminal, sync running replays, 24h TTL)
  - 4 boot_sweep cases — env gate, scope+trigger, exception swallow
  - 1 supervisor case — stalled listener restart under tight threshold
  - 2 boot-drain cases — claim every pending row at boot
- [x] Codex spec review across 8 rounds (clean)
- [x] Codex pre-push review across 2 rounds (clean)

## Migration

Hard cut, no env flag. After merge:

1. `git pull` + `uv sync` + `uv run python -m scripts.migrate` (4 new migrations: pending_job_requests, job_runs.linked_request_id, sync_runs.linked_request_id, job_runtime_heartbeat).
2. Restart all three VS Code tasks (`stack: backend`, `stack: frontend`, **new** `stack: jobs`).

## Spec

`docs/superpowers/specs/2026-04-30-jobs-out-of-process-design.md` (spec passed Codex review across 8 rounds before any code was written).

## Settled-decision + prevention-log

- `docs/settled-decisions.md` "Process topology" entry locked.
- `docs/review-prevention-log.md` "Don't add scheduling or job execution to the API process" entry — `tests/smoke/test_app_boots.py` asserts `app.state.job_runtime` is NOT set so future regression is caught immediately.

## Known follow-ups (out of scope)

- `tests/smoke/test_jobs_process_boots.py` end-to-end smoke for the entrypoint — would require driving the full 13-step startup against a real DB; supervisor + boot_drain + boot_sweep + listener already have direct unit coverage. File a follow-up if integration regression risk is deemed worth the test surface.


Luke-Bradford:
BLOCKING [`app/jobs/runtime.py::_run_manual`]: **FIXED** `da51af9`. Dropped the `mark_request_dispatched` call before `invoker()`. Row now transitions `claimed` → `completed`/`rejected` directly. The recovery query in `reset_stale_in_flight` already accepts both `'claimed'` and `'dispatched'` in its WHERE clause, so removing the intermediate state is forwards-compatible.

WARNING [`app/jobs/listener.py::_run_sync_with_request_lifecycle`]: **FIXED** `da51af9`. Same root cause — `run_sync` returns with the `sync_runs` row already terminal, so `dispatched` → `completed` in a single autocommit block was unobservable. Dropped the dispatched write here too.

WARNING [`app/jobs/listener.py` LISTEN f-string]: **FIXED** `da51af9`. Switched to `psycopg.sql.SQL("LISTEN {}").format(psycopg.sql.Identifier(NOTIFY_CHANNEL))` so identifier quoting is the driver's job.

PREVENTION (`mark_request_dispatched` before run-row open): **EXTRACTED** `da51af9`. New regression test `tests/test_jobs_queue_recovery.py::test_dispatched_with_no_run_row_is_replayed` pins the contract: a row at `'dispatched'` with no linked `job_runs` row must be replayed by `reset_stale_in_flight`. Structurally impossible after this fix; the test catches future regressions that re-introduce the broken ordering.

PREVENTION (LISTEN channel quoting): **EXTRACTED** `da51af9`. The fix itself is the canonical pattern; comment in code names `Identifier` as the safe quoter for SQL-identifier-level strings.

Issue #726, opened 2026-04-30
perf: bounded-concurrency SEC fetcher — sequential ~1 req/s vs allowed 10 req/s


Luke-Bradford:
## Problem

SEC fair-use is 10 req/s per UA. Our SEC fetch path runs at **~1 req/s** in production. The ``ResilientClient(min_request_interval_s=0.11)`` floor is correct (allows 9 req/s theoretical max), but the actual rate is dominated by sequential execution + SEC's ~700-900ms response time per Form 4 XML.

\`\`\`
sequential rate ≈ 1 / (min_interval + response_time)
              ≈ 1 / (0.11 + 0.8)
              ≈ 1.1 req/s
\`\`\`

We're using ~10% of our allowed bandwidth.

## Affected jobs

- \`sec_insider_transactions_ingest\` (Form 4 XML)
- \`sec_8k_events_ingest\` (8-K item bodies)
- \`sec_filing_documents_ingest\` (per-filing index.json)
- \`business_summary_ingest\` (10-K business sections)
- \`dividend_calendar.ingest_dividend_events\` (8-K Item 8.01 bodies)

All share the same single-threaded fetch loop.

## Proposed fix

Replace the sequential \`for cik in candidates: fetcher.fetch_document_text(...)\` loop with a bounded-concurrency executor:

\`\`\`python
with concurrent.futures.ThreadPoolExecutor(max_workers=8) as pool:
    futures = [pool.submit(fetcher.fetch_document_text, url) for url in candidates]
    for future in concurrent.futures.as_completed(futures):
        body = future.result()
        # ... parse + upsert
\`\`\`

The existing process-wide rate-limit clock at \`_PROCESS_RATE_LIMIT_CLOCK\` ([sec_edgar.py:70](../blob/main/app/providers/implementations/sec_edgar.py#L70)) already enforces the 10 req/s ceiling across threads — concurrency overlaps the response wait time, doesn't bypass the ceiling.

## Expected impact

5-8x throughput on SEC ingest paths. The remaining ~221k Form 4 backfill (after #FOLLOWUP date-floor lands) drains in ~7 hours instead of ~5 days.

## Risks / considerations

- DB upsert path is currently single-connection; concurrent fetchers feeding ONE upsert thread is fine. Don't parallelise the writes — concurrency is on the network leg only.
- Per-CIK SEC throttling — if SEC throttles individual UA-IPs above 10 req/s aggregate, the existing 0.11s floor already protects.
[...]

PR #728, opened 2026-04-30
feat(#726): bounded-concurrency SEC fetcher — 7.5x throughput

Luke-Bradford:
## What

Replace the sequential ``for url in urls: fetch_document_text(url)`` loop with a bounded-concurrency thread pool. Aggregate request rate stays bounded by the existing shared rate-limit clock (now lock-protected for thread safety).

## Why

SEC fair-use is 10 req/s per UA. Pre-PR the ingest paths ran at ~1 req/s — every request blocked on SEC's ~700-900ms response time, so we used 10% of the allowed budget. Operator surfaced the bandwidth gap in #726.

\`\`\`
Pre-PR:  rate ≈ 1 / (0.11 + 0.8) ≈ 1.1 req/s
Post-PR: rate ≈ rate-limit ceiling ≈ 9 req/s (8 concurrent workers overlap response wait)
\`\`\`

**Live verification (dev DB):** 40 Form 4 candidates in 5.3s = **7.53 req/s effective**. The 221k Form 4 backlog drains in ~8h instead of 5+ days.

## How

1. **Lock-protect the rate-limit clock**. ``ResilientClient._throttle_and_stamp`` combines the check + sleep + timestamp-write into one lock-protected critical section. New ``shared_throttle_lock`` parameter so providers sharing a clock list also share the lock. ``_PROCESS_RATE_LIMIT_LOCK`` companion to ``_PROCESS_RATE_LIMIT_CLOCK`` wired through both SEC clients.

2. **Add ``concurrent_fetch.fetch_document_texts``** — shared utility. Default 8 workers, dedupes inputs, caps workers to unique URL count, isolates per-future exceptions as ``None`` so one bad URL can't crash the batch.

3. **Wire into insider_transactions** — pre-fetch all XML bodies concurrently, then iterate parse + upsert + commit serially on the caller's psycopg connection (DB writes stay single-threaded; concurrency is on the network leg only).

## Test plan

- [x] ``pytest tests/test_concurrent_fetch.py`` — 10 unit tests including a parameterised throttle-safety check across 4/8/16 concurrent threads
- [x] ``pytest tests/test_insider_transactions*`` — 50+ existing tests pass against the new pre-fetch + serial-process flow
- [x] ``ruff/format/pyright`` clean
- [x] **Live SEC end-to-end**: 40 candidates in 5.3s = 7.53 req/s, zero throttle violations
- [x] Codex checkpoint-2: 3 design questions confirmed correct (throttle atomicity, shared lock plumbing, helper safety) + 1 Low finding (lost accession context in failure logs) addressed by re-emitting at caller scope

## Out of scope

Other ingest paths (8-K events, business_summary, dividend_calendar, filing_documents) can adopt the helper one-by-one. Same change pattern: replace the per-iteration ``fetch_document_text`` call with a pre-fetch step. Keeping this PR scoped to insider_transactions for reviewability.

### DIFF
--- a/app/providers/concurrent_fetch.py
+++ b/app/providers/concurrent_fetch.py
@@ -1,0 +1,92 @@
+"""Bounded-concurrency text-fetch helper for SEC ingest paths (#726).
+
+Sequential ``for url in urls: fetch_document_text(url)`` loops are
+bottlenecked by SEC's per-request response time (~700-900ms) rather
+than the rate-limit floor — at ~1 req/s, every SEC ingest job uses
+about 10% of the allowed 10 req/s budget.
+
+This helper runs ``fetch_document_text`` in a ``ThreadPoolExecutor``
+so a single ingest job can keep multiple requests in flight against
+SEC, overlapping response wait time. The shared
+``_PROCESS_RATE_LIMIT_CLOCK`` + ``_PROCESS_RATE_LIMIT_LOCK`` in
+``sec_edgar.py`` keep the aggregate request rate under SEC's
+fair-use ceiling regardless of how many threads are active —
+concurrency overlaps wait time, it does NOT bypass the floor.
+
+Per-future exceptions are caught and surfaced as ``None`` results so
+one bad URL cannot crash the whole batch — callers handle ``None``
+the same way they handle a 404 (typically: tombstone + continue).
+
+Worker count default 8: at 0.11s floor + 800ms response time, 8
+workers is enough to saturate the rate ceiling. More workers wouldn't
+help; fewer would leave throughput on the table.
+"""
+
+from __future__ import annotations
+
+import concurrent.futures
+import logging
+from collections.abc import Iterable
+from typing import Protocol
+
+logger = logging.getLogger(__name__)
+
+
+# Default thread count. Bigger isn't better — the floor is the
+# bottleneck above this. Keeps memory + connection-pool pressure
+# bounded.
+DEFAULT_FETCH_WORKERS: int = 8
+
+
+class _TextFetcher(Protocol):
+    def fetch_document_text(self, absolute_url: str) -> str | None: ...
+
+
+def fetch_document_texts(
+    fetcher: _TextFetcher,
+    urls: Iterable[str],
+    *,
+    max_workers: int = DEFAULT_FETCH_WORKERS,
+) -> dict[str, str | None]:
+    """Fetch every URL concurrently and return a ``{url: body}`` map.
+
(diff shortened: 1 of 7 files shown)


Luke-Bradford:
## Review reply

### WARNING #1 (lock-while-sleeping kills throughput) — **REBUTTED**

The math is wrong. The HTTP RTT happens **outside** the lock (after `_throttle_and_stamp` returns). So with N concurrent workers:

- Each lock holder spends `min_interval` sleeping (since the previous holder just stamped).
- After release, the holder fires the request and waits ~RTT for response *outside the lock*.
- Meanwhile the next thread acquires, sleeps `min_interval`, stamps, releases, fires.

Total wall-clock for N requests = `N * min_interval + RTT` (only the *last* request's RTT bounds total), NOT `N * (min_interval + RTT)` which is what the sequential pre-PR loop took.

**Live SEC verification on this PR**: 40 Form 4 candidates in 5.3s = **7.53 req/s**, vs ~1 req/s sequential = 7.5× speedup. The bot's claim of "0.88 s/request effective ceiling" would predict ~35s for 40 reqs — empirically falsified.

### WARNING #2 (test doesn't prove race safety) — **REBUTTED**

If the lock were removed, multiple threads simultaneously fail the elapsed check (all read the same stale `last_request_at`) and all sleep `interval - 0` ≈ `min_interval`. They wake near-simultaneously. They all execute the unsynchronised stamp-write. Without the lock, the actual call timestamps would cluster within microseconds of each other — exactly the burst the lock prevents.

The current test sleeps for the floor between threads precisely *because* the lock serialises them — that's the property under test. Without the lock, threads would NOT take turns sleeping the floor; they'd race past the elapsed check together.

I added a deterministic wall-clock regression alongside the rebuttal — `TestConcurrencyAchievesActualThroughput.test_concurrent_total_time_smaller_than_sequential` fires 16 requests through 8 workers with simulated 100 ms RTT, asserts wall-clock ≤ 50% of the sequential estimate. So if a future change DID serialise the RTT under the lock, CI would catch it.

### PREVENTION (sleep-under-lock + throughput regression)

Both predicated on the inverted throughput claim. The wall-clock regression now in the test suite covers the legitimate "future change accidentally serialises RTT" hazard.

### CI failure — **FIXED** in 68a36d3

Legitimate. The new `concurrent_fetch.py` helper trips the pinned-caller-set guard from #448/#453. Updated the allow-list with documentation that the helper is transport-only — the SQL-normalisation contract is preserved at the caller (insider_transactions etc. still own the upsert).

@anthropic-claude please re-review.

PR #733, opened 2026-04-30
fix: jobs process must bootstrap master key — daily_portfolio_sync crashed at boot

Luke-Bradford:
## What

The jobs process entrypoint at \`app/jobs/__main__.py::serve\` never called \`master_key.bootstrap()\`. The API process lifespan does. Every scheduled job that called \`decrypt(...)\` via \`_load_etoro_credentials\` crashed on its first tick with \`MasterKeyNotLoadedError\`.

## Why

Operator surfaced 2026-05-01: jobs process restart log showed \`daily_portfolio_sync\` and \`daily_candle_refresh\` both raising:

\`\`\`
File \"app/security/secrets_crypto.py\", line 122, in _get_aesgcm
    raise MasterKeyNotLoadedError(...)
\`\`\`

Root cause: \`serve()\` opened the pool → acquired the singleton fence → started the scheduler → ran the boot-drain. The master-key bootstrap that the API lifespan does at \`app/main.py:125\` was missing here. The very first job that touched encrypted credentials crashed.

Existing tests didn't catch it: API smoke tested the API lifespan; unit tests mocked the crypto layer. No test drove the jobs entrypoint's actual boot flow.

## How

1. New \`_bootstrap_master_key(pool)\` helper mirroring the API lifespan's bootstrap step. Runs **after** \`open_pool\` + fence acquisition, **before** \`runtime.start()\` and boot-drain.
2. Helper is module-level (not inline in \`serve()\`) so the smoke test can drive it without acquiring the singleton lock — a full \`serve()\` against a live daemon would SystemExit on the fence.
3. New \`tests/smoke/test_jobs_process_boots.py\` with \`test_jobs_bootstrap_master_key_loads_aesgcm\` — invokes the helper, asserts AES-GCM cipher is callable post-boot. Same shape as \`test_app_boots.py\` for the API.

## Test plan

- [x] \`pytest tests/smoke/test_jobs_process_boots.py\` — pass against dev DB
- [x] \`ruff/format/pyright\` clean
- [x] Verified the API process's bootstrap flow at \`app/main.py:117-137\` is the canonical pattern; this PR mirrors it
- [ ] Post-merge: operator restarts the jobs process, confirms \`daily_portfolio_sync\` / \`daily_candle_refresh\` execute without \`MasterKeyNotLoadedError\`

## Out of scope

The boot-drain step at \`app/jobs/__main__.py::_drain_pending_at_boot\` runs **before** \`runtime.start()\` per #719. The new bootstrap call sits between fence acquisition and boot-drain so any drained job that touches credentials sees the loaded key. If the order is ever rearranged, the smoke test catches a regression where bootstrap is removed entirely; ordering changes within the boot sequence would need their own test.

### DIFF
--- a/app/jobs/__main__.py
+++ b/app/jobs/__main__.py
@@ -57,6 +57,8 @@ from app.jobs.listener import ListenerState, listener_loop
 from app.jobs.locks import JOBS_PROCESS_LOCK_KEY
 from app.jobs.runtime import JobRuntime
 from app.jobs.supervisor import supervise
+from app.security import master_key
+from app.security.secrets_crypto import set_active_key as set_broker_encryption_key
 from app.services.sync_orchestrator.dispatcher import (
     claim_oldest_pending,
     reset_stale_in_flight,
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
## Review reply

WARNING (double-bootstrap soundness) — **FIXED** in 6bfb191. Replaced both branches with mock-based wiring contract tests that don't rely on bootstrap's idempotency:

- `test_jobs_bootstrap_master_key_installs_returned_key` patches `master_key.bootstrap` to return a known fake key, asserts `_get_aesgcm()` is callable post-helper.
- `test_jobs_bootstrap_master_key_no_op_when_bootstrap_returns_no_key` patches bootstrap to return `broker_encryption_key=None`, asserts the cipher stays unloaded.

Both reset the module-global cipher with `clear_active_key()` so prior tests' state can't trivialise the assertion.

CI failure (settings.database_url destructive-paths guard) — also fixed in 6bfb191. Same allowlist treatment as `test_app_boots.py`: the probe runs `SELECT 1`, the bootstrap is read-only against `broker_credentials.ciphertext`.

@anthropic-claude please re-review.

Issue #753, opened 2026-05-01
fix: filename probe burns SEC budget on pre-2003 filings (0001.txt → edgar.txt fallback)


Luke-Bradford:
## Symptom

Production logs (2026-05-01) for CAT (CIK 30697) backfill show alternating 404s + 200s:

\`\`\`
GET .../000003069701000020/0001.txt   404
GET .../000003069703000042/edgar.txt  200
GET .../000003069703000036/edgar.txt  200
GET .../000003069700000079/0001.txt   404
GET .../000003069701500020/edgar.txt  200
\`\`\`

Pattern: each accession is probed first with \`0001.txt\`, falls back to \`edgar.txt\` on 404. The \`0001.txt\` path 404s reliably for pre-2003 filings (filename convention changed); \`edgar.txt\` is the canonical filename for that era.

## Why this is wasteful

- Every pre-2003 filing burns one SEC fair-use request on a 404 before falling back.
- 404 noise floods production logs, drowning real diagnostics.
- The bounded-concurrency budget added in #728 is shared with this wasted spend.

The filings ARE being ingested correctly (the fallback works). This is purely an efficiency issue.

## Fix options

1. **Filing-date gate.** Skip \`0001.txt\` probe when \`filing_date < 2003-01-01\`. Requires plumbing the filing date into the document-fetch path.
2. **Discover via index.json first.** Walk the per-accession \`index.json\` (already implemented for \`filing_documents\` ingest) to discover the actual primary-doc filename, then fetch that directly. Slightly more bandwidth on modern filings (one extra index.json hit) but zero 404 noise overall.
3. **Probe-order swap.** Try \`edgar.txt\` first, fall back to \`0001.txt\`. Solves the pre-2003 case but flips the noise to post-2003 filings.

Option 2 is the cleanest — \`filing_documents\` already does this and the index.json hits are cached / re-used across the same accession.

## Linked

- Adjacent to #728 (bounded-concurrency budget being burned on these probes).
- Adjacent to #195 (broad filing-ingest tech-debt).

Issue #752, opened 2026-05-01
fix: external_identifiers contaminated with filing-agent CIKs (downstream of master-index ingest)


Luke-Bradford:
## Symptom

Production logs (2026-05-01) still show 404s on \`Archives/edgar/data/{CIK}/{accession}/index.json\` after PR #745 + PR #748 fixed both \`fetch_filing_index\` and \`coverage.py:1810\` to thread \`issuer_cik\` correctly:

\`\`\`
GET .../1493152/000149315226020513/index.json   404  (GlobeNewswire)
GET .../1193125/000119312526197194/index.json   404  (Donnelley)
GET .../1437749/000143774926014225/index.json   404  (Computershare/Filing Services Canada)
GET .../1628280/000162828026028946/index.json   404
\`\`\`

Multiple distinct agent CIKs cycling through means **multiple instruments** are misconfigured.

## Root cause hypothesis

\`coverage.py:1810\` reads \`cik_padded\` from \`external_identifiers (provider='sec', identifier_type='cik', is_primary=TRUE)\`. The fix routes the URL under that CIK. So if the URL is the **agent's** CIK, then \`external_identifiers\` itself contains the agent CIK as the issuer mapping.

Likely contamination path: \`app/services/fundamentals.py:1769\` buckets master-index entries by \`entry.cik\` which is the **column-1 CIK** of the SEC master.idx file — the filer-of-record CIK, NOT the issuer CIK. For agent-filed accessions, that's the agent. Intersecting against \`_load_covered_us_ciks(conn)\` then expanding via \`_upsert_filing_from_master_index\` would write \`filing_events\` rows whose CIK lookup back into \`external_identifiers\` may have returned an agent CIK instead of the issuer.

Need to confirm whether ingestion actually plants agent CIKs into \`external_identifiers\` or whether they leaked in some other path.

## Diagnostic SQL

\`\`\`sql
-- Find instruments whose primary SEC CIK is a known filing-agent CIK
SELECT i.instrument_id, i.symbol, i.company_name, ei.identifier_value AS agent_cik
FROM external_identifiers ei
JOIN instruments i USING (instrument_id)
WHERE ei.provider = 'sec'
  AND ei.identifier_type = 'cik'
  AND ei.is_primary = TRUE
  AND ei.identifier_value IN (
    '0001213900',  -- EdgarOnline
    '0001493152',  -- GlobeNewswire / Issuer Direct
    '0001193125',  -- Donnelley R.R. & Sons
[...]

Issue #740, opened 2026-05-01
feat: backfill CUSIP→instrument mapping in external_identifiers (gates #729 ownership card)


Luke-Bradford:
## Context

Discovered while implementing #730 PR 2 (13F-HR ingester). The ticket body for #730 stated *"Resolve CUSIP → instrument_id via existing external_identifiers (CUSIPs already mapped for US issuers)"* — but a 2026-05-01 audit shows \`external_identifiers\` carries **zero** rows with \`identifier_type='cusip'\`. Only CIK mappings exist (5,095 rows).

Without CUSIP mapping the 13F-HR ingester can populate \`institutional_holdings\` rows but cannot resolve any to a tradable \`instrument_id\`. The ownership reporting card (#729) ships dark until this gates closes.

## Source options

1. **SEC's quarterly Official List of Section 13(f) Securities** — published at sec.gov/divisions/investment/13flists.htm as XLS. Each row carries CUSIP, issuer name, class label, status. Authoritative for 13F-eligible securities. ~25,000 rows; refreshes quarterly. Limitation: name-based join to our instruments is fuzzy (string matching).

2. **SEC company-facts XBRL** — \`dei\` taxonomy occasionally surfaces \`Cusip\` for filers that supply it. Exact CIK→CUSIP join, but coverage is partial (issuers don't all report CUSIP in cover-page facts).

3. **OpenFIGI / external paid mapping API** — clean CUSIP↔ticker map. Off-table per the settled provider strategy (free regulated-source-only, #532).

## Recommended path

Option 2 (XBRL DEI) **plus** Option 1 (SEC 13F list) as a fallback for issuers whose XBRL doesn't surface a CUSIP. Both are SEC-published and free; both fit the regulated-source posture. Option 1 needs an issuer-name → instrument_id matcher (deterministic on company_name, fall back to manual review for ambiguous matches).

## What

1. New service \`app/services/cusip_backfill.py\` that:
   a. For every instrument with a known CIK in \`external_identifiers\`, look up the SEC company-facts payload and extract \`dei:Cusip\` if present.
   b. Backfill the SEC 13F securities list quarterly into a staging table; on each ingest, walk unresolved CIKs and try a name-based join.
   c. Upsert the resulting \`(provider='sec', identifier_type='cusip', identifier_value=<9-char>, instrument_id=<id>, is_primary=TRUE)\` into \`external_identifiers\`.
[...]

Issue #735, opened 2026-05-01
feat: project EntityPublicFloat into financial_periods (DEI cover-page handling)


Luke-Bradford:
## Context

Split out from #731 during implementation. `EntityPublicFloat` is a DEI cover-page fact emitted on 10-K with `period_end` = issuer's most recent Q2-end (the SEC-prescribed "as of" date for public-float computation), **not** fiscal year-end.

The current `_derive_periods_from_facts` logic in `app/services/fundamentals.py:802-818` anchors each period row to `max(period_end)` of mapped facts and then drops facts whose `period_end` differs from that anchor. So projecting `EntityPublicFloat` through the existing pipeline silently drops virtually every value.

#731 ships the four other ownership-card-relevant columns (`treasury_shares`, `shares_authorized`, `shares_issued`, `retained_earnings`) which are pure us-gaap balance-sheet facts on the fiscal period_end date and route through the existing pipeline cleanly. `public_float_usd` is held back to the next ticket so the DEI period-end mismatch can be designed properly without expanding #731's surface.

## What

Two viable approaches — pick one in design:

1. **Separate normaliser path for DEI cover-page facts.** In `_derive_periods_from_facts`, after the canonical-FY row is built, layer an additional pass that pulls DEI annual facts (frame matches `CY{year}Q2I` for public-float) and attaches them to the FY row even when their `period_end` is earlier than the FY anchor. Mark such columns explicitly so they don't get pulled into `_BALANCE_SHEET_COLUMNS`-style TTM aggregation.
2. **Dedicated `dei_annual_facts` table.** Store DEI annual cover-page facts (`EntityPublicFloat`, `EntityCommonStockSharesOutstanding` as-of-date, `EntityNumberOfEmployees`) in a parallel narrow table keyed `(instrument_id, fiscal_year)`. Reader joins to it from `financial_periods` for FY-only views. Cleaner separation, but a second source of truth.

Default to (1) unless there's a reason DEI deserves a separate table.

## Acceptance

- After re-derive, `SELECT COUNT(*) FROM financial_periods WHERE public_float_usd IS NOT NULL` is non-zero across covered FY rows.
- IEP / AAPL / TSLA spot-check: `public_float_usd` matches the issuer's 10-K cover page.
- Reader API exposes the column on the `balance` statement.

## Linked

Splits from #731 (which ships the four us-gaap-aligned columns).

PR #754, opened 2026-05-01
fix(#729): distribute float gap across unknown wedges + persistent click selection

Luke-Bradford:
## What

Two operator-flagged issues from the latest screenshot review.

### 1. Sunburst still rendering as one solid purple

Pre-fix: unknown-status wedges had synthetic \`shares=1\` in the renderer. Recharts proportions wedges by value, so on AAPL (4 of 5 categories status=unknown, Insiders=ok with ~9.1M shares = 0.06% of float), the middle ring data was \`[1, 1, 9_104_843, 1]\` — Insiders ate ~99.99% of the visible arc, the three gap wedges collapsed to sub-pixel slivers regardless of their tinted accent color.

Fix: \`SunburstCategory\` + \`SunburstLeaf\` now carry a separate \`display_shares\` field. The renderer reads \`display_shares\` for wedge sizing while \`shares\` stays at 0 for arithmetic correctness on unknown wedges. \`buildSunburstRings\` distributes the inner-ring \`gap_shares\` equally across unknown categories so each gets a visible arc.

Equal split because we have no per-category signal — operator sees "unknown gap of size X" rather than a weighted estimate that we cannot defend.

### 2. Click selection didn't stick

Click on L1 sunburst navigates to L2 with \`?category=...&filer=...\`, but the L2 sunburst didn't know which wedge was selected — only the table row got the highlight treatment. Hover-brightness on the wedge was the only chart-side feedback, which faded on mouse-out.

Fix: \`OwnershipSunburst\` now accepts \`selectedCategory\` + \`selectedLeaf\` props. Selected wedges render at full opacity (overrides the gap-opacity reduction). L2 wires these from \`categoryFilter\` / \`filerFilter\` URL params; L1 doesn't pass the props (no URL state to drive selection).

## Visual sanity (operator validates)

Reload \`/instrument/AAPL/ownership\`. Expected:
- Middle ring: 4 visible wedges — cyan-tinted Institutions (~25% of ring), blue-tinted ETFs (~25%), amber-tinted Treasury (~25%), purple Insiders (hairline 0.06%, distinct from the gap wedges). Each unknown wedge is ~1/3 of the gap arc.
- Click on Institutions wedge → URL becomes \`?category=institutions\` → wedge stays at full opacity (sticks visually).
- Click on a different wedge (e.g. Treasury) → previous selection drops back to gap opacity, new one sticks.

## Test plan

- [x] \`pnpm typecheck\` — clean
- [x] \`pnpm test:unit\` — 794 tests pass
- [x] \`node scripts/check-dark-classes.mjs\` — clean (173 files)

## Linked

- Builds on PR #751 (saturation + click rect).
- Effective slice fill still gated on #740 / #735 / #752 (data-corruption audit).

## Out of scope (filed as separate tickets during this work)

- **#752** — \`external_identifiers\` contaminated with filing-agent CIKs. Production logs show 404s on agent CIKs (1493152, 1193125, 1437749) downstream of #745 / #748 fixes; URL-routing code is correct, but the upstream data has agent CIKs mapped as primary issuer CIKs. Audit query + fix path documented in the ticket.
- **#753** — \`0001.txt\` filename probe burns SEC budget on pre-2003 filings. Pre-2003 filings reliably 404 on \`0001.txt\` before falling back to \`edgar.txt\`. Wasted budget + log noise.

### DIFF
--- a/frontend/src/components/instrument/OwnershipPanel.tsx
+++ b/frontend/src/components/instrument/OwnershipPanel.tsx
@@ -120,7 +120,15 @@ export function OwnershipPanel({ symbol }: OwnershipPanelProps): JSX.Element {
       if (target.kind === "category") params.set("category", target.category_key);
       if (target.kind === "leaf") {
         params.set("category", target.category_key);
-        params.set("filer", target.leaf_key);
+        // Synthetic gap leaves (key ``${category}-unknown`` or
+        // similar) carry no real filer identity — clicking them
+        // should drill to the category, not filter to a non-
+        // existent filer key. Detect via the ``-unknown`` suffix
+        // emitted by ``buildSunburstRings`` for unknown-category
+        // sentinel leaves.
+        if (!target.leaf_key.endsWith("-unknown")) {
+          params.set("filer", target.leaf_key);
+        }
       }
       const qs = params.toString();
       const suffix = qs.length > 0 ? `?${qs}` : "";
--- a/frontend/src/components/instrument/OwnershipSunburst.tsx
+++ b/frontend/src/components/instrument/OwnershipSunburst.tsx
@@ -298,7 +298,8 @@ export function OwnershipSunburst({
                 key={`inner-${idx}`}
                 fill={d.fill}
                 fillOpacity={d.opacity}
-                stroke={wedgeStroke}
+                stroke={d.is_selected ? "#cbd5e1" : wedgeStroke}
+                strokeWidth={d.is_selected ? 3 : 1}
               />
             ))}
           </Pie>
@@ -316,7 +317,8 @@ export function OwnershipSunburst({
                 key={`middle-${idx}`}
                 fill={d.fill}
                 fillOpacity={d.opacity}
-                stroke={wedgeStroke}
+                stroke={d.is_selected ? "#cbd5e1" : wedgeStroke}
+                strokeWidth={d.is_selected ? 3 : 1}
               />
             ))}
           </Pie>
@@ -334,7 +336,8 @@ export function OwnershipSunburst({
                 key={`outer-${idx}`}
                 fill={d.fill}
                 fillOpacity={d.opacity}
-                stroke={wedgeStroke}
+                stroke={d.is_selected ? "#cbd5e1" : wedgeStroke}
+                strokeWidth={d.is_selected ? 3 : 1}
               />
             ))}
           </Pie>
--- a/frontend/src/components/instrument/OwnershipSunburst.tsx
+++ b/frontend/src/components/instrument/OwnershipSunburst.tsx
@@ -38,6 +38,15 @@ export interface OwnershipSunburstProps {
   readonly onWedgeClick?: (target: WedgeClick) => void;
   /** Pixel size of the chart canvas (square). Default 280. */
   readonly size?: number;
+  /** When set, renders the matching middle-ring wedge with a
+   *  persistent "selected" treatment (brighter stroke + full
+   *  brightness). Wired to the L2 ``?category=`` query param so a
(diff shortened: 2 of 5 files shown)


Luke-Bradford:
## Resolution for review round 1 (16:02:59)

> WARNING: `is_selected` JSDoc says "Bumps stroke-width + caps opacity to 1" but no stroke-width change is applied anywhere in the added rendering code

**REBUTTED** — `OwnershipSunburst.tsx` does bump `strokeWidth` to 3 when `d.is_selected` is true. The contract is wired on all three rings:

- inner ring: `strokeWidth={d.is_selected ? 3 : 1}` at `frontend/src/components/instrument/OwnershipSunburst.tsx:302`
- middle ring: same at `:321`
- outer ring: same at `:340`

The JSDoc is accurate. No code change needed.

> PREVENTION: grep for `stroke-width` or `strokeWidth` in `OwnershipSunburst.tsx` as part of the PR checklist whenever a JSDoc mentions a visual property change.

**REBUTTED** — the bot's own grep would have found `strokeWidth={d.is_selected ? 3 : 1}` on three sites and confirmed the JSDoc. This was a false alarm and the rule it proposes (grep for the property when JSDoc mentions one) is already implicit in the existing self-review workflow; codifying it as a project-wide checklist item would add noise without catching a real class of bug.

## Resolution for review round 2 (16:15:19)

> WARNING: When a category is floored to `min_display_shares`, every leaf inside it is independently floored to `min_display_shares` as well; a category with N leaves then projects N×`min_display_shares` onto the outer ring while the middle ring shows only `min_display_shares`

**FIXED 5de3939** — extracted `rescaleLeaves` and call it from the floor pass. When the floor patches `cat.display_shares` to `min_display_shares`, `rescaleLeaves` distributes that target proportionally across the existing leaves (or evenly when their current total is zero), so the invariant `sum(leaf.display_shares) === cat.display_shares` holds. Realistic worst-case (Insiders 0.06% across 4 officers, was inflating the outer ring by 4×) is now correct.

> PREVENTION: add a test asserting that `sum(leaf.display_shares for leaf in cat.leaves) === cat.display_shares` for every non-empty category after `buildSunburstRings`, to catch future floor or gap-distribution steps that patch category and leaf sizes independently.

**EXTRACTED `frontend/src/components/instrument/ownershipRings.test.ts`** — added two tests in 5de3939:

1. Targeted regression on the exact failure shape (4-officer Insiders under floor, mixed with two unknown categories absorbing the gap).
2. Sweep invariant test that runs `buildSunburstRings` over three realistic input shapes (all-OK, mixed unknown + sub-floor multi-leaf, visible leaf + "Other" tail in a sub-floor category) and asserts the sum invariant holds for every non-empty category.

Codex flagged that my first sweep fixture (micro-cap 50k+8k = 58k vs 50k floor) actually short-circuited above the floor and never exercised `rescaleLeaves`; rewrote the fixture (float 10M, alice 300k visible + bob 30k tail = 330k below 500k floor) so the rescale path runs end-to-end on a visible-leaf + Other-tail shape.

Luke-Bradford:
Abandoning approach. Synthesizing wedges for unknown coverage + flooring sub-1% known wedges to a 5% minimum lies about the data — operator wants truthful arcs with empty space where data is missing. Coverage gaps today (#740 CUSIP backfill, #735 DEI projection) are temporary; new instruments will pull data from the start, so the steady state is all 5 categories known and the chart should be designed for that, not for the transient gap. New branch feature/729-truthful-ownership-pie will: (a) drop gap distribution, (b) drop the min-wedge floor, (c) add explicit open_market category as float - sum(institutions + etfs + insiders + treasury), (d) render unknown categories as empty arc via Pie startAngle/endAngle, (e) surface coverage gap in header badge only.

Issue #736, opened 2026-05-01
fix: SEC archive index.json 404 noise — filing-agent CIK mistakenly used as issuer CIK in URL builder


Luke-Bradford:
## Symptom

Production logs (2026-05-01) show repeated 404s on \`https://www.sec.gov/Archives/edgar/data/{CIK}/{accession_no_dashes}/index.json\` for CIK **1213900** (EDGAR Online — a filing agent, not an operating issuer):

\`\`\`
2026-05-01 01:35:56 GET https://www.sec.gov/Archives/edgar/data/1213900/000121390026050022/index.json "HTTP/1.1 404 Not Found"
2026-05-01 01:35:57 GET https://www.sec.gov/Archives/edgar/data/1213900/000121390026050027/index.json "HTTP/1.1 404 Not Found"
2026-05-01 01:35:58 GET https://www.sec.gov/Archives/edgar/data/1213900/000121390026050028/index.json "HTTP/1.1 404 Not Found"
2026-05-01 01:35:58 GET https://www.sec.gov/Archives/edgar/data/1213900/000121390026050030/index.json "HTTP/1.1 404 Not Found"
2026-05-01 01:35:59 GET https://www.sec.gov/Archives/edgar/data/1213900/000121390026050043/index.json "HTTP/1.1 404 Not Found"
\`\`\`

Five-in-a-row is not transient SEC race; it is a structural URL bug.

## Root cause hypothesis

SEC accession numbers \`{filer_cik}-{yy}-{seq}\` carry the **filing agent's** CIK in the prefix, not the **issuer's** CIK. EdgarOnline (CIK 1213900) is a registered agent and files on behalf of issuers; the archive directory always lives under the **issuer's** CIK, not the agent's.

Code paths that build the archive URL:

- \`app/providers/implementations/sec_edgar.py\` — walks submissions index and constructs \`Archives/edgar/data/{cik}/{accn_no_dashes}/index.json\`.
- \`app/services/filing_documents.py\` — same URL shape (recently rewritten in #725).

Likely bug: at least one of these uses the **submissions-shard CIK** (the iterator key) when constructing the URL, instead of the per-row **issuer CIK** that the submissions JSON exposes for each accession when the filer is an agent.

## Other affected filers

Any registered filing agent CIK is likely producing the same 404 cluster:

- 1213900 — EDGAR Online
- 1571049 — Donnelley Financial / DFIN
- 1185185 — Workiva
- 1387131 — RR Donnelley
[...]

Issue #195, opened 2026-04-12
tech-debt: rethink filing ingestion — multi-region providers, storage, observability, error handling


Luke-Bradford:
## Problem

The filing ingestion pipeline needs a fundamental rethink. It is currently SEC-only, but the tradable universe spans 30+ exchanges across multiple regions. A live run took 10+ minutes with no visibility, 3.2 GB of raw JSON accumulating on disk, and 7,006 out of 12,133 instruments logging "no CIK" warnings because they are on non-US exchanges that SEC will never cover.

### Current state

- **SEC-only**: The filing ingestion only knows about SEC/EDGAR. Non-US instruments (58% of the universe) are iterated, fail CIK lookup, log a warning, and achieve nothing.
- **Runtime**: 10+ minutes for a single `daily_research_refresh` catch-up — much of it wasted on instruments that can never produce results.
- **Raw file storage**: 15,788 files / 3.2 GB in `data/raw/sec/`. Every HTTP response is dumped to disk with **no cleanup, no rotation, no TTL**. This pattern exists across all providers (`data/raw/etoro/`, `data/raw/fmp/`, `data/raw/companies_house/`, `data/raw/sec_fundamentals/`).
- **No UI visibility**: The operator dashboard does not show that a job is running, what percentage is complete, or an ETA.
- **Error handling gaps**: Errors observed in logs during the run with unclear catch-and-handle behaviour.
- **Unclear value of stored data**: The frontend links out to SEC for actual filing documents. The raw JSON blobs on disk are write-only — nothing reads them back.

### Universe breakdown (live DB, 2026-04-12)

| Metric | Count |
|--------|-------|
| Total instruments in coverage | 12,133 |
| Has SEC CIK (US-listed) | 5,127 (42%) |
| No CIK (non-US exchanges) | 7,006 (58%) |

Top exchanges: IDs 4/5 (likely NASDAQ/NYSE) have good CIK coverage. 30+ other exchanges have zero or near-zero. The `country` column is NULL across the board — no metadata to filter on.

---

### Core principle

`filing_events` is the **single table** for all filings regardless of origin. Each regional provider populates the same normalised schema. Downstream consumers (thesis engine, rankings, UI) query one table and do not know or care where the filing came from.

### Architecture

```
Scheduler
  |-- US exchanges --> SEC EDGAR provider --> filing_events
  |-- UK exchanges --> Companies House provider --> filing_events
[...]

PR #765, opened 2026-05-02
fix(#752): agent-CIK defense-in-depth + diagnostic audit script

Luke-Bradford:
## Diagnosis

Live audit on 2026-05-02 dev DB: \`external_identifiers\` is **clean** — zero rows match any known filing-agent CIK across either \`is_primary\` value. The data plane is not corrupted; #752's contamination hypothesis is not the production issue.

**Real cause** of the 2026-05-01 20:51:44 production 404 burst:

- \`app.jobs\` worker process started **2026-05-01 01:12:14**
- PR #748 (\`issuer_cik\` thread-through) merged **2026-05-01 15:33:27**

The long-lived worker held pre-#736 code in-memory for 14 hours after the fix shipped. New accessions hit the legacy CIK-prefix fallback in \`fetch_filing_index\` and 404'd against the agent's archive directory. **Operator action: restart \`app.jobs\` so post-#736 routing is in-memory.**

## Defense-in-depth (this PR)

- **\`KNOWN_FILING_AGENT_CIKS\` block-list** in \`sec_edgar.py\` — EdgarOnline, GlobeNewswire, Donnelley etc. Codified in version control so adding a new agent is auditable.
- **\`fetch_filing_index\` legacy fallback rejects agent prefixes** with an operator-facing WARNING that names the stale-process failure mode and remediation. Saves SEC rate budget (no guaranteed-404 round-trips).
- **Behaviour unchanged** for non-agent prefixes (self-filers go through legacy as before) and when caller passes \`issuer_cik\` (production happy path unaffected).

## Diagnostic + remediation

- **\`scripts/audit_agent_cik_contamination.py\`** runs the original #752 SQL audit. Exits 0 clean / 1 contaminated, emits CSV on hit. Suitable for boot-time / scheduled invocation so a future regression that leaks agent CIKs into the table fails fast.
- **\`tests/test_agent_cik_defense.py\`** pins both contracts: block-list rejects without HTTP, audit splits primary vs. secondary, set-membership locked.

## Out of scope (deferred)

- \`upsert_cik_mapping\` proactive reject — master-index column-1 CIK path already validated as not contaminating.
- Boot-time audit hook in \`app/main.py\` lifespan — re-add when CI gains a production-DB smoke step.

## Test plan

- [x] \`uv run ruff check .\` — clean
- [x] \`uv run ruff format --check .\` — clean
- [x] \`uv run pyright\` — clean
- [x] 15 tests pass — \`test_agent_cik_defense.py\` (7 new) + \`test_sec_provider_filing_index.py\` (8 existing, no regression)
- [ ] Operator: restart \`app.jobs\` worker (NOT in PR scope)

## Related

- Followup to #736 / PR #745 / PR #748 — URL-routing code is correct; this PR adds belt-and-suspenders for the stale-process case.
- #195 broad filing-ingest tech-debt umbrella.

### DIFF
--- a/app/providers/implementations/sec_edgar.py
+++ b/app/providers/implementations/sec_edgar.py
@@ -80,6 +80,35 @@ _PROCESS_RATE_LIMIT_CLOCK: list[float] = [0.0]
 _PROCESS_RATE_LIMIT_LOCK: threading.Lock = threading.Lock()
 
 
+# Known SEC filing-agent CIKs (#752 defense-in-depth). These are
+# registered EDGAR submitters that file ON BEHALF of issuers — their
+# CIK appears as the prefix of the accession number for any filing
+# they submit, but the archive directory ALWAYS lives under the
+# issuer's CIK. Routing an archive fetch under any of these CIKs is
+# guaranteed to 404. The ``fetch_filing_index`` legacy fallback
+# rejects on hit so a stale long-running process running pre-#736
+# code surfaces a clear log line ("known agent CIK; pass issuer_cik")
+# instead of generating thousands of 404 round-trips against SEC.
+#
+# Source: cross-referenced from accession-prefix patterns observed in
+# the SEC master-index across 2024-2026. Adding a new agent here is
+# a code change rather than a DB migration so the block-list is
+# auditable in version control.
+KNOWN_FILING_AGENT_CIKS: frozenset[str] = frozenset(
+    {
+        "0001213900",  # EdgarOnline (Issuer Direct)
+        "0001493152",  # GlobeNewswire / Issuer Direct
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
Resolved review on PR #765:

- **BLOCKING** — duplicate-CIK test isolation hazard: **FIXED cc80aad**. Switched secondary-row test to a different agent CIK (``0001213900`` EdgarOnline) so each test owns a unique value. Avoids the ``ON CONFLICT (provider, identifier_type, identifier_value) DO NOTHING`` silent-skip across tests.
- **PREVENTION** — per-test unique-value rule: **EXTRACTED** in commit message. Future audit tests that seed ``external_identifiers`` must use per-test-unique CIKs.

Issue #756, opened 2026-05-01
Ownership over time — category share-shift chart on L2 page


Luke-Bradford:
## Why

Operator request from #729 review: visualise how category ownership composition changes per quarter so the operator can spot:

- Institutional accumulation / distribution
- Insider exits clustering before drawdowns
- ETF inflow / outflow trends
- Treasury repurchase pace

A static snapshot ring shows today; a time series shows the shift.

## Scope

L2 ownership page (\`frontend/src/pages/OwnershipPage.tsx\`): add a stacked-area or small-multiples chart below the ring + filer table showing per-category % of \`total_shares\` per \`period_of_report\`.

## Data

Persisted history already available:

- 13F: \`institutional_holdings.period_of_report\` quarterly cadence
- Form 4: \`insider_transactions.txn_date\` per transaction; aggregate post-transaction-shares per quarter
- XBRL: \`financial_periods.period_end\` for outstanding + treasury history

Compute \`(institutions + etfs + insiders + treasury) / total_shares\` per period; render with same color palette as ring.

## Definition of done

- Stacked-area chart on L2 below filer table, ~6-12 months of history.
- Same colour palette + legend coupling as ring.
- Hover surfaces per-category absolute shares for that period.

## Depends on

- #755 — without backfilled data, the chart is empty for the same reason the ring is.

PR #770, opened 2026-05-02
feat(#767): per-category freshness chips on ownership card

### LINKED ISSUE #767: feat: per-category freshness on ownership card
## Why

Ownership card today carries one card-wide \`as_of_date\`. Different categories have wildly different freshness:

- Treasury (XBRL): updates only on 10-Q/10-K filing — quarterly, sometimes 90+ days stale
- Insiders (Form 4): updates within 2 business days of any insider trade
- Institutions / ETFs (13F-HR): quarterly snapshot, 45-day filing window — can be 135 days stale
- Blockholders (13D/G when shipped): event-driven, near-real-time

Lumping these into one timestamp lets a chart that's "75% stale" read as fresh as one that's "5% stale".

## Scope

1. Reader endpoint adds per-category \`as_of_date\` (latest source-row period_end / filing_date used to compute that category).
2. \`SunburstCategory\` carries \`as_of_date\`.
3. Card header strip: \`Institutions 12d · ETFs 12d · Insiders 2d · Treasury 47d\` chips. Stale > review_frequency triggers an amber chip.
4. Tooltip on each category wedge surfaces its own \`as_of_date\`.

## Out of scope

- Recompute / refresh trigger UI — operator already has #677 force-refresh
- Cross-card freshness consolidation — separate concern

## Definition of done

- Operator can tell at a glance which category is the stalest on any instrument's ownership card.
- Stale-beyond-tier-cadence chips render amber.

Luke-Bradford:
## What

Per-category freshness chip strip on L1 ``OwnershipPanel`` + L2 ``OwnershipPage``. One chip per non-empty category showing source-row date + age delta + colour-coded freshness level (fresh / aging / stale / unknown).

Cadence thresholds tuned per upstream filing rhythm:

| Category | aging | stale |
|---|---|---|
| Institutions / ETFs (13F-HR) | > 135d | > 270d |
| Insiders (Form 4) | > 30d | > 90d |
| Treasury (XBRL 10-Q) | > 100d | > 200d |

## Why

Different ownership categories update at wildly different cadences. One card-level "as of" timestamp lets a 75%-stale card read as fresh as a 5%-stale one. Per-category chips give the operator a glanceable read on which slice is the bottleneck — "Institutions 12d" vs "Treasury 200d" surfaces the gap immediately.

## Test plan

- [x] ``classifyFreshness`` boundary tests pin every threshold (134d → fresh, 135d → aging, etc.)
- [x] ``formatAge`` tests cover days / months / years / future-date clamp
- [x] Chips render one per category, colour-coded via ``data-freshness-level``
- [x] Unparsable ``as_of_date`` collapses to "no date on file" without echoing the bad string (Codex regression)
- [x] ``SunburstInputs`` ``as_of`` fields propagate to ``SunburstCategory.as_of_date``
- [x] Full frontend suite: 814 passed
- [x] Pre-push gates: ruff / format / pyright / dark-class / typecheck — all green

## Notes

Pure frontend change. Every timestamp the chip needs is already exposed by existing reader endpoints (``totals.period_of_report``, ``summary.latest_txn_date``, ``balance.rows[].period_end``).

Codex review (CLAUDE.md checkpoint 2) caught one real bug: ``insiders_as_of`` derivation used a different row-eligibility predicate than the holders aggregator. A Form 4 row with a null ``post_transaction_shares`` would have advanced the chip ahead of the actual snapshot the ring renders. Fixed by extracting a shared ``isInsiderHoldingRow`` predicate. Regression test added.

Threads from #756 (Ownership over time): same per-category ``as_of_date`` field name will be reused by the L2 time-series chart's per-period freshness annotations.

Closes #767.

### DIFF
--- a/frontend/src/components/instrument/OwnershipFreshnessChips.test.tsx
+++ b/frontend/src/components/instrument/OwnershipFreshnessChips.test.tsx
@@ -1,0 +1,125 @@
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+
+import { OwnershipFreshnessChips } from "./OwnershipFreshnessChips";
+import type { SunburstCategory, SunburstRings } from "./ownershipRings";
+
+const TODAY = new Date("2026-05-02T00:00:00Z");
+
+function category(
+  key: SunburstCategory["key"],
+  as_of_date: string | null,
+  shares: number = 100,
+): SunburstCategory {
+  const labels: Record<SunburstCategory["key"], string> = {
+    institutions: "Institutions",
+    etfs: "ETFs",
+    insiders: "Insiders",
+    treasury: "Treasury",
+  };
+  return {
+    key,
+    label: labels[key],
+    shares,
+    reported_total: shares,
+    resolved_leaf_shares: shares,
+    leaves: [],
+    within_category_gap: 0,
+    as_of_date,
+  };
+}
+
+function rings(categories: SunburstCategory[]): SunburstRings {
+  return {
+    total_shares: 1_000_000_000,
+    reported_total: 1_000_000_000,
+    categories,
+    category_residual: 0,
+  };
+}
+
+describe("OwnershipFreshnessChips", () => {
+  it("renders nothing when there are no categories", () => {
+    const { container } = render(
+      <OwnershipFreshnessChips rings={rings([])} today={TODAY} />,
+    );
+    expect(container.firstChild).toBeNull();
+  });
+
+  it("renders one chip per category in input order", () => {
+    render(
+      <OwnershipFreshnessChips
+        rings={rings([
+          category("institutions", "2026-03-31"),
+          category("etfs", "2026-03-31"),
+          category("insiders", "2026-04-30"),
+          category("treasury", "2026-03-28"),
+        ])}
+        today={TODAY}
+      />,
+    );
+    const chips = screen.getAllByRole("listitem");
+    expect(chips).toHaveLength(4);
+    expect(chips[0]!.textContent).toMatch(/Institutions/);
+    expect(chips[3]!.textContent).toMatch(/Treasury/);
+  });
+
+  it("colour-codes by freshness level via data-freshness-level", () => {
+    render(
+      <OwnershipFreshnessChips
+        rings={rings([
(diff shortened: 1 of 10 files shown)


Luke-Bradford:
## Resolution for PR #770 round 1 (2026-05-02 01:28)

> WARNING: `isInsiderHoldingRow` is defined independently in both files; the PR's own motivation is preventing predicate drift between the aggregator and the freshness chip, yet the fix lands two separate copies of the predicate that can drift again.

**FIXED 3bd604c** — extracted to `frontend/src/components/instrument/ownershipInsiders.ts` as the single source of truth. Both `OwnershipPanel.tsx` and `OwnershipPage.tsx` now import `isInsiderHoldingRow` + `InsiderRowShape` from it; no other file defines or duplicates the predicate.

You're right — leaving two copies undid the structural guarantee. Bot caught what my own "4 lines, fold when a third consumer appears" rationale missed.

> NITPICK: `today={new Date()}` allocates a new `Date` on every render

**FIXED 3bd604c** — `OwnershipPanel` body extracted into a child `PanelBody` component that captures `today` via `useMemo(() => new Date(), [])`. `OwnershipPage` gets the same `useMemo` inline. Chip strip now memoises cleanly across parent re-renders.

> PREVENTION: add a lint rule or shared-module extraction so the predicate lives in one place

**EXTRACTED `frontend/src/components/instrument/ownershipInsiders.test.ts`** — 4 unit tests pinning the predicate's behaviour (non-derivative + parseable shares = include; derivative or null shares = exclude). Locks against future drift via behavioural tests rather than a CI grep gate. Grep verification: `isInsiderHoldingRow` now appears in exactly 4 files — definition + test + 2 importers.

Luke-Bradford:
## Resolution for PR #770 round 2 (2026-05-02 01:44)

> NITPICK: `parseIsoDate("YYYY-MM-DD")` resolves to UTC midnight while `new Date()` in production resolves to local time, so `ageInDays` can be ±1 day near local midnight; harmless at the cadence scales used (30-270 days) but worth a comment.

**FIXED** — added an inline caveat in `parseIsoDate` documenting the UTC-vs-local skew + the cadence-scale rationale that makes it harmless. If we ever surface hour-precision freshness, the comment flags where to align both sides on UTC.

PR #772, opened 2026-05-02
feat(#768): Form 3 ingester — fetch, parse, upsert (PR 2/N)

Luke-Bradford:
## What

PR 2 of N for #768. The fetch + parse + upsert path on top of PR 1's parser. ``ingest_form_3_filings`` walks ``filing_events`` for Form 3/3-A accessions without an ``insider_filings`` row, fetches each XML through the shared SEC fetcher, parses, and upserts across four tables.

## Why

PR 1 shipped the parser + ``insider_initial_holdings`` table; without an ingester they're dead weight. PR 3 will fold the Form 3 baseline into ``get_insider_summary`` so the cumulative running total surfaces on the ownership card.

## Test plan

- [x] 9 integration tests covering: rich happy path (4-table writes), fetch-404 + parse-miss tombstones, Form 4 not picked up by Form 3 ingester, per-instrument scope, idempotency, stale-child cleanup on reparse, ``3/A`` amendment path, XSL URL canonicalisation.
- [x] Pre-push gates: ruff / format / pyright / dark-class — all green.

## Codex review (CLAUDE.md checkpoint 2)

- **MEDIUM (FIXED)**: ``replace-on-reparse`` was only true for holdings. Footnotes + filers were ``ON CONFLICT``-only — a parser version that stops emitting a footnote / secondary joint-filer would leave stale rows pinned forever. Added DELETE-then-INSERT for both, mirroring holdings.
- **HIGH (DEFERRED to PR 3 with explicit precondition)**: Some SEC ingest paths store ``primary_document_url`` as ``-index.htm`` rather than the raw XML. If Form 3 rows land in ``filing_events`` via that path, the ingester would tombstone valid filings. This is the **same shape risk the existing Form 4 ingester carries today**; both assume an XML primary doc URL. PR 3 wires the scheduler tick that decides which Form 3 URLs end up in ``filing_events``, so PR 3 is the natural home for the fix.

Refs #768.

### DIFF
--- a/app/services/insider_form3_ingest.py
+++ b/app/services/insider_form3_ingest.py
@@ -415,11 +415,11 @@ def ingest_form_3_filings_for_instrument(
               AND fe.filing_type IN ('3', '3/A')
               AND fe.primary_document_url IS NOT NULL
               AND fe.instrument_id = %s
-              AND fil.accession_number IS NULL
+              AND (fil.accession_number IS NULL OR fil.parser_version < %s)
             ORDER BY fe.filing_date DESC, fe.filing_event_id DESC
             LIMIT %s
             """,
-            (instrument_id, limit),
+            (instrument_id, _FORM3_PARSER_VERSION, limit),
         )
         for row in cur.fetchall():
             candidates.append((int(row[0]), str(row[1]), _canonical_form_4_url(str(row[2]))))
@@ -439,7 +439,10 @@ def ingest_form_3_filings(
     Candidate selector:
       1. ``fe.filing_type IN ('3', '3/A')``.
       2. ``fe.primary_document_url IS NOT NULL``.
-      3. No existing ``insider_filings`` row.
+      3. No existing ``insider_filings`` row, OR the existing row's
+         ``parser_version`` is below the current
+         ``_FORM3_PARSER_VERSION`` (re-parse trigger when the parser
+         shape changes — Codex round 1 / round 2 review of #768 PR2).
       4. No backfill floor — Form 3 volume per issuer is bounded
          (~5-30 lifetime), and a 10-year-old Form 3 for a still-
          serving officer IS the correct baseline.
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
## Resolution for PR #772 round 2 (2026-05-02 11:49)

> BLOCKING: `_FORM3_PARSER_VERSION` is stored on upsert but neither candidate selector has `OR fil.parser_version < %s`; bumping the constant never triggers re-ingestion, contradicting the module docstring.

**FIXED 4ce26c7** — both selectors now branch on `(fil.accession_number IS NULL OR fil.parser_version < %s)` with `_FORM3_PARSER_VERSION` as the parameter. Regression test added (`TestParserVersionRefresh.test_existing_filing_below_current_parser_version_is_re_picked`): pre-populates `insider_filings` with `parser_version=0`, asserts the next ingest re-parses + bumps to current.

> WARNING: Upsert exception path calls `conn.rollback()` then `continue` without writing a tombstone, so a persistently bad accession is re-fetched on every scheduler tick indefinitely.

**FIXED 4ce26c7** — except branch now calls `_write_form_3_tombstone` after rollback. Regression test added (`TestUpsertFailureTombstone.test_upsert_exception_writes_tombstone_so_scheduler_does_not_retry`): monkey-patches `upsert_form_3_filing` to raise, asserts tombstone written + second pass scans zero candidates.

> PREVENTION: assert `_FORM3_PARSER_VERSION` appears in both the upsert param list and the candidate-selector WHERE clause before merge.

**EXTRACTED in tests/test_insider_form3_ingest.py** — `TestParserVersionRefresh` is the behavioural enforcement. Locks the contract via test rather than a CI grep gate.

Note on Form 4 ingester: it carries the same parser-version-reparse gap (Codex flagged this in round 1). Out of scope for this PR — same pre-existing behaviour. Tracking as tech debt.

Luke-Bradford:
## Resolution for PR #772 round 3 (2026-05-02 11:57)

> NITPICK: Upsert exceptions increment `parse_misses` rather than a distinct counter, so scheduler metrics cannot distinguish parse failures from constraint violations.

**REBUTTED — deferred to PR3** with reason: PR3 wires the scheduler tick that consumes `IngestForm3Result`. Splitting `parse_misses` into `parse_misses` + `upsert_failures` is a public-API change that's only meaningful once a consumer reads the distinction. Folding it into PR3 alongside the scheduler-side metrics avoids churning the dataclass twice. Tracked as a precondition for the PR3 metrics surface.

PR #774, opened 2026-05-02
feat(#768): API endpoint + ownership-card wiring for baseline insiders (PR 4/N)

Luke-Bradford:
## What

PR 4 of N for #768 — closes the loop end-to-end. Insiders with a Form 3 baseline grant but no Form 4 activity now surface on the ownership card's per-officer ring + L2 drilldown.

## Components

- **Backend endpoint** ``GET /instruments/{symbol}/insider_baseline`` reading the PR 3 reader. Same SEC-coverage 404 gate + empty-state contract as ``insider_transactions``.
- **Frontend API client** ``fetchInsiderBaseline`` with 404-tolerance for non-covered issuers.
- **L1 ``OwnershipPanel``**: fetches baseline alongside Form 4, merges into ``insider_holders``, factors latest ``as_of_date`` from BOTH sources into the freshness chip.
- **L2 ``OwnershipPage``**: same merge + ``buildFilerRows`` emits baseline rows with the ``baseline:{cik}:{n|d}`` key so a click on a baseline wedge from L1 lands on the right L2 row.

## Test plan

- [x] Backend: ``test_api_instrument_sec_gates`` parametrisation covers ``/insider_baseline`` (404 with no SEC CIK; 200 with SEC CIK).
- [x] Frontend: 7 ``OwnershipPanel.test.ts`` tests covering merge shape, empty-state, no-CIK-collision keying, null/zero-shares filter, freshness max-of-both branch (including Codex-flagged "baseline older than Form 4" branch).
- [x] Full frontend suite: 825 passed.
- [x] Pre-push gates: ruff / format / pyright / typecheck — all green.

## Codex review (CLAUDE.md checkpoint 2)

No findings on the three focus points (key prefix safety, L1→L2 highlight match, freshness merge logic). Coverage gap on "baseline older than Form 4" branch flagged + addressed.

Refs #768.

### DIFF
--- a/app/api/instruments.py
+++ b/app/api/instruments.py
@@ -1161,7 +1161,7 @@ class EightKFilingsResponse(BaseModel):
 # wires a single provider; per-region integration PRs add more.
 _EIGHT_K_PROVIDERS: tuple[str, ...] = ("sec_8k_events",)
 _DIVIDEND_PROVIDERS: tuple[str, ...] = ("sec_dividend_summary",)
-_INSIDER_PROVIDERS: tuple[str, ...] = ("sec_form4",)
+_INSIDER_PROVIDERS: tuple[str, ...] = ("sec_form4", "sec_form3")
 
 
 def _validate_provider(provider: str | None, allowed: tuple[str, ...]) -> None:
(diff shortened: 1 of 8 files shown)


Luke-Bradford:
## Resolution for PR #774 round 1 (2026-05-02 12:41)

> BLOCKING: `_INSIDER_PROVIDERS` is shared with the pre-existing `get_instrument_insider_transactions` endpoint, so adding 'sec_form3' to it silently makes `GET /insider_transactions?provider=sec_form3` pass `_validate_provider` validation against a Form-4-only reader; the two endpoints need separate provider tuples.

**FIXED** (next push) — split into `_INSIDER_PROVIDERS = ('sec_form4',)` and new `_INSIDER_BASELINE_PROVIDERS = ('sec_form3',)`. Two regression tests added asserting `/insider_transactions?provider=sec_form3` and `/insider_baseline?provider=sec_form4` both 400.

> PREVENTION: shared provider tuple cross-contamination — grep for `_validate_provider(provider, _INSIDER_PROVIDERS)` in CI and assert each call site's allowed-tuple contains only the providers the underlying reader actually consumes.

**EXTRACTED in tests/test_api_instrument_sec_gates.py** — two parametrised regression tests pin both directions of the contamination. A future merge of the tuples is caught structurally rather than via CI grep.

Luke-Bradford:
## Resolution for PR #774 round 2 (2026-05-02 12:46)

> WARNING: `insiders_as_of` iterates all `baseline.rows` for `latest_baseline` without the `shares > 0` guard, so a null-shares row (filtered out of `baselineToInsiderHolders`) can advance the freshness chip to a date for a wedge that never renders. Same divergent-predicate issue the PR comment cites for Form 4 in #767.

**FIXED** (next push) — extracted `isBaselineHoldingRow` + `InsiderBaselineRowShape` to `frontend/src/components/instrument/ownershipInsiders.ts` (same module that owns `isInsiderHoldingRow`). Single source of truth for baseline-row eligibility, used by all 5 sites that previously had the predicate inlined or missing: holders builders on L1 + L2, `buildFilerRows` table writer, both `latest_baseline` scans for the freshness chip.

> PREVENTION: add a lint rule or test assertion that any helper computing `latest_baseline` date applies the same `parseShareCount > 0` guard as the corresponding `baselineToHolder*` function.

**EXTRACTED in tests/test files** — 4 new cases in `ownershipInsiders.test.ts` pinning `isBaselineHoldingRow` behaviour (positive shares, null, zero, unparseable, derivative). Extended `OwnershipPanel.test.ts` "drops baseline rows with null or non-positive shares" to assert `insiders_as_of` stays null. The shared module pattern (same as the PR #770 `isInsiderHoldingRow` extraction) makes structural drift impossible — the only way to drift would be to bypass the helper, which the new tests would catch immediately.

Issue #730, opened 2026-04-30
feat: SEC 13F-HR institutional holdings ingest (Plan C.2)


Luke-Bradford:
## Context

Per memory + settled-decisions.md, Plan C.2 (13F ingest) is on the build queue, paused under the 2026-04-18 product-visibility pivot until #313 + #314 ship. Re-prioritised by operator 2026-05-01 to unblock the ownership reporting card (#729).

13F-HR is filed quarterly by every institutional manager with >\$100M discretionary AUM. SEC publishes the list of all 13Fs daily via the master-index (already ingested by our SEC pipeline) and the bodies live at the per-accession archive URL.

Single biggest unlock: 13F filers include ETFs, mutual funds, hedge funds, and pension managers. Aggregating across filers gives institutional ownership per stock; filtering by filer type gives ETF ownership separately. One ingest, two slices of the ownership card.

### 1. Schema

\`\`\`sql
CREATE TABLE institutional_filers (
    filer_id           BIGSERIAL PRIMARY KEY,
    cik                TEXT NOT NULL UNIQUE,
    name               TEXT NOT NULL,
    filer_type         TEXT,  -- 'INV', 'ETF', 'INS', 'BD', etc. from form CRD
    aum_usd            NUMERIC,
    last_filing_at     TIMESTAMPTZ,
    fetched_at         TIMESTAMPTZ NOT NULL DEFAULT NOW()
);

CREATE TABLE institutional_holdings (
    holding_id         BIGSERIAL PRIMARY KEY,
    filer_id           BIGINT NOT NULL REFERENCES institutional_filers,
    instrument_id      BIGINT NOT NULL REFERENCES instruments,
    accession_number   TEXT NOT NULL,
    period_of_report   DATE NOT NULL,  -- quarter end
    shares             NUMERIC NOT NULL,
    market_value_usd   NUMERIC,
    voting_authority   TEXT,  -- 'SOLE', 'SHARED', 'NONE'
    is_put_call        TEXT,  -- options vs underlying
    filed_at           TIMESTAMPTZ NOT NULL,
    UNIQUE (accession_number, instrument_id)
);

CREATE INDEX idx_holdings_instrument_period ON institutional_holdings (instrument_id, period_of_report DESC);
\`\`\`

### 2. Parser

- 13F-HR primary doc is XML (\`primary_doc.xml\`) for filer metadata + a separate \`infotable.xml\` for holdings.
[...]

PR #783, opened 2026-05-02
feat(#781): 13F CUSIP resolver via fuzzy issuer-name match

Luke-Bradford:
## What

Re-scopes #781 from PDF-list parser (SEC publishes the 13F Official List as PDF only — no machine-readable feed) to mining the 13F-HR ingest stream we already fetch.

- ``sql/099_unresolved_13f_cusips.sql`` — tracking table + CHECK enum + indexes
- ``app/services/institutional_holdings.py`` — ingester now records unresolved CUSIPs
- ``app/services/cusip_resolver.py`` — fuzzy-match resolver + probe-then-write conflict detection
- ``tests/test_cusip_resolver.py`` — 18 unit + integration tests

## Why

Every CUSIP that appears in a 13F-HR is by definition a 13F-eligible security. We already fetch each filing's holdings during the #730 ingest, so the unresolved-CUSIP set populates naturally from the data we have rather than parsing a PDF. Closes the #740 backfill bottleneck for the ownership card.

## Test plan

- [x] ``uv run ruff check / format / pyright`` — clean
- [x] ``uv run pytest tests/smoke/test_app_boots.py tests/test_cusip_resolver.py tests/test_institutional_holdings_ingester.py`` — 39 pass
- [x] Codex pre-push review — clean after 1 round of fixes (2 findings)

## Codex pre-push findings (all addressed)

1. **High** — ``_best_match`` picked arbitrary first-wins on top-score ties. ``"Alphabet Inc CL A"`` and ``"Alphabet Inc CL C"`` both normalise to ``"ALPHABET"`` after share-class strip; an unresolved CUSIP for either class has two equally-good candidates. Now returns ``is_ambiguous=True`` and the resolver tombstones with status=``ambiguous``.
2. **Medium** — ``ON CONFLICT DO NOTHING`` silently masked pre-existing different mappings as success. Replaced with probe-then-write that distinguishes ``already_resolved`` (same iid) from ``conflict`` (different iid; existing row preserved, backlog tombstoned for operator audit).

## Tradeoffs

- **Conservative threshold (0.92)** — false positives in ``external_identifiers`` corrupt every downstream join. Lower recall is preferable; rejected names go to ``unresolvable`` for operator review.
- **No new dep** — stdlib ``difflib.SequenceMatcher`` is the resolver. ``rapidfuzz`` would be the incremental upgrade if recall tightens.
- **Linear candidate scan** — fine at the candidate counts involved (low thousands of instruments × hundreds of unresolved CUSIPs per pass = sub-second).

### DIFF
--- a/app/services/cusip_resolver.py
+++ b/app/services/cusip_resolver.py
@@ -1,0 +1,613 @@
+"""CUSIP → instrument_id resolver via fuzzy issuer-name match (#781).
+
+Walks ``unresolved_13f_cusips`` (populated by the 13F-HR ingester
+on each unresolved CUSIP) and tries to promote rows into
+``external_identifiers`` by fuzzy-matching the filer-supplied
+``name_of_issuer`` against ``instruments.company_name``. Successful
+matches are removed from the unresolved table; rejections are
+tombstoned with ``resolution_status='unresolvable'`` so the next
+run skips them.
+
+This is the practical alternative to parsing the SEC's quarterly
+Official List of Section 13(f) Securities (PDF-only, no
+machine-readable feed). Every CUSIP that appears in any 13F-HR
+filing is by definition a 13F-eligible security; we already fetch
+the holdings during the #730 ingest so the unresolved-CUSIP set
+is populated naturally.
+
+Match strategy (deliberately conservative — false positives in
+``external_identifiers`` corrupt every downstream join):
+
+  1. **Normalise** both sides:
+     - Strip common corporate-form suffixes (``"INC"``, ``"CORP"``,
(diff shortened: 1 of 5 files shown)


Luke-Bradford:
NIT addressed in fbe354e — comment updated to say 'every trailing corporate-form suffix token' with reasoning. **FIXED fbe354e**

Luke-Bradford:
Both NITs addressed in this push:

- `cusip_resolver.py:191` — docstring step 5 now says 'every trailing corporate-form suffix token in turn' with example. **FIXED**
- `institutional_holdings.py:_record_unresolved_cusip` — added explicit note that ON CONFLICT DO UPDATE deliberately preserves resolution_status; tombstoned CUSIPs need operator clear before retry. **FIXED**

Issue #792, opened 2026-05-03
Ownership provenance: per-slice freshness footer + audit trail surface


Luke-Bradford:
Parent: #788

Every number on the ownership card is in the DB with full provenance (accession + period + filed_at) but **none of it is shown to the operator**. The operator must be able to answer ""where did this number come from?"" without leaving the card.

## Problem

DB columns already populated:
- ``institutional_holdings.accession_number, period_of_report, filed_at``
- ``blockholder_filings.accession_number, submission_type, date_of_event, filed_at``
- ``insider_transactions.accession_number, txn_date``
- ``insider_initial_holdings.accession_number, as_of_date``
- ``def14a_beneficial_holdings.accession_number, period_end, filed_at``
- ``share_count_history.period_end, latest_form_type, latest_filed_date``

Frontend shows none of this. Operator who wants to verify a wedge has to leave the app and search EDGAR by hand.

## Acceptance criteria

1. **Per-slice provenance footer.** Each slice on hover/click shows:
   - ``As of YYYY-MM-DD`` (period_of_report or as_of_date)
   - ``Source: SEC accession <N>``, link to https://www.sec.gov/cgi-bin/browse-edgar?action=getcompany&filenum={N} (or full archive URL)
   - ``Filed: YYYY-MM-DD HH:MM`` (filed_at)
   - ``Filer count: N`` for aggregated slices

2. **Per-holder accession link** in L2 holder tables — clicking a row opens the underlying SEC accession page in a new tab.

3. **Shares-outstanding provenance.** Header strip on the card: ``Shares outstanding: 448,375,157 (as of 2026-03-18, source: SEC 10-Q filing <accession>, XBRL dei:EntityCommonStockSharesOutstanding)``.

4. **Coverage methodology disclosure.** ``About this data`` link expands to:
   - Which forms feed which slice
   - Which CIKs are seeded vs auto-discovered
   - The dedup rule applied
   - The denominator definition
   - Last successful ingest run per source

5. **Per-source freshness chip.** Reuse the freshness chip pattern from [project_chart_redesign.md](#) (PR #670). Card shows ""13F: 12 days behind"", ""13D/G: live"", ""Form 4: 2 hours behind"" etc.

## API additions

[...]

Issue #788, opened 2026-05-03
EPIC: Ownership card production-trustworthy for investment decisions


Luke-Bradford:
## Goal

Make the per-instrument ownership card a chart an operator could base a real investment decision on. Today it's not.

## Why this is an epic

Two ship-blockers are conceptual (codex audit, 2026-05-03):

- **Sunburst is not a partition.** Insiders ↔ Blockholders ↔ Institutions ↔ DEF 14A overlap structurally. We append + sum as if mutually exclusive. Cohen on GME = ~75M displayed when reality is ~38M.
- **Wrong denominator.** Frontend uses ``shares_outstanding + treasury_shares``; backend contract + XBRL view say ``shares_outstanding``. Every non-treasury wedge understated when treasury exists.

Plus: ~2% institutional coverage on GME (7 of ~380 13F filers per quarter), no provenance surfacing, schema drift on insider baseline, ``filer_name`` vs ``reporter_name`` label bug, fake ""unresolved filers"" gap label.

### Beneficial ownership (SEC Rule 13d-3)

Holder is beneficial owner if they have **either**: (1) voting power, (2) investment power, or (3) right to acquire within 60 days. This is what every regulated filing reports.

### Filing forms — what each says, when, how to parse

| Form | Reports | Threshold / cadence | Reference |
|---|---|---|---|
| 13F-HR | Institutional holdings of Section 13(f) securities | AUM ≥ \$100M, quarterly within 45 days | https://www.sec.gov/divisions/investment/13ffaq.htm |
| 13D | Active beneficial owner | ≥5%, within 10 days | https://www.sec.gov/fast-answers/answerssched13htm.html |
| 13G | Passive beneficial owner | ≥5%, annual + amendments | same |
| Form 3 | Insider initial statement | Within 10 days of becoming insider | https://www.sec.gov/about/forms/form3.pdf |
| Form 4 | Insider transaction | Within 2 business days | https://www.sec.gov/about/forms/form4.pdf |
| Form 5 | Insider annual catch-up | Annual | https://www.sec.gov/about/forms/form5.pdf |
| DEF 14A | Proxy beneficial ownership table | Annual, Item 6 of Schedule 14A | https://www.sec.gov/files/proxy-rules-quick-reference.pdf |
| N-PORT | Mutual fund / ETF monthly holdings | Monthly, public quarterly | https://www.sec.gov/files/forms-nport.pdf |

### Slice math (denominator)

[...]

PR #800, opened 2026-05-03
feat(#792): provenance footer + per-holder edgar_url + shares-outstanding source (Batch 3 of #788)

Luke-Bradford:
## Summary

- **Per-holder `winning_edgar_url`** — derived from the accession's CIK segment + the accession number, formatted as the SEC archive index URL. Surfaces on each canonical holder + dropped source so the operator can click through to the source filing without leaving the card.
- **Shares-outstanding source** — `shares_outstanding_source` now carries the real producing accession + form_type from `financial_facts_raw`, mirroring the ordering the `instrument_share_count_latest` view uses internally. Frontend renders a click-through link next to the "as of" date in the panel header.
- Out of scope (separate PR): per-slice provenance summary block, freshness chip strip, methodology disclosure expandable. They land alongside the Batch 4 first-run UX.

## Test plan

- [x] `tests/test_ownership_rollup.py::TestProvenance` — 3 cases pinning the URL derivation, holder threading, and shares-outstanding accession provenance.
- [x] All existing 17 rollup tests still pass.
- [x] Frontend typecheck + 8 panel tests pass; existing test fixtures threaded through with `winning_edgar_url: null`.
- [x] All 4 local gates: ruff check, ruff format --check, pyright, pytest.


Luke-Bradford:
BLOCKING — backend-computed shares-outstanding edgar_url: `FIXED b3ba222`. `SharesOutstandingSource.edgar_url` is now derived in `_read_shares_outstanding` via `edgar_archive_url(accession)` and shipped as a pre-computed string in the response payload. The frontend no longer rolls its own URL (the prior `browse-edgar?action=getcompany&filenum=` was wrong because `filenum` expects a SEC file number, not an accession). New regression test `test_shares_outstanding_source_edgar_url_backend_computed` pins the threading.

Issue #793, opened 2026-05-03
First-run UX: ingest progress strip + backfill-on-activation pattern


Luke-Bradford:
Parent: #788

User quote (2026-05-03): ""we also need to be mindful of the first start up for a user, that once they have got set up with at least one api key for etoro, we should have visibility of the data being ingested, so they know how far the updates are, how long it will take or anything, to make that a good user experience. So... why am I missing data, is there still a lot more to do or what?""

Today there is **no operator-visible signal** that ingest is in progress, where it is, how much remains, or which sources are healthy. ``data_ingestion_runs`` rows accumulate silently. Coverage grows over a week-long warm-up window with no UX feedback.

## Problem

1. New user finishes operator setup → frontend shows charts that look ""half-baked"" because backfill is mid-flight.
2. New ingest pipeline ships → operator can't see existing data, only data filed since pipeline went live (no historical pass).
3. Failed ingest runs are buried in ``data_ingestion_runs.error`` text, not surfaced.
4. No estimate of time-to-coverage so operator can decide whether to wait or come back tomorrow.

## Acceptance criteria

1. **Persistent ingest-status strip** — top of every page, dismissible. Shows aggregate state:
   - ``Ingesting: 7 sources running, 12 complete, 1 failed``
   - Click expands to per-source rows: source name, last-run-status, last-success time, queue depth, ETA.

2. **Per-instrument coverage breadcrumb** — when user opens an instrument page, header shows ``Ownership coverage: 23% (improving — 4 ingest passes pending)`` until coverage reaches threshold.

3. **First-run wizard.** After operator setup completes:
   - Enumerate all seeded instruments × all ingest pipelines.
   - Display estimated wall-clock for first-pass coverage.
   - Stream real-time progress as backfill runs.
   - On completion, show ""Coverage complete — go explore"".

4. **Backfill-on-activation pattern.** New table ``ingest_backfill_queue (instrument_id, pipeline_name, priority, status, queued_at, completed_at, error)``. New ingest pipeline migrations populate this for every already-seeded instrument. Boot-time worker drains it. Idempotent.

5. **Failed-run surface.** ``/api/operator/ingest-failures`` lists failures from the last 7 days with link-out to the failing accession. Auto-retry for transient errors with backoff; permanent flag for parse failures.

## Data model

New tables:

    CREATE TABLE ingest_backfill_queue (
      instrument_id BIGINT REFERENCES instruments(instrument_id),
[...]

PR #801, opened 2026-05-03
feat(#793): operator ingest-health page (Batch 4 of #788)

Luke-Bradford:
## Summary

- Provider-grouped operator-facing visibility into every data source the app ingests from. Answers the user's 2026-05-03 ask: "we need to be mindful of the first start up… we should have visibility of the data being ingested."
- Migration 105: \`ingest_backfill_queue\` table with per-(instrument, pipeline) PK, partial indexes for the drainer (jobs process — separate PR), CHECK on \`triggered_by\`.
- \`app/services/ingest_status.py\`: grouped-provider rollup + recent-failures + idempotent backfill enqueue. Provider taxonomy curated (SEC fundamentals / SEC ownership / eToro / Other regulated / Uncategorised). New sources surface under \`other\` so the operator sees them rather than silently joining a wrong group.
- \`app/api/operator_ingest.py\`: 3 GETs + 1 POST under \`/operator/ingest-*\`, auth-gated.
- \`frontend/src/pages/IngestHealthPage.tsx\`: 3 sections (provider group cards + recent failures + backfill queue), auto-refresh every 30s.

## Test plan

- [x] \`tests/test_ingest_status.py\` — 31 cases: provider grouping (parametrised), state machine across never_run / red / amber / green / queue-failed, queue-backlog folding, recent-failures surface, idempotent enqueue.
- [x] All 4 local gates pass: ruff check, ruff format --check, pyright, pytest.
- [x] Frontend pnpm typecheck passes.

## Codex pre-push review caught (all fixed)

- **Auth**: bare router exposed ingest history + failure text + a public POST. Fixed by adding \`require_session_or_service_token\` dependency to the router.
- **State fold**: failed-only sources misclassified as \`never_run\`. Fixed; now \`red\` when any source has attempted but never succeeded. Regression test added.
- **Auto-refresh**: tick-based pattern was a no-op (effect dep was \`[]\`). Fixed via direct setInterval on the three refetch callbacks.

## Out of scope

- Drainer worker that consumes \`ingest_backfill_queue\` rows — lives in the jobs process per #719 settled topology; ships alongside per-pipeline wiring in later batches.
- Persistent ingest-status strip on every page.
- Guided first-run wizard variant — the page itself is usable as the first-run surface.

### DIFF
--- a/app/api/operator_ingest.py
+++ b/app/api/operator_ingest.py
@@ -1,0 +1,295 @@
+"""Operator-facing ingest health endpoints (#793, Batch 4 of #788).
+
+Three GETs + one POST surface the data feeding the
+``/admin/ingest-health`` page:
+
+  * ``GET /api/operator/ingest-status`` — grouped-provider rollup
+    with per-group state + per-source last-run summary + queue
+    backlog counts. Drives the operator card grid.
+  * ``GET /api/operator/ingest-failures`` — last-7-days failed /
+    partial runs for the "needs attention" list. Bounded by
+    ``limit`` so a flap-storm can't swamp the UI.
+  * ``GET /api/operator/ingest-backfill-queue`` — queue rows for
+    the per-pipeline drilldown view (current status, attempts,
+    last_error). Bounded by ``limit``.
+  * ``POST /api/operator/ingest-backfill`` — enqueue a backfill
+    request from an operator click. Idempotent.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime
+from typing import Literal
+
+import psycopg
+import psycopg.rows
+from fastapi import APIRouter, Depends, HTTPException, Query
+from pydantic import BaseModel, Field
+
+from app.api.auth import require_session_or_service_token
+from app.db import get_conn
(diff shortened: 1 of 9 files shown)


Luke-Bradford:
BLOCKING — `queue_total` semantic mismatch: `FIXED 712f370`. Total now = pending + running + failed (complete excluded; those are done). New regression test `test_queue_total_includes_pending_running_and_failed` pins the contract; existing `test_queue_backlog_counts_fold_into_groups` updated to expect 2 (pending + failed) instead of 1.

Luke-Bradford:
Round 3 — both medium findings: `FIXED ba86731`.

* **useEffect interval dep array**: replaced the unstable `[statusState, failuresState, queueState]` deps with the canonical ref-capture pattern. A ref holds the latest refetch closure (refreshed every render); a single stable interval calls `ref.current()`. No more flicker / timer leaks.
* **FK race in enqueue**: dropped the pre-check SELECT; rely on the `ingest_backfill_queue.instrument_id` FK and catch `ForeignKeyViolation` to return a clean 404. Race-safe + one fewer round-trip.

PR #816, opened 2026-05-03
feat(#809): cik_raw_documents sibling table + reconciliation cache

Luke-Bradford:
## What
Per-CIK raw-document store sibling of \`filing_raw_documents\` (per-accession). For SEC docs keyed by CIK rather than accession — \`submissions.json\` and \`companyfacts.json\` are rolling per-issuer documents.

- \`sql/109_cik_raw_documents.sql\` — \`(cik, document_kind)\` PK, CIK CHECK regex enforcing 10-digit zero-padded form at DB layer.
- \`app/services/cik_raw_filings.py\` — store/read/summary helpers with \`max_age\` cache-miss semantics.
- \`app/services/reconciliation.py\` — wires companyfacts.json fetch to the cache (24h TTL). Cache reads + writes both go through SHORT-LIVED dedicated connections so durability is independent of the caller's transaction lifecycle.
- 11 new integration tests.

## Why
Operator audit 2026-05-03: every reconciliation spot-check was re-fetching \`companyfacts.json\` from SEC at 10 req/s. Cache turns repeated CIKs in a sweep into hot reads, plus gives operators a "what did SEC say last time?" audit trail.

## Architecture decision: dedicated cache connection
Codex review caught two real bugs in earlier iterations:
1. Piggy-back on caller's connection → durability lost when caller rolls back (zero-finding sweeps) or transaction state corrupted on cache-write failure.
2. Same-cluster assumption from \`settings.database_url\` → tests against \`ebull_test\` would write cache to dev DB.

Fix: derive host/port/user/dbname from \`conn.info\` (authoritative for where the connection is talking to) + password from settings (psycopg strips it from \`conn.info.dsn\`). Open a dedicated short-lived connection per cache read/write.

## Test plan
- [x] \`uv run pytest tests/test_cik_raw_filings.py tests/test_reconciliation.py\` — 30/30 pass
- [x] \`uv run ruff check .\` clean
- [x] \`uv run ruff format --check .\` clean
- [x] Codex review (5 rounds) — all findings addressed; final pass clean

## Follow-ups
- Race tolerated for v1 → tech-debt #815 (low-severity; operator-triggered single-sweep workflow makes it benign).
- Re-wash workflow (the OTHER half of #809) — separate PR once we have raw filings flowing.


Luke-Bradford:
Fixed in 0b810ea.

- WARNING: `conn.info.host` None-guard for Unix sockets — `FIXED 0b810ea`. Falls back to `settings_parsed.hostname` / `5432` when `info.host` / `info.port` is None.
- PREVENTION: extracted to docs/review-prevention-log.md — `EXTRACTED docs/review-prevention-log.md`. New entry "Interpolating conn.info.host into URLs without a None guard" with self-review grep prompt.

PR #820, opened 2026-05-03
feat(#813): .RTH suffix stripping in CIK discovery

Luke-Bradford:
## What
First sub-task of #813. Adds \`.RTH\` suffix stripping to \`cik_discovery\` so eToro's regular-trading-hours operational duplicates fold onto the underlying SEC CIK during the discovery sweep.

- \`_OPERATIONAL_SUFFIXES\` closed tuple (\`.RTH\` only — warrants/preferreds deliberately excluded).
- Two-pass strategy in \`discover_ciks\`: direct matches first (underlying wins shared CIK), suffix-stripped fallback second.
- 3 new tests including the regression for the underlying-loses bug Codex caught.

## Why
Operator audit 2026-05-03 found 7,281 of 12,379 instruments with no CIK row. \`.RTH\` instruments are 559 of that cohort. Live sweep: 14 net new inserts (underlyings discovered through their \`.RTH\` counterparts where both were no-CIK).

The remaining ~545 \`.RTH\` instruments still have no CIK row of their own because the underlying already owns it — that's the right behaviour for filing attribution. UI redirect for \`.RTH\` rendering tracked as follow-up #819.

## Test plan
- [x] \`uv run pytest tests/test_cik_discovery.py\` — 10/10 pass
- [x] \`uv run ruff check .\` clean
- [x] Codex pre-push review (2 rounds) — high-severity underlying-loses bug fixed; final pass clean
- [x] Live sweep: \`scanned=7239 matches=592 inserted=14 misses=6647\`

## Follow-ups
- #819 — operational-duplicate redirect for .RTH instruments without separate CIK rows.
- #813 remaining sub-tasks: ETF map (\`company_tickers_exchange.json\`, \`company_tickers_mf.json\`), ADR resolver, manual override surface, scheduler entry, discovery report UI.

### DIFF
--- a/app/services/cik_discovery.py
+++ b/app/services/cik_discovery.py
@@ -211,6 +211,44 @@ def upsert_cik(
     return affected > 0
 
 
+# eToro-side operational suffixes that wrap the underlying common
+# stock ticker (no separate SEC entry — the RTH listing is the same
+# issuer as the underlying). Stripping the suffix and re-trying the
+# lookup catches ~560 no-CIK instruments on the dev cohort without
+# any false positives, because each suffix maps deterministically to
+# a single underlying.
+#
+# Add new suffixes ONLY when:
+#   1. The suffix is a deterministic operational duplicate of the
+#      underlying (not a separate security like a warrant or pref).
+#   2. The match rate is non-trivial against the operator audit
+#      cohort.
+#
+# Warrants (``-W``, ``-WT``) and preferreds (``-PA``, ``-PB``) are
+# DELIBERATELY not in this list — they are separate securities with
+# their own filings, and folding them onto the common-stock CIK
+# would mis-attribute filings on the pie chart.
+_OPERATIONAL_SUFFIXES: tuple[str, ...] = (
+    ".RTH",  # eToro "regular trading hours" duplicate
+)
+
+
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
REBUTTED — false positive.

Pass 2 of `discover_ciks` DOES increment `matches` at [app/services/cik_discovery.py:321](https://github.com/Luke-Bradford/eBull/blob/feature/813-cik-suffix-stripping/app/services/cik_discovery.py#L321):
```python
        matches += 1
        if upsert_cik(...):
            inserts += 1
```

The bot likely missed the line because the assignment is below the `if entry is None: misses += 1; continue` guard. Verified empirically: `pytest tests/test_cik_discovery.py::test_discover_strips_rth_suffix_and_matches_underlying` passes (asserts `matches_found == 1` against the `.RTH` resolution path which goes entirely through pass 2). Codex independently confirmed the bot is wrong.

PREVENTION: would file as `REBUTTED` rather than `EXTRACTED` since the suggested rule ("assert matches_found == rows_inserted") is a false equivalence — `upsert_cik` returns False on `uq_external_identifiers_provider_value` conflicts (the underlying-already-has-this-CIK case which is THE point of the suffix-stripping pass), so a positive resolution that no-ops on conflict still counts as a match but not an insert. The existing tests cover both branches separately.

PR #818, opened 2026-05-03
feat(#809): re-wash workflow + Form 4 wiring

Luke-Bradford:
## What
Re-wash framework that re-parses stored \`filing_raw_documents\` bodies under the current parser version. No SEC re-fetch when a parser bug ships.

- \`app/services/rewash_filings.py\` — ParserSpec registry + \`run_rewash\` orchestrator. Eager-fetches (accession, fetched_at) cohort identifiers; iterates per-row reading body + applying parser; bumps \`parser_version\` on success.
- \`app/services/insider_transactions.py\` — \`is_rewash\` flag on \`upsert_filing\` preserves original \`fetched_at\` on the conflict branch.
- \`scripts/rewash.py\` — operator CLI (\`--kind\`, \`--since\`, \`--dry-run\`).
- 10 integration tests.

## Why
Closes the second half of #809. PR #808 + #810/#811 retained raw bodies. This PR makes them re-parsable in place. Form 4 wired first (most common ownership kind, ~440k rows). Other kinds follow in separate PRs.

## Test plan
- [x] \`uv run pytest tests/test_rewash_filings.py tests/test_insider_transactions.py tests/test_insider_form3_ingest.py\` — 64/64 pass
- [x] \`uv run ruff check .\` clean
- [x] \`uv run ruff format --check .\` clean
- [x] \`uv run pyright\` clean
- [x] Codex pre-push review (2 rounds) — both findings (cursor+commit interaction, fetched_at mutation) addressed; final pass clean

## Follow-ups
- Concurrent re-wash + live-ingest duplicate-work race → tech-debt #817 (no corruption; idempotent upserts).
- Wire 13F / 13D/G / Form 3 / DEF 14A — separate PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #825, opened 2026-05-03
feat(#788/#809): register 13D/G blockholder rewash spec

Luke-Bradford:
## What
Third ownership kind in rewash registry (Form 4 = #818, Form 3 = #824, 13D/G here).

\`_apply_blockholders\` re-resolves instrument_id from \`filing.issuer_cusip\` via the same path the ingester uses (NOT re-using the stale typed-row value — Codex caught), then DELETE-then-INSERT replace pattern because the existing per-reporter upsert uses ON CONFLICT DO NOTHING.

3 new tests including the fresh-CUSIP regression.

## Test plan
- [x] \`pytest tests/test_rewash_filings.py\` 14/14
- [x] \`ruff\` / \`pyright\` clean
- [x] Codex pre-push review (2 rounds) — stale-instrument_id bug fixed; final pass clean

## Follow-ups
- 13F infotable + DEF 14A rewash specs (separate PRs).
- Historical CIK writer routing (separate PR).

### DIFF
--- a/app/services/rewash_filings.py
+++ b/app/services/rewash_filings.py
@@ -606,9 +606,14 @@ def _apply_blockholders(
 
     # Resolve canonical filer name + filer_id (preserved across
     # re-wash via ON CONFLICT (cik) DO UPDATE in _upsert_filer).
+    # ``reporting_persons`` is non-empty here — the empty-list guard
+    # above raises before this point — so the ``next()`` default
+    # falls back to the first reporter's name without needing the
+    # ``CIK {primary_filer_cik}`` branch the prior version had.
+    # Claude PR #825 review (round 2) caught the unreachable else.
     filer_name = next(
         (p.name for p in filing.reporting_persons if p.cik == filing.primary_filer_cik),
-        filing.reporting_persons[0].name if filing.reporting_persons else f"CIK {filing.primary_filer_cik}",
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
Fixed in 8c93a6d.

- BLOCKING: empty `reporting_persons` after re-parse silently destroyed existing rows — `FIXED 8c93a6d`. Added guard before DELETE that raises RewashParseError, transaction rolls back on raise → existing rows preserved.
- PREVENTION: regression test test_blockholders_apply_raises_on_empty_reporting_persons asserts existing rows survive when parser returns empty list (`EXTRACTED tests/test_rewash_filings.py`).

Luke-Bradford:
Rebased on main (97bb921). Both prior commits squashed: feature wiring + BLOCKING fix (empty reporting_persons → RewashParseError before DELETE). Form 3 + DEF 14A wiring from main preserved; blockholder section appended at end of file. `uv run pytest tests/test_rewash_filings.py` 20/20 passing.

Luke-Bradford:
Addressed in 204b2db.

- NITPICK: unreachable `f"CIK {primary_filer_cik}"` else branch — `FIXED 204b2db`. Empty-list guard above raises before this point, so the next() default just falls back to `filing.reporting_persons[0].name`.

PR #830, opened 2026-05-03
feat(#788): per-instrument ownership ingest drillthrough (Chain 2.5)

Luke-Bradford:
## What
Chain 2.5 admin substrate: per-instrument state across all 5 ownership pipelines in one read.

- \`app/services/ownership_drillthrough.py\` — \`get_instrument_drillthrough\` returns \`PipelineState\` per kind (13F / 13D/G / Form 4 / Form 3 / DEF 14A) with typed_row_count, latest_event_at, raw_body_count, tombstone_count, notes.
- \`app/api/operator_ingest.py\` — new \`GET /operator/ownership-drillthrough/{instrument_id}\` inside snapshot_read.
- 9 integration tests.

## Why
When the ownership card is sparse, the operator currently has to join 6 tables to know why. This service folds the answer into one read keyed on instrument_id.

## Test plan
- [x] \`pytest tests/test_ownership_drillthrough.py\` 9/9
- [x] \`ruff\` / \`pyright\` clean
- [x] Codex pre-push review (2 rounds) — header-count bug + null-instrument-partial bug + JOIN fanout all fixed; final pass clean

## Follow-ups
- Frontend admin page consuming this endpoint (Chain 2.6/2.7/2.8 lanes).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #833, opened 2026-05-03
feat(#788 Chain 2.7): DEF 14A drillthrough + CSV export

Luke-Bradford:
## What
Two operator endpoints for DEF 14A beneficial-ownership state:

- \`GET /instruments/{symbol}/def14a_holdings/drill\` — latest-filing holders + pipeline state with stale-holders surface when newer filing exists but didn't produce typed rows.
- \`GET /instruments/{symbol}/def14a_holdings/export.csv\` — historical CSV across all on-file accessions.

Provider gate (\`?provider=sec_def14a\`), 404 on unknown / no-SEC-CIK.

## Test plan
- [x] \`pytest tests/test_def14a_drill.py\` 11/11
- [x] \`ruff\` / \`pyright\` clean
- [x] Codex pre-push review (4 rounds) — stale-holders date-vs-date bug, "no holders" note suppression, accession tie-break all fixed; final pass clean

## Notes
Pipeline-state SQL inlined rather than calling \`ownership_drillthrough.get_instrument_drillthrough\` so this PR is independent of #830 merge order.

### DIFF
--- a/app/api/instruments.py
+++ b/app/api/instruments.py
@@ -21,6 +21,8 @@ DB-only.
 
 from __future__ import annotations
 
+import csv
+import io
 import logging
 from datetime import date, datetime
 from decimal import Decimal, InvalidOperation
@@ -1170,6 +1172,7 @@ _INSIDER_PROVIDERS: tuple[str, ...] = ("sec_form4",)
 # rather than silently passing through to a reader that doesn't
 # consume it. PR #774 review caught the cross-contamination risk.
 _INSIDER_BASELINE_PROVIDERS: tuple[str, ...] = ("sec_form3",)
+_DEF14A_PROVIDERS: tuple[str, ...] = ("sec_def14a",)
 
 
 def _validate_provider(provider: str | None, allowed: tuple[str, ...]) -> None:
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
Addressed in c34f200 (rebased + fixed).

- WARNING: tombstone overcount on retries — `FIXED c34f200` with COUNT(DISTINCT log.accession_number). Note: today `def14a_ingest_log.accession_number` is PK so retries update in place via ON CONFLICT — the bare COUNT(*) is mathematically equivalent. The DISTINCT is defense-in-depth: documents filing-level intent + stays correct if the schema later moves to a retry-attempt model. New regression test `test_drill_tombstone_query_is_filing_distinct` pins filing-level counting.
- NITPICK: `import csv` + `import io` moved to module-level imports — `FIXED c34f200`. Also extracted the same inline imports from the existing insider-baseline CSV path.
- PREVENTION: lesson — when an aggregate query intent is "distinct N filings", spell it as COUNT(DISTINCT accession_number) regardless of current PK shape; stays correct under schema evolution. — `EXTRACTED app/api/instruments.py` (comment in code)

Luke-Bradford:
Addressed in 55634e5.

- BLOCKING: duplicate CIK 0000111110 between TOMB14 and STALE — `FIXED 55634e5`. TOMB14 now seeds cik='0000111120' (unique). The PREVENTION lesson is captured in the inline comment so future reviewers see the gotcha.

PR #834, opened 2026-05-03
feat(#788 Chain 2.8): canonical ownership rollup CSV export

Luke-Bradford:
## Summary
- New endpoint `GET /instruments/{symbol}/ownership-rollup/export.csv` flattens the canonical deduped rollup into a flat CSV.
- One row per surviving holder + treasury memo (when treasury > 0) + residual memo. Header always emitted.
- Backend half of Chain 2.8: the L2 OwnershipPage frontend will rewire to consume this in a follow-up PR (currently 5-fetch client-side assembly).

## Why
The L2 OwnershipPage currently rebuilds the deduped rollup client-side from 5 separate fetches (financials, institutional, insiders, baseline, blockholders) and serialises a `?view=raw` CSV from the assembled state. The server-side `app.services.ownership_rollup.get_ownership_rollup` already produces the canonical, deduped view consumed by `/ownership-rollup` — this PR adds the matching CSV face so the frontend rewire (next PR) collapses to a single fetch.

## Test plan
- [x] `uv run pytest tests/test_ownership_rollup_csv.py` — 8/8 passing
- [x] `uv run pytest tests/test_ownership_rollup.py` — 23/23 (no regression)
- [x] `uv run pyright`
- [x] Codex pre-push review (CLAUDE.md checkpoint 2) — clean after one round (caught missing formula-injection guard on `filer_type` and `edgar_url`; fixed)

### DIFF
--- a/app/api/instruments.py
+++ b/app/api/instruments.py
@@ -3639,3 +3639,57 @@ def get_instrument_ownership_rollup(
             instrument_id=int(inst_row["instrument_id"]),  # type: ignore[arg-type]
         )
     return _rollup_to_response(rollup)
+
+
+@router.get(
+    "/{symbol}/ownership-rollup/export.csv",
+    response_class=PlainTextResponse,
+)
+def get_instrument_ownership_rollup_csv(
+    symbol: str,
+    conn: psycopg.Connection[object] = Depends(get_conn),
+) -> PlainTextResponse:
+    """CSV export of the canonical deduped ownership rollup
+    (Chain 2.8 of #788).
+
+    Same data shape the operator sees in the L2 ownership card,
+    flattened to one row per surviving holder + treasury memo +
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
Addressed in d18e59b.

- WARNING: module-level `pytestmark = pytest.mark.integration` silently gated 7 unit tests behind the integration bucket — `FIXED d18e59b`. Now per-test `@pytest.mark.integration` only on the DB-touching TestClient case. Verified: `pytest -m "not integration"` runs 7/8; integration bucket still picks up the 1 DB test.
- PREVENTION: lesson — when a file mixes pure unit tests with DB-backed integration tests, the marker MUST be per-test, not module-level. The naming convention here (`test_build_csv_*` for pure, `test_csv_endpoint_*` for integration) makes the split visible at glance.

PR #901, opened 2026-05-05
feat(#893): pytest perf redesign — xdist + per-worker DB + template

Luke-Bradford:
## What

Reworks pytest to satisfy issue #893: full suite ≤ 10 min wall-clock,
no concurrent-pytest deadlock, no test writes to operator dev DB.

## Why

Operator pain on `main`: shared `ebull_test` DB, no parallelism, 73-table
TRUNCATE per test, 6 dev-DB writers bypassing the fixture entirely.
Concurrent pytest invocations could deadlock for 10+ min.

## How

- pytest-xdist `-n 4` + per-worker private DB
  `ebull_test_<run_id>_<worker_id>` so concurrent invocations no longer
  share state. Per-invocation `run_id` (epoch-seconds + 6 hex) prevents
  cross-invocation collisions.
- Postgres `CREATE DATABASE … TEMPLATE ebull_test_template` for
  sub-second cold start. Migration-hash invalidation triggers rebuild
  when `sql/*.sql` changes. Controller-only build under
  `EBULL_TEMPLATE_LOCK`. Workers acquire the same lock briefly during
  `CREATE FROM TEMPLATE` so a concurrent rebuild can't mid-copy.
- 6 dev-DB-writer tests migrated off `settings.database_url` onto the
  per-worker test DB (Pattern A direct + Pattern B `monkeypatch`
  for helpers like `publish_manual_job_request`).
- `tests/smoke/test_app_boots.py` keeps its dev-DB lifespan contract
  (CLAUDE.md smoke gate) but is now wrapped in a cluster-wide Postgres
  advisory lock (`EBULL_SMOKE_LIFESPAN_LOCK`) so concurrent invocations
  serialise on lifespan migrations instead of racing them.
- `--basetemp` routed to `${TEMP}/ebull_pytest/<run_id>` outside the
  repo (kills the `tmp_pytest/` Windows lock smell).
- TRUNCATE chunked at 20 tables to stay under
  `max_locks_per_transaction` when 4 workers run simultaneously.
- New helper `tests.fixtures.cleanup_test_dbs` for operator-driven
  recovery after a crashed run.
- Smoke guard at `tests/smoke/test_no_settings_url_in_destructive_paths`
  tightened — allowlist now contains only the documented smoke
  exception + the guard itself.

## Test plan

- [x] `time uv run pytest`: **6m48s, 3786 passed, 1 skipped**.
  Down from baseline >10 min. Within the 5-10 min target.
- [x] `uv run ruff check`, `uv run ruff format --check`,
  `uv run pyright` all green.
- [x] Pre-push gate (lint + format + pyright + pytest + dark-mode
  hygiene) green.
- [x] Codex reviewed across spec v1→v5 (8 + 4 + 1 + 0 BLOCKING)
  plus pre-push review of branch diff (4 findings: ensure_worker_database
  drop+recreate, missing template lock on CREATE FROM TEMPLATE,
  loose assert_test_db prefix match, missing boot-drain cleanup —
  all folded into this commit).
- [x] Concurrent-invocation deadlock test passed: two `uv run pytest`
  shells finish independently, neither hangs.
- [x] Operator dev DB inspected after run — only the documented
  smoke gate writes ever land there (and only its own rows, cleaned
  up by lifespan).

## Out of scope

- Truncate-by-need / opt-in fixture API — high blast radius across
  ~130 files. Follow-up.
- Soft perf-regression gate (durations baseline + 2x slowdown fail) —
  measurement is in (`--durations=20`); enforcement is a follow-up.
- Pre-existing failures #875 + #876 — unrelated to this PR.

Closes #893.

### DIFF
--- a/docs/superpowers/specs/2026-05-05-pytest-perf-redesign.md
+++ b/docs/superpowers/specs/2026-05-05-pytest-perf-redesign.md
@@ -1,0 +1,391 @@
+# Pytest perf redesign — concurrent isolation + xdist + template DB
+
+**Date:** 2026-05-05
+**Issue:** #893 — tech-debt: pytest infra slowness + integration test contention
+**Status:** brainstorm + 2x Codex-reviewed (8 findings folded); awaiting operator sign-off
+
+## Problem
+
+Operator-observed pain on `main` as of 2026-05-05:
+
+- Full pytest suite is the pre-push gate (`.githooks/pre-push:34-35`). Single
+  shared `ebull_test` Postgres database is consumed serially.
+- `tests/fixtures/ebull_test_db.py::ebull_test_conn` truncates **73 tables**
+  (`_PLANNER_TABLES`) before and after every test that opts into the fixture.
+  ~260 fixture invocations across ~130 test files.
+- 121 SQL migrations under `sql/`. Migrations are skipped per-process if
+  already recorded in `schema_migrations` (cheap path), but a fresh DB pays
+  the full 121-file replay cost.
+- Concurrent pytest invocations (developer running multiple shells, or
+  pre-push hook firing while another suite is mid-flight) **block on row
+  and table locks** in the shared DB. Operator has observed 10+ minute
+  hangs from a stuck process holding locks while siblings pile up.
+- No `pytest-xdist` configured. Sequential-only execution.
+- Some tests write directly to `settings.database_url` — i.e. the
+  **operator's dev DB** — bypassing the `ebull_test` fixture entirely.
+  Audit (grep `psycopg.connect(settings.database_url`) finds 6 writer
+  files: `tests/test_jobs_listener.py`,
+  `tests/test_jobs_heartbeat.py`,
+  `tests/test_jobs_queue_recovery.py`,
+  `tests/test_jobs_queue_boot_drain.py`,
+  `tests/test_sync_orchestrator_dispatcher.py`, and one read-only
+  reachability probe in `tests/test_jobs_locks.py` (the test itself
+  is read-only against the DB; the lock writes happen inside
+  `JobLock` which receives a URL parameter). These violate
+  `feedback_test_db_isolation.md` and would deadlock under concurrent
+  pytest invocations.
+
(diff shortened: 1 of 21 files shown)


Luke-Bradford:
Review findings resolved on commit `2178cdc5a009f57d68e24a9fc81201e180250e4a`:

**BLOCKING — `test_db_available()` calling `build_template_if_stale()` from workers**: `FIXED 2178cdc5a009f57d68e24a9fc81201e180250e4a`
- `test_db_available()` no longer calls `build_template_if_stale()`. Workers do only `ensure_worker_database()` + the `SELECT 1` probe.
- Added a runtime guard at the top of `build_template_if_stale()` that raises `RuntimeError` if `PYTEST_XDIST_WORKER` is set, so a future regression can't silently reintroduce the bug.
- The `bench_fundamentals_upsert.py` script (the only other caller) explicitly invokes both helpers in sequence, so it's unaffected.

**WARN — nested advisory-lock release on broken connection**: `FIXED 2178cdc5a009f57d68e24a9fc81201e180250e4a`
- Both unlock paths in `ensure_worker_database` are now wrapped in `try/except: pass` so a secondary failure doesn't mask the primary error.

**WARN — `_dev_db_lifespan_lock` raises on postgres-maintenance-DB unavailability**: `FIXED 2178cdc5a009f57d68e24a9fc81201e180250e4a`
- If the admin `psycopg.connect` itself fails, the smoke test now falls back to running without cross-invocation serialisation. The within-invocation `xdist_group("dev_db_smoke")` pin still serialises workers; only concurrent pytest invocations lose protection on that one path, which is no worse than pre-#893.

**PREVENTION — assert worker-side build is impossible**: `EXTRACTED tests/fixtures/ebull_test_db.py`
- The runtime guard inside `build_template_if_stale()` enforces the contract; review-bot's recommendation is now structurally enforced.

Re-run: `3786 passed, 1 skipped in 5m47s` (faster than the previous 6m48s — workers no longer chain through `build_template_if_stale()` at collection time).

PR #958, opened 2026-05-06
feat(#936): targeted rebuild walks filings.files[] secondary pages

### LINKED ISSUE #936: tech-debt(#935): targeted rebuild must follow filings.files[] secondary pages
Parent: #935 (finding 1).
Debt against: PR #886 (#872 targeted manifest rebuild).

### Problem
`app/jobs/sec_rebuild.py` calls `check_freshness(... last_known_filing_id=None ...)`. `check_freshness` reads only the `recent` submissions array. Targeted rebuild therefore cannot recover accessions that have aged out of `recent` into `filings.files[]` secondary pages — yet the rebuild job claims full-history discovery.

### Fix
- Extend `check_freshness` (or add a sibling) to walk `filings.files[]` when invoked from rebuild.
- `app/jobs/sec_rebuild.py` opts into full-history mode.
- Atom / per-CIK polling stay on `recent` (cheap path).

### Tests
- Fixture filer with `recent` truncated + one accession only present in `filings.files[]/0001.json`.
- Assert rebuild discovers + manifests it; assert non-rebuild paths do not pay the cost.

### Done
- Targeted rebuild discovers an accession present only in a secondary `filings.files[]` page.
- Per-CIK / Atom paths unchanged.

Luke-Bradford:
Closes #936
Refs #935

## Summary
Targeted rebuild claimed full-history discovery but `check_freshness` reads only `recent[]`. Older filings in `filings.files[]` secondary pages could never be repaired by rebuild. Fix: when `delta.has_more_in_files`, walk every secondary page + UPSERT scope-matching rows.

## Changes
- `app/jobs/sec_rebuild.py`:
  - New `_walk_secondary_pages` helper. Mirrors pagination from `sec_first_install_drain._drain_secondary_pages` but filters by rebuild's scope (`sources_set`) so a `sec_def14a`-scoped rebuild doesn't touch every 8-K on the same page.
  - Per-page failure isolation (Codex pre-push catch): `http_get` + `parse_submissions_page` exceptions caught + logged + continue.
- Atom feed (#867) and per-CIK polling (#870) unchanged — they intentionally stay on `recent` (cheap path).

## Test plan
- [x] `test_discovery_walks_filings_files_secondary_pages`: aged-out accession in secondary page recovered + manifested.
- [x] `test_discovery_isolates_secondary_page_failures`: transport error on page 1 doesn't abort rebuild; page 2 still discovers its accession.
- [x] 7 tests pass; lint + format clean; pyright clean.

## Code duplication note
`_walk_secondary_pages` mirrors `sec_first_install_drain._drain_secondary_pages`. Cleanup into a shared helper deferred — the call sites differ on source filtering (rebuild scopes; drain takes everything).

### DIFF
--- a/app/jobs/sec_rebuild.py
+++ b/app/jobs/sec_rebuild.py
@@ -35,7 +35,6 @@ missed an accession.
 
 from __future__ import annotations
 
-import json
 import logging
 from dataclasses import dataclass
 from typing import Any
@@ -273,7 +272,7 @@ def _discovery_pass(
         # otherwise an aged-out accession can never be repaired.
         # Atom / per-CIK polling intentionally stay on ``recent`` (cheap
         # path); only rebuild pays the secondary-page cost.
-        if delta.has_more_in_files:
+        if delta.has_more_in_files and delta.files_pages:
             new_rows += _walk_secondary_pages(
                 conn,
                 http_get=http_get,
@@ -282,6 +281,7 @@ def _discovery_pass(
                 subject_id=sid,
                 instrument_id=instrument_id,
                 sources_set=sources_set,
+                files_pages=delta.files_pages,
             )
 
     return new_rows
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
**FIXED b837de5**

WARNING addressed: FreshnessDelta now carries files_pages: list[str] populated from the same primary fetch in check_freshness. _walk_secondary_pages consumes that list directly — the primary CIK JSON is fetched exactly once per rebuild discovery pass. No more redundant fetch, no more independent failure mode.

NITPICK addressed: test now asserts len(primary_fetches) == 1 to actually prove the no-re-fetch contract, plus a comment explaining why >= 1 is the right floor here (the recent payload's only DEF 14A is already seeded so contributes 0; the secondary page contributes 1).

Luke-Bradford:
**FIXED ec2d2ce**

Three log warnings added so the silent-skip path the bot worried about is now loud:

1. `check_freshness` extract path logs WARNING on JSONDecodeError/TypeError (was `except: pass`).
2. `check_freshness` logs WARNING when `has_more=True` AND `files_pages` ended up empty — catches the structurally-surprising-files-block path.
3. `sec_rebuild._discovery_pass` logs WARNING on its own end when the guard fires (defence in depth: rebuild's own log stream surfaces the gap regardless of upstream noise).

Also normalised `except (E1, E2)` to explicit-parens form. The parens-optional Python 3.14 syntax (PEP 758) is valid but unhelpful for readers.

PR #960, opened 2026-05-06
feat(#956): inline freshness seed in record_manifest_entry

Luke-Bradford:
Closes #956
Refs #935 #937 #959

## Summary
Pre-#956 only the first-install drain (#937) seeded `data_freshness_index`; Atom fast-lane / daily-index reconcile / per-CIK poll / targeted rebuild left new triples scheduler-invisible until the next bulk seed. Fix: `record_manifest_entry` invokes a new `seed_freshness_for_manifest_row` helper inline. Single boundary — every manifest write maintains the index automatically.

## Changes
- `app/services/data_freshness.py`: new `seed_freshness_for_manifest_row` with latest-row preservation (newer-gate on watermark fields) + state-machine rules (only escalate `never_filed` → `current`; don't clobber `error` / `expected_filing_overdue`).
- `app/services/sec_manifest.py`: `record_manifest_entry` calls helper inline. Lazy import to avoid the cycle.
- 7 new tests across `test_data_freshness.py` + `test_sec_atom_fast_lane.py`.

## Codex pre-push findings applied
- **State preservation** — initial draft unconditionally set `state='current'` on every UPSERT, clobbering legitimate poll errors. Fixed with conditional CASE.
- **Latest-row uniformity** — initial draft used `GREATEST(EXCLUDED, current)` for `expected_next_at`, which could keep stale poll-derived values when a newer accession arrived. Fixed with the same newer-gate as `last_known_*`.
- **Drain double-seed perf** — filed as **#959** follow-up. Drain now seeds inline per-row + bulk at end. Idempotent (correctness fine), redundant on the largest write path. Defer to #959 since dropping the bulk call needs careful behaviour analysis.

## Test plan
- [x] 87 tests pass across `test_data_freshness.py` + `test_sec_atom_fast_lane.py` + `test_sec_first_install_drain.py` + `test_sec_per_cik_poll.py` + `test_sec_rebuild.py` + `test_sec_daily_index_reconcile.py` + `test_sec_manifest.py`.
- [x] Atom discovery → freshness index has scheduler-visible rows in `state='current'`.
- [x] Older accession does not clobber newer watermark.
- [x] Re-discovery of same accession preserves poll-error state.
- [x] `never_filed` state correctly escalates to `current` on first manifest evidence.
- [x] `ruff check`, `ruff format --check`, `pyright` clean.

### DIFF
--- a/app/services/data_freshness.py
+++ b/app/services/data_freshness.py
@@ -129,6 +129,115 @@ def predict_next_at(
 # ---------------------------------------------------------------------------
 
 
+def seed_freshness_for_manifest_row(
+    conn: psycopg.Connection[Any],
+    *,
+    subject_type: ManifestSubjectType,
+    subject_id: str,
+    source: ManifestSource,
+    cik: str | None,
+    instrument_id: int | None,
+    accession_number: str,
+    filed_at: datetime,
+) -> None:
+    """Single-row scheduler seed for one manifest write (#956).
+
+    Companion to ``seed_scheduler_from_manifest`` (the bulk full-table
+    rebuild). Called inline from ``record_manifest_entry`` so every
+    manifest discovery write — Atom fast-lane, daily-index reconcile,
+    per-CIK poll, targeted rebuild, first-install drain — leaves the
+    scheduler queryable. Pre-#956 only the drain seeded; the others
+    relied on the next full ``seed_scheduler_from_manifest`` to pick
+    up new triples, leaving them invisible until that ran.
+
+    Latest-row semantics: the freshness ``last_known_*`` columns must
+    reflect the LATEST filing per (subject, source). Discovery writers
+    can call this with an older accession (rebuild walking
+    ``filings.files[]`` secondary pages, daily-index reconcile picking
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
**REBUTTED.**

`record_manifest_entry` declares `cik: str` (not Optional) at `app/services/sec_manifest.py:141`. Line 177 validates `if not cik or not cik.strip(): raise ValueError(...)` BEFORE the seed call at line 244. A `None` would raise `ValueError` before the alleged `AttributeError` could fire.

`seed_freshness_for_manifest_row` declares `cik: str | None` because the field is genuinely nullable in the DB (`institutional_filers` rows have CIK from a different lookup path). The widening matches the column shape. The caller passes a guaranteed-non-None value derived from the validated `record_manifest_entry` arg.

PREVENTION: the bot's grep heuristic ("`cik.strip()` on nullable") would need to walk the call site's input typing + validation block to be sound. Adding "`cik.strip()` is safe iff the surrounding function's `cik: str` (not `str | None`) AND there's a prior non-empty check" would catch real bugs without false-positiving on this pattern.

Codex independently verified the rebuttal. Merging.

Issue #976, opened 2026-05-06
feat(#974/B): LISTEN/NOTIFY ebull_credential_health channel + operator-health cache


Luke-Bradford:
Refs #974. Foundational PR: pub-sub channel + in-memory cache.

## Spec
docs/superpowers/specs/2026-05-06-credential-health-precondition-design.md (section: NOTIFY contract).

## Files
- New: \`app/jobs/credential_health_listener.py\` (mirrors app/jobs/listener.py pattern)
- Edit: \`app/jobs/__main__.py\` — start listener thread alongside existing job listener
- New: \`app/services/credential_health_cache.py\` — in-memory cache; INITIALIZING state; full-scan on startup with retry-with-backoff
- New: \`tests/test_credential_health_listener.py\`

## Acceptance
1. NOTIFY on ebull_credential_health arrives at subscriber within 1s.
2. 5s poll fallback recovers from dropped notify (simulated by skipping listener).
3. Startup full-scan populates cache for operators not in any pending notify.
4. Startup retry-with-backoff (1, 2, 5, 10, 30s cap) until initial scan succeeds; cache reports INITIALIZING during this window.
5. Consumers reading cache during INITIALIZING get \`MISSING\` (fail-safe — no credential-using layers run).
6. Cache returns MISSING for an operator with no rows.
7. INITIALIZING state surfaces on /system/status so admin UI can show transient banner.

## Codex
Pre-push diff review.

## Dependencies
Depends on: #974/A (uses get_operator_credential_health, listens for record_row_health_transition's NOTIFY).

PR #981, opened 2026-05-06
feat(#975): credential health state machine + atomic PUT /replace endpoint

Luke-Bradford:
Refs #974. Closes #975. Foundational PR for the credential-health-as-scheduling-precondition umbrella.

## What

- Schema migration `sql/128_broker_credentials_health_state.sql`: new health_state column + index + operator_credential_health_transitions table.
- New service `app/services/credential_health.py`: operator-level aggregate computation (REJECTED-first precedence), side-tx write-through helpers, REJECTED-stickiness, NOTIFY on aggregate movement only, recovery-timestamp upsert on any move out of REJECTED.
- New endpoint `PUT /broker-credentials/replace`: atomic revoke + insert in one transaction (no transient MISSING window for subscribers). Identical-secret short-circuit (`{"changed": false}`) avoids spurious flap on idempotent re-saves.
- `POST /broker-credentials/validate-stored` write-through: probes flip both rows' health to VALID (canonical recovery path); 401/403 flips to REJECTED.

## Why

Operator with bad keys today: the scheduler keeps firing every credential-using job on schedule, the WS subscriber spams 5s reconnect loops with stale in-memory keys, and the admin Problems panel cascades N copies of the same root 401. After saving correct keys nothing auto-resumes — operator has to click Sync now manually. This PR is the foundation: row-level health + operator-level aggregate + LISTEN/NOTIFY signal that subsequent children (#976-#980) consume to gate scheduling, reload the WS, and clean the admin UI.

## Test plan

- [x] 42 new tests at `tests/test_credential_health.py` covering: truth-table for stickiness, all aggregate states + missing-label cases + precedence, side-tx semantics, NOTIFY emission and payload shape, recovery timestamp via both _do_health_transition AND notify_aggregate_if_changed, environment scoping, NULL/missing transitions row.
- [x] Adjacent suites pass (broker_credentials_*, smoke).
- [x] ruff check + ruff format --check + pyright clean.
- [x] Codex pre-push review: 3 rounds; 5 findings raised, all fixed before push (validate-stored pool attribute, environment scoping, PUT /replace NOTIFY, conn.commit() flush, recovery timestamp on REJECTED → UNTESTED).

## Spec

- docs/superpowers/specs/2026-05-06-credential-health-precondition-design.md
- 5 Codex spec-review rounds: 26 findings raised, all addressed before any code landed.

### DIFF
--- a/app/api/broker_credentials.py
+++ b/app/api/broker_credentials.py
@@ -36,6 +36,7 @@ from uuid import UUID, uuid4
 
 import httpx
 import psycopg
+import psycopg.rows
 from fastapi import APIRouter, Depends, HTTPException, Request, status
 from pydantic import BaseModel, Field
 
@@ -60,6 +61,7 @@ from app.services.broker_credentials import (
     revoke_credential,
     store_credential,
 )
+from app.services.credential_health import record_health_outcome
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


Luke-Bradford:
## Round 1 review resolved

**WARNING #1** (last_health_error wipe on sticky skip) — `FIXED 6f4f72c`. _do_health_transition now splits the column updates: last_health_check_at bumps on every call (reflects "we checked"), but last_health_error is only rewritten when an actual transition happens. Sticky-skip preserves the operator-visible error string.

**WARNING #2** (transitions FK missing) — `FIXED 6f4f72c`. `operator_credential_health_transitions.operator_id` now has `REFERENCES operators(operator_id) ON DELETE CASCADE`. Dev DB FK applied via manual ALTER (test DB rebuilds from the migration hash).

**PREVENTION** (test for the wipe regression) — `EXTRACTED tests/test_credential_health.py`. New test `TestDoHealthTransition.test_sticky_skip_preserves_last_health_error` seeds a non-NULL error string, drives an incidental success against a REJECTED row, asserts the string is preserved. Pins the contract.

44 tests passing locally. ruff + format + pyright clean.

PR #1052, opened 2026-05-08
fix(#1045): wire PipelinedSecFetcher into D1/D2/D3 bootstrap entrypoints

Luke-Bradford:
## What

D-stage bootstrap entrypoints (`sec_business_summary_bootstrap`, `sec_def14a_bootstrap`, `sec_8k_events_ingest`) now pre-fetch cohort primary documents via `PipelinedSecFetcher` (4-way concurrent at the shared 7 req/s ceiling). Per-filing TCP+SSL handshake overlaps with adjacent fetches — latency hides on large backfills without breaching the rate floor.

New helpers in `app/services/sec_pipelined_fetcher.py`:
- `prefetch_document_texts(urls, *, user_agent, target_rps=7, concurrency=4)` returns `dict[url, str | None]`.
- `_CachedDocFetcher` wraps a sync provider with the prefetch cache. Cache-hit-with-string returns the body. Cache-hit-with-None = permanent 404/410. Cache-miss falls back to the underlying sync provider.

Each `ingest_*` function gained optional `prefetch_urls` + `prefetch_user_agent` kwargs (default False); `bootstrap_*` wrappers thread them through; the scheduler entrypoints opt-in with `prefetch_urls=True`.

## Why

Per-filing serial fetches were the dominant per-CIK cost in the D-phase. Pipelining gives ~1.5-2x latency hiding without changing the rate-clock contract.

## Test plan

- [x] `TestPrefetchDocumentTexts` — 200 / 404 / 503 cache behavior via `httpx.MockTransport`.
- [x] `TestCachedDocFetcher` — hit-string / hit-None / miss semantics pinned.
- [x] All 69 impacted tests pass (business_summary, def14a, eight_k, smoke).
- [x] Codex pre-push: APPROVE after addressing MED failure-mode parity + LOW test coverage + LOW docstring.

Closes #1045.

### DIFF
--- a/app/services/business_summary.py
+++ b/app/services/business_summary.py
@@ -1395,6 +1395,8 @@ def bootstrap_business_summaries(
     *,
     chunk_limit: int = 500,
     max_runtime_seconds: int = 3600,
+    prefetch_urls: bool = False,
+    prefetch_user_agent: str | None = None,
 ) -> IngestResult:
     """One-shot drain of the entire business_summary candidate set.
 
@@ -1423,7 +1425,13 @@ def bootstrap_business_summaries(
     total_parse_misses = 0
 
     while time.monotonic() < deadline:
-        chunk = ingest_business_summaries(conn, fetcher, limit=chunk_limit)
+        chunk = ingest_business_summaries(
+            conn,
+            fetcher,
+            limit=chunk_limit,
+            prefetch_urls=prefetch_urls,
+            prefetch_user_agent=prefetch_user_agent,
+        )
         total_scanned += chunk.filings_scanned
         total_inserted += chunk.rows_inserted
         total_updated += chunk.rows_updated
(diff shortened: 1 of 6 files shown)


Luke-Bradford:
## Review resolution (commit 2b9c51c)

**BLOCKING** — `c[2]` urls not filtered for None
**REBUTTED** — both candidate selectors already filter `AND fe.primary_document_url IS NOT NULL` in their `WHERE` clause:
- `app/services/business_summary.py:1530`
- `app/services/eight_k_events.py:720`

`c[2]` (the third tuple element = `primary_document_url`) cannot be None when a row enters `candidates`. Codex confirms the rebuttal sound.

**PREVENTION** — None URL grep
**EXTRACTED** — agreed in principle but the SELECT-side filter is the structural guarantee. Adding a redundant Python-side `if c[2]` would be defense-in-depth; deferring as low-value follow-up if the pattern recurs.

All findings reach terminal state. Ready to merge.

Issue #1064, opened 2026-05-08
epic: admin control hub rewrite — unified processes table + cooperative cancel + watermark resume


Luke-Bradford:
## Why

Operator feedback driving a full rewrite of the admin page. Spec at [docs/superpowers/specs/2026-05-08-admin-control-hub-rewrite.md](https://github.com/Luke-Bradford/eBull/blob/main/docs/superpowers/specs/2026-05-08-admin-control-hub-rewrite.md) (post-Codex round 6).

Verbatim operator quotes (the design constraints):

1. *"There is still no real progress indicators yet there is a progress field. If there is nothing that can go in there, why are we showing it."*
2. *"The error section is a waste of space."*
3. *"Make the gui for this page nicer, simpler, not just for bootstrap but for all services its jumbled up mess."*
4. *"There's no cancel button on the page either."*
5. *"Restarting jobs but the jobs are still running."*
6. *"Per-archive sublist breaks the page and makes it look janky."*
7. *"There must be a programmatic rule for those gaps."*
8. *"Iterate of a fresh check should be able to pick up from where it left off… watermarking would be important for each step."*
9. *"Errors should be clearly visible and not hidden but if a run on a failed process is running we should hide the visible page errors because each run should be regetting anything that was a prior issue."*
10. *"Optimised, slick, not heavy, careful about how we spend processing and download resources, self healing."*

## What

Single unified `/admin` control hub: ProblemsPanel + FundDataRow + **Processes table** (replacing BootstrapPanel + LayerHealthList + SyncDashboard + SeedProgressPanel + Background-tasks). Bootstrap is one row with custom drill-in. Cancel is cooperative-with-watermark-resume. Errors auto-hide while a retry is in flight. Failures float to top.

## PR sequence

- [ ] PR1 — schema (sql/135-139) + cancel signal helpers + boot recovery + smoke
- [ ] PR2 — bootstrap cooperative cancel + cancelled state + observation points
- [ ] PR3 — backend `/system/processes` envelope + bootstrap + scheduled adapters + endpoints + `_tracked_job` extension
- [ ] PR4 — watermark resolver + iterate/full-wash semantics
- [ ] PR5 — frontend ProcessesTable + drill-in route + auto-hide-errors
- [ ] PR6 — ingest_sweep adapter + orchestrator drill-in + sync cancel + finalizer guard
- [ ] PR7 — bootstrap timeline drawer + decommission BootstrapPanel
[...]

PR #1084, opened 2026-05-09
feat(#1083): PR8 — four-case stale model + visible-motion polish

Luke-Bradford:
## Summary
- Replaces the PR3 placeholder `is_stale` + `expected_p95_seconds` fields with the operator-amendment §A1 four-case stale model: `schedule_missed` / `watermark_gap` / `queue_stuck` / `mid_flight_stuck`. `ProcessRow.stale_reasons: tuple[StaleReason, ...]` carries them; multiple can fire on one row.
- FE chips per reason next to status pill, banner above ProcessesTable summarising stale rows, amber pulsing-left-border on stale (overrides sky on running). All with `motion-reduce:animate-none`.
- Sort: rows with non-empty `stale_reasons` float to just below `failed`.

## Why
PR3 left `is_stale=False` stubs because the spec's original "rolling p95" rule was reframed in operator-amendment round 1 to a four-case model anchored on existing telemetry (no heartbeat infra needed). PR8 ships that model. Read-only — no producer-side writes, no schema migration, no backfill.

## Test plan
- [x] `tests/test_stale_detection.py` — 22 pure-logic tests pin every reason path + ordering + boundary semantics
- [x] `tests/test_stale_thresholds.py` — registry default + per-process override + grep-validated keys vs SCHEDULED_JOBS
- [x] `tests/test_processes_envelope.py` — `is_stale` / `expected_p95_seconds` removed from ActiveRunSummary; `stale_reasons` defaults to empty tuple; `last_progress_at` added
- [x] `tests/test_bootstrap_adapter.py` — queue_stuck via `pending_job_requests` process_id='bootstrap'; mid_flight_stuck via MAX(stage `last_progress_at`); never schedule_misses or watermark_gaps
- [x] `tests/test_scheduled_adapter.py` — schedule_missed integration (cadence-anchored to last terminal), watermark_gap, queue_stuck (incl. NULL claimed_at fallback + recently-claimed boundary), mid_flight_stuck (incl. first-tick lag negative)
- [x] `tests/test_ingest_sweep_adapter.py` — watermark_gap when freshness overdue; sweeps never schedule_miss or queue_stuck; running underlying job suppresses gap
- [x] `tests/test_processes_endpoints.py` — `/system/processes` payload carries `stale_reasons`; deleted fields absent
- [x] `frontend/src/components/admin/ProcessRow.test.tsx` — chip per reason, multiple chips, mid_flight_stuck elapsed-since-heartbeat suffix, amber-overrides-sky on stale-while-running, motion-reduce
- [x] `frontend/src/components/admin/ProcessesTable.test.tsx` — banner conditional render, stale rows sort above ok rows, failed still outranks stale
- [x] `frontend/src/components/admin/StaleBanner.test.tsx` — empty-state, single-cause vs multiple-causes copy, +N more truncation, View deep-link
- [x] `uv run ruff check .` + `ruff format --check .` + `pyright` + scoped pytest (118 tests, 61s)
- [x] `pnpm typecheck` + scoped vitest (38 tests, 1.2s)
- [x] `node frontend/scripts/check-dark-classes.mjs` — 180 files, no violations
- [x] Smoke test `tests/smoke/test_app_boots.py` green
- [x] Codex pre-impl plan review (3 BLOCKING + 8 WARNING addressed pre-implementation)
- [x] Codex pre-push review (1 BLOCKING + 4 WARNING) — all fixed in follow-up commit; key fix: `expected_fire_at` derived from last terminal run replaces unreachable `next_fire_at`

Refs #1064. Closes #1083.


Luke-Bradford:
## Review round 1 resolution

### WARNING — f-string SQL in test helper
**FIXED bac6723** — `_insert_freshness_with_expected_next_at` rewritten to use `make_interval(mins => %s)` so the SQL stays a literal string, the offset still flows through psycopg parameters. Pushed pre-review (the bot's first comment was on the original commit; round 2's verdict on bac6723 was APPROVE).

### PREVENTION — f-string SQL interpolation in tests
**EXTRACTED docs/review-prevention-log.md** — added entry "f-string SQL interpolation in test helpers" (commit 45450ba). Self-review prompt to grep `tests/` for f-string fragments inside SQL.

Issue #1509, opened 2026-06-06
Jobs: job-level retry/backoff for transient failures (next_retry_at on job_runs)


Luke-Bradford:
## What
Per-job retry/backoff. On failure, classify **transient** (timeout/network/429/lock-contention) vs **permanent** (logic/validation). Transient → set `next_retry_at` on short capped exponential backoff; permanent → no retry (→ Needs-attention). Bounded attempts.

## Why
Today a failed job has NO retry — it waits a full cadence (daily ≤24h, weekly ≤7d) for its next natural fire. Foundation for the "Self-healing / will retry HH:MM" verdict.

## Where
`app/workers/scheduler.py::_tracked_job`, `app/services/ops_monitor.py::record_job_finish`; schema `next_retry_at`/`attempt` on `job_runs` (or sidecar); runtime re-fire path.

## Acceptance
- Transient failure auto-retries on backoff without a full-cadence wait; capped attempts then → permanent.
- Permanent failure does not retry. Terminal-state correct (prevention 1217). Classifier table-tested.

Part of #1508. Caveat: never retry into a held rate-limit/gate (coordinate #1484 + T4).

PR #1522, opened 2026-06-07
fix(#1516): widen finra short-interest ratio cols to NUMERIC(20,4)

Luke-Bradford:
## What
`days_to_cover` + `change_percent` were `NUMERIC(10,4)` (max 999,999.9999) in both `finra_short_interest_observations` and `finra_short_interest_current` (`sql/152_finra_short_interest.sql:34-35,115-116`). FINRA legitimately publishes values ≥ 10^6 for edge instruments (near-zero ADV → huge days-to-cover; near-zero prior short interest → huge % change), so `finra_short_interest_refresh` aborted the whole settlement file with `NumericValueOutOfRange` — failing every day on dev and turning `/system/status` red (the one red Processes row, surfaced while diagnosing #1508).

`sql/184` widens both ratio columns to `NUMERIC(20,4)` on both tables. No parser change — the values are valid, just wider than the column. Shares columns are already `NUMERIC(20,0)` (correct); only the two ratio cols were too narrow.

## Why this shape
- **Idempotent**: DO-block shape guard (`numeric_precision = 10` check per table) makes a manual re-run a no-op; the runner records the file once anyway.
- **Partition propagation**: `ALTER COLUMN ... TYPE` on the partitioned `observations` parent propagates to every partition (PG14+); verified all partitions now `p20`.
- **Lock/rewrite**: verified empirically on PG17 — widening `NUMERIC(10,4)→(20,4)` is metadata-only, no table rewrite (`pg_relation_filenode` unchanged across the ALTER). Brief `ACCESS EXCLUSIVE` lock for the catalog update only.
- **No inline COMMIT**: the migration runner wraps body + `schema_migrations` insert in one tx (prevention-log: tx-bound migrations).

## Test
- New `tests/fixtures/finra/shrt20260430_overflow.csv` + `test_ingest_persists_ratio_values_over_one_million`: a row with `days_to_cover=1234567.89` and `change_percent=9876543.21` (both ≥ 10^6) must upsert into observations + current rather than abort the file. DB-tier (`-m db`).

## Dev verification (DoD), at commit ef305f5
- **Migration applied on dev** (`scripts/migrate.py`) — all 4 cols + every partition now `NUMERIC(20,4)`.
- **Backfill executed**: re-ran `run_finra_short_interest_refresh` on dev. The 3 previously-failing settlement dates now upsert, `failed_files=0`:
  - 2025-09-15 → 5490 rows
  - 2025-11-28 → 5576 rows
  - 2026-04-15 → 5620 rows
- **Operator-visible figure**: rows whose ratio cols exceed the old 10^6 ceiling now persist, e.g. `change_percent=171206000.0000` (i=15534, 2026-04-15) and `57598300.0000` (i=1051020, 2025-09-15) — the exact values that aborted the file pre-184.
- **Cross-source order-of-magnitude sane** vs public FINRA/marketbeat: GME 57.9M shares short / dtc 4.16, AAPL 138.8M / 2.74, MSFT 77.3M / 2.34 (settle 2026-05-15).

## Gates
ruff check ✓ · ruff format ✓ · pyright ✓ · `pytest -m "not db"` 3437 passed ✓ · smoke 129 passed ✓ · finra db tier 18 passed ✓ · Codex ckpt-2 no findings.

## Note
#1509's permanent-failure classification (DB_CONSTRAINT → no retry) already stopped the retry-storm; this removes the underlying daily failure so the row goes green.

Closes #1516

### DIFF
--- a/sql/184_finra_short_interest_widen_ratio_cols.sql
+++ b/sql/184_finra_short_interest_widen_ratio_cols.sql
@@ -53,26 +53,32 @@
 -- an inline COMMIT would split them (prevention-log: tx-bound migrations).
 
 DO $$
+DECLARE
+    rec RECORD;
 BEGIN
-    IF EXISTS (
-        SELECT 1 FROM information_schema.columns
-        WHERE table_name = 'finra_short_interest_observations'
-          AND column_name = 'days_to_cover'
-          AND numeric_precision = 10
-    ) THEN
-        ALTER TABLE finra_short_interest_observations
-            ALTER COLUMN days_to_cover  TYPE NUMERIC(20, 4),
-            ALTER COLUMN change_percent TYPE NUMERIC(20, 4);
-    END IF;
-
-    IF EXISTS (
-        SELECT 1 FROM information_schema.columns
-        WHERE table_name = 'finra_short_interest_current'
-          AND column_name = 'days_to_cover'
-          AND numeric_precision = 10
-    ) THEN
-        ALTER TABLE finra_short_interest_current
-            ALTER COLUMN days_to_cover  TYPE NUMERIC(20, 4),
-            ALTER COLUMN change_percent TYPE NUMERIC(20, 4);
-    END IF;
+    -- Guard EACH (table, column) independently so a partial prior state
+    -- (one column already widened, its sibling not) still widens the
+    -- narrow one. The shape check fires per column, never coupling
+    -- change_percent's widening to days_to_cover's precision.
+    FOR rec IN
+        SELECT v.tbl, v.col
+        FROM (VALUES
+            ('finra_short_interest_observations', 'days_to_cover'),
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
**WARNING** — FIXED f5418e5. Restructured the `sql/184` DO block to loop over all 4 `(table, column)` pairs and guard each column's `numeric_precision` independently before its own `ALTER`. `change_percent`'s widening is no longer coupled to `days_to_cover`'s precision, so a divergent partial state widens the narrow column correctly. Verified on dev: parses + no-ops when already widened.

Luke-Bradford:
**NITPICK** — FIXED 52a8c90. Moved the new wide-ratio-columns test to the end of the file as `# 10` so section numbering stays monotonic (was a duplicate `# 5`).

PR #1610, opened 2026-06-13
feat(#1593): trade_events ledger — eToro trade-history ingest, closed-position archive, Activity groundwork (PR-1)

### LINKED ISSUE #393: Investigate eToro portfolio-equity-over-time API (live-only)
## Context

PR #392 hides the dashboard Portfolio-value chart when the reconstructed series is flat, because demo eToro doesn't backfill per-fill history and our local ledger can only reconstruct NAV for days we have fills on. On a fresh demo account the chart reduces to cash-only and is just noise.

User decision (2026-04-21): the fills-replay approach is fine; we're not going to fabricate synthetic fills or build a `portfolio_nav_daily` snapshot table. Instead, if eToro exposes portfolio equity over time on live accounts, we should use that as the data source for live accounts and accept that demo shows nothing.

## Scope for this ticket

Pure research. Answer these:

1. Does the eToro public or internal API expose a per-day portfolio-equity (or account-equity / NAV) endpoint? Capture the exact endpoint, auth scope, response shape.
2. Is it live-only, or does demo also have it? (Expectation: live-only, hence the need for this ticket.)
3. What lookback does eToro give? (1M? 1Y? unbounded?)
4. Can we pull it through the same path as our existing portfolio/fx sync (`fx_rates_refresh`, `portfolio_sync`), or does it need a new sync layer?

## Deliverables

- Findings comment on this ticket: endpoint + auth + shape + gotchas.
- If viable, file a follow-up implementation ticket. Scope would be: new `portfolio_equity_daily` table + sync layer + swap `/portfolio/value-history` data source for live accounts (keep the fills-replay path as fallback for demo).
- If not viable, close with "no-op, fills-replay stays the only source".

## Why this isn't urgent

- Chart now hides cleanly on demo (PR #392).
- Live accounts with even minor daily movement already get a useful chart via the existing fills-replay path.
- Only relevant when live trading begins and the operator wants pre-eBull history to show on the chart.

---

Related: #204 (value-history endpoint), #391 (advanced-chart spike).
Labels: \`research\`, \`backend\`, \`etoro\`.

Luke-Bradford:
Part of #1593 (PR-1, backend slice — the epic stays open; PR-2 adds the activity API + FE tab). Closes #393 (capability findings posted on the ticket; balances endpoint 403 + 12-month cap → fills/ledger path adopted).

Spec: `docs/proposals/etl/2026-06-13-etoro-trade-ledger.md` (committed in this PR; operator-signed 2026-06-13; Codex ckpt-1 ran 3 rounds — 3 BLOCKING/3 HIGH/2 MED found+folded, final verdict SOUND).

## What changed

- **sql/194**: `trade_events` — append-only broker-observed ledger, one row per position open/close event. Partial unique keys: `(position_id) WHERE open`, `(position_id, executed_at) WHERE close`. `CHECK (position_id >= 0)`, price `> 0` or NULL, `realized_pnl_usd` close-only. Two instrument columns: `etoro_instrument_id NOT NULL` (raw) + nullable FK `instrument_id` (deep-history delistings must land, not drop). Plus `broker_positions_closed` archive (LIKE broker_positions + `closed_detected_at`, PK (position_id, closed_detected_at)).
- **Provider** (`etoro_broker.py`): `get_trade_history(min_date, page_size)` — paged GET `/api/v1/trading/info/trade{/demo}/history` (note the env segment sits between `trade` and `history`, unlike the other info endpoints), `_parse_closed_trade` pure normaliser, `TradeHistoryParseError`. `BrokerClosedTrade` dataclass in `app/providers/broker.py`.
- **Service** (`app/services/trade_events.py`): transform (units contract below), `merge_events` (portfolio-sourced open beats history-derived), `ingest_trade_events` (ON CONFLICT DO NOTHING + closed-set counters + `conflict_anomaly` invariant check), sink-derived watermark (`max(close executed_at) − 7d`, deep epoch 2017-01-01 on empty ledger), `fetch_trade_history_safely` (spec §7 error posture), `enqueue_post_trade_sync`.
- **Chokepoint** (`portfolio_sync.py`): `sync_portfolio(conn, portfolio, trade_history=None)` is the ONLY ledger writer. Takes `pg_advisory_xact_lock` at entry (serializes its two callers). Archive-INSERT before the disappeared-DELETE sweep, real ids only.
- **Callers**: `scheduler.py::daily_portfolio_sync` + WS `_default_reconcile_runner` both read the watermark on a short-lived conn, fetch portfolio + all history pages in one provider session, and pass rows in. History-fetch failure → `None` → positions still sync, watermark unmoved.
- **Order paths** (`orders.py`, `order_client.py`): a persisted fill enqueues an immediate manual `daily_portfolio_sync` (same transaction — NOTIFY at commit).
- Fixture from the live demo probe, pure-logic test suite, ONE db-marked dedup/anomaly/null-price integration test, re-resolve runbook (`app/runbooks/trade_events_reresolve_instruments.py`, dry-run default).

## Why

Closed positions vanished at sync (DELETE sweep) and eToro-side history was never ingested — no auditable trade record, and #1594 (value-over-time v2) needs an exact units timeline. Probe (2026-06-13) proved the demo account exposes full closed-trade history and that `minDate` filters on closeTimestamp, so a sink-derived watermark is loss-proof.

**Deviation from the issue sketch (spec §21):** the order path does NOT write the ledger directly. Its `broker_positions` rows use synthetic negative position ids (#227, both writers) that the next sync replaces with real ids — direct writes would orphan open events. Intent-side audit already exists (`orders`/`fills`/`decision_audit`); the ledger records broker-observed reality, and the post-fill enqueue closes the latency gap to seconds.

**Units contract (spec §1.7):** open event units = ORIGINAL position size (`initialUnits`, or Σ slices for positions only ever seen in history); close event units = slice deltas. eToro partial closes reduce the same positionId (sql/024:19 `isPartiallyAltered`), so Σ(closes) ≤ open is a monitored invariant (`conflict_anomaly`), never a silent merge.

## Security model

No new auth surface. The ledger writer runs in the jobs process / WS reconcile with existing credentials handling (`_load_etoro_credentials`, audited decrypt). The order-path enqueue runs behind the existing operator session / service-token gate on `/portfolio/orders` and the guard-approved `execute_order` path. Raw API payloads land in JSONB as received (no secrets in trade-history bodies). The runbook is read-only by default (`--apply` to write, FK-existence join only).

## Schema / migration impact

sql/194 (new tables only, no rewrites). Applied to dev via the smoke-test boot. Test-fixture TRUNCATE list extended (`trade_events`, `broker_positions_closed`).

## Invariants checked

- Append-only, immutable: no UPDATE path, DO NOTHING conflicts, first observation wins; disagreement = loud WARNING + `conflict_anomaly`.
- ON CONFLICT predicates attached to both partial unique indexes (settled-decisions: partial-unique inference needs the predicate).
- No HTTP inside transactions: history rows fetched in the caller's provider session, passed as data.
- External timestamps only from API payloads (`openDateTime`/`openTimestamp`/`closeTimestamp`), never `now()`.
- `Decimal(str(x))` for every third-party float; NUMERIC(20,8) units/rates, NUMERIC(20,4) USD.
- Non-positive rates are not marks → NULL price + landed-only `null_price` counter.
- Synthetic negative ids (#227) excluded from ledger AND archive by filter + CHECK.
- cash_ledger untouched (broker_sync reconcile stays the cash truth — no double-count).
- Concurrent reconciles serialized (`pg_advisory_xact_lock`, xact-scoped).
- No ON DELETE CASCADE on the audit-grade tables.

## Failure paths considered

- History fetch 429/5xx/network → WARNING, sync proceeds, unmoved watermark re-covers next tick. 403/400/non-JSON 200 body → loud ERROR, same degrade (`fetch_trade_history_safely` + `TradeHistoryParseError` wrap).
- Empty ledger → deep backfill (bootstrap = steady-state with wide window, 1-3 HTTP calls).
- Unresolvable instrument (delisted) → row lands with NULL FK, counted, runbook re-resolves later.
- Missing `openDateTime` / non-positive units → degraded-skip with reason, counted, never fabricated.
- Order-path enqueue lost → next 5-min scheduled tick covers.

## Codex second opinion (ckpt-2, pre-push)

1 HIGH + 3 MED, all FIXED in 4fc1a97: reconcile serialization (advisory xact lock); `response.json()` ValueError wrapped as parse error; `null_price` counted at ingest on landed rows only (×2 findings).

## Verification (DoD clauses 8–12, at 4fc1a97)

- **Smoke (§18 panel = dev demo account, account-scoped source):** ran the real `daily_portfolio_sync` twice against dev DB. First run: `inserted=9` — 7 open events from live positions (VOO, QQQ, IEP×2, GME×2, BBBY; units = initialUnits; source `etoro_sync`) + ILMN open+close from history. Second run: `inserted=0 duplicate=9` (idempotency). `position_id < 0` count = 0; archive rows = 0 (nothing disappeared).
- **Cross-source (§19):** ILMN closed trade — ledger `realized_pnl_usd=1910.4700`, units `82.135523`, open 2025-08-12 @ 97.30 → close 2025-11-14 @ 120.56 — exactly matches the eToro API trade-history row captured in the probe (`tests/fixtures/etoro/trade_history_demo.json`); operator can eyeball the same figures on the eToro web UI history view.
- **Backfill (§10):** executed — the first dev sync above IS the deep backfill (empty-ledger watermark → 2017 epoch fetch).
- **Operator-visible figure (§11):** SQL-level verification this PR (figures above); the browser surface is PR-2's `GET /portfolio/activity` + Activity tab — recorded as the epic's next slice, not silently skipped.
- Local gates at HEAD: ruff ✓ format ✓ pyright 0 ✓ `-m "not db"` 3911 ✓ smoke 129 ✓ scoped db tier (trade_events_db, portfolio_sync ×2, order suites ×4) ✓.

## Conscious tradeoffs

- Always-fetch history every sync tick (1 GET/5min) vs disappearance-triggered: stateless simplicity, 60/min budget barely dented (spec §21).
- `POST_TRADE_SYNC_JOB` name duplicated in `trade_events.py` (scheduler import would be circular) — pinned by `test_post_trade_sync_job_name_matches_scheduler_constant`.
- History-derived opens are Σ-of-slices best-effort; correct for backfill (window spans account lifetime) and superseded by portfolio-sourced opens for watched positions. §22.1 partial-close row shape stays an open question with a loud anomaly counter watching it.

### DIFF
--- a/app/api/orders.py
+++ b/app/api/orders.py
@@ -35,6 +35,7 @@ from app.providers.broker import BrokerOrderResult
 from app.services.ops_monitor import get_kill_switch_status
 from app.services.quote_marks import positive_decimal_or_none
 from app.services.runtime_config import get_runtime_config
+from app.services.trade_events import enqueue_post_trade_sync
 
 logger = logging.getLogger(__name__)
 
@@ -477,6 +478,12 @@ def _persist_order_and_fill(
                 "ev": Jsonb({"order_id": order_id, "raw_payload": broker_result.raw_payload}),
             },
         )
+
+        # 7. Filled trade → queue an immediate portfolio sync so the
+        # broker-observed event lands in trade_events within seconds
+        # (#1593). Same transaction: NOTIFY fires only at commit.
+        if passed:
+            enqueue_post_trade_sync(conn, requested_by="orders_api")
 
     return order_id
 
--- a/app/providers/broker.py
+++ b/app/providers/broker.py
@@ -142,6 +142,38 @@ class BrokerPortfolio:
     mirrors: Sequence[BrokerMirror] = ()
 
 
+@dataclass(frozen=True)
+class BrokerClosedTrade:
+    """One closed-position slice from the broker's trade history.
+
+    eToro returns one row per closed slice carrying BOTH legs
+    (open + close). Partial closes reduce the same positionId, so a
+    position may appear in several rows (#1593 spec §4).
+
+    Money fields (net_profit, fees, investment, initial_investment)
+    are in the account currency (USD). Rates are in the instrument's
+    native currency.
+    """
+
+    position_id: int
+    instrument_id: int
+    is_buy: bool
+    units: Decimal
+    open_rate: Decimal | None
+    open_timestamp: datetime
+    close_rate: Decimal | None
+    close_timestamp: datetime
+    net_profit: Decimal | None
+    fees: Decimal | None
+    investment: Decimal | None
+    initial_investment: Decimal | None
+    leverage: int
+    order_id: int | None
+    social_trade_id: int | None
+    parent_position_id: int | None
+    raw_payload: dict[str, Any]
+
+
 class BrokerProvider(ABC):
     """
     Interface for broker operations.
@@ -194,3 +226,13 @@ class BrokerProvider(ABC):
 
         Returns all open positions and available cash.
         """
+
+    def get_trade_history(self, min_date: datetime, page_size: int = 200) -> Sequence[BrokerClosedTrade]:
+        """
+        Fetch closed-trade history rows with close timestamp >= min_date.
+
+        Non-abstract with a NotImplementedError default so existing test
+        fakes that implement only the abstract surface keep working;
+        the eToro implementation overrides it.
+        """
+        raise NotImplementedError
--- a/app/providers/implementations/etoro_broker.py
+++ b/app/providers/implementations/etoro_broker.py
@@ -467,7 +467,10 @@ class EtoroBrokerProvider(BrokerProvider):
                 headers=self._request_headers(),
             )
             response.raise_for_status()
-            rows = response.json()
+            try:
+                rows = response.json()
+            except ValueError as exc:  # 200 with a non-JSON body
+                raise TradeHistoryParseError(f"trade history page {page}: response body is not JSON: {exc}") from exc
             if not isinstance(rows, list):
                 raise TradeHistoryParseError(
                     f"trade history page {page}: expected a JSON array, got {type(rows).__name__}"
--- a/app/providers/implementations/etoro_broker.py
+++ b/app/providers/implementations/etoro_broker.py
@@ -14,7 +14,7 @@ from __future__ import annotations
 import decimal
 import logging
 from collections.abc import Sequence
-from datetime import datetime
+from datetime import UTC, datetime
 from decimal import Decimal
 from types import TracebackType
 from typing import Any
@@ -23,6 +23,7 @@ import httpx
 
 from app.config import settings
 from app.providers.broker import (
+    BrokerClosedTrade,
     BrokerMirror,
     BrokerMirrorPosition,
     BrokerOrderResult,
@@ -72,6 +73,15 @@ class PortfolioParseError(Exception):
 
     See spec §2.2.1 for the hierarchy rationale and §2.3.3 for the
     strict-raise sync contract that depends on it.
+    """
+
+
+class TradeHistoryParseError(Exception):
+    """Raised when a trade-history row cannot be parsed safely.
+
+    Same design as PortfolioParseError: directly subclasses Exception
+    so it is never swallowed by the incidental-exception handlers
+    inside the parse loop.
     """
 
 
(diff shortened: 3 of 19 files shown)


Luke-Bradford:
Round-3 resolution:

- **[WARNING] NotImplementedError escapes `fetch_trade_history_safely`** — FIXED d556ee03: `except NotImplementedError` branch (loud ERROR naming the provider class, returns None → positions still sync). Added to the parametrized error-posture test table.
- **[PREVENTION]** — EXTRACTED `docs/review-prevention-log.md` ("Non-abstract ABC default methods that raise must be caught by every safety wrapper around the call"), same commit d556ee03.

PR #1920, opened 2026-07-04
fix(#1906): header native-primary + display-currency companion; Activity P&L → display currency

Luke-Bradford:
## Summary

Closes #1906.

Two currency-consistency bugs from the 2026-07-04 audit.

### 1. Instrument header flipped GBP/USD between tabs
`GET /instruments/{symbol}/summary` and the SSE live-tick path
(`app/api/sse_quotes.py`) disagreed on currency semantics, so
`SummaryStrip.tsx` (which merges `livePrice ?? price`) showed whichever
resolved first — the header flipped by load-order, not tab identity.

**Fix (per operator decision on #1906, 2026-07-04): native price is
PRIMARY — the tradable number — with the display-currency (GBP) worth a
secondary, muted companion, and never dependent on quote path.**

- `InstrumentPrice.current`/`currency` stay **native**; new
  `display_current`/`display_currency` carry the FX-converted companion
  (both null when no FX rate is available or native == display — the
  native price still renders, never a converted number under a wrong label).
- SSE tick already emits a native triple + a `display` block; added
  `liveTickNativePrice` (primary) + `liveTickDisplayCompanion` (secondary)
  so both REST and live-tick paths expose the SAME native/companion split.
  The primary is now always native regardless of which resolves first — the
  flip is structurally impossible, not just papered over.
- `SummaryStrip` renders native primary + `≈ GBP …` muted companion.
- `identity.currency` stays native (the #1845 chart-axis contract), so
  header + chart axis now agree in the primary number.

**Codex ckpt-2 (fixed before this push):** source-couple primary +
companion. When a live tick drives the primary, its OWN display block (or
null) drives the companion — never the REST snapshot's stale conversion for
an older price (e.g. live `USD 110` with no display block must NOT pair with
a REST `GBP 78` computed for the old `USD 100`). Added two live-tick tests
covering the override + no-display-block edge.

### 2. Activity tab hardcoded USD
`fees_usd`/`realized_pnl_usd` rendered under a hardcoded `"USD"` while every
other Portfolio money figure is display-currency. Now converted via the same
`_convert_value` helper `GET /portfolio` uses, renamed `fees`/`realized_pnl`
with a `display_currency` field. Per the operator decision, account/position-
level book values stay display currency (GBP) — this is correct and unchanged
by the header rework. DB columns stay USD at rest (auditable). Missing-FX-rate
labels the response `display_currency: "USD"` to match what the numbers are.

## Settled decision / skill

Updated `.claude/skills/frontend/operator-ui-conventions.md` — the #1845
convention now reads native-primary + display companion (was header-FX'd).
No `docs/settled-decisions.md` entry governs header currency direction.

## Security model

No new auth surface; both endpoints keep their existing session/service-token
dependency. No new writes — read-path FX conversion using `live_fx_rates` +
`app.services.fx` helpers.

## Test plan

- Backend `tests/test_api_instrument_summary.py`: native-primary + companion,
  companion-null-on-missing-rate. 16 pass.
- Backend `tests/test_api_portfolio.py`: Activity conversion (GBP, null,
  USD no-op, missing-rate fallback). 41 pass.
- Frontend `useLiveQuote.test.ts`: new `liveTickNativePrice` /
  `liveTickDisplayCompanion` helper tests.
- Frontend `SummaryStrip.test.tsx`: native-primary+companion render, companion
  omitted when no conversion, live-tick override, live no-display-block drops
  stale REST companion.
- Full frontend suite: 115 files / 1213 pass. `typecheck` clean.
- `ruff check` / `ruff format --check` / `pyright` clean. Fast tier (4737) +
  smoke (141) green. Dark-mode gate clean.


Luke-Bradford:
Reworked to honour the operator decision on #1906 (2026-07-04): header is now **native-primary + display-currency (GBP) companion**, not FX-to-GBP-primary. The prior commits FX-converted the header to display currency (GBP primary) — that fixed the *flip* but chose the currency the operator said should be secondary. `a77fe75a` reverses to native-primary and source-couples the live-tick companion (Codex ckpt-2 finding). Activity realized P&L stays GBP (account-level book value) — aligned with the operator's "position-level values stay GBP" note.

PR #2372, opened 2026-08-07
feat(#2240): phase 5e-3 — the Deflated Sharpe, and the axis two correct numbers were joined on

Luke-Bradford:
Criterion 6's multiple-testing control (`Deflated Sharpe Ratio`), on a declared trial count. Stage 5e-3 of the phase-5 spec.

## Source rule

**Bailey, D. H. & López de Prado, M. (2014), *The Deflated Sharpe Ratio*, JPM 40(5):94-107, SSRN `2460551`** — read at implementation time, not recalled. Every constant is the paper's:

| choice | fixed by |
| --- | --- |
| DSR statistic | **eq. (2)** — `Z[(SR − SR₀)√(T−1) / √(1 − γ₃SR + ((γ₄−1)/4)SR²)]` |
| Threshold `SR₀` | **eq. (1)/(6)** under `H₀: SR = 0`, Euler-Mascheroni weighting of two Normal quantiles |
| Trials' correlation → `N̂` | **Appendix A.3 eqs. (8)/(9)** — `N̂ = ρ̂ + (1 − ρ̂)M`. This is criterion 6's "their correlation" input; a published rule, not a heuristic |
| `γ₄` convention | **RAW** fourth moment (paper's example: `γ₄ = 10`; Normal is `γ₃ = 0, γ₄ = 3`) |

**No new dependency.** scipy is absent from this project; stdlib `statistics.NormalDist` reproduces all three of the paper's published values (0.9004, 0.9505, 0.9505) to 4 dp. Those literals are the reference test and are **transcribed, never imported** — the 2026-08-06 tautology lesson.

## Full-population verification

`scripts/verify_2240_statistics.py --curve` — 5,266 series, 16,236 dates, 516s, **0 property violations, exit 0**.

```
M = 11 declared trials     rho 0.126345 (MEASURED over 8,201 shared dates)
N = 9.7365 independent     V[SR_n] 0.00285033     threshold SR_0 0.083344

S-1  DSR 0.017689  on T = ESS 104,291.8    (nominal 3,133,100 → 0.000000)
S-3  DSR 0.082309  on T = ESS   1,539.2    (nominal    27,782 → 0.000000)
```

Neither strategy is distinguishable from the best of ~9.7 independent noise trials — consistent with 5e-2, where S-3's 95% interval already straddled zero.

New harness properties: **P8** the DSR computes or refuses visibly; **P9** `T` is the ESS, asserted by contrast against the nominal count; **P10** A.3's interpolation bounds; **P11** the correlation is measured, not declared.

## ⚠⚠ The full population falsified my own assertion

P9 originally required the nominal-*n* DSR to come out **higher** — i.e. that a nominal *n* always flatters. **It does not.** Eq. (2) multiplies `(SR − SR₀)` by `√(T−1)`, so a larger `T` amplifies whatever *sign* that difference already carries. Both sleeves sit **below** their threshold (S-1's per-trade Sharpe is negative — expectancy −0.44%/trade), so the nominal count drove both DSRs to ~0 and the one-sided claim **failed on correct code**.

My unit test had passed because I picked a Sharpe *above* the threshold. The invariant is sign-agnostic: a nominal *n* overstates **confidence**, pushing the DSR further from 0.5 whichever way it leans. Both signs are now tested.

## ⚠⚠ The axis is the load-bearing decision

Eq. (2) divides a Sharpe by the standard error *of that same Sharpe*, so `SR`, skew, kurtosis and `T` must describe **one** series. §5.2 says the DSR consumes criterion 3's ESS, and that ESS is in units of **trades** — so the DSR is computed on the trade axis.

`dsr_trade_sharpe` is a per-trade Sharpe and is **not** criterion 7's annualised `sharpe`; `sql/266` stores them under different names so no reader can substitute one for the other. The rejected alternative — keep 5d's per-period Sharpe and divide its period count by the design effect — carries a design effect measured by clustering **trades** onto a series of **periods**, and no test on either side could see it. The paper's own worked example is where the per-observation units are visible (`2.5/√250`, `V = 1/(2·250)`, `T = 1250`).

Found by the round-trip test, not by review: the DSR has **no `effective_sample_size` column of its own** — it consumes criterion 3's — so a row deflated against one sample size and stored declaring another would be silently corrected by a save/load cycle. `StrategyResult` now refuses the disagreeing pair.

## The trial register

`app/services/trial_register.py` is a **declaration** and a documented **floor** — no query can produce criterion 6's count, since a variant eyeballed in a session left no row anywhere.

What counts is a **search of price data**: the four shipped strategies and every #2260 RSI arm, **including the withdrawn non-causal one** (an artefact is still a search). **S-5 and S-6 are absent** — specified in the catalogue, blocked on #2279, never run against data, so they cannot have produced a chance winner.

- Under-counting `M` lowers `N̂`, lowers `SR₀` and **raises** the DSR → a stored value is an **upper bound** on the honest one. The converse must not be read into a pass.
- `V[{SRₙ}]` is estimated only from trials that survived to be measured — the same flattering direction.

Both stated in the module header. **No entry carries a measured Sharpe** (that would be a hardcoded derived statistic); the caller supplies what it measured this run, and the register **raises** on an undeclared key.

## Schema

`sql/266` binds the DSR and its ten inputs **all-or-nothing** (11 columns). ⚠ `trial_count` is deliberately **outside** that set, governed by a one-way dependency instead: a DSR requires a declared count; a declared count does **not** require a DSR. Folding it in would have made unwritable a row the gate is designed to describe, and it keeps both of `check_promotable`'s criterion-6 refusals reachable. 17 new rejection cases cover every constraint.

The `strategy_results` view is re-created — `SELECT *` was expanded at `sql/264`'s creation time, so a column added to the store does not appear in the view.

## Security model

No new surface. Pure computation over already-stored research data plus one additive migration; no endpoint, no auth path, no order path, no external I/O. The hold-out mechanism (`sql/264`) is unchanged and a DSR row is subject to the same namespace trigger as any other result.

## Codex (checkpoint 2)

- **P2 — fixed.** Perfectly correlated trials give `N̂ = 1`, where `Z⁻¹[1 − 1/N]` is `−∞`, and the helper's raise escaped instead of failing closed. Reachable: two register entries running one rule over two corpora are exactly that shape. Now a refusal; the raise is kept in `expected_max_sharpe` where a direct caller passing nonsense is a programming error.
- **P3 — already fixed** by the sign rewrite, which uses `<=` and so flags equality.

## Probes

**18/18 caught.** One was NOT CAUGHT first time, for a **selector fault**: the `ddof` probe was aimed at the paper's reference arm, which builds `TradeMoments` directly and never runs that denominator — and a full-population fixture cannot see it either, since `√(n/(n−1))` at 3.1M trades is 1.00000016. Pinned on n=4, where the same defect moves the Sharpe 2.2361 → 1.9365.

## Gates

`ruff check` · `ruff format --check` · `pyright` · `pytest -m "not db"` · `pytest tests/smoke` — all exit 0, each run separately. 245 tests pass across the six touched suites. Migration applied to dev.

Refs #2240.


Luke-Bradford:
## Review resolution — `aa9c2e81` (both NITPICKs fixed)

Both were dead paths rather than wrong answers, and both are now gone.

### [NITPICK] `expected_max_sharpe`'s `trial_sharpe_mean` is unreachable flexibility
**FIXED `aa9c2e81`** — removed the parameter. Eq. (2)'s `SR₀` is evaluated under `H₀: SR = 0`, which is the only form this module needs; the paper carries the mean through eq. (1) for the general statement and drops it in `SR₀`. Removed rather than documented, per "no what-if abstractions" — if a non-null form is ever wanted it should arrive with the caller that needs it. The docstring now says the mean is *fixed* at zero rather than defaulted.

### [NITPICK] `_SLEEVE_TRIAL_IDS[label]` raises a bare `KeyError`
**FIXED `aa9c2e81`** — now a reported problem, consistent with the rest of `_criterion6`'s refuse-visibly pattern. It also names the consequence that matters rather than just the missing key: a measured trial with **no register entry is uncounted in `M`**, which under-counts the search and therefore **raises** the DSR — the flattering direction, and exactly what criterion 6 calls decorative.

### Gates on `aa9c2e81`
`ruff check` · `ruff format --check` · `pyright` · `pytest -m "not db"` · `pytest tests/smoke` — all exit 0, run separately. 43 pass in `test_deflated_sharpe.py`.

Rebased onto `770f3f65`. ⚠ For the record: `main` moved three times during this review cycle (#2378, #2380 and one earlier), and each time the branch went `CONFLICTING`, which produces **no CI run at all** rather than a failure — the shape the prevention log describes. Diagnosed each time with `gh pr view --json mergeable` plus `gh run list --limit 5` (sibling branches getting runs → not an outage), never with an empty commit.

Luke-Bradford:
## Review resolution — `13543803`

### [NITPICK] `_SLEEVE_TRIAL_IDS[label]` "still" raises a bare `KeyError`
**REBUTTED** — it is unreachable. `verify_2240_statistics.py:452` computes `unregistered = sorted(label for label in usable if label not in _SLEEVE_TRIAL_IDS)` and **returns** at 458 if any is found; the comprehension at 460 is eight lines below that guard, and `usable` is a freshly built local dict that is not mutated in between.

Per the merge rules, a rebuttal round is not merged on my say-so, so I ran **Codex checkpoint 3** on this specific claim. Verdict:

> **Reviewer is wrong on the stated `KeyError`.** … every key in local `usable` is checked against `_SLEEVE_TRIAL_IDS`, and the function returns before line 453 if any are missing. `usable` is a newly built local dict and is not mutated between the guard and comprehension. No other code path reaches that comprehension without the guard.

Only contrived cases remain (another thread mutating the module global mid-call), which I have not guarded and have said so.

### ⚠ But the same Codex pass found five more, and four are the very class you were pointing at
**FIXED `13543803`.** I asked it for every gap rather than just a verdict, and the finding behind the nitpick was better than the nitpick: `_criterion6` had four *other* paths that crash instead of refusing visibly.

1. A sleeve trial id the register no longer declares → `sharpe_variance` **raises** on its undeclared-key guard instead of reporting. Now a problem naming the drift.
2. Two sleeves mapped to one trial id → their Sharpes **collapse silently** in the dict, shrinking `measured_trials` with nothing saying so. Now refused.
3. `np.corrcoef` returns **NaN for a constant series** — reachable when a trial's per-date mean return never varies over the shared dates — and `average_trial_correlation` then raised on its range check. Now refused.
4. A measured `rho` outside A.3's `(-1/(M-1), 1]` bound made `implied_independent_trials` **raise before P10 could describe it**. Now a reported P11 failure. ⚠ Not hypothetical on other data; this run measures +0.126.

5. Also: **both P9 docstrings still said the nominal-*n* DSR "must come out strictly higher"** — the one-sided claim the full population falsified earlier in this PR and which the code no longer asserts. Corrected to the sign-agnostic form. That is stale documentation of a *known-wrong* claim, which is the worst kind, and I had missed it when fixing the code.

Codex's sixth item (concurrent mutation of a module global) is not a real path; not guarded, stated.

### Full-population re-run after the change
5,266 series, 416s, **0 violations, exit 0**, and the figures are **identical** — `rho 0.126345`, `N 9.7365`, S-1 DSR `0.017689`, S-3 `0.082309`. The guards refuse without moving any measurement, which is the property that matters for a change to the harness that produces the evidence.

### Gates on `13543803`
`ruff check` · `ruff format --check` · `pyright` · `pytest -m "not db"` · `pytest tests/smoke` — all exit 0, run separately.

PR #2378, opened 2026-08-07
fix(2176): Item 403's other subsection is admitted on the Item's structure, not a fifth vocabulary patch

Luke-Bradford:
Closes #2176.

Class 1 of the three residual classes. **Class 3 was already fixed** and class 2 landed with #2373 — both re-measured this session, see below, so this closes the ticket.

## The model, not a fifth patch

`.claude/CLAUDE.md`: *"a recurring per-case patch is the trigger to question the MODEL"*. #2088 / #2094 / #2097 / #2169 were all vocabulary patches on this surface and a fifth was due.

**Source rule.** 17 CFR 229.403 is ONE Item with TWO subsections — (a) beneficial owners of more than five percent, (b) directors and management plus the Instruction 5 group row. Issuers render them as two tables under one heading, which is already why `parse_beneficial_ownership_table` concatenates every eligible table in the winning window rather than taking the best-scoring one. Both subsections carry the reg's same prescribed captions, and Rule 13d-3(d)(1)(i) (17 CFR 240.13d-3) DEEMS a person the beneficial owner of securities they have the right to acquire — so 403(b) is commonly 403(a)'s header **plus component columns**.

ExlService (`0001193125-25-103261`, `0001193125-26-181891`) renders exactly that:

```
403(a)  KEPT       | Name and address (1) | Shares | % (2) |
403(b)  DROPPED    | Name and address (1) | Shares | % (2) | Vested but unsettled RSUs (3) | Total
```

`_item403_value_signature` joins every header cell and searches `_COMP_PCT_RE`, so the single word `Vested` in column 5 vetoes columns 3 and 4 — 13 holders per filing, including the Instruction 5 group row.

`_subsection_sibling_tables` admits a table whose normalised captions are a **superset** of an independently eligible table's, in the same window. It takes the eligible set as a precondition, so it **cannot re-rank window selection**; the row-identity floor still applies over it, because it admits a table the value-signature gate rejected.

**Reuse check (`edgartools`, tested on our failing case, not on the skill's summary):** `edgar.proxy.html_extractor.extract_beneficial_ownership` returns **`None`** on both accessions. Its selector is heading-anchored, picks ONE table, has no Item 402 veto and no sibling concatenation, so it cannot express this shape at all. **Structured source:** none — `sec-edgar.md` §"DEF 14A" records no structured-XBRL mandate for proxies as of 2026-05, so HTML table classification is the only path.

## Arms measured, and why the obvious one is wrong

`scripts/analyse_2176_comp_veto_scope.py`, over 25,954 accessions (accession order) of the window-table dump. Control is the shipped `_item403_value_signature`, **imported**; only candidates are simulated. Fidelity of the shipped replay against the real parser: **99.992%** — the 2 divergences are the dump's own `_MAX_HEADERS = 12` clip on an 18-cell header, documented in the script.

| arm | tables newly admitted | tables **de-admitted** | verdict |
| --- | ---: | ---: | --- |
| `c1` comp veto scoped to the column (needs a comp-free amount **and** percent cell) | 16 (14 shapes) | **2** | rejected |
| `c2` same, percent cell only | 40 (36 shapes) | **9** | rejected |
| `c3` a `vested but unsettled` strong arm (the fifth vocabulary patch) | 1 | 0 | works, does not generalise |
| `s1` sibling superset, anchor ≥2 captions | 1 | 0 | — |
| **`s1` sibling superset, anchor ≥3 captions** | **1** | **0** | **ships** |
| `s2` sibling subset either direction, ≥3 | 1 | 0 | — |

⚠ **c1 reads as +59 names — a small, plausible widening — and is a regression.** Its 14 newly-admitted shapes are essentially all Item 402 (`Total Dollar Value of Long-Term Incentive Award (100%)`, `Base Salary ($) | Incentive Target Percentage (%)`, `Time-based Awards (#) (50%) | Performance-based Awards (#) (50%)`), and it **de-admits two literally-prescribed 229.403 tables on `0000100493-25-000114`** (Tyson, `Title of Class | Name and Address of Beneficial Owner | Amount and Nature of Beneficial Ownership | Percent of Class`, 3 + 22 holders) — not by rejecting them, but because a newly-eligible comp table won an EARLIER window and selection never reached the real one. Same window-flip shape as #2164's zero-width scrub. A net row count cannot see this; the selected-SET diff can.

⚠ `_SUBSECTION_CAPTION_FLOOR = 3` is **not** separated by the data — 2 and 3 admit the same one table. It is the tighter of two equal arms, and the constant's comment says so rather than implying a measurement.

## Full-population A/B — `origin/main` worktree vs this branch

`scripts/ab_2358_def14a_line_structure.py`, all **42,705** stored `def14a_body` payloads re-parsed per arm. Control ran in a detached worktree at `0ee3db18` (`git status` clean, parser untouched); arms ran concurrently, which the prevention log already sanctions.

```
holders control             109176      holders treatment  109202
distinct holders LOST            0  (0 accessions)
distinct holders GAINED         26  (2 accessions)
share drift 0   percent drift 0   role drift 0

arm 2 (parse vs STORED):  REGRESSIONS (control yes, branch no)  0
```

All 26 gains are ExlService directors / officers and the two Instruction 5 group rows (`all current directors and executive officers as a group (16 people)` / `(15 people)`). Enumerated in full — nothing else moved.

## Codex checkpoint 2 — one P2, rebutted with a measurement

> *[P2] Require value evidence before admitting sibling tables … the sibling path should still require clean value-column evidence, not just caption superset plus identity.*

Right instinct, wrong threshold, and it would delete the case this PR exists to fix. Measured on the fixture: the 403(b) table has **13 holders, 13 with a share count, 6 with a percent** — the NEOs carry a percent, the directors do not — so `_has_item403_value_rows` returns `False` at 6 < 6.5 and the data-row fallback cannot rescue the table either. 229.403 does not require a percent on every row.

The guard that *does* stand over this path already: `_score_table_headers` returns **0** for any header carrying an `_ITEM_402_AWARD_MARKERS` caption and `window_qualifying` filters `score >= _WINDOW_SCORE_FLOOR`, so grant-date / exercise-price / payout-value / value-realized tables can never reach the sibling path. The residual exposure is a comp table that repeats ALL of an eligible table's captions and carries no award marker — **0 instances across 42,705 accessions**.

## Premise re-measured — class 3 was already fixed, class 2 landed with #2373

Control arm (`origin/main`, this session), holders per accession:

| class | accession | control | note |
| --- | --- | ---: | --- |
| 1 | `0001193125-25-103261` | 3 | the residual — 16 on this branch |
| 1 | `0001193125-26-181891` | 3 | 16 on this branch |
| 2 | `0000821130-25-000055` | 16 | fixed before this PR |
| 2 | `0000821130-26-000025` | 15 | fixed before this PR |
| 3 | `0001140361-25-046376` | **19** | **the exact count the issue records as LOST** |
| 3 | `0001140361-25-046469` | **19** | already fixed |

The issue's class-3 baseline is `main` @ `7b2ae45f`; header extraction has improved since, and that table now carries `Percent (2)` in its (18-cell) header rather than "no percent column anywhere". Nothing to do.

## Tests — 5 cases, 6/6 revert-probes

`TestItem403SubsectionSiblings`. ⚠ **Two probes were NOT CAUGHT first, and both were TEST faults**, recorded in the test docstrings so the shape is not repeated:

- the comp-table negative scored **2**, below `_WINDOW_SCORE_FLOOR`, so it never reached the superset test — the assertion passed for the wrong reason. It now asserts `_score_table_headers(...) >= 3` inline.
- the identity-floor negative used class-label rows, which #2373's per-row guard deletes at extraction whatever the floor does. Re-modelled on bare single-token entity names (`Fidelity`, `Wellington`), which fail `_is_beneficial_owner_identity` and survive `_is_instrument_not_owner`.

## Definition of Done — clauses 8-12

- **8 — golden panel**, over both arms: `AAPL` 13 accessions / 31 holders, `GME` 26 / 20, `HD` 5 / 43, `JPM` 11 / 37, `MSFT` 13 / 39. **0 accessions changed**, all five identical in both arms — no regression, and no evidence the fix works either; that is the A/B's job.
- **9 — cross-source, SEC EDGAR direct.** `https://www.sec.gov/Archives/edgar/data/1297989/000119312525103261/d141102ddef14a.htm` fetched live this session: **byte-identical to the stored payload** (2,076,184 bytes), and parses to the same 16 rows. Against the document's own text: `Rohit Kapoor 3,595,145 (4) 2.21 %` ✓ and `All current directors and executive officers as a group (16 people) (10) 6,583,751 4.04%` ✓ — both newly recovered. The 403(a) control row `Blackrock Inc. (11) 23,308,871 14.33` ✓ is unchanged.
- **10 / 11 — operator.** Parser-only; stored rows do not move until re-parsed. `## OPERATOR VERIFY` block is on #2176.
- **12** — every step above ran at `732c1524` (pre-rebase; head after rebase in the PR checks).

## Filed while in here, not folded in

**#2376** — the 403(a) percent column on this same filing is stored as NULL (`14.33` in the document). The data rows offset the issuer's spacer cells from the header row, so `_resolve_columns` gives `percent_idx = 3` and the value sits at index 4. **Present on both arms**, so it is not this change. Population unmeasured, deliberately, with the census to run first written on the ticket.

## Security

No new input surface, no authz boundary, no SQL. A parser widening whose blast radius is which table's rows get stored.


Luke-Bradford:
**NITPICK** — `scripts/analyse_2176_comp_veto_scope.py` function-local `re` import: `FIXED 143ec544`. Removed; `re` is imported at module scope. Re-ran the analyser over the same 25,954-accession dump after the edit — arm table unchanged (c1 16 admitted / 2 de-admitted, c3 and s1/s2 1 admitted / 0 de-admitted), so the script still runs and still reports the same ranking. `ruff check` · `ruff format --check` · `pyright` · `pytest -m "not db"` · `pytest tests/smoke -n 0` all re-run separately, all exit 0.

Branch re-checked against `origin/main` (`035bdcb0`) with `git merge-base --is-ancestor` — current, no rebase needed.

Issue #2379, opened 2026-08-07
sec_rebuild: a filer_cik scope on an issuer-scoped source resets nothing and reports completed


Luke-Bradford:
## What

`POST /jobs/sec_rebuild/run` with `{"filer_cik": "<cik>", "source": "sec_def14a"}` resets **nothing**, logs `no-op`, and the request row ends `status='completed'`. An operator reading the job status sees success.

Hit live today while running #2358's operator backfill. Six CIKs, six `completed` responses, zero rows reset — caught only because I checked `last_attempted_at` on one affected accession and found it unchanged since 2026-07-19.

## Why

`app/jobs/sec_rebuild.py::_resolve_scope`:

```python
if scope.filer_cik is not None:
    where_clauses.append(
        sql.SQL("(subject_type IN ('institutional_filer', 'blockholder_filer') AND subject_id = %s)")
    )
```

`filer_cik` matches **only** `institutional_filer` / `blockholder_filer` subjects. DEF 14A is **issuer**-scoped — its `data_freshness_index` rows are `subject_type='issuer'`, `subject_id=<instrument_id>`:

```text
dfi   : [('issuer', '1938', '0000043920', 'sec_def14a')]
manif : [('issuer', '1938', '0000043920')]
```

So `filer_cik` + an issuer-scoped source is **structurally incapable** of matching. It is not a data gap — no CIK will ever work for `sec_def14a`, `sec_10k`, `sec_10q`, `sec_8k`, `sec_424b` or any other issuer-scoped source.

The module docstring documents the payload without that restriction, which is what I followed:

```text
{ "filer_cik": str, "source": str }
    # all filings under that filer's CIK for the source
```

That reads as "any filer's CIK, any source".

## What it looks like when it happens

```text
INFO app.jobs.sec_rebuild sec rebuild: scope resolved to 0 triples — no-op
INFO app.workers.scheduler sec_rebuild: triples=0 manifest_reset=0 scheduler_reset=0 discovery_new=0
WARNING app.workers.scheduler Row-count spike detected: sec_rebuild: row_count dropped from 4040 to 0 (ratio=0.00 < threshold=0.5)
```

The spike detector **did** fire — the only thing that noticed. It is a WARNING in a log; the request status is still `completed`, and nothing surfaces to the caller. This is the class `.claude/CLAUDE.md` names: *"a job that no-ops and reports success is invisible to every automated check we have."*

[...]

Issue #2358, opened 2026-08-07
bug(def14a): a `<br>` with no source newline glues two Item 403 amounts into one number — 118,028,165,426 shares stored


Luke-Bradford:
Found by the mandated Codex checkpoint-2 pass on #2169's branch, then measured. Filed
separately rather than folded in, for the two reasons at the bottom.

## The shape

`_strip_inline_html` replaces every HTML tag with a **space**, `<br>` included. When an
issuer's markup carries no literal source newline after the tag, two values land on one
line:

```
'486,340<br>658,400'  ->  _strip_inline_html  ->  '486,340 658,400'
```

`_parse_share_count` then strips spaces AND commas before `Decimal()`, so that cell parses
to **486,340,658,400**. This is a WRONG STORED VALUE, not a dropped row — strictly worse
than #2169's variant, which at least fails loudly by storing nothing.

#2169 fixes the variant where a literal newline survives (`<br>\n`), because the cell then
carries `\n` and can be segmented. It is deliberately inert on this one.

## Measured — full population of the stored table

`def14a_beneficial_holdings`, 110,748 rows, 2026-08-07:

```sql
SELECT count(*) FILTER (WHERE shares > 1e12),
       count(*) FILTER (WHERE shares > 1e11),
       count(*) FILTER (WHERE shares > 1e10),
       count(DISTINCT accession_number) FILTER (WHERE shares > 1e10),
       count(*)
FROM def14a_beneficial_holdings;
-- 9 | 20 | 30 | 9 | 110748
```

Worst rows:

```
Karen David-Green George Damiris Rodney Eads   3,183,454,219,115,736   0001193125-26-148976
Kevin P. Reilly, Jr. † °                          50,018,311,362,250   0001193125-26-140058
BlackRock, Inc. 55 East 52nd Street New York, NY 10055
  Staley Capital Advisers, Inc. One Oxford Centre …  38,611,272,515,000  0001628280-19-000796
```

Note the first and third: the NAME side collapses the same way, gluing three people
(`Karen David-Green` / `George Damiris` / `Rodney Eads`) or two institutions into one
holder identity. So the defect costs both a wrong amount and a wrong holder count.

⚠ **Lower bound, and the direction is known.** `def14a_beneficial_holdings` still holds
pre-#2175 rows — that backfill has not drained (see #2176) — so this counts what the OLD
parser stored. A parse-time census over candidate Item 403 tables found **8 rows in one
[...]

PR #2380, opened 2026-08-07
docs: cost a corpus job before firing it, and never trust a job's completed status

Luke-Bradford:
Two prevention-log entries from running #2358's operator backfill on the live dev stack. Doc-only.

## 1. Cost a corpus operation before firing it

#2358's OPERATOR VERIFY block said to trigger a source-wide `sec_rebuild` and "allow hours". Measured, it is days:

| | |
| --- | --- |
| `sec_def14a` manifest rows | 117,075 |
| worker budget | `max_rows=200` per 5-min tick (`app/workers/scheduler.py:5583`), shared across sources |
| ceiling if def14a takes ~every tick | ~49 h |

The ticket's actual acceptance criterion was **9 accessions**. Six `instrument_id`-scoped rebuilds reset **216 rows in ~20 seconds** and are draining now. The corpus-wide re-parse (the corpus sits at `def14a-v4` against a `def14a-v13` parser) is split out as **#2377** with its cost attached, rather than riding invisibly behind a bug closure.

The entry also re-commits the existing "pgrep for a sibling corpus job first" rule, which I skipped: the source-wide rebuild ran while the ownership loop held a full-population A/B on the same corpus. Harmless only by mechanism — the A/B reads through a named cursor with a fixed MVCC snapshot and writes to a file. Under fresh short transactions it would have been contaminated mid-run.

## 2. A `completed` job status is not evidence it did anything

`{"filer_cik": "<cik>", "source": "sec_def14a"}` returned `completed` six times and reset zero rows: `_resolve_scope` matches `filer_cik` against `subject_type IN ('institutional_filer','blockholder_filer')` only, while DEF 14A is issuer-scoped. No CIK can ever match for any issuer-scoped source. The module docstring documents the payload without that restriction. Filed as **#2379**.

Only `Row-count spike detected: ... dropped from 4040 to 0` noticed — a WARNING in a log, during an operator step. Caught because I checked `last_attempted_at` on a known-affected accession and found it unchanged since 2026-07-19.

Correct form verified: 6 rebuilds, `manifest_reset` 35+53+44+41+23+20 = 216, matching `pending=216` exactly.

## Security

No security surface. Documentation only — no code, schema, endpoint or auth path touched.

## Verification

`docs/review-prevention-log.md` is `merge=union` (#2369), so this append cannot conflict with either running loop. Pre-push gate green.

Refs #2358. Refs #2377. Refs #2379.


Luke-Bradford:
Rebased onto `02981a52` (#2378 landed an entry in the same file). Both entries kept — mine (2) plus the loop's Item 403 subsection entry (1), verified by grep. Branch scope unchanged: 6 insertions, one file.

⚠ **Correcting my own PR description above:** I wrote that `merge=union` meant this append "cannot conflict with either running loop". That is wrong, and this PR proved it — GitHub's server-side merge does **not** apply `.gitattributes` merge drivers, so the union strategy only helps on a local rebase/merge. The conflict was real; the resolution was cheap only because a local rebase could apply the driver.

Issue #2230, opened 2026-08-03
fix(ownership): Section 16 + 13D/G attribute the same shares to multiple reporting persons — the insiders/blockholders wedges are not additive


Luke-Bradford:
Split out of #2226 (see the partition comment there for the full-population evidence).

## Problem

`_compute_residual` and `_compute_concentration` sum every `pie_wedge` slice as if the
sources were additive. For Section 16 (Forms 3/4/5) and Schedule 13D/G they are not:
**the same shares are attributed to several reporting persons by rule.**

- **Rule 13d-5(b)(1)** — members of a group are EACH deemed to beneficially own all
  securities beneficially owned by the group.
- **Rule 16a-1(a)(2)** — a fund, its general partner and its managing members each report
  the fund's entire position, disclaiming beneficial ownership except to the extent of
  pecuniary interest.

`GNLN` carries three distinct holders each reporting exactly **67,864,878** shares
(Jacoby & Co. Inc., LoCascio Aaron, Schoenfeld Adam — all Form 3, all 2019-04-18) against
631,459 shares outstanding.

## This is #1659's question, asked of the channels #1659 did not cover

#1659 demoted DEF 14A to a non-additive memo overlay (`denominator_basis =
"proxy_disclosure"`) precisely because SEC Item 403 disclosures list the same shares under
multiple owners. Its prevention-log entry states the rule: *"before adding a heuristic to
de-dup a source feeding an additive aggregate, ask 'is this source additive AT ALL, or
does it list the same units under multiple owners?' If overlapping-by-design, DEMOTE the
whole source."*

**That argument does not transfer mechanically.** #1659 could demote DEF 14A because the
real holders were already counted via 13D/G + 13F + Form 4. Nothing else covers insiders,
so demoting Form 4 would empty the insiders wedge rather than de-duplicate it. This needs
a different answer, which is why it is a separate ticket rather than a #1659 follow-up.

## ⚠ Exact-value clustering is already FALSIFIED — do not propose it

#1659 tested it and it wrongly merges independent equal-grant executives (dev false
positives: BXP Koop+LaBelle @ 1,875,000; FOXA Ciongoli+Tomsic @ 2,750,000). The
disambiguating footnote ("also deemed to own shares held by X") is free-text prose, not
structured in EDGAR, edgartools or our parser.

The numbers below are evidence of the SHAPE, not a proposed discriminator:

- **2,178 instruments** have >=2 distinct holders at an identical share count in
[...]

PR #2384, opened 2026-08-07
fix(#2230): read the deemed-ownership chain off the Form 3/4 fields instead of inferring it from the share value

Luke-Bradford:
## What and why

`_reconcile_insider_control_groups` (#1652) collapses a Section 16 deemed block that several related CIKs each report in full, gated on the shared value being **non-round** and **≥ 1,000,000**.

Measured on the full population this session — **627** cross-accession, all-10%-owner, equal-value insider clusters exist; **402** pass those gates and **225** do not. The 225 are not marginal: a 16-CIK Blackstone chain on `LGN` at 958,692 (41,308 shares under the floor) and a 9-CIK Starwood chain on `TRTX` at 12,000,000 (an exact multiple of 100,000).

The guards are not truths about deemed ownership. `_reconcile_same_accession_groups` (#1764) already says so in its own docstring: a shared accession *"IS the group evidence #1652/#1645 must infer"*, so that pass *"needs **NO** magnitude floor or roundness proxy — those guards exist only to substitute for the membership evidence that a shared accession already provides."* Wherever equally direct evidence exists, they can be retired for that bucket.

## Source rule

Two **structured** Form 3/4 fields supply that evidence, and both are load-bearing:

| field | rule | what it excludes |
|---|---|---|
| `insider_filers.is_ten_percent_owner` (relationship box) | Rule 13d-5(b)(1) — group members are 10% owners by aggregation | Dropping it admits **26** further clusters, overwhelmingly equal DIRECTOR grants: `FLG` 11,220 × 11 CIKs, `BBT` 2,595 × 10, `NWFL` 825 × 8 — the #1659 class |
| `ownership_nature` (Form 4 Table I col 5, Direct/Indirect) | Rule 16a-1(a)(2) — a chain is ONE direct holder + N deemed owners reporting the same block **indirectly** | The FP class inverts the shape: four family trusts each *fbo* a different beneficiary hold **directly** (`GSHD` 9,787 × 4, `ASST` 150,000 × 4) |

Cluster size stays gated at **≥ 3 distinct CIKs**, matching the tier #1645/I17 already settled for the 13D channel. The 2-CIK all-natural-person pair is left double-counted **deliberately** — `TEAM`'s two co-founders hold near-identical stakes independently and by design, and merging them would erase half of Atlassian's reported insider ownership.

### Discriminators tested and FALSIFIED (recorded so they are not re-derived)

- **Entity name-shape** (the prior session's proposal) fails both ways. `XFOR`'s 13-identity cluster at exactly 2,000,000 was catalogued as the equal-grant FP class because it reads as a list of individuals — it is `New Enterprise Associates 16, L.P.` plus twelve NEA managing members, a **true positive**. Conversely `XPRO`'s four `Trust ulw Janice P. Mosing fbo <different child>` rows are entity-named **false positives**.
- **Shared filing address** fails because officers and directors file from the *issuer's* address. `TEAM`'s co-founders share `c/o atlassian corporation`; the genuine Starwood group on `TRTX` spans three addresses. 128 of the 225 share one address, 96 do not, and the split does not track truth.
- The Form 3/4 "Other (specify)" box **does** carry explicit declarations (`member of a group` 851, `member of 10% owner group` 895, `member of 13(d) group` 245; 3,044 rows / 66 forms) but reaches only **5 of the 627** clusters — a corroborator, not a gate.

## ⚠ Release hazard, found by widening the A/B to the whole pie

Folding an identity out of `survivors` removes it from the owner-once identity grouping, which **releases** that identity's 13F row into the institutions wedge. On `CQP`, folding `Blackstone Group L.P.` (CIK 0001393818) out of a 462,922 deemed block released its 13F row and **the pie GREW by 101,420,487 shares** — the opposite of this pass's purpose.

Claiming the stranded row is **not available**: #1652 settled consumption as exact-value-only precisely so a member's larger genuine 13D block is not deleted (spec, Codex ckpt-1 MED), guarded by `test_non_exact_13d_residual_stays_for_1645` and `test_rep_residual_split_documented`. I tried the claiming fix, both tests failed, and I reverted it rather than reverse a settled decision.

So the **new tier fails closed**: if folding would strand a non-rep member's other-channel rows, the whole cluster is left alone. That is the conservative direction and it costs most of the raw opportunity — 26 changed instruments became 14. The original value-proxy route is deliberately **not** gated on this; its behaviour is unchanged.

**This is the honest residual: the remaining ~211 clusters need the #1652 exact-value-consumption decision revisited, which is a bigger call than this ticket.** Not folded in.

## Full-population A/B

Both arms are **real code** — the control ran from a separate worktree checked out at `origin/main`, never a simulation — **concurrently** against the same dev DB, since the jobs daemon is live and sequential arms would be invalid.

Population: all **2,178** instruments carrying an equal-value insider cluster (the complete blast radius of the pass).

| | |
|---|---|
| compared | 2,178 / 2,178, **0 harness errors** |
| wedge-level INCREASES, any slice, any instrument | **0** |
| identities GAINED | **0** |
| same identity set, different total (re-attribution) | **0** |
| instruments that lost insider identities | **14** |
| double-counted shares removed from the pie | **39,611,250** |

Loss side in full: `XFOR` −24,000,000 (12 identities) · `ALOY` −10,000,000 · `DRIO` −1,785,716 · `ACI` −862,830 · `COUR` −654,000 · `MPAA` −640,600 · `DNLI` −625,000 · `LOAR` −418,084 · `RYTM` −264,402 (9 identities) · `BFRI` −176,652 · `EXE` −102,044 · `ESTC` −55,456 · `PK` −17,644 · `HGV` −8,822.

Harness: `scripts/ab_2230_deemed_chain.py` + `scripts/ab_2230_compare.py`.

## Tests

`tests/test_insider_deemed_chain_collapse.py`, 16 pure-logic cases, no DB, named after the real clusters. **Revert probes 8/8 CAUGHT**, baseline re-verified PASS after restore — including probes for the release-hazard guard and for it being written too broadly (checking the rep, which would refuse every cluster whose rep has a 13F row).

## Gates

`ruff check` · `ruff format --check` · `pyright` · `pytest -m "not db"` · `pytest tests/smoke` — each run separately and unpiped, **all exit 0**.

⚠ **Codex checkpoint 2 could NOT run.** `codex exec review --base origin/main` (files staged first) fails in this environment with `ERROR codex_models_manager::manager: failed to refresh available models: unknown variant `max`, expected one of `none`, `minimal`, `low`, `medium`, `high`, `xhigh`` and dumps ~4 MB of its model-manager JSON instead of a review; 47 such ERROR lines. Retried with `-c model_reasoning_effort=high`, same result. Not claiming the checkpoint passed — flagging it as a tooling breakage for the operator.

- **Clause 8 — smoke panel.** `AAPL`, `GME`, `MSFT`, `JPM`, `HD` are all in the 2,178-instrument A/B population and all five are **unchanged** across arms (they carry no qualifying cluster) — the intended control result for this change.
- **Clause 9 — cross-source, SEC EDGAR direct.** `XFOR`'s admitted cluster is `New Enterprise Associates 16, L.P.` plus twelve managing members, all at exactly 2,000,000, filing from NEA's two offices (1954 Greenspring Drive / 2855 Sand Hill Road per `insider_filers`). That is the Rule 16a-1(a)(2) fund-plus-managing-members shape, independent of any value coincidence.
- **Clauses 10 + 11 — OPERATOR.** See the `## OPERATOR VERIFY` block on #2230. This is a **pure read-path** change — no schema, no stored value, no backfill — so the operator step is the live endpoint only.
- **Clause 12** — every item above is at `e391d7c2` (pre-rebase) / this branch head.

## Security

No security surface. Pure read-path aggregation over already-ingested SEC data; no new input, no auth or authorisation change, no SQL built from user input (the one new query is a static LEFT JOIN with an existing bound `instrument_id` parameter).

## Prevention log

Four entries added: the proxies-are-evidence-substitutes lesson; the falsified name-shape/address discriminators; Postgres `\b` being **backspace** not a word boundary (a regex that silently matched 0 and read as a finding until `\y` gave 2,921); and a ⚠ **pending skill edit** — data-engineer I17's *"`<memberOfGroup>` is empirically ABSENT from modern 13D XML"* is false (38,974 of 107,947 rows carry it, all post-2024-12-18 mandate). Writes to `.claude/skills/**` were denied on this worktree again (#2376, third occurrence), so the correction is logged with a marker rather than applied. **Its conclusion — never depend on the column — is unchanged and was not re-tested.**

Refs #2230. Refs #2226.


Luke-Bradford:
## Review resolution — both findings actioned in `6518833f`

**[WARNING] duplicated rep tie-break** — `FIXED 6518833f`. Correct and worth fixing: the preview and the fold each computed the representative independently, so a future change to one would have made the guard protect one identity while stranding another — silently, since both would still pick *a* plausible rep. Extracted `_control_group_rep_key` and both call it.

**[NITPICK] extract into a shared helper** — `FIXED 6518833f`. Same change; this is exactly what the nitpick asked for, so it is fixed rather than deferred.

Also pinned the equivalence the two call sites rely on, which was implicit before: `sorted(..., reverse=True)[0]` (the fold) and `max(...)` (the preview) return the same element because Python's sort is stable. New test `test_rep_preview_and_fold_pick_the_same_holder` asserts both agree, including on a deliberately tie-heavy cluster with a 13D member.

### Gates on this commit

`ruff check` · `ruff format --check` clean; `pytest` on the four collapse/rollup suites — **41 passed**. Full `pyright` and the whole fast tier were run by the **pre-push hook, which reported `Pre-push gate: green`** — I could not complete them as standalone foreground runs because the box is saturated (load average 10.03 on 10 CPUs; a third headless loop plus a corpus job are running alongside this one). The hook is the deterministic gate and it passed.

⚠ Unchanged from the PR body: **Codex checkpoint 2 still could not run** (`unknown variant \`max\`` from `codex_models_manager`). Not claiming it passed.

The A/B numbers in the PR body were measured at `95f5d3d4`; `6518833f` is a pure refactor of the rep-selection key with no behavioural change, so they stand.

Issue #2721, opened 2026-08-14
Define the position-termination rule for delisted series + raise delisting_source coverage — gate 2 of 2 (the survivorship_free label)


Luke-Bradford:
The second of the two workstreams between firing strategies and funded paper trades. `universe_basis_not_survivorship_free` hard-refuses every stored result, and #2698 measured exactly what is missing (2026-08-14):

- The Intrader corpus HAS the dead names (~12,500 terminating series 2013-2024, fully quarantine-judged) — **but `delisting_source` is NULL on 22,842 of 22,879 series.** We know a series stops; not whether by bankruptcy (−100%), acquisition (often a premium), ticker change, or archive gap. Opposite signs for a held position.
- **The backtest must define what a held position realises when its series terminates.** That decision IS the survivorship treatment; stamping `survivorship_free` while it is undefined would claim handling that does not exist.

## Scope

1. Termination rule by evidence class (Form 25 provision where linked; an explicit `unknown_termination` arm where not — honest wide bounds beat a silent guess).
2. Raise `delisting_source` past 37/22,879 (Form 25 linkage resolved 259 symbols in #2597; register is 2023-only, so this also attacks the 2020-2026 span).
3. Only then: parameterise `BACKTEST_UNIVERSE` — a versioning event (every StrategyIdentity moves; pointer-not-splice for the 324 existing results).

## Hard bounds from #2698 — do not rediscover

- Capture date 2024-09-27 is absolute: windows ending after it can never earn the label on this corpus.
- The regime benchmark (`price_daily` SPY) starts 2022-05-10; a research-corpus benchmark closes the pre-2022 regime gap. Smaller, separable: the seam is `regime_provider`, already injectable on both backtest arm passes.

Refs #2437. Refs #2698. Refs #2597.

Luke-Bradford:
## Steps 1-3 SHIPPED — PR #2728 merged (`d7cb3767`) + fetcher fix PR #2729; register expansion running

**Merged (5 review rounds, all comments FIXED, final APPROVE clean):**
- sql/353: suspension-date-less Form 25 evidence representable (`sec_form25` only; other sources stay date-paired; bare dates still impossible; dated rows still need provision) + `delisting_filed_date`.
- Linkage writes undated evidence; identity gate (series starting after EVERY filing → refused: ALPS/ATCX/USX/DBD; straddling two filings → conflict).
[...]

PR #2728, opened 2026-08-14
feature(#2721): Form 25 evidence without a suspension date + termination rule (steps 1-3)

Luke-Bradford:
## What

Gate 2 of 2 between firing strategies and funded paper trades (`survivorship_free`). Implements the reviewed handoff on #2721 (evidence pass + Codex ckpt-1 already on the issue), in its stated order.

**Step 1 — persist suspension-date-less Form 25 evidence.** `sql/353` + `link_form25_delistings`. The old schema CHECK-paired `delisting_date` with `delisting_source` (sql/249) and biconditionally paired provision with source (sql/253), which made "linked to a Form 25 that states no suspension date" unrepresentable — and `(b)`, exchange-initiated failure, states one on **0 of 105** cohort rows (the date lives in the EX-99 exhibit; exchanges attach a stub). The exchange-failure class — the one a termination rule needs most — was held at zero coverage by the schema itself. Now: `delisting_source` + `delisting_provision` + new `delisting_filed_date` persist with `delisting_date` NULL, for `sec_form25` ONLY. A bare date stays impossible; a dated Form 25 row still must carry its provision; `filed_date` is never written into `delisting_date`.

**Identity gate (found by running the measurement).** The evidence write would have fired on 4 series whose first bar postdates every Form 25 on their symbol (ALPS filed 2023-07 / first bar 2025-10 — a later ticker occupant; DBD — post-Ch11 relisting). New classifier verdict `identity_unverified` refuses these; a series straddling two filings (relist-then-delist cycle, unattributable aggregate) is a `conflict`. Zero straddle cases in the 2023 register (measured: no resolved symbol carries two distinct filed dates); the branch exists before the 2013-2024 expansion makes it reachable.

**Step 2 — register builder hardened for 2013-2024.** Per-year failure gating + one retry pass (a partial year no longer reads as complete — `main` used to discard the failure count); `--years 2013-2024` strictly-sequential mode; hard guard against `--emit-fixture` overwriting the frozen 2023 acceptance fixture; `filings.files[]` pagination in `_resolve_symbol` (sec-edgar.md's own "always recurse" rule — the miss is biased along the failure-vs-acquisition axis, because failures stop filing and stay in `recent` while (a)(3) survivors push their pre-delisting cover pages out); stale GCRA header claim corrected (the local 0.2 s floor is a deliberate half-budget partition).

**Step 3 — `app/services/series_termination.py`, pure, versioned, UNWIRED.** No execution path calls it; `TERMINATION_RULE_VERSION` joins strategy identity only at the commit that wires it (the `BACKTEST_UNIVERSE` parameterisation, next stage). Class table per the ckpt-1-amended design: linked `(b)`/`(a)(4)` → last close × (1−0.55); `(a)(3)` → last close; linked-unparsed + unknown → two-armed bounds; Q-suffix a labelled last-close class; `(a)(4)` kept a distinct label so the expanded register can verify the asserted ≈`(b)` equivalence; `UnlinkedStratum` enum for the census split.

**Source rule:** Shumway, JF 52(1) 1997, Table V p.336 (−30% NYSE/AMEX missing performance-delist returns); Shumway & Warther, JF 54(6) 1999 (−55% Nasdaq). Venue unknown for dead names → adverse anchor, same pessimistic-end construction as `UNKNOWN_NOMINAL_PRICE_BAND`. Rule 12d2-2 provision semantics per sec-edgar.md §2.6.

## Verification (dev DB, full population, 2026-08-15)

- Migration applied; re-link run for BOTH vendors (this is the backfill — it is done, not queued):
  - Intrader: 258 overlap → 37 dated + 221 undated evidence, 0 conflicts, 0 identity-unverified. **(b) 0 → 90 series.**
  - HF archive: 17 overlap → 2 dated + 11 undated, 4 identity-unverified refused.
- A/B: dated arm unchanged (39 rows before and after — same 39 series, same dates); the gain side is undated-evidence-only; distinct-series metric.
- Terminating Intrader cohort (last_bar < 2024-09-01, n=12,468): evidence coverage **37 → 249**, of which (b) **0 → 89**.
- Resolver regression after the pagination change: 2 stored cohort issuers re-resolved to identical symbols (ESTE, CTG).
- `research_corpus_census` view now counts any evidence as `series_with_delisting` and exposes `series_with_delisting_date` as the dated subset.

Not run here: the 2013-2024 harvest itself (hours against the shared SEC budget; the hardened builder is what this PR ships — the run is the post-merge step and its per-year resolution rate is measured, not extrapolated from 2023's 91/105).

## Review

Codex ckpt-2 ran three rounds on this diff: R1 found the non-Form-25 constraint weakening (fixed: undated state scoped to `sec_form25`) and the earliest-vs-latest filing identity comparison (fixed: `latest_filed` + straddle-as-conflict); R2 found the census/classifier disagreement on straddles (fixed: census keys on the verdict); R3 clean.

Skill updated in-PR: `.claude/skills/data-sources/research-price-corpus.md` (evidence model, identity-gate behaviour).

## Security

No security surface: research-corpus schema + one-shot scripts; no auth, no user input, parameterised SQL throughout.

Refs #2721. Refs #2437. Refs #2698.

### DIFF
--- a/.claude/skills/data-sources/research-price-corpus.md
+++ b/.claude/skills/data-sources/research-price-corpus.md
@@ -213,14 +213,31 @@ on March 02, 2023"* — and LIN trades now.
 truncates — at ingest or at read time. `link_form25_delistings` in
 `app/services/research_corpus_ingest.py`.
 
+**#2721 extended this to the EVIDENCE model (sql/353):** `delisting_source` +
+`delisting_provision` + `delisting_filed_date` now persist WITHOUT a suspension date —
+`(b)` states one on 0 of 105 cohort rows, so the date-paired schema held the
+exchange-failure class at zero coverage by construction. `delisting_date` is still never
+back-filled from `filed_date`, a bare date is still unrepresentable, and a DATED
+'sec_form25' row still must carry its provision. Measured on the 2023 register alone the
+re-link moved terminating-Intrader coverage 37 → 249 series, `(b)` 0 → 89. What a held
+position REALISES at a termination is `app/services/series_termination.py` (pure,
+versioned, UNWIRED until the `BACKTEST_UNIVERSE` parameterisation wires it): linked
+`(b)`/`(a)(4)` → last close × (1 − 0.55) (Shumway JF 1997 Table V −30% NYSE/AMEX;
+Shumway & Warther JF 1999 −55% Nasdaq; adverse anchor, venue unknown), `(a)(3)` → last
+close, unlinked → two-armed bounds.
+
 **What is actually detectable, and it is the OTHER half of the guard.** "First bar precedes
 the known listing date" was thought unimplementable because the schema holds no listing date.
 A proxy needs neither a listing date nor a threshold: a series whose first bar postdates a
 Form 25 on the same symbol cannot be the series of the security that filing removed. Four
 today — `ALPS` (filed 2023-07-27, first bar 2025-10-31), `ATCX` (2023-04-19 → 2026-01-09),
-`USX` (2023-07-03 → 2026-06-03), `DBD` (2023-06-20 → 2023-08-14). ⚠ It is reported, never
-acted on: `DBD` is Diebold Nixdorf relisting after Chapter 11, the SAME company. The honest
-output is "identity unverified" — which is the FIRST guard (label it), not the second.
+`USX` (2023-07-03 → 2026-06-03), `DBD` (2023-06-20 → 2023-08-14). ⚠ Since #2721 the
+linkage REFUSES to write evidence onto these (`classify_form25_match` →
+`identity_unverified`): the filing removed a security whose price history the series
+demonstrably is not, and under the evidence model an unrefused write would mark a live,
+running series as delisted. `DBD` is Diebold Nixdorf relisting after Chapter 11, the SAME
+company — which is exactly why the verdict is "identity unverified" (censused), never
+"different company" (asserted).
 
 ⚠ **No corpus series is currently known to be welded.** `REED` and `SRAX` span their `(b)`
 filings with no gap at all (last bar the day before, first bar the day of) — both moved to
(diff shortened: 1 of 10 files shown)


Luke-Bradford:
Both NITPICKs FIXED 796cdee4:

- `--year` + `--years` together now `parser.error` ("--year and --years contradict; pass one"); `--year` default moved to post-parse so the contradiction is detectable.
- Malformed `--years` (`20x3-2024`, backwards ranges) now surfaces as an argparse error, not a traceback.

Both covered by new tests in `tests/test_build_form25_register_cli.py`.

Luke-Bradford:
Round-2 comments:

- WARNING (`_resolve_symbol` pagination gated on `not candidates`): FIXED {see head commit}. Partially rebutting the framing while fixing the substance: `recent` holds the NEWEST filings, so when it yields candidates none in the strictly-older `files[]` pages can outrank them for the newest-at-or-before-delisting scan — "older/better" cannot occur. The real gap was scan-capacity: the top-5 loop can exhaust on fetch/parse failures, where older pages ARE the legitimate fallback. Pagination now fires whenever `recent` yields fewer than the scan's five-candidate capacity. Resolver regression re-run (ESTE → identical symbol + accession).
- NITPICK (page missing `filingFrom` silently skipped): FIXED same commit — default flipped so a span-less page is fetched as unknown-but-possibly-relevant.

Luke-Bradford:
Round-3 WARNING: FIXED {head commit}. `_resolve_symbol` restructured — the recent-window scan runs first; the `files[]` pagination now fires whenever that scan RESOLVES NOTHING (no candidates at/before the cutoff, or every attempted candidate failed to fetch/parse), with a `tried` set so the fallback scan attempts only untried accessions. The static pre-loop count is gone. Resolver regression re-run: ESTE and CTG both resolve to identical symbol + accession.

Luke-Bradford:
Round-4 NITPICK: FIXED {head commit} — dead `if span is not None` replaced with the repo's `assert` idiom (aggregate always returns one row; the guard exists for typing only).

PR #2731, opened 2026-08-15
feature(#2721): wire survivorship-free backtest universe

Luke-Bradford:
## Summary

- pin each backtest universe to one research vendor and admit series rather than live instrument ids
- add the Intrader survivorship-free universe, terminating-series treatment, immutable membership record, and termination census
- remove the structural survivorship refusal only when the result actually uses the survivorship-free basis
- include universe-selection and termination rules in every strategy identity
- preserve the survivor-only default and keep the live paper scanner unchanged

## Why

The corpus now has two vendors linked to many of the same instruments. The old instrument-based query could load both and silently clobber per-name maps. More importantly, every result remained structurally unpromotable because no run could honestly produce the `survivorship_free` basis. This change makes that basis reproducible and auditable without weakening the promotion gate.

## Review correction

Self-review found and fixed a promotion replay defect before push: `evaluated_instrument_count` counts linked instruments plus unlinked dead series, while the recovered checker counted linked instruments only. Regression tests now cover mixed and unlinked-only result records.

## Development database verification

`uv run python -m scripts.verify_2721_survivorship_universe --smoke-run 100`

- survivor-only: exact restoration, 5,265 series, zero Intrader series
- survivorship-free vendor total: 22,879
- admitted: 17,290 (4,649 live, 12,641 terminating)
- excluded unlinked-alive: 5,589
- all strata reconcile exactly
- 100-series rolled-back run wrote four result arms and termination census rows, then persisted nothing
- migration SHA-256 matches the hash recorded by the development database

## Validation

- focused backtest/universe/position/registry suite: 259 passing
- promotion/universe/control-plane regression suite: 62 passing
- `ruff check .`
- `ruff format --check .`
- `pyright`
- all 25 repository invariant lints
- full fast-test tier
- application boot/schema smoke tier
- frontend dark-mode, status-pill, and chart-integrity gates

Refs #2721. Refs #2437.


Luke-Bradford:
Post-merge correction: the hosted review warning was posted as an issue comment while the check itself concluded success and no formal review object was returned. I missed that comment before merge. Both substantive findings are addressed in follow-up PR #2732, with write-time census reconciliation, actual writer-boundary coverage, full pre-push validation, and a dev rollback smoke. #2721 remains open pending that review and the historical relink census.

PR #2744, opened 2026-08-15
fix(#2743): retain source archives used by TA evidence

### LINKED ISSUE #2743: Retain content-addressed source archives used by TA evidence
## Incident

The sealed #2476 PEAD result declares Company Facts SHA-256 `126056a9…`, but the scheduled SEC refresh replaced the only local ZIP. The current cache is `0c5b0d0b…`; no older copy was found. #2493 correctly labels its census as a refreshed source-rule run, but its own current bytes will also disappear on a later refresh.

A digest in a result document proves identity only while the identified bytes remain retrievable. Today it is a checksum without a reproducible artifact.

## Acceptance

- Before a TA evidence script reads a mutable bulk archive, create/resolve a content-addressed retained artifact under the eBull data directory.
- Snapshot by hard link on the same filesystem so the current 1.4 GB archive consumes no duplicate blocks until the mutable path is replaced; fail closed rather than silently copy or read moving bytes when safe retention is unavailable.
- Hash the retained inode, not merely a mutable sidecar, and return its immutable path/digest.
- Concurrent callers are idempotent; an existing digest target is verified.
- Pin metadata sufficient to distinguish source path, digest, size and capture time without storing secrets.
- Integrate the #2493 verifier and retain its currently declared `0c5b0d0b…` capture.
- Tests cover refresh-after-hardlink preservation, bad existing targets, idempotence and cross-device/link failure.

The missing historical #2476 bytes cannot be reconstructed and remain disclosed as unavailable.

Refs #2476. Refs #2493.

Luke-Bradford:
## Outcome

TA evidence now retains the exact mutable source archive before parsing it, under a content-addressed path:

`<retention-root>/sha256/<digest>/<filename>`

This prevents a later cache refresh from making a declared evidence digest impossible to reproduce.

Closes #2743

## Integrity boundary

- Creates a hard link first, then hashes and parses the retained inode.
- Has no copy fallback: cross-device retention fails closed.
- Publishes the retained archive, digest sidecar, and capture metadata without replacement.
- Verifies an existing target and refuses symlinks, corruption, or incomplete/conflicting metadata.
- Preserves the first capture timestamp across idempotent and concurrent calls.
- The verifier records and reports the retained artifact path.

The hard link adds no duplicate data blocks until the mutable cache path is replaced. The old #2476 archive bytes remain unrecoverable; no recovery claim is made here.

## Evidence

The current 1,404,609,804-byte Company Facts archive was retained and re-read through the integrated verifier:

- SHA-256: `0c5b8755431ea133d693982712db09e626954281567e48db6f5a70c4018e2aaf8`
- source and retained path shared inode `50316251`
- link count was `2`
- the verifier reproduced the committed feasibility census and retained metadata/sidecar

## Verification

- 8 focused artifact-retention tests
- 4 PEAD feasibility tests
- Ruff and Pyright
- full pre-push fast suite
- smoke app boot against the dev database
- frontend dark-mode, status-pill, and chart-integrity gates

No trading runtime, capital control, promotion decision, or production path is changed.

### DIFF
--- a/app/services/research_artifact.py
+++ b/app/services/research_artifact.py
@@ -1,0 +1,144 @@
+"""Content-addressed retention for mutable source files used as evidence.
+
+Bulk-source refreshes replace their cache path atomically.  A result checksum
+is not reproducible evidence if those bytes then disappear.  This module pins
+the current inode by hard link, hashes the retained link, and publishes it
+under its measured SHA-256.  It deliberately has no copy fallback: retention
+must be atomic, same-filesystem and cheap, or fail closed.
+"""
+
+from __future__ import annotations
+
+import errno
+import hashlib
+import json
+import os
+from dataclasses import dataclass
+from datetime import UTC, datetime
+from pathlib import Path
+from typing import Final
+from uuid import uuid4
+
+ALGORITHM: Final = "sha256"
+
+
+@dataclass(frozen=True)
+class RetainedResearchArtifact:
+    path: Path
+    sha256: str
+    size_bytes: int
+    metadata_path: Path
+
+
+def _sha256(path: Path) -> str:
+    with path.open("rb") as handle:
+        return hashlib.file_digest(handle, ALGORITHM).hexdigest()
+
+
+def _publish_text_once(path: Path, value: str) -> None:
+    """Atomically create immutable small metadata, or verify the incumbent."""
+    temporary = path.with_name(f".{path.name}.retaining-{uuid4().hex}")
+    try:
+        with temporary.open("x", encoding="utf-8") as handle:
+            handle.write(value)
+            handle.flush()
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
Review follow-up in c732d829:

- Added directory fsync after first publication of the retained archive, digest sidecar, and metadata link, with a regression test covering all three publication points.
- Kept full SHA-256 verification when a content-addressed target already exists. This is intentional: the path is an offline evidence-integrity boundary, and detecting altered bytes is more important than avoiding a sequential read. The code now records that trade-off explicitly.

Verification after the change: 13 focused tests passed, Ruff clean, canonical Pyright 0 errors, and the complete pre-push gate (fast suite, smoke boot, invariants, and frontend integrity checks) passed.

Issue #2745, opened 2026-08-15
Freeze TA in-sample shortlist and one-shot hold-out admission protocol


Luke-Bradford:
## Why this is frozen now

Run 98349 is active on the detached runtime commit `ef206555`. At the time this issue was authored:

- `job_runs.status = running`;
- the process was CPU-active;
- both result relations contained zero `survivorship_free` rows;
- the invocation declared 40 rows: 10 strategies × 2 ambiguity arms × 2 quarantine arms × `in_sample`.

This issue freezes the admission rule before any outcome is visible. It is not permission to open hold-out data or allocate capital.

## Frozen run identity

- universe: `survivorship_free`
- cost model: `static-p75-insession-v3+split-adjusted-max+carry-fx-structural-zero-long-x1-real-usd`
- trial register: `trial-register-2026-08-15-r6`, M = 272
- namespace: `in_sample`
- synthetic control requested
- strategy versions:
  - `s1-time-series-momentum / strategy-registry-v1+cd8a60d57047`
  - `s2-cross-sectional-momentum / strategy-registry-v1+b06b3560c9e3`
  - `s3-mean-reversion-in-trend / strategy-registry-v1+4b466c368e9c`
  - `s4-volatility-compression-breakout / strategy-registry-v1+3b9ed4d738fa`
  - `s5-support-bounce / strategy-registry-v1+878c7474c928`
  - `s6-resistance-breakout / strategy-registry-v1+61ff19557688`
  - `s7-trend-pullback / strategy-registry-v1+9240e675a1ed`
  - `s8-range-mean-reversion / strategy-registry-v1+9052ecd5fb62`
  - `s9-squeeze-expansion / strategy-registry-v1+a1af530c6fbb`
  - `s10-relative-strength-leader / strategy-registry-v1+42e74c7725a1`

## Invocation-integrity gate

Before looking at performance:

1. Run 98349 must finish successfully.
2. Exactly 40 parent rows must publish atomically, with the frozen identities above and no other strategy/version/window.
3. Every strategy must own the complete Cartesian set of `best_case|worst_case × masked|admitted`.
4. Every in-sample row must own its complete walk-forward split, frozen evaluated-universe record, ambiguity record, regime-cohort reconciliation, termination census, and synthetic-control provenance.
5. Any missing, duplicate, partial, mismatched, superseded, or internally unreconciled evidence rejects the invocation. It is repaired or rerun without reading performance.

## Strategy-level shortlist rule

[...]

PR #2746, opened 2026-08-15
feat(#2745): freeze conjunctive in-sample admission audit

Luke-Bradford:
## Outcome

Adds the read-only, outcome-blind verifier for the frozen run-98349 admission protocol in #2745.

The verifier reads `job_runs.status` first and exits with code 2 unless the invocation is successful. While the live job is still running it returns only:

`{"error": null, "run_id": 98349, "state": "running"}`

No result metric is opened on that path.

Refs #2745

## What is pinned

- all 10 exact strategy versions;
- the 40-row Cartesian result shape;
- survivorship-free corpus, cost, rule, metric, bootstrap, DSR, trial-register, synthetic-control, and walk-forward identities;
- recomputed result-version hashes and synthetic-control verdicts;
- frozen-universe, ambiguity, termination-census, regime-cohort, and fold attachments;
- conjunctive strategy admission across all four ambiguity/quarantine rows.

A single failed row rejects the whole strategy. Invocation-integrity failure suppresses every survivor. The script never ranks strategies, chooses an arm, opens hold-out data, changes purpose, or enables capital.

## Verification

- 4 focused tests:
  - complete 40-row population;
  - one bad arm rejects the entire strategy;
  - a missing row invalidates the invocation and yields no survivors;
  - a favourable stored synthetic-control boolean is rejected when recomputation disagrees.
- Ruff and Ruff format clean.
- repository-wide Pyright: 0 errors.
- full pre-push fast suite and smoke app boot passed.
- all invariant and frontend integrity gates passed.

The dev database remains above its non-blocking size warning (72 GB versus 60 GB); that pre-existing operational warning is preserved, not hidden.

### DIFF
--- a/scripts/audit_2745_in_sample_run.py
+++ b/scripts/audit_2745_in_sample_run.py
@@ -1,0 +1,381 @@
+#!/usr/bin/env python3
+"""Read-only audit of run 98349 against the outcome-blind #2745 protocol.
+
+The job status is read before any result metric.  A non-successful job exits
+without opening the result relations, preserving the invocation's atomic
+evidence boundary.  Once successful, every strategy is judged conjunctively
+across best/worst ambiguity and masked/admitted quarantine arms; this script
+never ranks rows or chooses a favourable arm.
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+from dataclasses import dataclass
+from decimal import Decimal
+from itertools import product
+from pathlib import Path
+from types import MappingProxyType
+from typing import Any, Final, cast
+
+import psycopg
+from psycopg.rows import dict_row
+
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
Review follow-up in c2648348:

- NULL evaluated-population counts now reach the named population/universe refusals instead of being converted before validation.
- NULL trade counts now emit `trade_count_missing` and continue the reconciliation audit instead of raising.
- Added a regression test covering both missing-count paths.

After the fix: 5 focused tests pass, Ruff is clean, repository-wide Pyright reports 0 errors, the complete fast suite and smoke boot pass, and every pre-push invariant/frontend gate is green.

Issue #2747, opened 2026-08-15
Complete §3.4 ambiguity threshold and version it into result identity


Luke-Bradford:
## Finding

The shipped ambiguity record is deliberately versioned `v1-no-cohort-threshold`. Every genuinely different best/worst Sharpe pair therefore refuses as `ambiguity_arms_not_compared`, even when the same run computed §9 synthetic controls for both arms.

This issue is opened while survivorship-free run `98349` is still `running` and before any of its performance rows are readable. It is an outcome-independent correction, not a survivor rescue.

## Frozen interpretation

The governing sentence in `2026-08-07-bounded-backtester.md §3.4` defines materiality as:

> the two arms' Sharpe differ by more than the gap between the strategy and the random cohort's 95th percentile

For a real two-arm measurement, each arm has its own matched random-entry cohort. Define:

- `best_margin = best_case_sharpe - best_case_cohort_sharpe_p95`
- `worst_margin = worst_case_sharpe - worst_case_cohort_sharpe_p95`
- `cohort_gap_threshold = min(best_margin, worst_margin)`

Both margins must be finite and strictly positive; otherwise the ordinary synthetic-control refusal already fails and the ambiguity comparison also remains not-comparable. The shared ambiguity verdict is:

`abs(best_case_sharpe - worst_case_sharpe) > cohort_gap_threshold`

Equality is non-material because the source says “more than”. Shared measurements remain non-material without a cohort.

Using the smaller margin is required: the arm pair has one shared verdict, and borrowing the more favourable sibling's headroom could admit a pair whose weaker arm has less evidence above its own matched null than the ambiguity spread.

## Identity and immutability

The ambiguity rule changes what a result means and must be a member of `ResultIdentity`, stored as a first-class result column and included in the result-version hash. A rule change must mint a new result identity; it must not collide with or reinterpret existing rows.

The auxiliary `strategy_result_ambiguity.ambiguity_rule_version` must equal its owning result's identity field. Existing v1 rows retain their old identity semantics and remain fail-closed.

## Acceptance

- bump the ambiguity rule version and add it to result identity/schema/read/write paths;
- derive the threshold only from the two controls computed in the same run for the exact strategy, namespace, and quarantine arm;
- verify both controls' model id, size, seed and percentile agree before comparison;
- store the two arm Sharpes and the derived shared threshold immutably; recompute the verdict at replay;
- reject missing, non-finite, non-positive-margin, mismatched, or hold-out control inputs;
[...]

PR #2748, opened 2026-08-15
Complete matched-control ambiguity gate

Luke-Bradford:
## Outcome

Completes the missing in-sample §3.4 ambiguity comparison without reading run 98349's outcomes or opening holdout data.

For a genuine best/worst arm pair, the writer now freezes the smaller of the two strictly-positive Sharpe margins above each arm's own matched random-entry cohort p95. The arm gap is material only when it is strictly greater than that shared ceiling. Missing or non-clearing controls remain not-compared; mismatched control provenance raises as an integrity failure.

## Identity and storage

- adds `ambiguity_rule_version` to `ResultIdentity`, the result ledger, API provenance, and database schema;
- preserves old v1/v2 result hashes under the historical no-threshold rule;
- corrected measurements mint `strategy-result-v3` identities;
- binds each immutable ambiguity record to its owner's rule version with a composite foreign key;
- prevents the holdout control-support view from composing evidence across ambiguity-rule versions;
- leaves legacy ambiguity records unrecognised by current replay rather than reinterpreting them.

## Operator surface

The strategy overview now derives ambiguity refusals from the frozen record. Four rows existing no longer masquerades as a passed ambiguity check: the page distinguishes unrecorded, not-compared, material, and clean verdicts, and exposes the rule version with the other result pins.

## Outcome blindness

Issue #2747 was opened while run 98349 was still running and before its result relations were readable. Any corrective campaign must rerun the complete same 10-strategy Cartesian set under v3; this change does not select a strategy and does not authorise holdout access or capital.

## Verification

- full pre-push gate green on the exact commit, twice before push and once in the push hook;
- complete fast suite and smoke boot green;
- affected Python/Postgres suite: 476 tests green;
- frontend TypeScript check and 30 strategy-page tests green;
- existing result-model revert harness: 27/27 caught;
- existing result-ledger revert harness: 24/24 caught after repairing two raw-insert baselines exposed by the new required column;
- new #2747 harness: 3/3 caught (favourable-margin substitution, inclusive boundary, identity-member omission), restored suite green.

Refs #2747


Luke-Bradford:
Addressed the review nit in 327c9f3f by wrapping the §3.4 source-rule docstring. The full local pre-push gate passed on the replacement head; waiting for fresh CI/review rather than relying on the superseded green checks.

PR #2754, opened 2026-08-15
Expose exact strategy trade lifecycle

### LINKED ISSUE #2753: Expose exact strategy trade close and reconciliation lifecycle in operator activity
## Gap

#2751 makes fired signals, allocation and entry execution visible on `/strategies`, but a funded row stops at the entry-side `execution_status`. Durable close/reconciliation evidence already exists in `strategy_position_ownership`, `strategy_position_operations`, `strategy_trade_orders`, reconciliation state and `trade_events`; none of it is attached to the operator activity row. A completed trade therefore cannot be audited end to end from the strategy workspace.

## Required outcome

- Extend the bounded current-version strategy activity contract with exact strategy-trade status and unambiguous ownership/position identity.
- Surface the latest material position operation (type, trigger, state, timestamps/error) and exact close/reconciliation state without joining by instrument.
- Surface observed close-event count, realised P&L and fees only when the broker-history ledger is complete; otherwise return explicit incompleteness reasons/nulls.
- Prevent one-to-many joins from duplicating signals; ambiguous ownership must be visible/fail-closed rather than arbitrarily selected.
- Render lifecycle status on `/strategies`, clearly separating rule-defined signal outcomes from broker-observed realised P&L.
- Add DB/API and focused UI coverage for open, closing, closed, rejected, missing-history and ambiguous ownership states.
- Do not add an order bypass, alter promotion criteria, or loosen paper/live controls.

## Acceptance

For each funded entry, the page can trace signal → allocation → trade → exact broker position → material close/reconciliation state → broker-observed close/P&L completeness, while rejected signals remain orderless and live remains fail-closed.

Luke-Bradford:
Closes #2753.

## Outcome

Extends the current-version `/strategies/signals` activity contract and strategy workspace from entry-side visibility through the exact broker-position lifecycle:

- durable strategy-trade state and exact ownership identity;
- latest material close/ratchet/repair operation and its immutable operation/order IDs;
- close-order reconciliation state, broker state, attempts and errors;
- broker-observed close count, realised P&L and fees;
- clear lifecycle and close-history completeness states in the UI.

## Fail-closed properties

- Aggregated/lateral CTEs retain one API row per signal; one-to-many ownership and close ledgers cannot duplicate pagination rows.
- More than one ownership record is surfaced as ambiguous; no broker position or operation is selected.
- Ownership always joins through `strategy_trade_id`, and broker history through the exact owned `broker_position_id`; there is no instrument inference.
- Realised P&L and fees are returned only as a complete pair when every observed close event contains both. Partial broker history returns null money fields plus stable reasons.
- A failed entry with no broker position is a legitimate terminal/no-position lifecycle, distinct from a funded open/closing/closed trade that failed to reconcile.
- Rejected funding decisions remain orderless; no execution, promotion, paper or live gate changes.
- Rule-defined signal outcome remains explicitly separated from broker-observed realised P&L.

## Verification

- 26 DB-backed strategy monitoring/lifecycle tests, including open, submitted close, complete close, failed entry, missing history, partial money and ambiguous ownership.
- 37 focused StrategiesPage tests.
- 70 broader API, monitoring, position-manager, order-reconciliation and paper-runtime tests.
- Full frontend suite: 137 files / 1,524 tests.
- Complete local pre-push gate: Ruff, format, Pyright, 288 mutation-probe anchors, 1,522 fast tests, smoke suite and frontend hygiene checks.

The existing non-blocking dev-DB size warning remains 72 GB; no validation gate failed.

### DIFF
--- a/app/api/strategies.py
+++ b/app/api/strategies.py
@@ -721,6 +721,43 @@ class StrategyPositionCloseResponse(BaseModel):
     operation_id: int | None
 
 
+StrategyTradeStatus = Literal["planned", "submitted", "open", "closing", "closed", "failed", "reconcile_required"]
+StrategyOperationStatus = Literal["intent_persisted", "submitted", "applied", "rejected", "reconcile_required"]
+StrategyReconciliationState = Literal[
+    "unresolved", "pending", "resolved", "rejected", "not_found", "ambiguous", "error"
+]
+StrategyCloseHistoryStatus = Literal["not_applicable", "not_closed", "complete", "incomplete", "unavailable"]
+
+
+class StrategyTradeLifecycle(BaseModel):
+    trade_status: StrategyTradeStatus | None
+    ownership_count: int
+    broker_position_id: int | None
+    ownership_status: Literal["active", "released"] | None
+    position_claimed_at: datetime | None
+    position_released_at: datetime | None
+    position_release_reason: str | None
+    latest_operation_type: Literal["fixed_exit_repair", "stop_ratchet", "close"] | None
+    latest_operation_id: int | None
+    latest_operation_order_id: int | None
+    latest_operation_trigger: str | None
+    latest_operation_status: StrategyOperationStatus | None
+    latest_operation_created_at: datetime | None
+    latest_operation_submitted_at: datetime | None
+    latest_operation_resolved_at: datetime | None
+    latest_operation_error: str | None
+    latest_reconciliation_state: StrategyReconciliationState | None
+    latest_reconciliation_broker_status: str | None
+    latest_reconciliation_attempt_count: int | None
+    latest_reconciliation_updated_at: datetime | None
+    latest_reconciliation_error: str | None
+    close_event_count: int | None
+    realised_pnl_usd: Decimal | None
+    observed_fees_usd: Decimal | None
+    close_history_status: StrategyCloseHistoryStatus
+    incomplete_reasons: list[str]
+
+
 class FiredSignal(BaseModel):
     signal_id: int
     strategy_id: str
@@ -745,6 +782,7 @@ class FiredSignal(BaseModel):
     execution_status: str | None
     actual_fill_price: Decimal | None
     slippage_pct: Decimal | None
+    trade_lifecycle: StrategyTradeLifecycle | None
 
 
 class FiredSignalsResponse(BaseModel):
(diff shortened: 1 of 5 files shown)


Luke-Bradford:
Addressed the Python vocabulary-drift nit in 70bb7930 by introducing `StrategyTradeStatus`, `StrategyOperationStatus`, `StrategyReconciliationState`, and `StrategyCloseHistoryStatus` aliases shared by the Pydantic model and reconstruction casts. The TypeScript declarations remain the explicit cross-language API contract; replacing those with generated client types would be a separate repository-wide change.

Confirmed the second point directly: `BadgeTone` in `frontend/src/components/ui/Badge.tsx` includes `info`, with its own semantic blue class mapping. The replacement head passed the complete local pre-push gate again; waiting for fresh CI/review.

Luke-Bradford:
Reproduced the reviewer warning against the ambiguous-ownership fixture and fixed it in a61eb152. `trade_lifecycle.close_event_count` is now nullable and is exposed only when ownership resolves to exactly one broker position (or in the legitimate no-position planned/submitted/failed states, where it is zero). Ambiguous/missing ownership no longer exposes the aggregate count, just as it already suppresses position identity, operation, realised P&L, and fees.

The DB regression now creates two competing owned positions with a close event on each and asserts: one signal row only; ownership count 2; position/operation/count/P&L/fees all null; stable ambiguity reason. Focused backend/frontend tests and the complete pre-push gate are green on this exact commit. Requesting fresh CI and review; I will not rely on the prior green checks.

Luke-Bradford:
I also treated the fresh APPROVE nits as actionable rather than waving them through. Fixed in d4967fec:

- a valid terminal reconciliation state of `rejected`, even with no `last_error_code`, now emits the stable fail-closed reason `position_operation_reconciliation_rejected`; the DB regression exercises the schema-valid rejected shape (`reconciled_at` set)
- trade states `failed` and `reconcile_required` now explicitly use the red risk badge; the UI regression verifies both even when `incomplete_reasons` is empty

The complete pre-push gate is green on this exact head after the additions. Requesting one more fresh CI/review pass.

Luke-Bradford:
Reviewed the final nit against `_trade_lifecycle` rather than accepting or deleting it mechanically. The backend returns a non-null lifecycle for every `funding_status == "funded"` row, including a missing trade join (which is surfaced as `funding_not_reconciled_to_trade`). I am retaining the frontend `funded + null lifecycle => Trade unavailable` branch as a deliberate fail-visible API-boundary fallback for malformed/stale responses; it does not claim a trade exists and is safer than rendering `No funded trade` for that inconsistent state. No change required.

PR #2757, opened 2026-08-15
fix(strategies): bind metrics to causal panel axis

### LINKED ISSUE #2697: The 64-year hold-out result rows are internally inconsistent — trade count and CAGR say ~7 years, the label says 64.5, and vs-buy-and-hold is computed across the mismatch
Found while answering "do we have a viable strategy" from the stored evidence, 2026-08-14.

## The inconsistency

For `s1`, `s2`, `s3` the `hold_out` row labelled `1962-01-02 -> 2026-07-08` does not reconcile with itself. The annual folds in the same table *do*, which is what makes this a defect rather than a convention I have not read.

| strategy | total_return | stored `cagr_pct` | CAGR implied by total_return over 64.5y | years implied by stored CAGR |
| --- | --- | --- | --- | --- |
| s1 | -49.12% | **-8.34%** | -1.04% | 7.8 |
| s2 | +2230.48% | **+51.40%** | +5.00% | 7.6 |
| s3 | -58.44% | **-15.92%** | -1.35% | 5.1 |
| s4 (4.5y window) | -81.48% | -31.20% | -31.17% ✅ | 4.5 ✅ |

`s4` reconciles exactly. The three 64.5-year rows are out by ~10x.

**Trade counts corroborate the short reading.** `s2` records 6,677 trades on the "64.5-year" row — but 1,069 trades in the single year 2022 and 516 in the 2026 half-year. ~1,000 trades/year × 6,677 total ≈ **6.7 years of trading**, not 64.5.

`cagr_pct` derives from `years = (len(dates) - 1) / ppy` (`app/services/strategy_statistics.py:403`) with `ppy ≈ 251`, so ~1,900 curve points ≈ 7.6 years. The CAGR is correct *for the curve it was given*. The window label is not.

**`buy_and_hold_return_pct` reads the full 64.5 years** — 9,817% ≈ 7.39%/yr, which is right for US equities over that span. So the benchmark spans 64.5 years while the strategy curve spans ~7.

### Consequences

1. **`return_vs_buy_and_hold_pct` is meaningless on these rows.** `s2`'s -7,587% is a ~7-year strategy result subtracted from a 64.5-year benchmark.
2. **`cagr_pct` reads as spectacular where it is not.** `s2` at "51.40% CAGR" is the single most promotion-attractive number in the store. Anything keyed on it — a gate, a summary, an operator glance — is reading a figure whose denominator is not the declared window.
3. **The same declared window carries two different benchmark values** — 9,817.72% and 11,695.56% appear on rows with identical `window_start`/`window_end`. At least one is wrong.

### Reproduce

```sql
SELECT DISTINCT ON (strategy_id) strategy_id, window_start, window_end,
       total_return_pct, cagr_pct, trade_count, buy_and_hold_return_pct
FROM strategy_results_store
WHERE namespace='hold_out'
  AND window_end=(SELECT max(window_end) FROM strategy_results_store)
ORDER BY strategy_id, window_start ASC, created_at DESC;
```

### What needs deciding first, before any fix

Which of the two is the truth — is the **window label** wrong (the run genuinely covered ~7 years and is mislabelled), or is the **equity curve** truncated (the run should have covered 64.5 years and silently produced 7)? They imply very different fixes, and the second is the more serious: it would mean the long-horizon backtests never actually ran to their declared span.

Do not "fix" `cagr_pct` to match `total_return_pct`. The arithmetic is not the defect; the disagreement about what period the row describes is.

⚠ This does not change the viability conclusion — see the per-year folds, which reconcile and show 20/20 losses to buy-and-hold. It changes whether the headline row can be cited at all.

Refs #2437. Refs #2240.

Luke-Bradford:
Closes #2697

## Outcome

Binds every corrected strategy result to a predeclared causal date axis and pre-mask opportunity population. Legacy or incomplete axis evidence remains readable but fails closed for promotion.

## What changed

- replaces raw/custom evaluation-window calls with registered evidence-window IDs
- derives and persists the complete metric-axis tuple, digest, opportunity-set digest, and evidence-window ID
- computes strategy, comparator, and synthetic-control metrics on the exact dated axis, including cash tails
- requires exactly one comparator leg per opportunity name and prevents cross-sectional ranking from using names outside the frozen population
- reconciles annualisation provenance through the result ledger and promotion replay
- validates promotion evidence-window IDs and dates against the authoritative registry
- bounds strategy-price and regime reads at the sealed evaluation-window end
- adds forward-only migrations 359 and 360; no legacy backfill
- adds derivation and old/new A/B verification harnesses, including 1,000-member synthetic-control threshold deltas
- runs declared 1,000-member controls through a bounded four-process spawn pool; seeds remain pure functions of member index, results are restored to canonical index order, and missing or duplicate members abort
- emits transient member-count heartbeats only; no partial performance figures are exposed

## Verification completed on current head

- exact head: 7b87e567d90707fe41d8ae3ee0577c887f1e3f56
- full repository pre-push gate green: Ruff, formatting, full Pyright, shellcheck, all mirrored static invariants, full fast tests, app boot smoke tests, frontend integrity checks, and 289 revert-probe anchors
- expanded focused metric-axis, backtest, synthetic-control, regime, quarantine, promotion, paper-execution, and MT1 contract suites green
- result-model mutation probe: 28/28 mutations caught with restored pass
- dedicated metric-axis probe: 7/7 structural reverts caught with restored pass, including fixed strategy/member axes, sealed corpus/regime reads, exact final head, and SQL ceilings
- exact serial-versus-spawned cohort equality covered for current and script-local legacy A/B paths
- a 50-series S1 wiring smoke on earlier functional head 162f859f completed all four ambiguity/quarantine A/B rows and both 1,000-member controls per row in 225.66 seconds; this is a wiring benchmark only, not current-head acceptance or strategy evidence

## Acceptance still required before merge

1. Wait for exact legacy retry run 99585, bound to request 408 and the declared r6 payload.
2. Run the structural-only legacy auditor first. Every row must be refused as metric_axis_unproven; no legacy performance value may be inspected or interpreted.
3. Execute the full-population old/new A/B on exact head 7b87e567d90707fe41d8ae3ee0577c887f1e3f56.
4. Run the structural A/B auditor before reading any performance output; require the complete declared current and legacy populations, sealed window, and final-head recheck.
5. Inspect final CI and obtain substantive review of the final SHA before marking ready or merging.

The corrected preregistered MT1 evaluation and metric-axis derivation replay are programme gates after this infrastructure change is accepted; they are not evidence that can be retrofitted into this PR. This PR makes no profitability or promotion-ready claim and remains draft until the acceptance sequence above is complete.


Luke-Bradford:
## Acceptance step 1 does not resolve by waiting, and step 2's property is already provable

#2773 is merged into this branch. New combined head: `057dbb56f48212b8278acda56da66505c2f6fa68` (41 ahead of `main`, 0 behind). The `7b87e567` pin in the body above is superseded.

### Step 1 — "wait for exact legacy retry run 99585" is finished, in failure

Measured on the dev DB, 2026-08-19:

| run | request | status | finished | error |
| --- | ---: | --- | --- | --- |
| 98349 | 396 | `failure` / `internal_error` | 2026-08-15 19:09 UTC | `orphaned: reaped at boot (owning worker thread died without a terminal status)` |
| 99585 | 408 | `failure` / `internal_error` | 2026-08-16 11:39 UTC | same |

```sql
select count(*) from job_runs
 where job_name = 'strategy_backtest_run'
   and params_snapshot->>'trial_register_version' = 'trial-register-2026-08-15-r6'
   and status = 'success';
-- 0
```

Exactly two runs ever carried the declared r6 params and **both failed**; none succeeded. So step 1 cannot be satisfied by waiting — run 99585 ended three days ago.

Nor is the failure a fail-closed refusal, which was the other reading on the table. A refusal would be a *completed* run whose rows are stamped `metric_axis_unproven`; these are orphan reaps, an infrastructure death. Request 408's own row records the cause: `operator terminate: oversized backtest stopped before jobs-daemon restart` — the oversized-cohort problem #2772 exists to bound.

`scripts/audit_2697_legacy_metric_axis.py` agrees, and refuses to proceed rather than interpreting:

```
$ uv run python -m scripts.audit_2697_legacy_metric_axis --run-id 99585
{"error": "orphaned: reaped at boot ...", "run_id": 99585, "state": "failure"}   # exit 2
```

It requires `status = 'success'`, and `_EXPECTED_PARAMS` pins `trial-register-2026-08-15-r6`, so no other stored run can be substituted — the last `success` (92744, 2026-08-12) is a hold-out refresh with different params.

**Step 1 therefore requires a re-run, not a wait — and that re-run is only affordable behind #2773's compiled path.** That is an independent confirmation of the stacked merge order, arriving from the opposite direction to the usual one: not "the A/B needs the compiled path" but "the legacy arm itself died of the thing the compiled path fixes".

### Step 2 — the property it exists to prove already holds on the full stored legacy population

The `--run-id` gate is blocked; the claim underneath it is not. Every stored result row predates this change, and the refusal is derivable without any run:

| check | result |
| --- | ---: |
| rows in `strategy_results` | 24 |
| rows with a non-null `metric_axis_*` / `opportunity_set_digest` / `evidence_window_id` field | **0** |
| rows whose `result_version` digest reconstructs from its identity | **24 / 24** |
| `metric_axis_promotion_refusals` verdict | **`('metric_axis_unproven',)` on 24 / 24** |

Derived with the auditor's own `_identity` and `_NoMetricAccess` guard, so a refusal that ever started consulting an outcome value would have raised instead of returning. It did not: **no legacy performance value was read**, per step 2's constraint.

This is evidence for the property, not a substitute for the step. The declared-population half of the auditor (`EXPECTED_KEYS`, `EXPECTED_TRIAL_COUNT`) still needs the re-run.

### Carried in from #2773 — a launch prerequisite

#2775 (fixed on the merged branch) corrected two under-counts in the scale gate's memory arm. **Both previously published projected-memory figures are superseded and were under-statements**; the wall-time figures are unaffected, because `time.monotonic()` differences are intervals. The S-1 launch pilot must be re-run so the 8 GiB ceiling is checked against the corrected projection before any bounded invocation.

No performance, profitability or promotion claim is made here.

Luke-Bradford:
## Unblocked — it was the draft flag, not the conflict

This PR reported `CONFLICTING` and its `review` check reported `skipping`. Both were misleading, and the second was the actual blocker.

**The conflict was stale.** `git merge origin/main` applied across all 61 commits with **zero conflicts**. Merged in as `05081c07` (a merge of `main` into the branch, not a rebase, so the push is fast-forward and no remote history is rewritten). Now **0 commits behind `main`**.

**The blocker was `isDraft=true`.** `.github/workflows/*review*.yml` gates its job on:

```yaml
if: github.event.pull_request.draft == false
```

so the review never ran, `gh pr checks` showed `review  skipping`, and `.autonomy/config.yaml`'s `merge_gate.strategy: bot_comment` waited for a `github-actions` "Claude Code Review" comment that could not be produced. `safe_merge.sh` would have refused this PR forever with *"no review comment ... yet"* — the same indistinguishable-from-pending message #2783 records for the stacked-base case. Marked ready for review; the bot can now run.

⚠ This is #2783's failure class with a different key. There the gate keyed on the PR's **base**; here on its **draft** flag. Both fail closed, silently, and both surface as an ABSENCE of checks — which reads as "not finished yet" rather than "will never happen". Worth folding into #2783's fix direction 2: `safe_merge.sh`'s `bot_comment` gate should name *why* a review is unobtainable (draft, or non-default base) rather than emit the generic wait message.

**Local gate green on `05081c07`** before push: ruff, ruff-format, pyright (whole repo, tests included), `pnpm --dir frontend typecheck`, all 25 chokepoint lints, probe-anchor validation, CI/pre-push parity, fast tier, smoke.

⚠ What this PR is carrying, for whoever reviews it: 61 commits including the backtest evaluation going **465.4s → 68.6s** on the s5/300 profile (end-to-end bit-identical, 4 arms × 64 fields compared with `==`, 0 differences) and the #2779 close-tie fix, which corrected a phantom loss booked on every affected S-7 exit (+1.6 CAGR points, +0.09 Sharpe). None of that is on `main` yet, and the TA autonomy loop builds from `main`.

Luke-Bradford:
## Review resolution

**[BLOCKING]** — `INTRADER_CAPTURE_DATE` never imported, a `NameError` waiting to fire.

`REBUTTED` — it is imported. `app/services/backtest_run.py:200-206`:

```python
from app.services.universe_selection import (
    INTRADER_CAPTURE_DATE,
    SURVIVORSHIP_FREE_VENDOR,
    ...
)
```

and it resolves at runtime:

```
$ python -c "from app.services import backtest_run as b; print(b.INTRADER_CAPTURE_DATE)"
2024-09-27
```

⚠ The finding is a **diff-vs-file** confusion, and worth recording as such: that import block is unchanged code, so it was not in the diff under review, and "not present in the diff" was read as "not present in the file". The name is used at four other sites in the same module (`:305`, `:1492`, `:2737`), any of which would have failed long before the `# pragma: no cover` line if the import were genuinely missing.

**[WARNING]** — `evaluated += 1` moved ahead of the `len(indices) < 2` skip, inflating `series_evaluated`.

`FIXED 738d843b` — correct and a real defect. `series_evaluated` is logged and feeds `evaluated_instrument_count`, so it is reported evidence rather than a diagnostic, and counting series that contributed nothing makes an arm look broader than it was. Moved back after the skip in **both** `evaluate_arm` and `evaluate_level_arms`, with a comment recording that the position is load-bearing.

**[NITPICK]** — `sql/359`'s hardcoded `DATE '2021-06-29'` has no coupling test.

`FIXED 738d843b` — a third copy of the boundary after `HOLDOUT_BOUNDARY` and the spec pin, with nothing tying it to either; a CHECK constraint that disagrees with the code either rejects valid rows or admits invalid ones. Added `test_the_SQL_constraint_carries_the_same_boundary_as_python`, which **reads the literal out of the migration** rather than restating it, so it cannot become a fourth copy.

⚠ Scoped to that one constraint — the same migration registers the evidence windows, which carry their own different and correct dates.

⚠ Probed rather than assumed: mutating the SQL literal to `2021-06-30` makes it fail with `carries [2021-06-30] but HOLDOUT_BOUNDARY is 2021-06-29`, and it passes again on restore. A coupling test that cannot fail would have been worse than none.

**[PREVENTION] undefined-name-guard** — grep changed lines for identifiers absent from the import block, especially under `# pragma: no cover`.

`REBUTTED` — the specific instance is a false positive (above), and the rule as stated cannot be applied by grepping the diff, which is exactly what produced the wrong finding: the import may be unchanged and therefore invisible. The check that *would* work is resolving the name against the whole file, which `pyright` already does on every push and which is green here. Adding a weaker grep-based ritual on top would give false confidence in the direction that just failed.

⚠ The real lesson is worth keeping though, and it is about `# pragma: no cover` rather than imports: that annotation suppresses coverage evidence, so a line carrying it has no test telling you it ever executed. Findings against such lines deserve runtime verification before they are called blocking — which is what `print(b.INTRADER_CAPTURE_DATE)` did here in one command.

Luke-Bradford:
## Review resolution — round 2

**[WARNING]** — `sql/360` hardcodes `EVALUATION_WINDOW_START` / `HOLDOUT_BOUNDARY` with no coupling test, the same class fixed for `sql/359` one round earlier.

`FIXED c245335f` (test in `738d843b`'s successor) — correct, and the repetition is the point. Rather than add a second copy of the guard, **generalised the existing test to cover both migrations**: `test_the_SQL_migrations_carry_the_same_dates_as_python` now reads `sql/359`'s `strategy_results_metric_axis_namespace` block and `sql/360`'s control-support predicate, comparing each against the Python constants.

⚠ Probed, not assumed: mutating `sql/360`'s `DATE '1962-01-02'` → `'1962-01-03'` fails with

```
sql/360 control-support carries [1962-01-03, 2021-06-29], python says [1962-01-02, 2021-06-29]
```

and it passes again on restore.

**[NITPICK]** — `rel_tol=1e-12` is extremely tight for a cross-recomputed float; document why it is safe given the pairwise-summation prevention entry.

`FIXED c245335f` — documented, because the question is a fair one to ask again. The short answer: **this is not a reduction.** `periods_per_year` is `(len(dates) - 1) / (span_days / DAYS_PER_YEAR)` — two exact integers and one division — so recomputing it from the same axis is bit-identical, and 1e-12 is looser than the equality that actually holds. The pairwise lesson governs SUMS and MEANS over arrays, where accumulation order is a free variable; there is no accumulation here. The tolerance exists to absorb a decimal round-trip on the stored value, not arithmetic noise.

**[PREVENTION] sql-literal-coupling**

`EXTRACTED docs/review-prevention-log.md` — added as its own entry. The rule as recorded: read the literal **out of** the `.sql` rather than restating it (a restated date is just a fourth copy), scope the regex to the predicate that owns it (a migration that also registers reference data carries other, legitimately different dates), and **probe that the test can fail** before trusting it.

⚠ The entry also records why this one was worth generalising rather than patching: the bot raised the same class twice in consecutive rounds, one file over. That repetition is the signal.

Luke-Bradford:
## Review resolution — round 3

**[NITPICK]** — the `sql/359` case is scoped to its CHECK block but the `sql/360` case scans the *entire* file, contradicting the prevention-log entry added in this same PR.

`FIXED` — correct, and worth fixing rather than deferring: a rule falsified by its own reference implementation is worse than no rule, and this PR is where both were written. The `sql/360` case is now scoped to the `candidates` CTE, which is what owns both literals.

Probed in **both** directions, because scoping can fail either way:

| probe | expected | result |
| --- | --- | --- |
| drift **inside** the CTE (`2021-06-29` → `2021-06-30`) | fail | fails ✓ |
| unrelated `DATE` literal **outside** the CTE | ignored | passes ✓ |

⚠ The second probe failed on its first attempt, and the probe was wrong rather than the code: `SELECT h.result_id AS holdout_result_id,` appears at line 13 **inside** the CTE as well as at line 46, so a `replace(..., count=1)` inserted the "unrelated" literal *into* the block under test — and the test failed correctly. Re-run with explicit line placement it passes.

That is this repo's own documented trap (a mutation probe that does not land where its author thinks it does), hit while verifying a guard against a different one. Recording it here because the near-miss is the interesting part: had I read the first red as "the scoping is broken" I would have loosened a guard that was working.

PR #2758, opened 2026-08-15
feat(strategies): preregister capped volatility overlay

Luke-Bradford:
## Outcome

Creates a reviewable, outcome-sealed first capital-candidate design without granting it capital authority.

- renames the ambiguous C-1 pre-spec to the unique MT-1 identity;
- corrects the evidence transfer: published results are long-short factor/anomaly portfolios, while MT-1 is a capped long-only variant;
- makes the leverage mismatch and the S-8 clock mismatch explicit;
- freezes the signal lineage, causal 120-usable-month expanding normalisation, 1x cap, complete-session contract, paired difference-in-differences inference, turnover disqualifier and 36-date/157-week forward-shadow floor;
- adds a pure source-hashed portfolio-level sizing contract with no database or outcome access.

## Fail-closed boundary

This PR deliberately does not add a manifest capital candidate, trial-register entry, database declaration, evaluator, promotion, forecast or broker path. The trial count is not spent until the implementation is complete and the experiment is ready to run. S-10 remains a permanent harness control.

## Verification

- 11 focused volatility-overlay tests;
- Ruff and formatting clean;
- Pyright clean;
- full repository pre-push gate green twice, including all fast tests, app/DB smoke boot and frontend integrity gates.

## Remaining ordered work

1. review and merge this scientific contract;
2. implement the fresh MT-1/S-8 reference books and paired evaluator without reading outcomes;
3. add the two exact trial entries and freeze declarations immediately before the first gated run;
4. run the in-sample experiment, then the sealed hold-out only if every prior gate passes;
5. require the mandate-driven batch allocator before any bounded demo capital.

Refs #2437.

### DIFF
--- a/app/services/strategy_volatility_overlay.py
+++ b/app/services/strategy_volatility_overlay.py
@@ -1,0 +1,235 @@
+"""Causal portfolio-level inverse-variance exposure for MT-1 (#2437).
+
+Contract: ``docs/proposals/ta/2026-08-15-mt1-volatility-managed-relative-strength-
+preregistration.md``.  This module is deliberately pure and sees only completed
+monthly returns.  It does not know a strategy, instrument, signal, price row or
+outcome namespace, so it cannot silently turn the portfolio overlay into
+per-name inverse-volatility weighting.
+
+Published core: Cederburg et al. (2020), equations (3), (4), and their real-time
+normalisation: prior-month realised variance, a 120-month initial training
+period, then expanding history.  eBull's leverage ban adds the by-construction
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
Reviewed the zipapp/frozen-bundle nit. The import-time source hash is the repository's established strategy identity pattern (S-1 through S-10 and the outcome/structure rule sets use the same Path(__file__).read_bytes() mechanism), and eBull's supported runtime is a filesystem checkout/container rather than a zipapp. Changing only MT-1 would create identity inconsistency without adding a supported deployment mode, so no change is warranted in this slice.

PR #2759, opened 2026-08-15
feat(#2437): add frozen MT-1 trial evaluator

Luke-Bradford:
Closes part of #2437

## Outcome

Implements the pure, outcome-source-independent statistic for the preregistered MT-1 four-arm controlled experiment. This PR does not open price outcomes, alter the trial register, freeze a database declaration, promote a strategy, or authorise capital.

## Frozen contract enforced

- evaluates MT-1 scaled/unscaled and S-8 scaled/unscaled unconditionally on one common complete-month axis
- computes CER with gamma=5 and sample variance
- reports the preregistered difference in differences: MT-1 CER improvement minus S-8 CER improvement
- uses the same paired moving-block indices across all four arms, 12-month blocks, 10,000 replications, seed 243715082026, and a two-sided 95% linear-percentile interval
- requires at least 120 common months
- requires both scaled arms to match one frozen month-end decision clock, reconcile every exposure change, and remain at or below 600% annualised turnover before calculating outcome statistics
- reports positive-difference lower bounds, drawdown improvement, and 5% expected-shortfall improvement as conjuncts
- labels the aggregate only as historical statistical conjuncts; no promotable or capital-eligible field exists
- refuses non-finite inputs and derived arithmetic
- pins exact synthetic interval outputs so changes to the seed, blocks, pairing, CER or percentile tails fail regression

## Verification completed

- exact current commit: a69450f256d3b91b4419af5dfe8f3a7cf6645ad2
- 18 focused evaluator tests green on the exact current commit
- 147 overlay/evaluator/bootstrap/trial-register/declaration-gate tests were green before the test-only interval-regression amendment
- full repository pre-push gate green on the exact current commit: Ruff, formatting, full Pyright, 25 static invariants, 288 mutation-probe anchors, full fast tests (including all 18 evaluator tests), app/DB smoke, and frontend integrity checks

## Explicitly still required

- build the four fresh portfolio books from the causal backtest path, including exact daily/session, cost, universe, ambiguity, termination and quarantine provenance
- derive rather than accept the structural turnover/reconciliation audit inputs
- review and merge this pure evaluator before adding the two trial-register entries
- freeze the MT-1 capital-candidate and S-8 falsification-only database declarations
- pass the sealed access gate before any outcome read
- run in-sample first; a failed arm ends this version
- separately pass synthetic-control, DSR, benchmark, recent-window and all promotion gates

This PR deliberately makes no claim that MT-1 is profitable or evidence-qualified.

### DIFF
--- a/app/services/strategy_mt1_trial.py
+++ b/app/services/strategy_mt1_trial.py
@@ -1,0 +1,341 @@
+"""Pure evaluator for the frozen MT-1 four-arm controlled trial (#2437).
+
+This module does not load prices, query the result ledger, or authorise an
+outcome read.  It accepts four already-built, after-cost monthly return books
+and applies the estimand frozen in
+``docs/proposals/ta/2026-08-15-mt1-volatility-managed-relative-strength-
+preregistration.md``.  Keeping the statistic here allows the implementation to
+be reviewed before the trial register and sealed-outcome gate are opened.
+
+The primary comparison is a difference in differences: volatility scaling's
+CER improvement on MT-1 less the same improvement on the S-8 negative control.
+All four arms use identical resampled month indices in every replication.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import math
+from dataclasses import dataclass
+from datetime import date
+from pathlib import Path
+from typing import Final, cast
+
+import numpy as np
+import numpy.typing as npt
+
+TRIAL_ID: Final = "mt1-capped-volatility-managed-relative-strength-v1"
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
Addressed the tuple-shape review finding in a69450f256d3b91b4419af5dfe8f3a7cf6645ad2: replaced the broad type ignore with cast(tuple[int, int, int, int], excluded). Ruff, formatting, Pyright, all 18 focused tests, and the full local pre-push gate are green on that exact commit.

PR #2761, opened 2026-08-15
feat(#2437): add costed target exposure curve

Luke-Bradford:
Closes part of #2437

## Outcome

Adds the opt-in holdings-level portfolio engine needed to construct MT-1 and its S-8 negative-control scaled books without using an `exposure × return` shortcut. Existing strategy-curve callers retain their prior full-exposure behavior.

## Mechanics frozen

- distinct identity: `capped_target_exposure_after_decision_close_v1`
- begins in cash until the caller supplies the first causal target
- accepts only finite targets in `[0, 1]`; leverage is impossible
- marks the declared decision bar before applying the new target, so only subsequent returns use it
- preserves source entries/exits at their stored open fills
- applies the last exposure target to source events between monthly decisions
- performs one closing uniform rebalance over post-fill holdings when a source event and target decision share a date
- charges each synthetic buy/sell with the existing leg-specific half-spread, sells before buys, and caps buys by cash
- refuses an untradeable frozen leg because uniform scaling would be unreachable
- reports the actual costed equity, invested capital, traded notional and rebalance costs through the existing `EquityCurve`

The preregistration now explicitly freezes that the volatility input ends at the prior completed month, the target is fixed before the first-session monthly bar, the sizing trade occurs after that bar's close mark, and the new target cannot consume that bar's outcome.

## Verification

- exact commit: 4326bd4a6b4c18f7070ab6960c2a2e88de8fa52e
- 57 equity-curve tests green
- 239 equity/backtest/statistics/overlay/MT-1 integration tests green
- full repository pre-push gate green on the exact commit: Ruff, formatting, full Pyright, 25 static invariants, 288 mutation-probe anchors, full fast tests, app/DB smoke and frontend integrity checks

## Still outside this PR

This does not derive the monthly exposure schedule, expose outcomes, add trial-register entries or declarations, persist strategy evidence, qualify a strategy, allocate capital, or authorise trading. The next book-builder layer must construct the fresh S-10/S-8 references, derive exposures from complete causal monthly histories, and produce the structural audit inputs.

### DIFF
--- a/app/services/equity_curve.py
+++ b/app/services/equity_curve.py
@@ -60,7 +60,7 @@ recomputed fill could be expressed. The caller resolves them from the ledger.
 from __future__ import annotations
 
 from array import array
-from collections.abc import Sequence
+from collections.abc import Mapping, Sequence
 from dataclasses import dataclass, field
 from datetime import date
 from typing import Final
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
Review disposition: accepted the alphabetical `__all__` nit. Also corrected an independently identified overstatement in the rule comment: source fills remain distinct, while event and exposure changes share one closing rebalance; synthetic trades still pay the holding-specific half-spread. Focused checks and the full pre-push gate are green on the new head; awaiting fresh hosted checks.

PR #2762, opened 2026-08-15
feat: build causal MT-1 four-arm monthly books

Luke-Bradford:
Closes part of #2437

## Outcome

Adds the pure four-arm book constructor between the frozen MT-1 exposure/evaluator components and the future paved research runner. It freshly walks both MT-1 and S-8 source books, derives each capped exposure only from its own completed prior-month history, builds holdings-level costed scaled curves, and returns the four monthly arms only after both structural audits pass.

## Pre-outcome session-axis clarification

The preregistration said trading days on the frozen panel calendar. Source inspection found that the union research-price axis contains stray weekend rows. Counting those rows changes `J` in the declared variance formula. Before trial registration or outcome access, this PR binds daily observations to the repository NYSE calendar and refuses any portfolio activity on a closed date. The union axis may retain an inert contaminant, but it cannot become a volatility observation or trade date.

## Fail-closed construction

- exact post-warm-up S-10 first-session decision clock; no omitted later decision
- complete consecutive NYSE session months only; missing sessions and partial terminal months refuse
- 120 usable-pair expanding history, prior completed month only
- decision-bar outcome cannot affect its own target
- binary-float equity ratios cross into deterministic decimal arithmetic through `Decimal(str(value))`
- exposure replay is derived internally; no caller-owned reconciliation flag
- annualised turnover is derived from costed traded notional, mean equity and elapsed years, with the frozen 600% ceiling
- both source clocks must match before any four-arm result is returned
- source and synthetic trades use the existing holdings engine and half-spread mechanics

## Verification

- exact commit: 1e0c7eb007e62f0db8cd52f1b3467723c3f1c450
- 95 focused equity/overlay/book/evaluator tests green
- direct 21-year synthetic construction feeds the frozen 10,000-resample four-arm evaluator with at least 120 common months
- full repository pre-push gate green: Ruff, formatting, full Pyright, 25 static invariants, 288 mutation-probe anchors, full fast tests, app/DB smoke, and frontend integrity checks

## Still outside this PR

This does not read the research corpus, create trial-register/declaration rows, open outcomes, persist evidence, qualify a strategy, allocate capital, or authorise broker execution. The next integration must construct the two source `LegBook`s from the exact manifest identities on the corrected metric axis and bind this constructor through the declaration/access gate.

### DIFF
--- a/app/services/strategy_mt1_books.py
+++ b/app/services/strategy_mt1_books.py
@@ -296,9 +296,12 @@ def _turnover_audit(
             f"{label} annualised turnover {annualised:.12g} exceeds the frozen 600% ceiling"
         )
     decision_dates = tuple(decision.decision_date for decision in decisions)
+    expected_decision_dates = tuple(
+        when for when in sorted(s10_rebalance_dates(dates)) if when >= decisions[0].decision_date
+    )
     return ScaledBookStructuralAudit(
         decision_dates=decision_dates,
-        expected_decision_dates=decision_dates,
+        expected_decision_dates=expected_decision_dates,
         annualised_turnover=annualised,
         traded_notional=traded_notional,
         exposure_reconciled=True,
--- a/app/services/strategy_mt1_books.py
+++ b/app/services/strategy_mt1_books.py
@@ -1,0 +1,398 @@
+"""Fresh, causal four-arm portfolio construction for MT-1 (#2437).
+
+The public constructor is deliberately four-arm: both source books are walked
+afresh and both structural audits pass before monthly outcome arms are exposed.
+Daily volatility observations use declared NYSE sessions.  Closed dates on the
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
Review disposition: fixed the independent-clock nit by re-deriving `expected_decision_dates` directly from `s10_rebalance_dates(dates)` inside the turnover audit. I did not add a float tolerance to the closed-day invariant: on an inert date this engine repeats the identical held-value sum and therefore produces bit-identical equity/invested values; a tolerance would admit small real activity. Added a prevention test carrying a live holding through an inserted Sunday `NaN` mark, which passes exact equality, plus a derived >600% turnover refusal test. Full pre-push is green on the amended head; awaiting fresh hosted checks.

PR #2764, opened 2026-08-15
Bind MT-1 identities and preregistration declarations

Luke-Bradford:
Refs #2437

## Outcome
- gives MT-1 and the S-8 negative control distinct dependency-complete strategy identities
- versions the calendar, equity engine, four-arm book, overlay, source strategies, evaluator, contract, universe, and cost model
- adds reproducible 36-decision-month / 157-week declarations with MT-1 capital-candidate and S-8 falsification-only purposes
- freezes the pair atomically and accepts only digest-identical retries
- deliberately keeps both identities out of the ordinary strategy manifest until the dedicated four-arm runner exists

## Safety boundary
This PR does not access outcomes and does not freeze production declarations. Run the dry-run and irreversible freeze only from merged code. No live-money authority is introduced.

## Verification
- repository pre-push gate green
- 8,606 fast tests passed
- focused declaration and atomic database tests passed
- dry-run produced two distinct coherent digests and matched the structural policy on refreshed origin/main

### DIFF
--- a/app/services/equity_curve.py
+++ b/app/services/equity_curve.py
@@ -59,10 +59,12 @@ recomputed fill could be expressed. The caller resolves them from the ledger.
 
 from __future__ import annotations
 
+import hashlib
 from array import array
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass, field
 from datetime import date
+from pathlib import Path
 from typing import Final
 
 import numpy as np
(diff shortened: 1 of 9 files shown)


Luke-Bradford:
Reviewed both optional points. Keeping the source-hash helpers local is deliberate: each version remains derivable from the module it identifies without adding a shared-helper dependency whose own movement would need separate treatment. The whole equity_curve.py hash is also intentionally conservative. MT-1 depends on both constructors plus their shared dataclasses and helpers; narrowing the hash creates an omission surface. A comment-only change can move a not-yet-frozen identity, but changed executable arithmetic cannot reuse old evidence, and the exact version/digest is printed before the irreversible freeze. No code change made for these non-blocking nits.

Issue #2781, opened 2026-08-20
s4_exit_levels_batch raises where its five siblings refuse: one bad bar aborts the whole outcome batch


Luke-Bradford:
## Summary

`s4_exit_levels_batch` **raises** where every other bracket strategy returns a typed refusal. An uncaught exception in an exit-level factory aborts the WHOLE outcome batch for one bad bar instead of recording one unresolved outcome — the exact failure mode `tests/test_2437_exit_level_refusal_surface.py` exists to prevent, and S-4 is excluded from that test.

Found by generalising #2780's batch-equivalence suite across all six bracket strategies.

## What differs

`s4_exit_levels_batch` has no `try`/`except`. It **raises `ValueError`** for:

* an out-of-range `signal_index`
* a non-finite or non-positive `entry_price`
* a missing or non-positive ATR at the signal bar

and returns `"unorderable_exit_levels"` only for a non-finite ATR value or an unorderable bracket.

`_s4_exit_levels` does not catch, so those propagate to the caller.

S-5, S-6, S-7, S-8 and S-9 all wrap their bracket in `except ValueError, IndexError:` and return `"unorderable_exit_levels"` for every one of those conditions. Measured on a shared fixture: bar 0 has no prior close and therefore no ATR — five strategies refuse it, S-4 raises.

## Why it has not bitten

`_exit_levels_for_entries` only requests indices where an entry actually fired, and the registry refuses to fire on an unevaluable bar, so signal indices have a valid ATR in practice. The divergence is latent rather than live.

⚠ It is one masked bar from being live. The quarantine `masked` arm exists precisely to remove bars, and the refusal-surface test's own docstring calls an uncaught exception here *"the one that matters"*.

## Why the test does not simply normalise it

The batch adapter's contract is that it does whatever its scalar oracle does. Forcing S-4 to refuse would have changed shipped behaviour to make a test pass, on a strategy whose adapter carries a deliberate runtime equivalence check (`_s4_exit_levels` re-runs the scalar and raises `RuntimeError` if the two disagree). So #2780's suite asserts **agreement with the oracle**, not uniformity across strategies, and this ticket records the divergence.

[...]

Issue #2780, opened 2026-08-20
price_levels._cluster is 77% of s5 runtime: np.average called 30M times; bounded fix is 11x and bit-identical below 8 elements


Luke-Bradford:
## Summary

`price_levels._cluster` is **77% of s5-support-bounce's runtime**, and `numpy.average` inside it is a third of the whole evaluation on its own. s5 is the single most expensive strategy in a full-set backtest — 7.5 hours of a 13.6-hour run, 24x the cost of s2 on identical inputs.

A bounded change makes the hot path up to 11x faster with **bit-identical output**.

## Profile

`cProfile` over `verify_2697_metric_axis_ab --limit 300 --strategy s5-support-bounce`, 465.4s total:

| function | cumtime | tottime | ncalls |
| --- | ---: | ---: | ---: |
| `evaluate_level_arms` | 464.9s | 0.3s | 1 |
| `s5_support_bounce._support_below` | 396.3s | 0.4s | 460,544 |
| `price_levels.at` | 360.6s | 10.9s | 460,544 |
| `price_levels._cluster` | **335.3s** | **87.3s** | 921,032 |
| `numpy…_function_base_impl.average` | **153.4s** | 47.1s | **30,056,519** |

⚠ This refutes the premise carried in the #2772 readiness doc and repeated since — that the *"strategy-by-strategy `research_price_daily` fetch and Decimal object reconstruction"* is the remaining performance boundary. Neither appears. Separately measured: `numeric` decodes at **823k rows/s**, so the entire 76M-row table reads in ~92 seconds, and `float8` is *slower* because the cast costs more server-side than it saves client-side. Two strategies reading the identical corpus differ 24x, which alone rules the read out as the dominant cost.

## Cause

```python
for cluster in clusters:
    if volumes is None:
        price = float(np.mean([prices[i] for i in cluster]))
    else:
        weights = np.array([max(volumes[i], 0.0) for i in cluster])
        values = np.array([prices[i] for i in cluster])
        price = float(np.average(values, weights=weights)) if weights.sum() > 0 else float(values.mean())
```

Per cluster this builds **two numpy arrays from Python list comprehensions** and calls **`np.average` with weights**. Clusters are pivots within a price tolerance of each other, so they are typically a handful of elements — sizes where numpy's per-call overhead dwarfs the arithmetic entirely.

## Measured fix, and why it is bounded at 8

A plain accumulation loop against the same inputs:

| cluster size | old | new | speedup | bit-identical |
[...]

Issue #2778, opened 2026-08-19
Scale gate's cohort time projection is noise: 3-member sample swings >2x in both directions, and a refusal aborts the whole invocation


Luke-Bradford:
## Summary

The synthetic-control scale gate projects a 1,000-member cohort by measuring **three members serially in the parent** and multiplying. Measured against the same workloads it projected, that estimate swings by more than 2x **in both directions**. A gate that admits or refuses on it therefore refuses valid work at random, and the failure is expensive: a refusal aborts the whole invocation.

This is distinct from #2775, which was about the *memory* term being mismeasured. Here the *time* projection is measured correctly and is simply not a reliable estimate.

## Measured, 2026-08-19, full S-1 survivorship-free corpus

| arm | projected | actual | error |
| --- | ---: | ---: | ---: |
| `s1/best_case/admitted` (run 1) | 862.7s | 960.6s | **11% under** |
| `s1/worst_case/admitted` (run 1) | 1245.1s | — | refused, aborted the run |
| `s1/worst_case/admitted` (run 2) | 611.3s | 528.8s | 16% over |
| `s1/best_case/admitted` (run 2) | 868.4s | 871.8s | 0.4% over |

The same arm — identical strategy, identical corpus, identical code — projected **1245.1s** on one invocation and **611.3s** on the next, and actually completed in **528.8s**. The projection that aborted a full-set run over-stated the work by **2.35x**.

## Mechanism

`_run_members` computes:

```python
projected_s = pilot_elapsed + (
    pilot_elapsed / pilot_size * remaining / max_workers * SYNTHETIC_CONTROL_PROJECTION_SAFETY_FACTOR
)
```

with `SYNTHETIC_CONTROL_SCALE_PILOT_MEMBERS = 3`. Two independent problems compound:

1. **The sample is three.** Member cost varies with the seeds drawn at those indices — `member_seed(index)` is a pure function of the index, so members 0..2 are a fixed but arbitrary draw, and their cost is not representative of the other 997. This is the dominant term in the variance above.
2. **The extrapolation changes the execution model.** The pilot measures members run **serially in the parent**, then divides by `max_workers` as though workers were free. They are not: measured worker throughput is ~1.7x slower per member than the parent's, because eight workers contend for memory bandwidth on a ten-core host. The 1.5x safety factor does not cover this — it was still 11% under on one arm.

Note the two errors point in opposite directions, which is why the result is noise rather than a correctable bias.

## Why it matters

[...]

PR #2782, opened 2026-08-20
perf(#2780): backtest evaluation 465.4s -> 68.6s on the s5 profile, bit-identical

### LINKED ISSUE #2779: Full-set backtest dies at s7: _pick_close treats non-redundant exit sources tying on a date as fatal, defeating _SOURCE_PRECEDENCE
## Summary

A full-set backtest died after **13.6 hours**, seven of ten strategies complete, persisting nothing:

```
ValueError: signal 7694: close sources disagree on 2011-06-16
  (level=318.874053244667719, signal_pair=326.9)
  — a disagreement about the same bar is a failure, not a tie-break
```

`app/services/position_builder.py:527`, in `_pick_close`, reached from `build_positions` while starting `s7-trend-pullback`.

## The guard defeats the precedence list that exists to answer this

`_pick_close` takes the earliest exit date, then refuses if the tied candidates carry different prices. Its docstring is explicit that this was **generalised**:

> ⚠ Generalised to every pair rather than special-cased to C2/C3. Two sources naming the same bar at different prices is the same defect whichever two they are.

The rule it generalises from is narrower, and the docstring quotes it: C2-vs-C3 for S-4, *"where the two are redundant by design"* — *"a disagreement is a failure, not a tie-break — it means the resolver and position construction disagree about the window"*.

**Redundant-by-design sources should agree. `level` and `signal_pair` are not redundant — they are different exit rules.** A level exit fires when a stop or target is touched, priced at the level (318.874053244667719 — note the precision, it is a computed level). A signal-pair exit fills on a paired exit signal, priced at the fill (326.9, a market price). Both can be true on the same bar and there is no reason they should be equal.

The strongest evidence that the generalisation is wrong is one line up in the same file:

```python
_SOURCE_PRECEDENCE: tuple[CloseSource, ...] = ("level", "ambiguous", "max_hold", "calendar", "signal_pair")
...
return min(tied, key=lambda candidate: _SOURCE_PRECEDENCE.index(candidate.source))
```

`level` already outranks `signal_pair`. **The precedence ordering only has meaning if tied sources can disagree** — if every same-bar tie had to carry an identical price, the ordering would be decorative. As written, the equality check fires first, so the precedence list is unreachable for exactly the cases it was written for.

## Why this is not a one-line fix

It decides **which price a trade exits at**, so it changes every backtest result that contains such a bar. That is a data-treatment decision and wants the spec rather than a patch: the governing §3.2 text should say whether non-redundant sources tying on a date resolve by precedence, and the redundant pairs (C2/C3) should keep the strict equality check that was actually specified.

⚠ This is the repo's own "question the MODEL, not the case" pattern: the tempting fix is to special-case this instrument/date. The model question is which source pairs are redundant-by-design and which are merely simultaneous.

## Cost, and why not to just re-run

- 13.6 hours of compute, **nothing persisted** — the Deflated Sharpe is computed across all measured trials, so no row can be written until every strategy is measured (#2778's second half).
- A re-run reaches the same point after ~11 hours and fails identically. **Do not restart until this is settled.**

⚠ Note this failure was not a gate. Two runs were previously lost to invented resource thresholds; this one was lost to a genuine defect 13.6 hours in. Both share one cause: an invocation that cannot persist partial progress turns any error into a total loss. That makes partial persistence the higher-value fix, not a nicety.

## Acceptance

- §3.2 states explicitly which close sources are redundant-by-design and which resolve by precedence
- a full-set invocation completes past `s7-trend-pullback`
- a probe covering a same-bar tie between non-redundant sources, and one covering a genuine C2/C3 disagreement still failing closed

Refs #2778
Refs #2772

Luke-Bradford:
Twelve commits of backtest-evaluation performance work, all bit-identical, plus the two
correctness fixes that surfaced while measuring. The s5/300 profile goes **465.4s → 68.6s**.

Closes #2779.
Refs #2780. Refs #2775. Refs #2697.

## Why

A full-set backtest was ~20 hours, which is long enough that strategy validation happens
about once a day. #2780 was filed against one symptom (`np.average` called 30M times inside
`price_levels._cluster`); the real shape turned out to be the same defect at three layers —
**whole-series derived data recomputed per item**:

* `atr_series` and `_volumes` rebuilt per SIGNAL rather than per series;
* `price_levels._cluster` allocating two NumPy arrays per cluster;
* `LevelScan.at` materialising a Python list for every cluster and discarding 94.1% of them.

## What changed

| commit | what |
| --- | --- |
| `0fdcf6f3` | #2775 cohort time bound derived from the run budget, not a second constant |
| `8ee87974` | **#2779** close-source equality scoped to the redundant pair — the full-set run died at s7 |
| `2072193b` | #2697 A/B harness forwards `scale_budget` through its cohort monkeypatch; acceptance step 3 had been unrunnable |
| `b24d5e9e` | `_cluster` vectorised — argsort + diff + cumsum + bincount |
| `f517c35c` | bind `ordered_weights` unconditionally so the fallback narrows |
| `27996560` | #2779 position-builder probes re-anchored onto the scoped check |
| `b7f671b8` `aec0e61f` `4959c1ab` | S-5, S-6, then S-7/S-8/S-9 given the batch exit factory S-4 has had since #2623; coverage guard so a seventh bracket strategy cannot land without one |
| `7f609d3c` | bind the oracle to a local before closing over it |
| `b2626d75` | **`at` filters before materialising** — the largest single win |
| `67ed9dbe` | prevention-log entry |
| **`d0851b27`** | **#2779 close ties broken by WHEN they executed, not source alone** — the Codex P1 |

### The last one, in detail

`at` asked each cluster two questions — `touches` and `last_touch` — and built a Python list
of ints to answer them. `touches` is the cluster's size and `last_touch` is
`np.maximum.reduceat` over the price-ordered indices; both are already computed by the
vectorised pass, and `PriceLevel` stores neither the list nor anything derived from its order
except the volume share. Measured over one 300-series profile: **30,756,116 clusters built,
1,822,327 survived** `MIN_TOUCHES` and `MAX_TOUCH_AGE_BARS`.

`_segment` now returns the arrays, `at` filters on them, and only a surviving cluster takes
per-cluster work. `_cluster` keeps the materialising form — off the hot path, and it is what
the tests pin against a hand-transcribed oracle.

## A P1 that Codex caught, and no gate did

`codex exec review` found that `8ee87974` — this branch's #2779 fix — stopped an
ordinary `level`-vs-`signal_pair` tie being fatal (right) and then resolved it the
wrong way (wrong). It preferred `level`, reasoning that an intraday touch "closed
the position before any close-based exit fill could".

**That premise is false.** `signal_ledger.resolve_fills` states its rule outright —
*"the fill price is that bar's OPEN. There is no other path"* — and does not
distinguish entry fills from exit fills. A signal-pair exit executes at the FIRST
tick of its bar, before any intraday path exists. #2779's own reported case proves
it: `signal_pair=326.9` (that bar's open) against `level=318.874053244667719` (a
stop). Booking the stop records a ~2.5% loss on a position that had already closed.
The previous behaviour at least crashed; this branch would have made it silent.

Fixed in `d0851b27`. A reordered tuple cannot do it — three constraints form a
cycle: `level` must precede `max_hold` (§3.2 names `outcome_resolver.py:470` as the
expiry reference, and the two are price-equal by the redundancy check), `max_hold`
precedes `signal_pair` today, and `signal_pair` must precede an intraday `level`.
So the key is **two-tier**, and the tiers are different kinds of thing: the first
asks *when in the bar the close executed* (a fact about the market), the second is
the label preference among closes that executed at the same moment (a convention).

⚠ `level` sits on both sides of that line, so the test is not a source lookup: an
`expired` outcome exits at the next bar's open and is exactly the case carrying
`redundant_with_max_hold`; a `tp`/`sl` touch is intraday. A gap-through touch also
resolves at the open and is not detectable there — but it prices AT that open, so
it can differ only in the label, which makes sorting it late price-neutral.

### Blast radius, measured

The only hybrid `signal_pair` + `level_based` regime is S-7, so it is the only
strategy this can move. Both arms run unprofiled at `67ed9dbe` and `d0851b27`:

| S-7 `current`, masked arm | superseded | fixed |
| --- | ---: | ---: |
| CAGR % | -52.524 | **-50.848** |
| Sharpe | -2.633 | **-2.537** |
| Sortino | -3.259 | -3.168 |
| exposure time % | 4.0129 | 4.2289 |
| turnover | 4.1289 | 4.3599 |

Consistent across all four (ambiguity × quarantine) arms: **CAGR +1.6 to +1.7
points, Sharpe +0.09**. S-7 is a long strategy, so booking a stop below the open
understated every affected exit; removing the phantom loss moves returns up.
Exposure and turnover rise for the second half of the same fix — an `ambiguous`
close carries no price and DISCARDS the position, and an open-priced fill now
correctly beats it, so trades that were being thrown away are booked at the price
that actually happened.

⚠ **S-7 remains deeply unprofitable (CAGR ≈ -51%, Sharpe ≈ -2.5).** This makes the
measurement honest, not the strategy good.

⚠ This is the one part of the branch that is deliberately NOT bit-identical.
Everything under "Accuracy" below concerns the performance work.

## Measured

`cProfile` over `verify_2697_metric_axis_ab --limit 300 --strategy s5-support-bounce`:

| | wall | calls |
| --- | ---: | ---: |
| before | 465.4s | 2,215,369,726 |
| after clustering | 212.5s | 852,546,268 |
| after batching | 136.3s | 296,129,017 |
| re-anchored at `7f609d3c` | 128.8s | 296,127,986 |
| **after `b2626d75`** | **68.6s** | **165,079,331** |

⚠ **Quote the last leg as x1.88 and not a single combined multiple.** 128.8s → 68.6s is the
only same-machine, same-day pair; 465.4s came off a different session's machine state.

| | before | after |
| --- | ---: | ---: |
| `price_levels.at` cum | 101.57s | 45.52s |
| `_cluster` → `_segment` cum | 76.13s | 20.02s |
| `builtins.max` calls | 32,044,242 | 1,288,126 |
| `numpy.asarray` calls | 922,186 | 4,740 |
| `np.average` | 17.99s / 3,050,672 | 5.35s / 863,888 |
| `nansum` | 8.39s / 1,822,327 | 7.03s / **1,822,327** |

⚠ `nansum`'s call count is IDENTICAL either side. It runs once per surviving level, so that
equality is independent evidence that the same level set survives.

## Accuracy — the hard constraint

A level price feeds a threshold comparison, so a last-bit change flips which trades exist.
Everything below is `==` on floats, never `approx`.

| evidence | result |
| --- | --- |
| **End-to-end**: A/B harness run unprofiled at `7f609d3c` and at `b2626d75` | 4 (ambiguity × quarantine) records, **64 field comparisons, 0 differences**, `acceptance_summary.complete = true` both sides |
| **Full population**: `verify_2437_level_scan --equivalence --limit 200`, real series incl. quarantine-masked and the `volumes=None` arm | **319,668 comparisons, 0 mismatches**, 873,766 levels returned, 2,853,574 clusters rejected as thin + 584,722 as stale |
| `_cluster` vs a hand-transcribed oracle | 1,500 randomised cases + 4,000 ad hoc, zero price or membership mismatches |
| `at` vs a transcribed `_reference_at` | 25 randomised series walked at every bar, incl. zero-weight and NaN volumes |
| six batch exit factories vs their scalar oracles | `tests/test_2780_s5_exit_levels_batch.py`, 31 tests |

⚠⚠ NumPy switches from sequential to **pairwise** summation at 8 elements, so `bincount`
parts from `np.average` in the last bits at that size (7.1e-15 measured). Clusters of 8+ fall
back to `np.average` on exactly the old inputs, and a test verifies the boundary empirically
rather than trusting the comment.

## The harness that could not fail

`verify_2437_level_scan --equivalence` proved the #2437 hoist by comparing `scan.at` with
`levels_at`. That was a real check when written. #2437's own hoist then made `levels_at`
build a `LevelScan` and call `at` — and the module docstring recorded the collapse
approvingly: *"exactly one code path and no oracle to keep in agreement"*. From that commit
on, both sides of the comparison were the same function, and the arm would have reported a
clean sweep over any rewrite at all.

Nobody had touched the harness. **A correct refactor deleted an oracle and left a passing
harness behind.** It now compares against a transcribed `_reference_at`, and exits non-zero
if no level survives anywhere or if either filter never rejects anything. Logged in
`docs/review-prevention-log.md` as a distinct member of the tautology family — the existing
entries are authored couplings or mis-aimed probes; this one is neither.

## Security model

Not touched. No new input surface, no authz decision, no SQL, no user-controlled data — the
diff is pure in-process numerics over an existing read path, plus two test/verification
scripts.

## Strategy identity

⚠ `LEVEL_RULE_VERSION` is a sha256 of the whole `app/services/price_levels.py` and S-5/S-6
carry it in their `INPUT_RULE_SETS`, so their identity moves. That is unavoidable for any
edit to that file and the branch lands **one** identity move however many times it is
touched (it moved twice here, at `b24d5e9e` and `b2626d75`). The levels themselves are
bit-identical throughout, so nothing about the strategies' behaviour changes — only the hash
that declares which source produced them.

## Known-open, deliberately not in scope

* **#2780 is not closed.** `_segment` still re-sorts the whole pivot prefix on every bar
  (20.02s over 921,032 calls). The sort is tolerance-independent — only the cluster cuts move
  with per-bar ATR — so incremental price-ordering is next, and is a materially safer claim
  than the incremental *clustering* a prior review rejected.
* **The quarantine-pass fusion was measured and declined.** A prior handoff put psycopg
  `wait` at 13% and named it the next target. On the s5 profile that its own verification
  command uses, `wait` is 5.1% across all 273 queries — mostly `load_corpus`'s few big reads
  — and `load_arms` is 2.8%, so fusing recovers ~1.4%, for a signature change across six call
  sites including the A/B harness whose `len(captured) == len(arms)` assertion assumes one
  quarantine arm per call. Recorded on #2780 so it is not retried on the stale figure.
* **#2781** (`s4_exit_levels_batch` raises where its five siblings refuse) and **#2778**
  (cohort projection noise) remain open and untouched.

## Verification

```bash
uv run pytest tests/test_price_level_scan.py tests/test_2780_s5_exit_levels_batch.py -q
uv run python -m scripts.verify_2437_level_scan --equivalence --limit 200
uv run python -m cProfile -o /tmp/s5.prof -m scripts.verify_2697_metric_axis_ab \
    --limit 300 --strategy s5-support-bounce
```

⚠ `verify_2697_metric_axis_ab` refuses a dirty worktree, so its evidence has one exact
source head. ⚠ Under `python -m cProfile -m` it dies at the very end on
`_pickle.PicklingError` — cProfile rebinds `__main__`, so the multiprocessing cohort stage
cannot spawn. The evaluation-phase profile is valid; run without the profiler to see the
script complete.

Full pre-push gate green at `67ed9dbe`: ruff, ruff-format, pyright (whole repo, tests
included), every chokepoint lint, fast tier, smoke.


Luke-Bradford:
## Ready to merge — review satisfied, gate green, merge blocked on a local permission only

**State:** head `d0851b27`, `MERGEABLE` / `CLEAN`, base `fix/2697-metric-axis-integrity` unmoved
(0 commits behind), worktree clean.

### Review status

⚠ This PR will never report CI or a review bot. Every workflow gates on `branches: [main]`,
which matches a PR's **base**, so a PR stacked on a non-`main` base gets neither.
`gh pr checks 2782` reports "no checks reported" and that is the **terminal** state, not a
pending one. Per the repo's review decision tree, `codex exec review` is the reviewer here.

* `codex exec review --base origin/fix/2697-metric-axis-integrity`, first pass → **one P1**,
  against `8ee87974`. Fixed in `d0851b27` (see the PR body).
* Same command, second pass on the fixed branch → **no findings**. Its closing note: the
  changed behaviour is covered by focused equivalence and regression tests, and the failures
  it observed were its sandbox lacking PostgreSQL and shared memory, not defects.
* Full pre-push gate green on `d0851b27`: ruff, ruff-format, pyright (whole repo, tests
  included), all 25 chokepoint lints, probe-anchor validation, CI/pre-push parity, fast tier,
  smoke.
* `scripts/probe_2240_position_builder.py`: **33/33 caught**, restored suite passes.

### Remaining command

```bash
gh pr merge 2782 --merge --delete-branch
```

Then, per this repo's linking rules — close **only** what the PR resolves:

```bash
gh issue close 2779   # Closes; the tie-break defect and its correct resolution both land here
```

⚠ Leave **#2780**, **#2775** and **#2697** OPEN. They are `Refs`, not `Closes`:
* **#2780** — clustering, batching and the filter hoist are done; `_segment` still re-sorts the
  whole pivot prefix on every bar (20.02s over 921,032 calls) and that is the next win.
* **#2775** — its acceptance was met on the base via #2773/#2776/#2777; this branch adds one
  refinement commit (`0fdcf6f3`).
* **#2697** — owned by the base branch.

### Why the merge is not already done

`gh pr merge` was refused by the local auto-mode permission classifier. That is an environment
restriction on this session, not a failed check and not an unresolved review finding — every
gate this repo actually defines is green on `d0851b27`.

Issue #2785, opened 2026-08-20
fix(ownership): _select_control_group_rep clause 4 declines swaps on a stranded row that cannot release


Luke-Bradford:
Split out of #2230 rather than folded into it — the fix there is scoped to the FOLD gate
and deliberately leaves this call site alone.

## Problem

`_releases_other_rows` returns True whenever a holder's identity holds **any** row outside
the cluster. #2230 established that this is strictly wider than the hazard it is named
for: a stranded row is only a *release* if it changes what `_reconcile_owner_once` decides
about the identity, and that function branches on `present & _INSIDER_SOURCES`. An
identity retaining any `form4` / `form3` / `def14a` row outside the cluster is still a
Section-16 person after the demotion — same wedge, 13F still a `dropped_source` (#1640) —
so nothing is released.

The fold gate now asks the narrower `_releases_into_another_wedge`
(`app/services/ownership_rollup.py`). **Clause 4 of `_select_control_group_rep` still asks
the wide one**, so it declines rep swaps for the same non-hazard.

## Why it was not changed in #2230

Not because the argument differs — it does not — but because the arithmetic differs. #2385
measured clause 4 under the wide predicate (it "declines 24 of 75 otherwise-eligible
swaps") and a rep change moves wedge totals substantially in both directions: `USBC`
737,776,188 → 379,961,188, `ACDC` 225,951,558 → 298,774,575. Changing the fold gate and
the rep gate in one A/B would have left two variables in one arm.

## First step — measure, do not fix

Not measured in the #2230 run. `scripts/audit_2230_release_hazard.py` already has the
re-entrancy flag that separates the two call sites (`in_rep_selector`); invert it to record
the rep-selector's calls instead of ignoring them, and report:

- how many swaps clause 4 declines today;
- of those, how many the demoted incumbent's stranded rows are **all** `_INSIDER_SOURCES`,
  i.e. would flip under the narrow predicate;
- the per-instrument pie-total delta of the flips, gain side inspected separately.

Then decide on the numbers. ⚠ A rep swap is not symmetric with a fold: the fold removes
members and can only shrink the identity's contribution, whereas a swap changes **which**
identity survives into owner-once, so the non-inflation argument in
`_releases_into_another_wedge`'s docstring does NOT transfer unexamined.

Refs #2230. Refs #2385. Refs #2226.

Luke-Bradford:
[...]

Issue #2385, opened 2026-08-07
fix(ownership): insider control-group fold labels the block with the highest-CIK member, not the Rule 16a-1(a)(2) direct holder


Luke-Bradford:
Noticed while writing the `## OPERATOR VERIFY` block for #2230 (merged `9fc0062b`). **Display-only — the wedge total is unaffected**, so this is not a regression of that fix.

## What

`_control_group_rep_key` (`app/services/ownership_rollup.py:2511`) selects the cluster representative on `(winning_source in _INSIDER_GROUP_SOURCES, shares, filer_cik or "", winning_accession)`, all descending. Inside a deemed chain every member reports the **same** `shares` by construction, and all are insider-source, so the first two components tie for the whole cluster and the rep is decided by **highest CIK** — i.e. by which member registered with the SEC most recently.

The key does not read `ownership_nature`, even though the gate immediately above it does: `_is_deemed_chain` (`:2489`) admits a cluster only when `n_direct <= _DEEMED_CHAIN_MAX_DIRECT` (1) and `n_indirect >= _DEEMED_CHAIN_MIN_INDIRECT` (2). So wherever a `direct` member exists, the code already knows which holder actually holds the block, and then discards that knowledge when choosing the label.

## Measured (dev DB, on merged `9fc0062b`)

```sql
SELECT oc.holder_cik, oc.holder_name, oc.ownership_nature
  FROM ownership_insiders_current oc JOIN instruments i USING (instrument_id)
 WHERE i.symbol='XFOR' AND oc.shares=2000000
 ORDER BY oc.ownership_nature, oc.holder_cik DESC;
```

13 rows — **one `direct`**, `New Enterprise Associates 16, L.P.` (`0001694560`); **twelve `indirect`**, the NEA managing members. `get_ownership_rollup` on `XFOR` emits:

> `Control group: 13 related CIKs reporting 2000000.0000 (deemed ownership) counted once under Chang Carmen. Folded: … New Enterprise Associates 16, L.P. (0001694560) …`

`Chang Carmen` (`0001746342`, `indirect`) wins purely on CIK order. The operator sees a 2,000,000-share NEA fund position attributed to an individual, with the fund listed as folded *under* her.

## Why the direct holder is the right rep

[...]

PR #2787, opened 2026-08-20
fix(#2785): measure the rep-swap release gate — narrowing clause 4 is REFUTED

Luke-Bradford:
## What this is

#2785 asked whether `_select_control_group_rep` clause 4 should follow #2230's narrowing —
from `_releases_other_rows` ("does the incumbent hold any row outside the cluster") to
`_releases_into_another_wedge` ("would demoting it change which pie wedge it lands in").
It asked for the measurement first, and explicitly warned that a swap is not symmetric
with a fold.

**The measurement refutes the change.** Clause 4 stays wide. The service diff is
documentation only — no behaviour change, no migration, no stored value moves.

That is the deliverable: the census the ticket asked for, the A/B that killed the fix,
the three discriminators that also died, and the evidence written where the next session
will hit it before re-attempting this.

## Census — `scripts/audit_2785_rep_swap_gate.py` (new)

Runs the REAL selector three times per cluster with the clause-4 predicate rebound —
`ungated` (forced False, so "clause 4 was reached" is observable), `wide` (production),
`narrow` (#2230's rule) — and reads the answer off which `Holder` comes back. Re-spelling
the selector's clauses in the harness would reintroduce the drift PR #2384's review
WARNING calls out; running the real function under each arm cannot drift.

Full population, 3,108 instruments, 0 harness errors:

| | shard 0 | shard 1 | shard 2 | total |
|---|---:|---:|---:|---:|
| clusters reaching the rep selector | 1,895 | 1,773 | 1,615 | **5,283** |
| clause 4 reached | 77 | 68 | 58 | **203** |
| clause 4 declines today (wide) | 34 | 28 | 19 | **81** |
| incumbent keeps an insider row outside the cluster | 30 | 25 | 17 | **72** |
| would flip under the narrow predicate | 30 | 25 | 17 | **72** |

All three harness cross-checks clean on every shard. **On the census alone the change
looks obviously right** — 88.9% of refusals are for a hazard that cannot occur, the
identical finding #2230 made for the fold gate. The 9 that stay refused strand only
`13d`/`13g`/`13f`, e.g. `GPGI`'s `Resolute Compo Holdings LLC` stranding a **51,683,967**
share 13D against a block of **879,963**.

## The A/B is what refuted it

Paired full-population A/B, control = a separate worktree at `origin/main` `8fc1b4ad`,
treatment = that SHA plus the one predicate. Both arms 3,108 instruments, 0 harness
errors, 0 only-in-either.

```
net                     -385,131,434.5008 shares
  26 instruments SHRINK -468,017,523
  17 instruments GROW    +82,886,088     <-- the finding
  3,065 unchanged
collapse count changed on 0 instruments
UNEXPLAINED gains (not a fold rep): 0
```

The net looks like a clean win and the net is the trap — `full-population-ab.md` says
inspect the gain side, and here the gain side **is** the verdict.

**Mechanism.** While the incumbent is rep, its stranded rows share an identity key with
the block row, so `_reconcile_owner_once` groups them — additively for pooled Section-16
forms (#1941), by MAX across competing beneficial restatements. Demoting the incumbent
breaks that grouping. Two worked cases, as the two ends and **not** as a population
taxonomy:

- `WBD` — exactly pie-neutral. `Newhouse Steven O` 184,070,739 becomes the partnership's
  184,023,290 + his own 47,449. Same total, and the block finally sits on the entity that
  holds it. This is the case that makes the narrowing tempting.
- `AIRS` — **+14,038,819**, exactly the block value.

## Three discriminators tried, three falsified — this is what settles it

| candidate key | GROW | SHRINK | NEUTRAL |
|---|---|---|---|
| `max(stranded) >= block value` | 11/17 | — | 14/55 non-growers fire |
| instrument carries >1 collapse (the `AIRS` shape) | 14/17 | 25/29 | 13/26 |
| incumbent is the kept rep of a collapse | 17/17 | 29/29 | 25/26 — **tautology** |

`TTRX` grows by 15,416,260 on an instrument carrying exactly **one** collapse, which
refutes the `AIRS` story as the general explanation. The third key is true by construction
in the control arm and looked like a strong signal until that was noticed.

Route stratification does not rescue it either — `value_bucket` 20 neutral / 16 shrink /
7 grow, `same_accession` 6 / 13 / 10. Both routes grow, so the narrowing cannot be scoped
to one. (Cluster-level attribution of an instrument-level delta.)

Repeated failed keys is this repo's own signal to question the MODEL rather than guess
another. "Neutral-or-better" is a statement about the owner-once **grouping**, and no
predicate over one member's stranded rows can decide it.

⚠ **This is a refusal to ship, not a closure of the question.** The narrowing does improve
representative correctness. The blocker is downstream identity/component arithmetic, so
the follow-up is to model the owner-once grouping across connected clusters — not to keep
the arbitrary representative forever.

## Confounder checked, not assumed

The jobs daemon writes during a sequential A/B, and it did: `ownership_institutions_current`
`refreshed_at` moved 17:46:47Z → 18:00:01Z between the arms, share sum −2,018, row count
unchanged. So the control shard was re-run afterwards and diffed against itself:

```
control shard 0 re-run: 1,036 vs 1,036 compared
instruments whose pie_total differs between the two control runs: 2
    CEG  299,188,817 -> 299,188,200  (-617)
    BX   784,040,409.1613 -> 784,040,402.1613  (-7)
identity-set differences: 0
```

624 shares on two instruments, **neither among the 43 the change moved**. The A/B is
uncontaminated.

## Second opinion

Codex checkpoint 2 returned one P2 — the census printed its cross-check results and
returned 0, so a caller could not tell "clean population" from "harness measured
nothing". Correct, and an inconsistency in this branch's own work, since the probe was
already hardened that way. Fixed in `6399951f`; verified both directions (`--limit 200`
exits 0 with 4 clusters reaching clause 4, `--limit 3` exits 1 with "clause 4 never
reached").

A Codex pass on the **framing** also caught a generalisation from two cases — the first
draft asserted the `AIRS` topology as the mechanism. Scoring it across all 72 flips
falsified it, and the claim was weakened to what was measured (`793f93a3`). That is
#2282-2c's "a descriptive claim is a measurement too" landing on a causal claim rather
than a prose quantifier.

Clauses 10 and 11 **do not apply**: no parser, ETL, schema or endpoint change, nothing to
backfill, and no stored value moves. `_select_control_group_rep` runs at READ time and is
byte-identical to `origin/main`. **Do not run `POST /jobs/sec_rebuild/run` for this.** No
`## OPERATOR VERIFY` block, because there is nothing for the operator to verify.

- **Clause 8 — golden panel.** Unchanged by construction and confirmed by the A/B: `AAPL`,
  `MSFT`, `JPM`, `HD`, `GME` are all in the 3,065 instruments whose pie total is identical
  across both arms.
- **Clause 9 — cross-source, SEC EDGAR direct.** The `WBD` reading was checked against the
  filing structure rather than only our store: the block belongs to
  `ADVANCE/NEWHOUSE PROGRAMMING PARTNERSHIP` (CIK 0001455382) with `Newhouse Steven O`
  (CIK 0001913168) restating it indirectly — the Rule 16a-1(a)(2) shape clause 1-3 read
  off Table I, and the reason the swap is the correct attribution even though it is not
  safe to take yet.
- **Clause 12 — SHAs.** Census + revert + docs `ff43124a`; claim corrected to the measured
  width `793f93a3`; Codex P2 `6399951f`.

## Tests

`test_swap_is_declined_even_when_the_stranded_row_is_an_insider_row` pins the 72-flip
class as still-declined and names the evidence, so the next attempt at this change fails
a test that points at the measurement instead of silently shipping. Revert-probed:
injected the narrow predicate (anchor asserted unique first), test **CAUGHT** it, restored
and verified passing.

Gates each run separately and unpiped: `ruff check` · `ruff format --check` · `pyright`
(0 errors) · `pytest -m "not db"` · `pytest tests/smoke` · `pytest -m db` on
`test_ownership_rollup.py`, `test_ownership_rollup_record_holder_db.py`,
`test_insider_control_group_collapse.py`, `test_ownership_owner_reconcile.py` — all exit 0.

## Prevention log

Two entries: the gate-narrowing lesson (a fold removes, a swap re-parents, so a shared
predicate cannot be aligned without its own A/B), and a measurement-hygiene one — zsh does
not word-split an unquoted parameter, so `probe $SYMS` resolved nothing, exited 0 and
diffed clean on the run whose whole purpose was to show movement; plus macOS `pgrep` having
no `-c` flag, which made a wait-loop report "done" instantly.

Refs #2785. Refs #2230. Refs #2385. Refs #2226.


Luke-Bradford:
## Review resolution — `d52144b2`

Both NITPICKs actioned. Neither is rebutted.

**NITPICK 1 — `recording_rep` mutates module-level `orl._releases_other_rows` in-place.**
`FIXED d52144b2`. Documented at the mutation: it has to be a global (the selector resolves
the predicate by global lookup at call time, which is exactly what makes a three-arm
in-process A/B possible without re-spelling its clauses), it makes the script
single-threaded-only, and the way to parallelise is separate processes — which is what
`--shard` already does. Each arm already restored in a `finally`, so an exception cannot
leave production code pointing at `_never`; that is now stated rather than implied.

**NITPICK 2 — shared `route["name"]` dict; "worth confirming, unverified without seeing
the call graph".** `FIXED d52144b2`, and confirmed two ways rather than asserted:

- **Call graph.** `_reconcile_same_accession_groups` and `_reconcile_insider_control_groups`
  are invoked sequentially from a single site (`app/services/ownership_rollup.py:4880` and
  `:4883`) and neither calls the other, so they do not nest today. The save/restore of
  `prev` is stack discipline and stays correct if that ever changes.
- **Empirically.** `route` initialises to `"unknown"`, and across all **5,283** clusters in
  the census the only labels emitted were `same_accession` (4,291) and `value_bucket`
  (992) — zero `unknown`. So no selector call escaped a wrapper.

⚠ Chasing that one down surfaced a real imprecision in my own labelling, so thank you for
it: **`value_bucket` names the CALLER, not the admission route.** The #1652 value-proxy
route and the #2230 deemed-chain tier both live inside
`_reconcile_insider_control_groups`, and only the latter has a fold-release gate — so the
label does not separate the two things a reader would most want separated. Corrected in
the script's `wrap_route` docstring and in clause 4's docstring, which had published the
table as "by admission route" and now says "by calling pass" with the caveat attached.
It does not change the finding: both callers grow, so the narrowing cannot be scoped to
one of them either way.

Gates re-run separately and unpiped after the change: `ruff check` · `ruff format --check`
· `pyright` (0 errors) · `pytest -m "not db"` · `pytest tests/smoke` — all exit 0. Census
smoke re-run (`--limit 200`) exits 0.

Luke-Bradford:
## Review resolution — `8ab53168`

**NITPICK — "repeated `orl._releases_other_rows = real_wide` inline restores … redundant
but harmless".** `FIXED 8ab53168` — kept, with the reason now pinned in the file, because
they are load-bearing rather than harmless duplication and a future reader would otherwise
be right to delete them.

The module-level `finally` in `main` runs **once, at the end of the run**. These run **once
per cluster**, and both flows need them:

- **normal** — the next cluster's production arm (`wide_rep`, called before any rebinding)
  must see `real_wide`;
- **exceptional** — the population loop catches `Exception` per instrument and *continues*.
  Drop the inner `finally` and a raise inside the ungated arm leaves `_never` bound for
  every remaining instrument. `wide_rep` then silently becomes a second ungated arm, the
  two agree by construction, and the clause-4 decline count collapses toward zero — on a
  run that exits 0 with no error. That is the same silent-clean-measurement failure this
  module's docstring is about, and the one the #2386 census already burned on once.

Comment only, no behaviour change. Gates re-run separately and unpiped: `ruff check` ·
`ruff format --check` · `pyright` (0 errors) · `pytest -m "not db"` · `pytest tests/smoke`
— all exit 0.

Issue #2790, opened 2026-08-20
bug(insider): 20 ownership_insiders_current rows carry a period_end in the future (max 2047-05-24) — permanently un-supersedable


Luke-Bradford:
Noticed while measuring insider-row ages for #2230 / #2788. Small, but it interacts badly with any period-keyed release rule, which is exactly what #2788 will build.

## Measured (full population, 2026-08-20)

```sql
SELECT count(*), count(DISTINCT instrument_id), min(period_end), max(period_end)
  FROM ownership_insiders_current WHERE period_end > current_date;
-- 20 | 14 | 2027-02-17 | 2047-05-24
```

Worst five:

| instrument_id | holder | source | period_end | shares |
| ---: | --- | --- | --- | ---: |
| 10510 | `CIK:0001228922` | form4 | **2047-05-24** | 1,178,231.88 |
| 8940 | `CIK:0001807988` | form4 | 2033-12-11 | 14,033 |
| 9959 | `CIK:0001559864` | form4 | 2033-11-18 | 249,887 |
| 6177 | `CIK:0001207854` | form4 | 2031-11-18 | 159,029 |
| 6661 | `CIK:0001497463` | form4 | 2031-01-31 | 24,204 |

These are issuer typos in `<periodOfReport>` (2047 for 2017 or 2027, etc.), faithfully ingested.

## Why it is worth a ticket rather than a shrug

1. **A future-dated row can never be superseded** by anything a period-keyed rule can compare it against — it is permanently the newest row for its holder. Any release/expiry mechanism (#2788) inherits 20 rows it structurally cannot retire.
2. **It poisons an as-of anchor.** `OwnershipSlice.as_of_max` is computed from holder as-of dates, so one 2047 row makes every other holder on that instrument look stale by 20 years to anything anchored on it. The #2230 census hit exactly one such instrument in its 664-strong cohort (`RMCF`, `as_of_max = 2027-07-27`) — immaterial there, and only because the cohort was small.
3. There is precedent for the guard: `financial_facts_raw` already has an XBRL period sanity window (#1218, sec-edgar skill §7.16). The insider path has none.

## Scope

A sanity window at insider ingest (reject or clamp `period_end > filed_at`, which is the tighter and more defensible bound — a Form 4 cannot report a period after the day it was filed), plus a one-off correction for the 20 stored rows. Both need the usual full-population A/B; the population is small enough to enumerate and eyeball in full.

Refs #2230. Refs #2788.

[...]

Issue #2789, opened 2026-08-20
bug(tooling): 9 of 10 scripts render get_ownership_rollup outside snapshot_read, violating its documented contract


Luke-Bradford:
Found while writing the #2230 residual census; Codex checkpoint 2 flagged it on the new script and the same defect turned out to be in every existing sibling.

## The contract

`get_ownership_rollup`'s docstring ([`app/services/ownership_rollup.py:4776`](https://github.com/Luke-Bradford/eBull/blob/main/app/services/ownership_rollup.py#L4776)):

> The caller MUST already be inside :func:`app.db.snapshot.snapshot_read` … The function does NOT open its own transaction.

It issues many separate reads (denominator, each `ownership_*_current`, notices, families). Under the connection's default READ COMMITTED, a concurrent ownership refresh can hand one render a denominator from one commit and a wedge from another. **Nothing raises** — the row is simply wrong, which is the failure mode that looks like a clean run.

## Measured, this run

```
$ grep -l "get_ownership_rollup" scripts/*.py | while read f; do
    printf "%-45s snapshot_read=%s\n" "$f" "$(grep -c snapshot_read "$f")"; done
scripts/ab_2230_deemed_chain.py                snapshot_read=0
scripts/ab_2232_partial_class_denominator.py   snapshot_read=0
scripts/audit_2230_insider_oversubscription.py snapshot_read=4   <- fixed on the #2230 branch
scripts/audit_2408_nature_record_holder.py     snapshot_read=0
scripts/audit_2230_release_hazard.py           snapshot_read=0
scripts/audit_2385_control_group_rep.py        snapshot_read=0
scripts/audit_2386_deemed_chain_gate.py        snapshot_read=0
scripts/audit_2785_rep_swap_gate.py            snapshot_read=0
scripts/dq_audit.py                            snapshot_read=0
scripts/probe_2785_wedge_detail.py             snapshot_read=0
```

**9 of the 10 scripts that render the rollup violate the contract.** They include the full-population A/B harnesses that decided #2230, #2385, #2386 and #2785 — and the jobs daemon writes ownership continuously on dev, so the exposure is live rather than theoretical.

[...]

Issue #2788, opened 2026-08-20
fix(ownership): insider _current never releases a position — #2229 shipped M1 for 13F only, and it is the binding axis of the #2226 tracker


Luke-Bradford:
Split out of #2230 after the residual was re-measured. #2226 named this mechanism (**M1**, *"no `_current` table can ever release a position"*) across all three channels; **#2229 shipped it for 13F only** and the insider half was never re-filed.

## The gap

`refresh_*_current` deletes a `_current` row only `WHEN NOT MATCHED BY SOURCE`, so "latest observation per holder" means *forever* unless something supersedes it. #2229 built that something for institutions (`superseded_by_later_13f_hr`). Grep the ticket in the read path and every hit is institutional — `app/services/ownership_rollup.py:158`, `:991`, `:1149`, `:5005`. There is no insider equivalent.

## Measured (full population, 2026-08-20, this run)

```sql
SELECT source,
       count(*) FILTER (WHERE period_end >= current_date - interval '2 years') AS within_2y,
       count(*) FILTER (WHERE period_end <  current_date - interval '2 years') AS older
  FROM ownership_insiders_current GROUP BY 1;
-- form4  63,080 / 87,289
-- form3   7,424 / 13,148
```

**100,437 of 170,941 rows (58.8%)** carry a `period_end` older than two years, holding **320,316,309,513 of 561,235,592,698 shares (57.1%)**. Oldest `form3` row: `1994-01-27`; oldest `form4`: `1998-08-08`.

## Why this is the binding axis, not #2230's

`PYTHONPATH=. uv run python -m scripts.audit_2230_insider_oversubscription` (12,739 instruments rendered through the real read path, 0 harness errors) prices both mechanisms on ONE cohort with ONE criterion — *"would this, applied perfectly, bring the wedge back under `shares_outstanding`"*:

| | |
| --- | ---: |
| instruments whose rendered insiders wedge > `shares_outstanding` | **664** |
| cleared by folding every surviving equal-value cluster (#2230) | **16** (2.4%) |
| cleared by bounding insider rows to 2y of the slice's own newest as-of | **452** (68.1%) |
| cleared by a 5y bound | **319** (48.0%) |

On the 372 instruments where de-duplication is arithmetically possible at all, unfolded equal-value clusters carry **1,603,720,100** of the **14,468,619,737**-share overage — **11.1%**.

## Worked case — `CHRD`

[...]

PR #2791, opened 2026-08-20
measure(#2230): the insider-oversubscription residual is a STALENESS axis, not a deemed-attribution one

Luke-Bradford:
## What this does

Measures the #2230 residual and prices it against the alternatives, on one cohort with one
criterion. **No behaviour change** — a read-only audit script, its pure-logic tests, and a
prevention-log entry.

The ticket's headline (#2226's partition, 2026-08-03) was **378** instruments whose insiders
wedge exceeds `shares_outstanding`. Six changes have landed on that surface since — #2229,
#1652/#2230's deemed-chain tier, #1764, #2385, #2386, #2786 — and none of those runs
re-measured it. This does.

## The finding: #2230's own mechanism is not the binding axis

`PYTHONPATH=. uv run python -m scripts.audit_2230_insider_oversubscription`, full
population, every instrument rendered through the real `get_ownership_rollup` — **12,739
rendered, 0 harness errors**. Both arms answer the same question: *would this mechanism,
applied perfectly, bring the wedge back under `shares_outstanding`?*

| | |
| --- | ---: |
| instruments rendered | 12,739 |
| with a usable denominator | 4,424 |
| **insiders wedge > `shares_outstanding`** | **664** |
| … cleared by folding every surviving equal-value cluster (**#2230's mechanism**) | **16** (2.4%) |
| … cleared by bounding insider rows to 2y of the slice's own newest as-of | **452** (68.1%) |
| … cleared by a 5y bound | **319** (48.0%) |

Partition of the 664:

| | count | note |
| --- | ---: | --- |
| one holder ALONE exceeds the denominator | **292** | nothing to de-duplicate; median 2.52×, p90 16.7× |
| additivity-eligible | **372** | of which 191 carry an unfolded equal-value cluster, 181 carry none |

On those 372, unfolded equal-value clusters carry **1,603,720,100** of the
**14,468,619,737**-share overage — **11.1%**.

⚠ **The staleness arm is a diagnostic, not a proposal**, and the same run says so: it
removes **26,534,264,979** shares against that 14.47B overage — it over-corrects by 183%,
because a long-held stake nobody has had occasion to restate is indistinguishable from an
exited one *under a fixed window*. Lifting it into a rule would be the #2231 mistake
(a bound fitted to the symptom rather than derived from the source). The script's docstring
and `partition_insiders`' docstring both carry that warning inline.

## Root cause of the dominant axis, and what was filed

#2226 named mechanism **M1** — *"no `_current` table can ever release a position"* —
explicitly across institutions, insiders and blockholders. **#2229 shipped it for 13F
only.** Grep the ticket in the read path and every hit is institutional
(`ownership_rollup.py:158`, `:991`, `:1149`, `:5005`). The insider half was never re-filed,
and #2226's close-out (*"#2229 merged. This tracker moves 47.0% → 40.3%"*) is what buried
it: a tracker percentage that moves reads as progress on the whole mechanism.

Full population today:

```sql
SELECT source,
       count(*) FILTER (WHERE period_end >= current_date - interval '2 years') AS within_2y,
       count(*) FILTER (WHERE period_end <  current_date - interval '2 years') AS older
  FROM ownership_insiders_current GROUP BY 1;
-- form4 63,080 / 87,289 · form3 7,424 / 13,148
```

**100,437 of 170,941 rows (58.8%)**, holding **320,316,309,513 of 561,235,592,698 shares
(57.1%)**.

Filed rather than folded in:

- **#2788** — the insider release gap itself, with the source-rule research named. ⚠ It
  carries an explicit "do not port #2229 mechanically" section: a 13F exit is reported by
  OMISSION, a Section-16 disposal is an affirmative filing (sec-edgar §2.3 codes
  `S`/`D`/`G`/`U`), so a frozen insider row does not mean what a frozen 13F row means.
- **#2789** — **9 of the 10 scripts** that render the rollup do so outside `snapshot_read`,
  violating the contract in `get_ownership_rollup`'s own docstring
  (`ownership_rollup.py:4776`). Includes the A/B harnesses behind #2230, #2385, #2386 and
  #2785. Found by Codex on this diff; the same defect was in every sibling.
- **#2790** — 20 `ownership_insiders_current` rows carry a `period_end` in the future (max
  `2047-05-24`), so they can never be superseded by a period-keyed rule — which is exactly
  what #2788 will build.

## Codex checkpoint 2

Two P2s, both real, both fixed, then re-run clean:

1. **Leap-year drift** — `timedelta(days=365 * years)` is not an N-year boundary. Replaced
   with `_years_before`, calendar-year subtraction with the 29 Feb → 28 Feb clamp
   (`5a333d09`).
2. **`snapshot_read`** — `get_ownership_rollup` documents that the caller MUST already be
   inside it, and the census was not. Fixed (`5970d1fe`), **and the entire census re-run
   under snapshot isolation**: every figure above is from that run, and it is
   byte-identical to the pre-fix run.

## Definition of done

- **Clause 8 — golden panel** (`db041dee`..`5970d1fe`, from the final census): all five
  healthy and byte-identical to the values #2230's own OPERATOR VERIFY block recorded at
  `6b884d94`, which is an independent check that this harness renders what the read path
  renders.

  | | insiders | insiders / outstanding | pie / outstanding |
  | --- | ---: | ---: | ---: |
  | `AAPL` | 17,631,265.0000 | 0.12% | 0.676 |
  | `GME` | 98,026,742.0200 | 21.85% | 0.584 |
  | `MSFT` | 459,504,753.1060 | 6.19% | 0.813 |
  | `JPM` | 27,945,968.0221 | 1.05% | 0.771 |
  | `HD` | 40,988,847.3002 | 4.11% | 0.819 |

- **Clause 9 — cross-source, SEC EDGAR direct** (`5970d1fe`).
  `data.sec.gov/submissions/CIK0001731303.json`: `Forge Energy, LLC` has **5 filings in
  total** — Form 3 `2018-02-20`, Form 4 `2018-06-13`, SC 13G `2018-02-20`, two 13G/A ending
  `2019-01-30` — and nothing since. We render its Form 3 holding of **46,000,000** shares as
  a current `CHRD` insider position today, against **54,697,872** shares outstanding
  (`CHRD` insiders total **209,094,440** = 382.3%). Independent of both our DB and the audit.
- **Clauses 10 + 11 — NOT APPLICABLE.** No parser, ETL, schema or endpoint change; no stored
  value moves; nothing to backfill. **No `## OPERATOR VERIFY` block** — the only
  reproduction step is re-running the script (~9 min sharded 8 ways, then ~2 min to
  classify).
- **Clause 12** — each of the above is tagged with its SHA inline.

## Tests

`tests/test_audit_2230_insider_oversubscription.py`, **14 pure-logic cases**, no DB — the
partition arithmetic is what the verdict rests on, so it is table-tested rather than left to
the audit's own output. **Revert-probed 12/12 CAUGHT**, each probe asserting
`count(old) == 1` before mutating so a probe that matches nothing cannot pass as a probe
that caught nothing.

## Security

No security model touched: read-only script, no new endpoint, no auth surface, no
user-controlled input, no writes.

Refs #2226. Refs #2230. Refs #2229. Refs #2788. Refs #2789. Refs #2790.


Luke-Bradford:
### NITPICK — `--shard`/`--shards` not validated

`FIXED a3d14d14`

Agreed, and it is the exact failure mode this script exists to avoid making: an
out-of-range shard selects nothing, so the census exits 0 with an empty output file, which
is indistinguishable from a clean population. Same shape as the `audit_2386` census trap
and the `cmd | tail` exit-code gotcha already in the prevention log.

`main()` now rejects it before opening a connection:

```
$ PYTHONPATH=. uv run python -m scripts.audit_2230_insider_oversubscription --shard 5 --shards 2 --out /tmp/x.jsonl
error: --shard must satisfy 0 <= shard < shards, got shard=5 shards=2
EXIT=2, and /tmp/x.jsonl is not created
```

Pinned by `test_an_out_of_range_shard_is_an_error_not_an_empty_run`, parametrised over
`shard >= shards`, a negative shard, and `shards=0`. Revert-probed CAUGHT.

⚠ One thing worth recording, because the probe is what found it: my first version had **two**
guards (`shards < 1` then the range check) and the `shards < 1` branch was **unreachable** —
`0 <= 0 < 0` is already False, so the range check catches `--shards 0` on its own. The probe
came back `NOT CAUGHT` for it, which is how a redundant guard announces itself. Collapsed to
one check in `a3d14d14`; probes now 13/13 CAUGHT.

Issue #790, opened 2026-05-03
Ownership coverage expansion: seed list (top 150 13F), Form 3 historical backfill, Soros/Geode mis-seed fix


Luke-Bradford:
Parent: #788

Codex audit + SEC EDGAR full-text-search verification: GME has ~380 distinct 13F-HR filers per quarter; we have 7. ~2% coverage. Same gap will hit every other instrument. This ticket covers the data-side expansion.

### P1. 10-CIK seed list is too narrow

[scripts/seed_holder_coverage.py](scripts/seed_holder_coverage.py) seeds 10 institutional CIKs. SEC universe per quarter (CUSIP 36467W109 / GME, 2025-Q3+Q4): 761 filings, ~380 distinct filers. Coverage: ~1.8%. Need top ~150 13F filers covering ≥80% of float on Russell 1000 instruments.

### P2. Soros mis-seeded as Geode + tagged ETF

CIK 0001029160 is **Soros Fund Management LLC** per live SEC data. Seed file labels it ""Geode Capital Management"" and tags it ETF. Geode's real CIK is 0001572162. Soros is a hedge fund — every Soros position is currently routed to ETF bucket on the chart.

### P3. Form 3 baseline never backfilled

PRs #771-#774 shipped the Form 3 ingester but only fetch new filings. ``insider_initial_holdings`` has 0 rows for GME despite Cohen et al. being directors/officers. Need a one-shot historical walker.

### P4. No backfill discipline pattern

When a new ingest pipeline ships, operator can't see existing data — only data filed since the pipeline went live. Each new pipeline needs an explicit ""backfill on activation"" pass for already-seeded instruments.

## Acceptance criteria

1. **Top 150 13F filer seed list.** Source: SEC's 13F manager list — https://www.sec.gov/cgi-bin/browse-edgar?action=getcompany&type=13F-HR&dateb=&owner=include&count=40 — sorted by AUM. Includes Vanguard, BlackRock (correctly disambiguating fund vs adviser CIKs), State Street, Fidelity, Capital, Wellington, Geode (real CIK), JPMorgan, Bank of America/Merrill, Morgan Stanley, Goldman, T. Rowe Price, Invesco, Northern Trust, Bank of NY Mellon, Dimensional, AllianceBernstein, Franklin Resources, Citadel, Renaissance, Two Sigma, Millennium, D.E. Shaw, Bridgewater, AQR, Soros (real labelling), Tudor, Point72, etc.

[...]

PR #2792, opened 2026-08-20
fix(#2788): bound the insiders wedge by the Form 4 ingest retention cutoff

Luke-Bradford:
## What this changes

The insiders wedge rendered Form 4 rows that the ingest layer's own chokepoint refuses to store. `_collect_canonical_holders_from_current` now excludes `ownership_insiders_current` rows with `source = 'form4'` and `filed_at::date < form4_retention_cutoff()`, and `_read_beyond_retention_insiders` lists exactly what that removed as a new `CorrectionApplied` kind, `insider_beyond_form4_retention`.

**This is not an exit claim, and the code says so in three places.** #2229's 13F supersession rests on a source rule — Form 13F Special Instruction 5b makes a holdings report a *complete* statement, so omitting a security is affirmative evidence of an exit. **Section 16 has no equivalent.** Form 4 is transaction-triggered (the `S`/`D`/`G`/`U` codes in `data-sources/sec-edgar.md` §2.3 arrive as filings, not as silence), and Rule 16a-2's obligation simply ends when a person stops being an insider — which is why a frozen row stops moving rather than going to zero. The correction states only that no in-retention evidence remains, per the #790 posture that the truthful state beats a fabricated one.

**Form 3 is deliberately not gated.** #1233 §4.3 exempts it ("Form 3 rows are NOT gated here"), because the initial statement is the only evidence we hold for a holder who has never transacted. `test_form3_is_never_gated_by_the_form4_cutoff` pins that.

## Source rule

`INSIDER_FORM4_RETENTION_YEARS = 3` (`app/services/insider_transactions.py:94`), set by #1233 PR4 spec §4.3 and enforced at every Form 4 write chokepoint by `form4_within_retention` — which is itself lint-guarded by `scripts/check_form4_retention.sh`. This PR adds no new constant and fits no window to the symptom; it gives the existing, already-binding boundary a read-side counterpart. Deliberately **not** the 2y/5y staleness arms from #2230's census, which that script's own docstring calls diagnostics and which over-remove by 183% of the overage.

⚠ One honest note on scope: the write cap only ever refused *new* pre-retention rows, it never removed rows that aged past the rolling boundary. Applying it at read time is therefore stricter than any regime previously in force, and that is intentional — it is the release mechanism #2788 asks for. Priced below.

## Root cause — measured this run, not carried from the issue

`PYTHONPATH=. uv run python -m scripts.audit_2788_insider_retention --census` (new, read-only, computes every figure at run time so a later ingest moves the numbers instead of leaving a stale copy in prose):

```
form4_retention_cutoff() = 2023-08-20

retention split (form3 shown but NEVER gated)
  form3 | beyond=false | 20561 rows | 4190 instruments | filed 2006-01-04..2026-08-19
  form4 | beyond=false | 72296 rows | 4472 instruments | filed 2023-08-21..2026-08-19
  form4 | beyond=TRUE  | 78043 rows | 3579 instruments | filed 2006-01-03..2023-08-18

backing-observation ingest day
  2026-08-14 | beyond=TRUE | 76920 rows | 3559 instruments | 203,364,662,254 shares
```

**76,920 of the 78,043 beyond-retention rows (98.6%) arrived on 2026-08-14.** That is `scripts/backfill_2701_insider_research_ingest.py`, which ran `ingest_insider_dataset_archive` with `retention_cutoff_override = date(2006, 1, 1)`. That function's only INSERT target is `ownership_insiders_observations` — the sole source of `ownership_insiders_current`, which is the insiders wedge. Its own comment says the override "CHANGES NO DEFAULT" for "the operator ALERTING path"; both statements are about the alerting path, and the consumer actually exposed was the ownership rollup.

Before that run the form4 observation corpus was retention-bounded by construction — pre-2026-08-14 rows split 975,515 within-retention against 19,650 beyond, and the beyond side reached back only to `2023-06-05` (rows that had aged past the boundary, not rows written past it).

⚠ **#2701's own goal was not met.** Its consumer is `insider_transactions` / `insider_filings` (`scripts/verify_2437_insider_forward_returns.py`), which the DERA ingest never writes:

```sql
SELECT count(*) FILTER (WHERE period_of_report < date '2017-01-01') AS pre_2017,
       count(*) AS total FROM insider_filings;
-- 2,581 | 568,814      (pre-2017 = 0.45% of the corpus)
```

So the run moved ~4.5M rows into the ownership layer and left the research corpus where #2701 found it. Nothing failed. Filed separately — see below.

## The ticket's first question, answered

#2788 said "the holder ceased to be a reporting person" was the obvious story and **unmeasured**, and must not be inherited as a premise. Measured on the full population:

| | |
| --- | ---: |
| `ownership_insiders_current` rows with `period_end` > 2y old | 100,437 of 170,941 |
| … whose instrument's own insider slice is still fresh (< 2y) | **99,203 (98.8%)** — so the freeze is holder-level, not issuer-level |
| … with a LATER live observation at the same (instrument, holder, nature) | 336 keys — the projection is faithful; there is no supersession bug |
| … whose holder files later on the same instrument under a DIFFERENT nature | 10,690 keys |
| … with no row at all in `insider_transactions` for that (instrument, holder) | 87,041 (86.7%) — i.e. written by the DERA path, not the XML path |

So the answer is neither "they ceased reporting" nor a projection defect: 78.3% of the stale rows are DERA-dataset provenance, and 98.6% of the beyond-retention ones landed six days ago.

## Full-population A/B

Both arms rendered through the real read path by `scripts/audit_2230_insider_oversubscription.py` (4 shards, 12,739 instruments, **0 harness errors on both sides**). The control arm is genuine, not simulated: its four processes started 21:48:03–21:48:15 and `app/services/ownership_rollup.py` was first modified at 21:49:30.

| | control | treatment |
| --- | ---: | ---: |
| instruments rendered | 12,739 | 12,739 |
| **insiders wedge > `shares_outstanding`** | **664** | **410** |
| pie total > `shares_outstanding` | 2,243 | 1,903 |

Distinct-instrument diff over the 12,738 comparable instruments:

- **over → under (GAIN): 254**
- **under → over (REGRESSION): 0**
- insiders wedge changed on 3,174; shrank on 3,166, **grew on 7**

⚠ **FLNC is excluded as drift, not counted as a regression.** It has no `shares_outstanding` in the control arm and one in the treatment arm — a DEI fact landed on the live dev DB between the two runs. That single instrument is also the whole of the `4,424 → 4,425` "usable denominator" difference. Flagging it rather than quietly dropping it: the arms are ~40 minutes apart against a live database, so this is the drift bound, not zero.

Gain side, worst control ratios that cleared: `ATLX` 53.82→0.32 · `GOVX` 34.49→0.01 · `IVF` 21.26→0.44 · `PLUR` 19.91→0.85 · `TIL` 13.49→0.96 · `FKWL` 11.04→0.12 · `SIRI` 10.23→0.40.

### The 7 instruments where the wedge GREW — inspected, not waved past

Removing rows cannot grow a sum, so a fold changed. `IPAR` is the clean case: control folded `Philippe Benacin Holding SAS` (6,846,064, filed 2015) with `BENACIN PHILIPPE` (6,846,064) as an equal-value control-group cluster. The bound removes the 2015 row, the remaining holding-SAS row carries a *different* value (6,871,064), the equal-value match breaks, and both now count.

So #2230's equal-value collapse is anchored on an exact share value and is brittle to any row removal. **3 of the 7 were already over and stay over (`ACI`, `CAR`, `DWSN`); 4 stay under; none crosses the threshold.** Filed as a follow-up rather than folded in — it is #2230's mechanism, not this bound.

## Clause 8 — golden panel

Rendered through `get_ownership_rollup` inside `snapshot_read`, control figures taken from my own control-arm census rather than from any issue body:

| symbol | insiders (control) | insiders (treatment) | corrections | shares removed |
| --- | ---: | ---: | ---: | ---: |
| AAPL | 17,631,265 (0.12%) | **9,782,902 (0.07%)** | 28 | 7,848,363 |
| GME | 98,026,742 (21.85%) | **48,022,385 (10.70%)** | 56 | 91,432,341 |
| MSFT | 459,504,753 (6.19%) | **2,621,605 (0.04%)** | 48 | 456,887,271 |
| JPM | 27,945,968 (1.05%) | **9,096,971 (0.34%)** | 45 | 18,848,997 |
| HD | 40,988,847 (4.11%) | **555,475 (0.06%)** | 51 | 40,433,372 |

All five pie ratios stay under 1. ⚠ Treatment insiders + shares_removed does not equal control except on AAPL — the folds re-run over a different row set, which is the same effect as the 7 growths above and is why the audit script's docstring insists the rendered wedge is not the raw `_current` sum.

## Clause 9 — cross-source, SEC EDGAR direct

Largest removed holder on MSFT: `BALLMER STEVEN A` (CIK `0000902015`), 333,254,734 shares, our row dated `2014-08-19`.

`data.sec.gov/submissions/CIK0000902015.json` reports 23 Section-16 forms, **latest `2025-04-25`** — which reads as an eleven-year ingest gap, and was the claim I was about to write. It is not one. That filing is a **Form 3 for Stagwell Inc** (`STGW`, issuer CIK `0000876883`), resolved from `…/Archives/edgar/data/902015/000095017025058893/ownership.xml`. His last MSFT **Form 4** is `2014-08-20`, matching our stored row. A reporting person's submissions feed is per-FILER and carries no issuer field; extracted to the prevention log.

Second fixture, the cost side: `RC Ventures LLC` (CIK `0001822844`) on GME, 36,657,204 shares removed. EDGAR direct: 6 Section-16 forms, latest `2023-06-13` — genuinely before the `2023-08-20` cutoff, so the removal is correct under the rule. **But RC Ventures is a live holder**, and this is the honest cost of the bound: the wedge now understates rather than overstates.

That cost is bounded and measured. Of the 78,043 removed rows, **1,123 (719 instruments, 1,992,203,569 shares, 0.97% of removed shares) aged past the boundary naturally** rather than arriving on 2026-08-14, and every one was filed between `2023-06-05` and `2023-08-18` — within ~2.5 months of the cutoff. Those are the rows no prior regime removed, and they are the release mechanism working as intended: rolling, so they keep aging out.

## Codex checkpoint 2

One P2, real, fixed in this commit. `_read_beyond_retention_insiders` omitted the dual-pipeline `:NDT:`/`:NDH:` de-collision predicate the insiders read applies, so a stale accession present in **both** ingest pipelines would emit two corrections and report twice the shares that actually left. Both predicates are now shared module constants (`_INSIDER_BEYOND_RETENTION_SQL`, `_INSIDER_DUAL_PIPELINE_DECOLLISION_SQL`) interpolated by the read and the producer alike — #2229 kept its pair in step with a comment, and a comment does not fail when someone edits one side.

⚠ **Latent, not active**: the same predicate over the full population returns **0 rows / 0 instruments** today, because the manifest path holds no pre-cutoff Form 4 rows for the two pipelines to collide on. Reported as a guarded latent defect rather than as a figure that moved.

## Tests — 7 cases, 6/6 revert-probed

`tests/test_ownership_insider_retention_bound.py`. One pure structural case (both call sites interpolate both shared fragments) plus six DB-backed.

```
P1 exclusion removed                        : CAUGHT (2 failed)
P2 form3 also gated                         : CAUGHT (2 failed)
P3 boundary off-by-one (< -> <=)            : CAUGHT (1 failed)
P4 telemetry unwired                        : CAUGHT (2 failed)
P5 shared fragment inlined in producer      : CAUGHT (1 failed)
P6 de-collision dropped from producer (P2)  : CAUGHT (2 failed)
```

Two harness faults corrected mid-probe, both of the silent-false-CAUGHT class:

- the driver first gated on `returncode != 0`, and `-x` under xdist exits **2** (interrupted), which is also pytest's usage-error shape. It now gates on a real `FAILED ` line and prints them.
- P5 came back **NOT CAUGHT**: the structural test grepped `inspect.getsource`, and both docstrings *name* the fragment, so the assertion passed on prose. The docstring is now stripped via AST first.

The vocabulary contract test `test_correction_kind_vocab_contract.py` also caught the new kind missing from `_CorrectionAppliedModel.kind` and `OwnershipCorrectionKind` — both added. ⚠ It then failed a second time on `sold`, a quoted word inside my new frontend JSDoc: that test reads the union by scanning quoted literals in the block. Comment reworded, and a warning left in place for the next editor.

## Clauses 10 + 11

**Not applicable — nothing to backfill.** This is a read-path correction in the #1639/#2229 shape; no stored value changes, no migration, no parser version bump. Choosing the read path over a `refresh_insiders_current` predicate was deliberate: it keeps the change reversible, and it is the only placement that lets the A/B above run both arms against unmutated data.

An `## OPERATOR VERIFY` block is posted on #2788 with the endpoint check and the acceptance queries, since the live-endpoint step needs the dev stack this worktree cannot reach.

## Gates

`ruff check` · `ruff format --check` · `pyright` (0 errors) · `pytest -m "not db"` · `pytest tests/smoke` · `pytest -m db` on the touched module and its neighbours (`test_ownership_rollup`, `test_ownership_13f_hr_supersession`, `test_ownership_13f_nt_supersession`, `test_correction_kind_vocab_contract`, `test_ownership_rollup_csv`) · `pnpm typecheck` · `pnpm test:unit` (1,525 passed) — each run separately and unpiped, all exit 0.

## Security

No authn/authz surface, no user input, no new SQL parameter that is not bound. The one new parameter (`form4_cutoff`) is a `date` computed in Python and passed as a psycopg parameter, never interpolated — the interpolated fragments are static SQL text held in module constants.

## Follow-ups filed

- the #2701 write-target defect (a research ingest writing the operator-visible ownership layer, and its own research goal unmet)
- #2230's equal-value fold being anchor-brittle to row removal (the 7 growths)
- the parked `data-sources/sec-edgar.md` §7.4.1 skill edit, blocked by #2403 for the seventh time

Refs #2788. Refs #2226. Refs #2230. Refs #2701.


Luke-Bradford:
### Review resolution

**NITPICK — `scripts/audit_2788_insider_retention.py:INGEST_DAY_SQL`, 6-column join with no supporting composite index: `REBUTTED`.**

The six join columns are not an arbitrary tuple — they are `ownership_insiders_observations`'s PRIMARY KEY, in declaration order:

```sql
SELECT conname, pg_get_constraintdef(oid) FROM pg_constraint
 WHERE conrelid = 'ownership_insiders_observations'::regclass AND contype = 'p';
-- ownership_insiders_observations_pkey |
--   PRIMARY KEY (instrument_id, holder_identity_key, ownership_nature, source,
--                source_document_id, period_end)
```

(`sql/113_ownership_insiders_observations.sql:77`.) The join is deliberately keyed that way because `_current` holds exactly one backing observation per `(instrument, holder identity, nature)` and that PK is the only tuple that identifies it unambiguously — anything narrower would fan out. The supporting index is the PK's, so none was added in the diff.

Measured rather than argued — the whole census, all three queries including that join against 5,577,156 live observation rows:

```
$ /usr/bin/time -p env PYTHONPATH=. uv run python -m scripts.audit_2788_insider_retention --census
real 18.36
```

Two things I'd still grant from the finding, neither of which changes the code:

- it is a read-only census script, not a hot path, and its docstring says so;
- the table is `PARTITION BY RANGE (period_end)`, so the PK index is per-partition and this is a partition-wise join — which is why 18s holds rather than degrading, but it also means the plan is more sensitive to partition count than to row count if the range is ever widened.

No change made. Happy to reverse if a reviewer sees a plan I haven't.

CI: 6/6 green on `5333347e` (`build`, `lint`, `perf-claim-lint`, `review`, `supply-chain`, `verify-issue-link`), `mergeStateStatus = CLEAN`.

Luke-Bradford:
### Handoff — ready to merge, deliberately NOT merged

State at `5333347e`: bot **APPROVE**, CI **6/6 green**, `mergeStateStatus = CLEAN`, branch not stale (based on `2bc1cdcb`, which is still `origin/main` as of this comment).

The only finding on the round is the NITPICK above, and I **rebutted** it rather than changing code. This loop's operating rule is not to self-merge a round that contains a rebuttal of its own, so the PR is left open on purpose — this is not an unfinished state, it is the gate.

**To finish, in order:**

```bash
git fetch origin
gh pr view 2792 --comments        # read the rebuttal above
git merge-base --is-ancestor origin/main origin/fix/2788-insider-release   # expect exit 0; rebase if not
gh pr merge 2792 --squash --delete-branch
```

If you disagree with the rebuttal, the change it would call for is an index on `ownership_insiders_observations` — but the join is already on that table's PRIMARY KEY and the full census runs in 18.4s against 5,577,156 rows, so my read is that there is nothing to add.

After merge, the `## OPERATOR VERIFY` block on #2788 is the next step: four commands, no backfill, no migration. The sharpest single check is `MSFT` moving from `459504753.1060` (6.19%) to `2621605.1060` (0.04%) on `/instruments/MSFT/ownership-rollup`.

Follow-ups already filed and needing no action here: **#2793** (the #2701 write-target defect — the real root cause, still unfixed) and **#2794** (#2230's equal-value fold un-folding when its anchor row is removed).

Luke-Bradford:
## Review resolution

**NITPICK** — `scripts/audit_2788_insider_retention.py:INGEST_DAY_SQL`, 6-column join with no supporting composite index visible in the diff.

`REBUTTED` — the index exists; it is just not in the diff, because it is the table's primary key:

```
CREATE UNIQUE INDEX ownership_insiders_observations_pkey
  ON public.ownership_insiders_observations
  USING btree (instrument_id, holder_identity_key, ownership_nature, source,
               source_document_id, period_end)
```

That is exactly the six columns the join equates, in order, so the join is a PK lookup rather than the unindexed scan the note anticipated. The observation was right about the diff and the diff was right not to add anything.

⚠ Adding a *further* composite index to speed this script up would be a real cost in the wrong place — write amplification on every insider ingest — to accelerate a read-only census that is run by hand.

---

Also rebased onto `main` — the PR had gone `CONFLICTING` because five ownership PRs landed under it while it was open. Done as a **merge of `main` into the branch** rather than a rebase, so the push is fast-forward and no remote history is rewritten (`28caa5d8`). No conflicts; the substantive diff is unchanged at 6 files.

Re-verified on the merged result: ruff, ruff-format, pyright (whole repo), `pnpm typecheck`, `pnpm test:unit` (1525 passed), and this PR's own DB-backed suite `tests/test_ownership_insider_retention_bound.py`.

⚠ The new merge commit resets the review gate, so this needs a fresh bot pass on `28caa5d8` before it can merge.

Issue #2783, opened 2026-08-20
A PR stacked on a non-main base is unmergeable through safe_merge, and its Closes/board side effects silently no-op


Luke-Bradford:
## Summary

Three independent mechanisms all key off a PR's **base** branch, and all three fail closed —
silently — for any PR stacked on a branch other than `main`. Found while merging #2782.

## The three facets, one root

1. **The review bot never runs.** `.github/workflows/*review*.yml` triggers on
   `pull_request: branches: [main]`. That filter matches the PR's **base**, not the head, so a
   PR based on e.g. `fix/2697-metric-axis-integrity` never triggers it. `gh pr checks` reports
   *"no checks reported"* — indistinguishable, to a reader, from "checks not finished yet".

2. **`safe_merge.sh` can therefore never pass.** `.autonomy/config.yaml` sets
   `merge_gate.strategy: bot_comment` with `author_login: github-actions` and
   `marker: "Claude Code Review"`. For any non-doc-only stacked PR the gate reaches
   `REFUSE -- no review comment from github-actions on #N yet` and stays there **forever**, because
   the comment it waits for cannot ever be posted. The autonomy loop has no path off this.
   Verified on #2782: `[.comments[] | select(.author.login=="github-actions" and
   (.body|contains("Claude Code Review")))] | length` → `0`.

3. **`Closes #N` is inert, so `done_everywhere` does nothing.** GitHub registers
   `closingIssuesReferences` only for PRs targeting the **default** branch. On #2782 that array
   was empty despite a `Closes #2779.` line in the body. `safe_merge.sh::done_everywhere` reads
   exactly that array (correctly — #301 made it the authority instead of a prose regex), so on a
   stacked PR it closes nothing and boards nothing, without warning.

## Evidence that this is the live path, not a hypothetical

`#2773`, `#2776`, `#2777` and now `#2782` all merged into `fix/2697-metric-axis-integrity`, and
**every one was merged by hand** (`mergedBy=Luke-Bradford`), bypassing `safe_merge.sh` entirely.
So the repo's real practice for stacked work is already "skip the gate", and nothing records that
the gate was skipped or why.

## Why it matters

⚠ The failure direction is the dangerous one. Facet 1 produces an ABSENCE of checks, and
`.claude/CLAUDE.md` already warns about exactly this shape: *"Treating a documented skip as 'still
pending' makes those PRs permanently unmergeable; treating any absence of review as a skip is the
[...]

PR #2798, opened 2026-08-20
fix(#1976): retry the review at effort "low" when thinking leaves no text

Luke-Bradford:
Refs #1976. Refs #2783.

## Why

`max_tokens` caps thinking and text **combined**. On a large diff, adaptive thinking can consume the whole budget and the response carries zero text blocks — the workflow then fails with no review posted, which shows on the PR as `review  fail` and no explanation.

Raising the cap to 16000 (already on `main`) moved that boundary; it did not remove it. **PR #2757 hit it again today at `input_tokens: 282286`.** That PR carries 61 commits of strategy work — the backtest evaluation going 465.4s → 68.6s and the #2779 exit-pricing fix — and it is currently unmergeable because `merge_gate.strategy: bot_comment` waits for a review comment the workflow cannot produce.

## What changed

The retry the workflow's own error message already asks for:

- `extract.py` exits **42** when the response carried no text block
- the caller reruns the identical prompt once at `output_config.effort: "low"`, which thinks far less and leaves room for the review text
- a second 42 fails the job, so it cannot loop

The API call is factored into a `call_api()` shell function so both attempts share one implementation rather than a copy.

⚠ **Keeps `main`'s diagnostic rather than replacing it.** The `usage` dict is what tells a reader whether the cap or the prompt size was the binding constraint; the message now also names the retry, so the log explains the second call before it happens.

⚠ **One thing deliberately left as-is:** the `Fail workflow on truncated review` step still fails the job when `stop_reason == max_tokens` on a review that *did* emit text. A truncated review is not a passed review, and this change must not quietly convert one into the other.

## Provenance

The core of this was already written on the abandoned branch `fix/claude-review-max-tokens` (`a25520fe`) and never opened as a PR. `main` had since landed the 5000 → 16000 half independently, so this merges `main` in and resolves the two resulting conflicts — taking `main`'s comment wording and diagnostic, and the branch's retry mechanism.

## Verification

`yaml.safe_load` parses the workflow; both halves confirmed present (`"max_tokens": 16000` at line 311, `request_low.json` / `sys.exit(42)` / the `effort=low` retry at 333–406). Full pre-push gate green: ruff, ruff-format, pyright, all 25 chokepoint lints, fast tier, smoke.

⚠ The real proof is #2757 itself — its review is the case that fails today, so re-running it after this lands is the acceptance test.

## Security model

Not touched. No change to secret handling, permissions, or what the workflow posts — `ANTHROPIC_API_KEY` and `GITHUB_TOKEN` are used exactly as before, and the retry sends the same prompt to the same endpoint.


Luke-Bradford:
## Review resolution

**NITPICK 1** — `payload_low = dict(payload)` is a shallow copy.

`FIXED` — now `copy.deepcopy(payload)`, with `import copy` added to `build_prompt.py`. You are right that it is safe *today* (the only mutation replaces `output_config` wholesale rather than editing a nested value), but the retry's entire premise is that it is the **identical prompt at a lower effort** — so the two payloads differing in exactly one field should be structural, not a thing a future editor has to remember. Comment records why.

**NITPICK 2** — the full `response.json` body is echoed for both the primary and retry calls, doubling log verbosity on the retry path.

`REBUTTED` — that echo is the diagnostic this PR exists because of. #2757's review failed today with `review  fail` and **no comment**, and the only way to establish *why* was reading the echoed body to recover `usage={'input_tokens': 282286, ...}` and the `stop_reason`. Without it the failure is indistinguishable from a network error, a bad key, or a malformed prompt.

⚠ The retry path is also exactly where the second body matters most: if attempt 1 exhausted the budget and attempt 2 also returns no text, the two `usage` figures side by side are what say whether effort `low` helped at all. Halving the verbosity would remove the evidence at the moment it is most needed, to save log lines on a path that fires rarely.

⚠ It is also not a secret-exposure concern — the response body contains the model's own output, not the key; `ANTHROPIC_API_KEY` is passed as a header and never echoed.

---

Verified after the fix: `yaml.safe_load` parses the workflow, and `build_prompt.py` extracted from its heredoc parses under `ast.parse`.

PR #2799, opened 2026-08-20
fix(#2797): S-2 rebalances on the first WEEKDAY bar of a new month

### LINKED ISSUE #2797: S-2 rebalances on weekend junk dates — 13 of 73 rebalances destroyed, and it has NEVER fired a signal in production
## Symptom

`s2-cross-sectional-momentum` has fired **zero** signals in production, all-time, under every
version. Measured on dev at 2026-08-20:

```
select strategy_id, verdict, sum(row_count), max(signal_bar_date)
from strategy_signal_daily_counts
where strategy_id = 's2-cross-sectional-momentum' group by 1,2;
```

| verdict | rows |
| --- | --- |
| `not_fired` | 26,203 |
| `not_evaluable` (`insufficient_warmup`) | 20,301 |
| `fired` | **0** |

The prevention log already records this zero as a *legitimate* empty
(`docs/review-prevention-log.md`, "An output whose EMPTINESS is a legitimate outcome cannot be
aged"). That entry is correct about the freshness check it was written for, and it never asked
why the zero was there. This ticket answers that: it is a bug.

## Cause

`s2_cross_sectional_momentum.rebalance_dates` takes *"the first bar whose calendar month differs
from the previous bar's"* over the panel's **union** calendar. `load_union_calendar` is a union
over every loadable instrument, and `price_daily` carries weekend junk rows.

Measured on the validated universe (6,774 instruments, 3,673,648 bars):

- **3,669 weekend bars** across **389 instruments** and **329 distinct weekend dates** (0.0998% of all bars).
- **13 of 73** S-2 rebalance dates (17.8%) land on a Saturday or Sunday.
- On those dates the panel is **4–117 names**; every real rebalance date runs **3,629–5,380**.

Because the FIRST qualifying bar takes the month, a weekend row consumes the month and the real
first trading day never rebalances at all.

| rebalance date | day | eligible names (`close >= 1.0`) |
| --- | --- | --- |
| 2026-08-01 | Sat | **9** |
| 2026-08-03 | Mon | 5,380 |
| 2025-11-01 | Sat | 11 |
| 2025-11-03 | Mon | 3,629 |

The full weekend-landed set: 2021-05-01, 2022-05-01, 2022-10-02, 2023-07-01, 2023-10-01,
2024-06-01, 2024-09-01, 2024-12-01, 2025-02-01, 2025-03-01, 2025-06-01, 2025-11-01, 2026-08-01.

2026-08-01 is the ONLY rebalance date that has ever fallen inside a written scan window, and its
panel of 9 is below S-2's `MIN_CROSS_SECTION` of 10. That is the whole of the production zero.

**The failure is silent.** `strategy_signal_daily_counts` has no row of any kind for
2026-08-01 — not even a `thin_cross_section` refusal — because the junk instruments are not
frontier-eligible. An absent month is indistinguishable from a quiet one.

## The fix already exists, one module over

`s10_relative_strength_leader.s10_rebalance_dates` drops weekends before the month rule runs, and
carries the source rule: the validated universe is US-listed stock (§4.0) and US equity venues
hold no regular weekend sessions (NYSE/Nasdaq holiday-and-hours calendars). Its docstring says:

> S-2 shares the exposure and that is noted there, not silently fixed here — editing its calendar
> would move its identity

`s2_cross_sectional_momentum.py` contains no such note — `grep -n weekend` on it returns nothing.
The mechanism that was supposed to keep the exposure visible does not exist, which is why it sat
for two weeks.

## Blast radius of fixing it

Editing the module moves S-2's `source_hash` and therefore its `StrategyIdentity`. That cost is
measured, not assumed:

- `strategy_signals` fired rows for S-2: **0** — nothing is orphaned.
- `strategy_preregistration_declarations` pinning S-2: **0**.
- `strategy_promotions` / `strategy_live_gate_assessments`: **0** rows total.
- `strategy_results`: 8 S-2 rows, which stay pinned to the versions that produced them.

⚠ Those 8 results were produced under the contaminated calendar, and S-2's exit is
*"until the next rebalance"* (`strategy_manifest._s2_exit_regime` passes the same `decision_dates`),
so the bug mistimed entries **and** holds. Any prior S-2 backtest conclusion is superseded, not
carried forward.

## Scope

1. Weekday cut in `rebalance_dates`, with the source rule cited.
2. Correct the stale claim in `s10_rebalance_dates`'s docstring.
3. Pure-logic test pinning the weekend cut and the month-consumption case.
4. Prevention-log entry: a union calendar carrying junk dates silently consumes a
   period-boundary rule.

Refs #2437

Luke-Bradford:
Closes #2797. Refs #2437.

## What was wrong

`s2-cross-sectional-momentum` had fired **zero** signals in production, all-time, under every
version — 26,203 `not_fired` rows, 20,301 `not_evaluable`, no `fired` row.

S-2 rebalances on *"the first bar whose calendar month differs from the previous bar's"*,
evaluated over the panel's **union** calendar (`load_union_calendar` — every date on which ANY
instrument has a loadable bar). `price_daily` carries weekend rows. A union is a max and
"first of the period" is an argmin over it, so the two compose into *the corpus's rarest row
decides the boundary for the whole panel*.

Full population, validated universe, 2026-08-20:

```sql
SELECT count(*), count(DISTINCT instrument_id), count(DISTINCT price_date)
  FROM price_daily WHERE extract(isodow FROM price_date) >= 6;  -- 3669 / 389 / 329
```

**3,669 weekend bars** across 389 instruments on 329 distinct dates — 0.0998% of 3,673,648 bars
— controlling **13 of 73** rebalance dates (17.8%).

The most recent, Sat 2026-08-01, is the whole of the production zero: it is the only rebalance
date that has ever fallen inside a written scan window. **It wrote no row of any kind** — not a
`thin_cross_section` refusal, not a `not_fired` — because the junk instruments are not
frontier-eligible. An absent month and a quiet month render identically, which is why this
survived a prevention-log entry that had already measured the zero and reasoned it healthy
(*"An output whose EMPTINESS is a legitimate outcome cannot be aged"* — that entry is correct
about the freshness check it was written for; it never asked why this particular zero was there).

## Source rule

The validated universe is US-listed stock (spec §4.0) and US equity venues hold no regular
weekend sessions (NYSE/Nasdaq holiday-and-hours calendars), so a weekend row is a corpus
artefact and not a bar §4 can mean. Weekends are cut **before** the month rule runs — after is
too late, the boundary is already spent.

This is not a new reading. `s10_relative_strength_leader.s10_rebalance_dates` already does
exactly this and already carries the rule. Its docstring said:

> S-2 shares the exposure and that is noted there, not silently fixed here — editing its
> calendar would move its identity

It was **not** noted there — `grep -n weekend s2_cross_sectional_momentum.py` returned nothing.
The only record of S-2's exposure lived in the module that did not have it. That sentence is
corrected in this PR.

## Full-population A/B

`scripts/ab_2797_s2_weekday_rebalance.py`, over `price_daily` through `load_masked_bars` — the
same loader and universe stamp (`SCAN_UNIVERSE`) the production scan reads. The control arm
**re-derives** the pre-fix function verbatim rather than simulating it, so neither arm can
inherit the other's bug. `[ab] OK`, exit 0.

| | dates | panel range | entry signals |
| --- | --- | --- | --- |
| **control-only** (removed) | 13 | 0–11 | **7** |
| **treatment-only** (gained) | 13 | 0–3,346 | **2,419** |

Per date, the two that matter most:

| date | day | panel | selected |
| --- | --- | --- | --- |
| 2026-08-01 | Sat | 0 | refused |
| **2026-08-03** | **Mon** | **3,204** | **320** |
| 2025-11-01 | Sat | 11 | 1 |
| 2025-11-03 | Mon | 3,345 | 334 |

The script fails if the treatment leaves any weekend date standing, or if it moved a date whose
control panel was already healthy (≥1,000). Neither happened.

⚠ **Two of the 13 gained dates are still refused or thin, and that is not fixed here.**
2021-05-03 has a panel of 0 (corpus coverage/warm-up at that depth — the control was 0 too), and
2024-09-02 has 11, because it is **Labor Day**: a market holiday is a corpus-hole *weekday*, so
it still takes its month and is then refused by `MIN_CROSS_SECTION`. That case is deliberately
left alone for S-10's stated reason — teaching a pure calendar function participation counts it
cannot verify is worse than a self-healing hole. Measured incidence: 1 date in 73.

## ⚠ This does NOT make S-2 fire tomorrow

`write_window_indices` bounds a cold start to **one bar** by design — spec §11 forbids
backfilling signals, because deriving yesterday's signals from today's stored bars reintroduces
the look-ahead phase 5 removed. Every new `strategy_version` therefore restarts at the frontier.
S-2 rebalances monthly, so the first signal this fix can produce is at the **next month
boundary, Tue 2026-09-01**.

⚠ A second-order consequence worth stating rather than leaving to be rediscovered: S-2 moved
through **four** versions in 24 days (≈ one per 6 days) against a 30-day decision cadence, so a
version has a low probability of surviving to any rebalance at all — independently of this bug.
That is a property of active development rather than a defect to fix here, but it means "S-2
fired nothing" will keep being ambiguous while identities churn weekly. Recorded in the
prevention-log entry; no ticket filed.

## Identity impact

Both S-2 and S-10 move (`_source_hash` hashes the module file, so a docstring edit moves it too).
The cost was measured, not assumed:

| | S-2 | S-10 |
| --- | --- | --- |
| fired `strategy_signals` | 0 | 0 |
| `strategy_preregistration_declarations` | 0 | 0 |
| `strategy_promotions` | 0 | 0 |
| `strategy_live_gate_assessments` | 0 | 0 |
| `strategy_results` | 8 (stay pinned to their own versions) | 0 |

⚠ Those 8 S-2 results were produced under the contaminated calendar, and S-2's exit is
*"until the next rebalance"* (`strategy_manifest._s2_exit_regime` passes the same
`decision_dates`), so the bug mistimed entries **and** holds. Any prior S-2 backtest conclusion
is superseded, not carried forward.

## Duplication, on purpose

S-2 and S-10 keep two independent copies of the rule. Co-location was not ruled out on cost —
both identities move here regardless — but on evidence: a test over one shared import is the
tautology this repo has already shipped once. `TestNoDriftAgainstS10` compares the two
implementations behaviourally over four years of dense calendar (every date, weekends included)
and again over one punched with holes.

## Security

No security surface: two pure date functions, a read-only measurement script, a test and a doc.
No new input, endpoint, credential or query against user-supplied data.

## Verification

| gate | result |
| --- | --- |
| `ruff check .` / `ruff format --check .` | clean |
| `uv run pytest -m "not db"` | exit 0 |
| `uv run pytest tests/smoke` | exit 0 |
| pre-push hook (incl. chokepoint lints, frontend integrity) | green |
| Codex checkpoint 2 (`codex exec review --base origin/main`, staged) | no actionable defects |
| full-population A/B | `[ab] OK`, exit 0 |

Definition-of-Done clauses 8–12 do not apply: no parser, ETL, schema migration or ownership /
observations data path is touched, and no backfill is possible (§11 forbids it for signals).
The equivalent evidence for this rung is the full-population A/B above, at `cbdebba1`.


Luke-Bradford:
## Review resolution

**NITPICK 1** — S-2/S-10 docstring duplication must be updated in lockstep if `MIN_CROSS_SECTION`/eligibility changes; only `TestNoDriftAgainstS10` catches divergence.

`REBUTTED` — that test *is* the intended mechanism, and it is the stronger of the two options. A shared constant would couple two strategies whose identities are deliberately independent (each hashes its own module into `strategy_version`), so extracting the rule would make an S-10 edit silently move S-2's version. A behavioural anti-drift test keeps them independent while making divergence fail loudly. Noting it here so the next reader knows the duplication is load-bearing rather than an oversight.

**NITPICK 2** — `('refused' if picked < 0 else picked):>9` mixes str/int under one format spec.

`REBUTTED` — stylistic, in an ad-hoc A/B script, and `>9` is defined for both types. Not worth a commit that would reset the review gate on an APPROVE'd PR.

Merging: 5/5 checks green, `CLEAN`, APPROVE on the head commit.

Luke-Bradford:
Review resolution — `9ab9ad88`. Codex checkpoint 3 ran over the rebuttal-only round.

**NITPICK 1 (S-2/S-10 duplication lockstep) — `REBUTTED` in part, `FIXED 9ab9ad88` in part.**

REBUTTED on the constants: `MIN_CROSS_SECTION` and the eligibility rule are *not* shared and
carry no lockstep requirement. S-2 cuts at `MIN_CROSS_SECTION = DECILE` (10) with
`ELIGIBILITY_BARS = 273`; S-10 cuts at 1000 and has no warm-up constant, deliberately — its own
comment says *"much larger than S-2's 10 on purpose"*, because S-2's floor protects a one-shot
decile cut while S-10's protects a retention band. Codex confirmed this half.

FIXED on the part that was right, which I had also overstated: my docstring claimed the function
body was the *only* duplication. Codex pointed out the rationale prose is duplicated too, and
`TestNoDriftAgainstS10` covers the rule but not the prose. The docstring now says exactly that
rather than implying full coverage.

**NITPICK 2 (`-1` sentinel formatted alongside real counts) — `FIXED 9ab9ad88`.**

Not rebutted. Codex's point stands: "purely stylistic" and "read-only script" are not rebuttals,
and the real defect underneath the formatting is that `-1` and a genuine `0` were being carried
in the same `int` and summed with `max(picked, 0)`. `selected()` now returns `int | None`, and
one `_row()` helper does the formatting, so "refused" and "selected zero" cannot be conflated.

**Two further defects Codex found in my own text — `FIXED 9ab9ad88`, neither raised by the bot:**

1. The diff quoted three different panel-size ranges (`0–11`, `4–117`, `~11`) as if they measured
   one population. They do not. `4–117` counts raw `price_daily` rows; `0–11` is the **decision**
   population (masked bars, past the 273-bar warm-up, above the $1 floor) which is the one
   `MIN_CROSS_SECTION` is compared against; `~11` was S-10's pre-existing figure. Worse, `5,747`
   was carried across from S-10's docstring and never measured on this population at all. The
   docstring and test header now state the A/B's decision-population numbers and say explicitly
   why the raw row count is the wrong number to quote. This is the repo's own
   *"never hardcode a derived statistic whose subject the reader has to infer"* rule, and I broke
   it while writing a prevention entry about an adjacent failure.
2. The S-10 docstring's pytest node id omitted the class, so it did not resolve. Now
   `tests/test_2797_s2_weekday_rebalance.py::TestNoDriftAgainstS10`.

**Codex ckpt-3 verdict:** *"Nothing I found should block merge."*

Re-verified at `9ab9ad88`: ruff clean, `ruff format --check` clean, pyright 0 errors, targeted
tests exit 0, and the full-population A/B re-run is **byte-identical** to the pre-refactor run
(`diff` reports no change), `[ab] OK`, exit 0.

Issue #2800, opened 2026-08-20
bug(ownership): AVAL's cover-page share count is the PREFERENCE class alone — denominator ~3.15x too small, and I18's sibling test cannot catch it


Luke-Bradford:
Found incidentally while taking the #2232 residual. Not fixed there — #2232's arm 3 suppresses `AVAL` for a *different* and larger defect (the FY2025 cover tag reads `7`), which masks this one rather than resolving it. When the issuer corrects that tag, `AVAL` starts rendering again against a denominator that is still wrong.

## What is wrong

`AVAL`'s stored denominator comes from `dei:EntityCommonStockSharesOutstanding`, and for this issuer that element carries the **preference class alone**, not the entity-wide total.

Measured this session against SEC EDGAR direct — the FY2025 20-F, accession `0001104659-26-044493`, report `R297.htm` ("EQUITY ATTRIBUTABLE TO OWNERS OF THE PARENT - Total outstanding shares"):

| line | value |
| --- | ---: |
| Total shares outstanding (all classes) | **23,743,475,754** |
| … ordinary shares | 16,178,324,920 |
| … preference shares | 7,565,150,834 |

Both class figures sum exactly to the total (`16,178,324,920 + 7,565,150,834 = 23,743,475,754`).

The cover series in `financial_facts_raw` tracks the **preference** column, and the arithmetic that proves it is the issuer's own conversion disclosure. `R298.htm` reports `CommonStockConversionToPreferredStock` of **500,000** for FY2024 and **22,887,579** for FY2025:

```
cover FY2023  7,541,763,255
            +       500,000   <- R298 FY2024 conversion
cover FY2024  7,542,263,255   <- matches exactly
            +    22,887,579   <- R298 FY2025 conversion
              7,565,150,834   <- R297 preference-share line
```

So every `AVAL` ownership percentage we have rendered was computed against ~7.54bn instead of ~23.74bn — **overstated by a factor of ~3.15**.

## Why the existing detectors miss it

`_detect_dual_class_denominator` (data-engineer I18, #1646) requires three gates and this issuer fails two of them:

1. **Gate 1 — denominator concept must be `us-gaap:CommonStockSharesOutstanding`.** The multi-class fingerprint I18 keys on is companyfacts stripping *every* dimensional cover value, forcing a us-gaap fallback. That does not happen here: this issuer publishes an **undimensioned** `dei` cover value which is nonetheless scoped to one class, so the fallback never triggers and the gate excludes it.
[...]

PR #2801, opened 2026-08-20
fix(#2232): reach the residual with a 13F pigeonhole, not a magnitude cut

Luke-Bradford:
Refs #2232. Refs #2800.

⚠ #2232 stays OPEN after this merges — deliberately, see "What this does not fix".

Takes the #2232 **residual** the 2026-08-08 run handed over: *"`AVAL = 7` and the other issuer-side XBRL errors that DO file a cover fact. No discriminator exists for them yet and magnitude is falsified; the honest next step is to look for one, not to widen this guard."*

## The residual is 2 instruments, not a class

Measured first, before designing anything. `PYTHONPATH=. uv run python -m scripts.audit_2232_denominator_residual`, full population, 4,451 live denominators:

```
arm 2 surface — some holder exceeds the denominator: 347  (no cover fact 27 · cover fact present 320)

  max_holder / denominator, by decade (empty bands are the evidence):
               1x –            10x :  276
              10x –           100x :   54
             100x –         1,000x :    5
           1,000x –        10,000x :    0   <-- EMPTY
          10,000x –       100,000x :    2
         100,000x –     1,000,000x :    5
       1,000,000x –    10,000,000x :    3
      10,000,000x –   100,000,000x :    2
```

Ten of the twelve above that gap are already suppressed by the shipped scope arm. The other two — `HQ` and `AVAL` — are the whole residual. **The gap is the finding**: it says the shipped conjunction already isolates the right cohort, so what remained was a boundary case rather than a population.

## Three discriminators died first, each on the full population

1. **ASC 260-10-55-12 retroactive restatement.** A real reverse split restates the EPS weighted-average count in every period presented, so cover-vs-weighted-average should agree for a split and diverge for a tagging error. It does — `GWAV` cover 829,631 = weighted average 829,631, same accession. It reaches neither residual instrument: `AVAL`, `MOB` and `PAVS` are 20-F IFRS filers with no `us-gaap` weighted-average concept at all, and `HQ`'s weighted average is 1, agreeing with its wrong cover. It false-positives too — 22 of the 320 exceed 10× because a ±200-day window straddles the pre-split filing (`PW` at exactly 10.0, `XPON` 10.9, `DBGI` 11.5).
2. **"the holder exceeds every share count the issuer has EVER reported."** Threshold-free, and fires on **134 of 4,451** — 131 at ≤54×, i.e. short-fact-history noise. Misses `AVAL` outright: largest holder 1,000,000 against a maximum historical cover of 15,240,124,702.
3. **An exchange listing-standard floor** (Nasdaq 5405(b)/5505(b) ≥1,000,000 publicly held shares; NYSE American ≥500,000). Published, citable, and still the falsified magnitude axis wearing a citation: of the 10 live sub-500,000 denominators with a cover fact, **8 are real** — `PW` 367,120 (a 1-for-10 reverse split, weighted average 3,672,274), `SPRC`, `RUBI`, `NEXR`, `SVRN`, `CMND`, `WORXD`, `PAVS`.

## What ships — arm 3, a pigeonhole over the one channel that supports it

`denominator_is_partial_class` gains a third, independent arm: **more 13F managers than shares outstanding**.

**Source rule.** Form 13F is additive across filers by design — Special Instruction 5 routes a manager whose holdings another reports to a 13F-NT (5a) or to a COMBINATION report covering only its own disjoint subset (5c), and Column 7 lists only the managers *"on whose behalf this Form 13F report is being filed"*. Column 5 reports a whole number of shares. So N distinct managers occupy N distinct whole shares, and a share count below N cannot be the whole company's.

The floor is **not chosen**: it is however many managers filed at all, an independent per-instrument observation. And it is a COUNT test on purpose — #2229 staleness, #2230 deemed attribution and #2231 unadjusted pre-split numerators all inflate share AMOUNTS while leaving the manager COUNT correct, so none of them can produce it. That is what keeps those three tickets measurable, which the 2026-08-08 run identified as the constraint.

Full population: fires on **3 of 4,451** — `AVAL` 88/7, `GLXY` 493/100, `MWH` 199/100. The last two are already suppressed by arm 1, so arm 3 reaches exactly one new instrument. Nearest non-firing instrument arm 1 does not already take is `NVR` at `managers/shares = 0.000398`, ~31,600× below `AVAL`.

## ⚠ The first version of this guard was unsound, and Codex checkpoint 2 caught it

It counted holders across all three channels. Both legs of the P1 were then confirmed on the data rather than argued:

- **Overlap.** Rule 13d-5(b)(1) deems every member of a group to own all the group's securities, so N distinct Section-16 / 13D-G identities legitimately restate ONE block. That is #2230's entire mechanism — the guard's premise was contradicted by an open ticket on the same tracker.
- **Fractions.** `_build_slice` enforces `shares > 0`, not `>= 1`, and the fractions are real:
  ```sql
  SELECT count(*) FILTER (WHERE shares <> trunc(shares)), min(shares) FILTER (WHERE shares > 0)
    FROM ownership_insiders_current;
  -- 13,017 of 170,941 rows fractional; smallest positive 0.0015
  ```
  Institutions and blockholders have 0 fractional rows and a minimum positive of exactly 1. The shipped count uses `shares >= 1` so the whole-share premise holds by construction rather than by today's data.

**Cost of the correction, stated rather than absorbed:** arm 3 no longer reaches `HQ`. It has ZERO 13F managers — all nine of its disclosed holders are Section 16 / 13D-G. The census prints that line explicitly (`known-bad, cover fact present, NOT reached by any arm: 1`), the docstring names it, and the truth table pins `(0 managers, 1 share) -> False` so widening the count back across all channels is a visible decision rather than a silent one.

## What this does not fix

- **`HQ`** — above. A cover, balance-sheet-issued and balance-sheet-outstanding all reporting `1` for a predecessor entity, with no 13F coverage to reason over.
- **#2800, filed from this work.** Even the pre-typo `AVAL` cover values are the **preference class alone**: `R297` splits 23,743,475,754 into ordinary 16,178,324,920 + preference 7,565,150,834, and the cover series moves by exactly the `CommonStockConversionToPreferredStock` figures `R298` reports (500,000 FY2024, 22,887,579 FY2025). So the denominator has been ~3.15× too small for years, and arm 3 masks that rather than resolving it. `_detect_dual_class_denominator` cannot reach it — gate 1 needs a us-gaap fallback and this cover value is undimensioned `dei`; gates 2/3 need ≥2 in-universe siblings.
- **#2232 stays open.** `HQ` is unreached and #2800 is unmeasured.

## Security model

Not applicable. Read-path only: one pure predicate and one pure counter over data already loaded for the rollup, no new query against user-controlled input, no new endpoint, no authorisation surface.

## Full-population A/B — paired, real control, matched base

Control = a second worktree at `origin/main`; both arms ran the same probe script at **`71a2ce5a`**, and both arms' base was verified identical after they drained (`main` has since moved to `566df70b`, which narrows what the A/B covers but does not void it).

```
control rows 529 · treatment rows 529
harness errors: 0

instruments compared: 529
instruments CHANGED : 1
  → newly suppressed (partial_class_denominator): 1
  → changed for ANY OTHER reason (must be 0)    : 0

--- SUPPRESSED (the loss side) ---
symbol        denominator    largest holder %  wedges lost  banner was
AVAL                    7         14,285,714%            3  unknown_universe

--- golden panel (must be byte-identical) ---
AAPL/GME/MSFT/JPM/HD  IDENTICAL on all nine compared fields
```

Population is 529 of 4,652 denominator-carrying instruments; the script prints the 4,123 exclusion on every run in both arms. It is the union of all three conditions' surfaces taken independently, and arm 3's surface is deliberately **wider** than the guard (every channel's raw `_current` rows, before owner-once collapse) so over-firing would show.

⚠ **The first A/B run was void and is reported rather than dropped.** It was launched with the control detached at `origin/main` before PR #2792 merged, and the branch based on the commit before that. The compare returned **294 instruments changed for another reason, including all five panel instruments** — `MSFT` insiders 2,621,605 → 459,504,753 — which is #2792's retention bound read backwards, not this diff. Rebased, both arms re-run at one base, result above. Lesson landed in the prevention log.

## Revert-probes — 6/6 CAUGHT

| probe | verdict |
| --- | --- |
| arm 3 removed | CAUGHT |
| arm 3 boundary `>` → `>=` | CAUGHT |
| arm 3 moved below the per-class short-circuit | CAUGHT |
| institutional-category filter removed (Section 16 + 13D/G counted again) | CAUGHT |
| `pie_wedge` basis filter removed | CAUGHT |
| whole-share bound relaxed `>= 1` → `> 0` | CAUGHT |

⚠ Two of these needed the harness fixed before they meant anything, both logged:

- The first harness ran `pytest -x`, and under this repo's `-n 4` xdist config every mutation returned exit **2** — `EXIT_INTERRUPTED`, which is also what a collection error returns. Scored as CAUGHT, it would have been 5/5 on no evidence. Re-run without `-x`, gating on exit **1** exactly; `addopts = "-q … -n 4"` emits no `N failed` line to parse.
- The `pie_wedge` probe came back NOT CAUGHT because every memo overlay today carries a category outside `_INSTITUTIONAL_CATEGORIES`, so no fixture reached the filter. Corrected by constructing an `institutions`-category slice with `denominator_basis="institution_subset"` — the contract, not the coincidence. (An earlier `shares > 0` filter came back NOT CAUGHT for the opposite reason — unreachable behind `_build_slice`'s own drop since #1916 — and was deleted, with the upstream guarantee asserted instead.)

**Clause 8 — golden panel**, dev DB, on the guard's own input. Arm 3's ratio is ~6 orders of magnitude below firing on all five, and the A/B confirms every rendered field is byte-identical:

| symbol | shares outstanding | as of | 13F managers (≥1 share) | managers/shares |
| --- | ---: | --- | ---: | ---: |
| AAPL | 14,594,180,000 | 2026-07-17 | 6,930 | 0.000000475 |
| GME | 448,691,257 | 2026-06-05 | 655 | 0.000001460 |
| MSFT | 7,425,545,491 | 2026-07-23 | 7,156 | 0.000000964 |
| JPM | 2,658,186,195 | 2026-06-30 | 5,930 | 0.000002231 |
| HD | 997,116,682 | 2026-05-19 | 5,130 | 0.000005145 |

**Clause 9 — cross-source, SEC EDGAR direct.** Three independent fetches this session, none of them our store:
- `www.sec.gov/Archives/edgar/data/1504764/000110465926044493/R1.htm` — SEC's own rendering of the cover: `Entity Common Stock, Shares Outstanding` = **7**, undimensioned. So the value is the issuer's tag, not a §7.17 stripping artefact and not our parse.
- `data.sec.gov/api/xbrl/companyconcept/CIK0001504764/dei/EntityCommonStockSharesOutstanding.json` — serves the same **7** at 2025-12-31, against **7,542,263,255** at 2024-12-31.
- `…/R297.htm`, the same filing's audited equity note — total shares outstanding **23,743,475,754** (ordinary 16,178,324,920 + preference 7,565,150,834, summing exactly).

**Clauses 10 + 11 — operator.** No backfill and no migration: the guard runs at read time over data the rollup already loads, and no stored value changes. See `## OPERATOR VERIFY`.

**Clause 12 — SHA** recorded in the OPERATOR VERIFY block below.

## OPERATOR VERIFY

⚠ **Do NOT run `POST /jobs/sec_rebuild/run` for this.** There is nothing to re-ingest; a rebuild would cost hours and change nothing.

**0. Restart the jobs daemon** (`stack: jobs`). Not strictly required — the change is read-path only and the API runs `--reload` — but the daemon holds `ownership_rollup` in memory and would otherwise serve the old branch on any job that reads it.

**1. Panel — must be UNCHANGED.**

```bash
for s in AAPL GME MSFT JPM HD; do
  curl -s "http://localhost:8000/instruments/$s/ownership-rollup" \
    | jq -c --arg s "$s" '{sym:$s, reason:.no_data_reason, outstanding:.shares_outstanding,
        wedges:([.slices[]|select(.denominator_basis=="pie_wedge")]|length)}'
done
```

`reason` must be `null` on all five, with 4 wedges each. ⚠ The `outstanding` figures drift as filings land (`JPM` read 2,679,511,418 on 2026-08-07 and 2,658,186,195 now) — a different value is not a regression; a non-null `reason` is.

**2. The one instrument this PR moves.**

```bash
curl -s "http://localhost:8000/instruments/AVAL/ownership-rollup" \
  | jq -r '"\(.banner.state) \(.no_data_reason) slices=\(.slices|length) outstanding=\(.shares_outstanding) as_of=\(.shares_outstanding_as_of)"'
```

Expect `no_data partial_class_denominator slices=0 outstanding=null as_of=2025-12-31`. On `main` today it renders a three-wedge pie whose largest single holder is **14,285,714%** of the company.

**3. Eyeball one chart.** `http://localhost:5173` → `AVAL` → ownership. Expect the empty state and the existing suppressed-denominator caption ("…does not cover the whole company — a single disclosed holder already reports more shares than it contains…"). That copy is unchanged by this PR and is true for `AVAL`: 1,000,000 disclosed against 7.

**4. Acceptance query** — re-runnable, and the authority for every count above:

```bash
PYTHONPATH=. uv run python -m scripts.audit_2232_denominator_residual
```

Expect `arm 3 surface — more 13F managers than shares: 3 of 4451` (`AVAL` 88/7, `GLXY` 493/100, `MWH` 199/100), `→ of those, NOT already reachable by arm 1+2: 1`, and `known-bad, cover fact present, NOT reached by any arm: 1` (`HQ`).

---

_CI note (2026-08-21): the `verify-issue-link` red on this PR is a stale event payload, not the current body — the gate script passes locally against the live body. Re-firing the `edited` event to clear it. See the resolution comment for the reproduction._


Luke-Bradford:
Resolution round — TA loop picking this PR up; the ownership loop that authored it is paused (`.ebull-ownership/var/autonomy/PAUSE`, written 2026-08-20 23:11, i.e. minutes after the review posted).

### NITPICK — `Decimal(additive_institutional_holders)` for consistency

`REBUTTED` — `int > Decimal` is an exact comparison in CPython, not a float coercion. `decimal.Decimal.__gt__` converts the `int` operand exactly (`_convert_for_comparison`), so `additive_institutional_holders > outstanding` at `app/services/ownership_rollup.py:4377` cannot lose a digit however large either side is — which is the property this arm depends on, since the whole point is a share count small enough to be exceeded by a manager count. Wrapping it adds a conversion that reads as if it were load-bearing when it is not. The module's explicit `Decimal(...)` calls elsewhere are all sites that go on to do *arithmetic*, where the mixed-type result type matters; this one is a pure predicate.

### The two red signals on this PR are both STALE GitHub state, not defects

1. **`verify-issue-link` fail** — the current body passes. Reproduced locally against the live body with the gate's own script:

   ```bash
   PR_TITLE="$(gh pr view 2801 --json title --jq .title)" \
   PR_BODY="$(gh pr view 2801 --json body --jq .body)" \
   bash scripts/check_pr_issue_link.sh
   # PR body links every title-referenced issue.  exit=0
   ```

   The workflow reads `github.event.pull_request.body` (`.github/workflows/pr-issue-link.yml:33`), i.e. the body **as of the triggering event**. The negation was edited out after the failing run, and `gh run rerun` replays the original payload, so a rerun can never clear it — only a fresh `edited`/`synchronize` event can.

2. **`mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`** — not reproducible. Against the identical base (`232978cd`, local `origin/main` == `gh api .../commits/main`):

   ```bash
   git merge-tree --write-tree origin/main origin/fix/2232-pigeonhole-denominator
   # f450d7d4486ad644953c9cb7854ac9767f53d50a   exit=0
   ```

   A clean tree with exit 0 is a clean merge. GitHub's mergeability is computed lazily and this value is stale.

Re-firing the `edited` event now to clear (1) and nudge (2). The review APPROVE stands on `0eacab43`, which is still the head commit — a body edit does not move the SHA, so this does not reset the review requirement.

Luke-Bradford:
Correction + handoff — the `edited` re-fire did **not** work, and the remaining unblock is one push.

### What the re-fire actually did

Nothing. `gh pr edit --body-file` changed the body, but no `pr-issue-link` run was created for this branch:

```bash
gh run list --workflow=pr-issue-link.yml --limit 6 \
  --json databaseId,headBranch,event,createdAt --jq '.[] | "\(.databaseId) \(.event) \(.headBranch) \(.createdAt)"'
# ...no fix/2232-pigeonhole-denominator row, before or after the edit
```

The workflow does declare `types: [opened, edited, synchronize, reopened]` (`.github/workflows/pr-issue-link.yml:18`), so this is GitHub declining to schedule the run, not a trigger-config gap. The most likely cause is the second symptom below: GitHub cannot build `refs/pull/2801/merge` for a PR it believes is conflicting, and `pull_request` events for such a PR are dropped.

### `CONFLICTING` is wrong — verified with a real merge, not just `merge-tree`

The earlier comment cited `git merge-tree --write-tree` (exit 0). Because that only proves the absence of *content* conflicts, it was re-tested as an actual merge against current `origin/main` (`232978cd`, identical to `gh api .../commits/main`):

```bash
git worktree add --detach /tmp/mt2801 origin/main
cd /tmp/mt2801 && git merge --no-commit --no-ff origin/fix/2232-pigeonhole-denominator
# Auto-merging docs/review-prevention-log.md
# Automatic merge went well; stopped before committing as requested   exit=0
# 6 files staged, no conflict markers
```

So `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` is stale GitHub state on a branch that merges cleanly.

### The remaining work — one push, then the normal gate

Both symptoms clear on a `synchronize` event. The branch is checked out in the **paused** ownership worktree (`/Users/lukebradford/Dev/.ebull-ownership`, `PAUSE` sentinel written 2026-08-20 23:11), so this is that worktree's push to make — running it there also keeps the pre-push hook on the venv that has the deps:

```bash
cd /Users/lukebradford/Dev/.ebull-ownership
git fetch origin && git rebase origin/main          # clean; verified above
uv run ruff check . && uv run ruff format --check . && uv run pyright
uv run pytest -m "not db" && uv run pytest tests/smoke
git push --force-with-lease
```

Acceptance: `verify-issue-link` goes green (the current body already passes the gate script locally — see the previous comment for that reproduction), `mergeable` recomputes to `MERGEABLE`.

⚠ The force-push moves the head SHA, so the existing APPROVE on `0eacab43` no longer covers it — the review bot must APPROVE the new SHA before `safe_merge.sh 2801`. The one NITPICK from that round is already resolved as `REBUTTED` in the previous comment and needs no code change.

Not doing that push from here: this is the TA loop, the branch belongs to a sibling worktree, and force-pushing a ref another worktree has checked out would leave that worktree's local branch diverged for whoever resumes it. Everything above is the whole of the remaining work.

PR #2805, opened 2026-08-21
fix(#2274): defer a jobs-daemon reload past a job that is provably alive

Luke-Bradford:
Refs #2274

Scoped to item 2 of that ticket (`dev_reload` reaping long jobs). Item 1 (wall-clock ceiling / no-progress watchdog) stays open, so this uses `Refs`, not `Closes`.

## Why now

This is the live blocker on the strategy build queue's item 6. The first in-sample backtest at current strategy versions died three times inside one hour.

| UTC | event |
| --- | --- |
| 00:17:05 | `dev_reload` respawns the child |
| 00:17:41 | `strategy_backtest_run` **111447** starts (17,290 targets) |
| 00:57:25 | `app/api/strategies.py` mtime changes — merge `1ab1704c` re-detaching `~/Dev/eBull` at `origin/main` — SIGTERM |
| 01:00:25 | child cannot drain in 180s (43 min into a ~17h run) → **SIGKILL** |
| 01:00:28 | boot reaper marks 111447 `failure: orphaned` at 12,625/17,290 |

Log timestamps are BST and `job_runs` is UTC; the offset is applied above. Those two clocks living in different stores is a large part of why three consecutive reaps read as mysterious worker deaths rather than as one repeating cause.

**The run was healthy when it was killed.** 111447's `last_progress_at` is `01:00:20` — 8 seconds before the SIGKILL. Not a wedged job being cleaned up; a working one being preempted.

The trigger each time was an ordinary merge, which is exactly how a merged change is *meant* to reach this daemon. So the loop's own merge cadence made a long run unfinishable, and 7 of 10 strategies stay unmeasured until it is fixed.

## What changed

An automatic reload now **waits** for a job that is provably alive instead of killing it.

"Provably alive" is a claim the job volunteers, not one the supervisor infers: a `job_runs` row in `status='running'` whose `last_progress_at` is inside that job's own staleness threshold. Deliberately **not** an age cut — age is a proxy for liveness and would protect a wedged job forever. A job that stops heartbeating goes stale and stops blocking reloads on its own, which is why this needs no watchdog and cannot false-fire on a legitimately-long run — #2274's own stated constraint.

Also a **correctness** fix, not only an availability one: had a drain mid-run actually succeeded, the surviving rows would carry a `strategy_version` derived from code that changed underneath the run.

Load-bearing properties:

- **Fails open.** Any probe failure — DB down, slow, schema drift — reloads exactly as before. Bounded by both `connect_timeout` *and* a libpq startup `statement_timeout`, because the former bounds only the connect and a hung query on the supervisor's own thread is a wedge wearing fail-open's clothes.
- **Thresholds are not invented and not local.** They come from `app/services/processes/stale_thresholds.py`, the registry the admin console's `mid_flight_stuck` detection already reads (default 300s, 1800s for eight slow-tick producers). They travel into the query as a jsonb map so Postgres applies the per-job cut **before** any row limit.
- **Ownership is free.** The singleton advisory-lock fence means exactly one jobs process, so a freshly-heartbeating `running` row is necessarily the current child's. No PID plumbing.
- **Only automatic reloads defer.** An explicit stop still drains and kills on the budget.
- **A child that dies holding pending changes respawns on them**, and does not probe first — a dead child's own `running` rows stay inside the staleness window and would otherwise make a crash wait out a heartbeat that can never advance again.

## Answering the ticket's own open question

> worth checking whether the progress heartbeat is actually reaching `job_runs` for this job

Measured on the full table:

```sql
select count(*) from job_runs where last_progress_at is not null;  -- 8
select count(*) from job_runs;                                     -- 111409
```

8 of 111,409, four of them `strategy_backtest_run`. The heartbeat does **not** reach `job_runs` for the routine jobs, but it works reliably for the backtest — which is precisely the job that needed protecting. The other 111,401 rows behave exactly as they did pre-change.

⚠ My first attempt at this measurement was wrong and self-consistently so: `count(processed_count)` on a `NOT NULL DEFAULT 0` column returns the row count whatever the data says, so it read as "the heartbeat is universal".

## Why not the option the ticket called cheapest

#2274 lists "accept that multi-hour sweeps must run outside a dev session and write it in the runbook" as cheapest and possibly correct. It is no longer correct, and the evidence is the table above: the autonomy loop merges continuously, the backtest needs ~17h, three runs died in one hour. Under that option there is no window in which item 6 can ever complete — it would convert a fixable ops defect into a permanent block on measuring 7 of 10 strategies.

## Codex checkpoint 2 — six rounds

Every finding was real except the last. Recording them because several are non-obvious:

1. **P1 — a crash while deferred discarded the held changes**, leaving the daemon dead until an unrelated edit landed. Not hypothetical: the jobs child took a bare SIGKILL (`child exited -9`, no preceding SIGTERM) mid-backtest on the same night. Now respawns on the pending batch.
2. **P2 — `connect_timeout` bounds only the connect.** A hung query would block the supervisor's thread entirely. Added a libpq startup `statement_timeout`, verified empirically: `SHOW statement_timeout` → `3s`, and `SELECT pg_sleep(10)` → `QueryCanceled` (which the probe's guard turns into "no live job", so it still fails open).
3. **P1 — hard-coding the 300s default** would have made the eight slow-tick producers "healthy" to the admin console and "stale" to the supervisor at the same instant. Rewired to the existing registry.
4. **P1 — filtering per-job thresholds in Python after a `LIMIT`** let stale fast-tick rows mask a live slow-tick one, returning "nothing to defer". Pushed the per-job cut into SQL so `LIMIT 1` only ever picks *which* blocker to name, never *whether* one exists.
5. **P2 — the probe→SIGTERM race.** A job starting in that window would still be killed at the budget. Expiry now re-probes and extends while something is alive. This fixes the original bug from the other side as well: a fixed 180s cap is what turns a graceful shutdown into data loss, because the drain a long job needs is longer than any constant it is safe to wait on blindly.
6. **P1 — an operator stop must stay bounded.** The shutdown handler runs on the watcher's frame, so a Ctrl-C arriving while an automatic reload was already inside `_stop_child` was invisible to the extension loop and the stack became unstoppable. `stop_event` is threaded in.
7. **P1 — REBUTTED.** Codex read the registry's admin `process_id` keys against `job_runs.job_name` and called the mismatch a defect. It is not: `process_id` **is** the `ScheduledJob.name` verbatim for everything that owns a job_runs row. The three keys that do not match (`bootstrap`, `sec_13f_sweep`, `nport_sweep`) are exactly the three that are not scheduled jobs and can never produce one — the registry's own docstring says sweeps "have no own active_run". Measured against both `SCHEDULED_JOBS` and the distinct `job_name` values in `job_runs`: all five keys that can appear do match. The proposed translations `sec_13f_quarterly_sweep` / `sec_n_port_ingest` appear in neither, so adopting them would add dead keys implying coverage that does not exist. A reproduce command is recorded in the constant's comment.

## Tests

19 pure-logic tests on the fast gate; 6 DB-backed tests in a separate module.

The split is deliberate and was itself a caught regression: `tests/conftest.py::pytest_collection_modifyitems` applies the `db` marker **per module**, so adding one DB test to the pure file silently deselected all 9 pre-existing tests from the `-m "not db"` pre-push gate. It surfaced as `EXIT=5` ("no tests collected"), which greps for `FAILED` cannot distinguish from success. The pure module carries a `⚠ Keep this module DB-free` note so it does not recur.

The DB module exists because `live_job` swallows every exception by design — a typo in its SQL would degrade the probe to a permanent `None`, silently disabling the feature while every pure test still passed.

## Verification

- `ruff check` / `ruff format --check` / `pyright` (0 errors) / `pytest -m "not db"` / `tests/smoke` — all green at `12404364`.
- New SQL executed unguarded against the dev DB (outside `live_job`'s exception guard) to prove it parses and runs.
- jsonb per-job lookup verified read-only on dev: override → `1800`, default-threshold job → `300`, unknown job → `300`.
- `statement_timeout` verified enforced, not merely set (see round 2 above).
- DB-backed tests exercise real `job_runs` rows, including the stale-masks-live regression with the stale row seeded as the *fresher* heartbeat.

⚠ **Not yet verified live, and deliberately so.** The end-to-end acceptance is: restart the daemon on this commit, dispatch the backtest, touch an `app/**` file, and confirm the log says `DEFERRING reload` rather than `SIGTERM`. That cannot be done before this merges, because the daemon runs from `~/Dev/eBull` at `origin/main` — and dispatching a 17h backtest under the *current* code would just feed another run to the bug. Sequencing is: merge → re-detach → restart daemon → dispatch → verify. I will run it and report on the issue.

## Security

No security surface. The change adds one read-only `SELECT` against `job_runs` from a dev-only supervisor that is a documented pass-through outside `DEV_LIKE_ENVS`. Parameterised throughout — the threshold registry is passed as a jsonb parameter, never interpolated. No new credentials, endpoints, or external calls.

## Risk

The failure mode of this change is "the daemon runs slightly older code for a while", against a current failure mode of "hours of corpus work destroyed". Deferral is unbounded in wall-clock but bounded by liveness at both decision points, so nothing can hold the daemon on stale code indefinitely without also heartbeating.

### DIFF
--- a/app/jobs/dev_reload.py
+++ b/app/jobs/dev_reload.py
@@ -32,10 +32,53 @@ is fully reaped — releasing the lock — BEFORE the replacement is
 spawned. Never spawn-then-kill. After a SIGKILL escalation we also
 settle for ``_POST_KILL_SETTLE_S`` so Postgres notices the dead session
 and drops the lock, mirroring ``stack-restart.sh``'s ``kill_jobs_process``.
+
+Deferring a reload past a live job (#2274)
+------------------------------------------
+An automatic reload used to preempt the child unconditionally. A job
+that cannot drain inside ``_DRAIN_TIMEOUT_S`` — which a multi-hour
+corpus run or backtest never can — was therefore SIGKILLed, and the
+next boot's reaper wrote it off as ``orphaned: reaped at boot``. On
+2026-08-21 that destroyed three consecutive ``strategy_backtest_run``
+attempts inside one hour; the last died 8 seconds after its own
+heartbeat, i.e. while healthy and 12,625 of 17,290 targets through.
+The trigger each time was an ordinary merge re-detaching the checkout
+at ``origin/main``, which is exactly how a merged change is *meant* to
+reach this daemon — so the loop's own merge cadence made a long run
+unfinishable.
+
+So a reload now WAITS for a job that is provably alive instead of
+killing it. "Provably alive" is a claim the job volunteers, not one we
+infer: a ``job_runs`` row in ``status='running'`` whose
+``last_progress_at`` is inside that job's own staleness threshold, as
+already defined for the admin console by
+``app/services/processes/stale_thresholds.py``. Deliberately NOT an age
+cut — age is a proxy for liveness and would protect a wedged job
+forever. A job that stops heartbeating goes stale and stops blocking
+reloads on its own, which is why this needs no watchdog and cannot
+false-fire on a legitimately-long run (#2274's own constraint).
+
+This is a correctness fix as much as an availability one. Had a drain
+mid-run actually succeeded, the surviving rows would carry a
+``strategy_version`` derived from code that changed underneath the run.
+
+Three properties are load-bearing:
+
+* **Fail open.** Any probe failure — Postgres down, slow, schema drift —
+  reloads as before. A supervisor must never be blockable by its own
+  liveness check, or a DB blip freezes the daemon on stale code.
+* **Ownership is free.** The singleton fence means there is exactly one
+  jobs process, so a freshly-heartbeating ``running`` row is necessarily
+  the current child's. No PID plumbing.
+* **Only AUTOMATIC reloads defer.** An explicit stop (SIGTERM to the
+  supervisor, operator stopping the task) still drains-then-kills
+  immediately. Deferral is about not letting a file-watcher preempt
+  hours of work; it is not a veto on the operator.
 """
 
 from __future__ import annotations
 
+import json
 import logging
 import os
 import signal
@@ -48,6 +91,11 @@ from pathlib import Path
 from typing import Final
 
 from app.config import DEV_LIKE_ENVS, settings
+from app.services.processes.stale_thresholds import (
+    DEFAULT_THRESHOLD_S,
+    get_threshold,
+    overridden_process_ids,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
NITPICK — `FIXED d4c358ea`. The crash-respawn comment named `_LIVE_JOB_STALENESS_S`, which round 3 replaced with the per-job lookup against `stale_thresholds`. Now reads "its own threshold". Correct catch: a dead constant name in a comment is exactly the kind of stale reference that misleads the next reader, since it implies a single global window where the code applies eight different ones.

Issue #2807, opened 2026-08-21
fix: owned P&L drops off the strategy card after a scan-version rotation while staying in the capital base


Luke-Bradford:
A paper deployment's owned P&L vanishes from its `/strategies/overview` card as soon as the scan version rotates, while the same P&L stays counted in the capital base. Pre-existing; surfaced by Codex checkpoint 2 on #2806 and deliberately left out of that fix as a definitional change.

## The asymmetry

`load_owned_pnl` is called with `scan_pnl_versions` = current scan versions ∪ every `strategy_deployments.strategy_version`, and returns rows keyed by `strategy_signals.strategy_version`. Two consumers read that one dict:

- `app/api/strategies.py` — the per-strategy card, at a **single** key (`scan_key`, the current scan version).
- `app/services/strategy_monitoring.py::realised_pnl_for_keys`, at **every** `paper_deployment_key`.

So after a rotation the deployment's realised P&L still lands in the capital base — which is correct, and `load_paper_realised_pnl`'s docstring states the rule: *"Old strategy versions remain part of the shared pot after they are retired; limiting this calculation to the current manifest would make realised gains or losses disappear from the capital base."* — but the card silently shows `StrategyPnl()` for it.

## Why it will bite

Scan versions rotate often on this repo — `strategy_scan_watermark` holds 28 rows across 10 strategies, i.e. several rotations per strategy inside a fortnight, because the identity hash moves whenever the registry does. Any position opened before a rotation is orphaned from its own card the moment one happens.

Not yet observable: `strategy_deployments` has 0 rows.

## Fix shape

Sum `StrategyPnl` across every version returned for a `strategy_id` rather than reading one key. That needs a decision the basis fix deliberately did not take:

- the counts and cash sums (`strategy_trade_count`, `invested_capital`, `realised_pnl`, `unrealised_pnl`, `observed_fees`) add cleanly — dollars pool across rule-set versions in a way rates do not, so #2670's *"pooling two would average two arithmetics into one rate"* does not bar this;
- `complete` / `incomplete_reasons` need a stated combination rule (almost certainly `all()` and set-union, i.e. fail-closed), and that rule is the actual design question.

Acceptance: with a deployment at a prior scan version and an open position, the card's P&L equals the capital base's contribution for that strategy.

Refs #2806.

Luke-Bradford:
[...]

Issue #2803, opened 2026-08-21
The /strategies scan card can never show a scan: it filters live-scan watermarks by backtest-basis versions


Luke-Bradford:
## Symptom

`/strategies/overview` reports every strategy as `scan.status: "rotated"`, `frontier_date: null`,
`fired_entries: 0` — i.e. "this strategy has never scanned" — immediately after a
`strategy_signal_scan` run that wrote **81,485** rows (run `111454`, 2026-08-21).

The operator's Strategies page therefore shows a dead product while the scan is working correctly.

## Cause

`app/api/strategies.py:1003` `_current_versions()` builds identities on the **backtest** basis:

```python
entry.identity(universe=BACKTEST_UNIVERSE, cost_model_id=COST_MODEL_ID)   # survivorship_free
```

That one version set is bound as `%(versions)s` for both the result queries (correct — stored
results *are* on the backtest basis) **and** three queries that read **live-scan** artefacts:

| query | reads | filter |
| --- | --- | --- |
| `_SCAN_SQL` (1185) | `strategy_scan_watermark`, `strategy_signal_daily_counts` | `w.strategy_version = ANY(%(versions)s)` |
| `_EXCLUSIONS_SQL` (1253) | `strategy_signal_daily_counts` | `strategy_version = ANY(%(versions)s)` |
| `_PRIOR_VERSION_SCANS_SQL` (1242) | `strategy_scan_watermark` | `NOT EXISTS (… cur.version = w.strategy_version)` |

But the live scan stamps the **scan** basis —
`app/services/strategy_signal_scan.py:484`, `SCAN_UNIVERSE = "survivor_only"` (line 88).

`universe` is part of `StrategyIdentity`, so the two bases produce disjoint hashes.

## Full-population verification

```
SCAN_UNIVERSE = survivor_only | BACKTEST_UNIVERSE = survivorship_free
overlap between backtest-basis and scan-basis version sets: 0
watermark rows matching backtest-basis versions (what the panel queries): 0
watermark rows matching scan-basis versions (what the scan writes):      10
total watermark rows: 28
```

Not a sample — the intersection is empty by construction, so `_SCAN_SQL` can **never** match a
watermark. The scan card is permanently `rotated`/`never_run` with zero counts no matter how often
the scan runs, and `_PRIOR_VERSION_SCANS_SQL` then classes the genuinely-current scan version as
"prior", which is what synthesises the misleading `rotation` block.

## Fix

[...]

PR #2808, opened 2026-08-21
fix(#2806): read fire rate, attribution and owned P&L on the scan identity basis

### LINKED ISSUE #2806: fix: fire-rate, attribution and owned-PnL still read the result identity basis on live-scan relations (#2803 residual)
`/strategies/overview` reports `share_unavailable_reason: "never_scanned"` for **all 10 strategies**, hours after a scan that wrote to all 10. Same root cause as #2803 — the result identity basis bound to a live-scan relation — on the three loaders that #2803 did not reach.

## What the operator sees

```bash
curl -s -b "ebull_session=..." localhost:8000/strategies/overview \
  | jq -r '.strategies[] | "\(.strategy_id) \(.fire_rate.scanned_days) \(.fire_rate.share_unavailable_reason)"'
```

```
s1-time-series-momentum                0 never_scanned
s10-relative-strength-leader           0 never_scanned
s2-cross-sectional-momentum            0 never_scanned
s3-mean-reversion-in-trend             0 never_scanned
s4-volatility-compression-breakout     0 never_scanned
s5-support-bounce                      0 never_scanned
s6-resistance-breakout                 0 never_scanned
s7-trend-pullback                      0 never_scanned
s8-range-mean-reversion                0 never_scanned
s9-squeeze-expansion                   0 never_scanned
```

Every one of those had in fact scanned: `strategy_scan_watermark` carries frontier `2026-08-19` for all 10, written `2026-08-21T00:38Z`. #2803 fixed the *scan card* so the frontier and rotation block now read correctly; the *fire-rate* block on the same card still says the strategy has never run.

## Root cause

`_current_versions()` (`app/api/strategies.py:1004`) builds identities on `BACKTEST_UNIVERSE` (`survivorship_free`); `strategy_signal_scan` stamps `SCAN_UNIVERSE` (`survivor_only`, `strategy_signal_scan.py:88`). `universe` is part of `StrategyIdentity`, so the two version sets are disjoint.

#2803 added `_current_scan_versions()` and rebound the three SQL constants that live **inside** `app/api/strategies.py` and are executed as `cur.execute(_X_SQL, scan_params)`. It did not reach the three loaders in `app/services/strategy_monitoring.py`, which receive versions as a **function argument** rather than through the shared `params` dict, and are all still called with `version_values`:

| call site | loader | scan relation it filters |
| --- | --- | --- |
| `app/api/strategies.py:1553` | `load_attribution` | `strategy_signals s` — `s.strategy_version` |
| `app/api/strategies.py:1560` | `load_fire_rate` | `strategy_signal_daily_counts` |
| `app/api/strategies.py:1562` | `load_owned_pnl` | `strategy_signals s` — `s.strategy_version` |

The lookup key at `app/api/strategies.py:1707` (`key = (strategy_id, versions[strategy_id])`) is on the result basis too, so it must move for the same three.

`_CONTROL_SQL` reads `strategy_results_store` / promotions / deployments and is correctly on the result basis. It is not in scope.

## Full-population verification (dev, before any fix)

```
version-set intersection (backtest vs scan basis):                    0

strategy_signals                total= 56835  backtest_basis=     0  scan_basis= 33655
strategy_signal_daily_counts    total=   428  backtest_basis=     0  scan_basis=   216
strategy_deployments rows:      0
```

Zero, not "few" — the predicate cannot match, so this is a permanent empty rather than an intermittent one. `load_attribution` and `load_owned_pnl` are latently in the same state: both currently return nothing regardless, and with `strategy_deployments` at 0 rows there is no trade for them to mis-report yet, so the defect would first become visible as a silently-zero P&L on the first paper fill.

## Why the existing guard did not catch it

`tests/test_2803_scan_card_version_basis.py::test_every_scan_relation_query_is_executed_on_the_scan_basis` already lists `strategy_signal_daily_counts` in `_SCAN_RELATIONS`. It scans `inspect.getsource(get_strategy_overview)` for `cur.execute(NAME, params)` pairs — a shape the three loaders do not have, because they bind versions through a keyword argument in another module. The guard knows the relation matters and cannot see the call.

## Fix

1. Bind `_current_scan_versions()` to `load_attribution` / `load_fire_rate` / `load_owned_pnl`, and key their lookups on the scan version.
2. Extend the #2803 guard to cover loader call sites, not only `cur.execute` pairs — derive the scan-relation set from the loaders' own SQL so a fourth loader cannot be added on the wrong basis silently.

Refs #2803.

Luke-Bradford:
Closes #2806.
Refs #2803. Refs #2807.

`/strategies/overview` reported `share_unavailable_reason: "never_scanned"` for **all ten strategies** while `strategy_scan_watermark` carried frontier `2026-08-19` for all ten, written the same morning. #2803 fixed the scan card; the fire-rate block on that same card still told the operator the strategy had never run.

## Root cause — the #2803 defect, one module over

`_current_versions()` builds identities on `BACKTEST_UNIVERSE` (`survivorship_free`); `strategy_signal_scan` stamps `SCAN_UNIVERSE` (`survivor_only`). `universe` is part of `StrategyIdentity`, so the two version sets are **disjoint**.

#2803 rebound the three queries executed as `cur.execute(_X_SQL, scan_params)` **inside** `app/api/strategies.py`. It could not reach the three loaders in `app/services/strategy_monitoring.py`, which take their versions as a **keyword argument** — and all three filter a relation the live scan writes:

| call site | loader | scan relation filtered |
| --- | --- | --- |
| `app/api/strategies.py:1561` | `load_attribution` | `strategy_signals s` — `s.strategy_version` |
| `app/api/strategies.py:1569` | `load_fire_rate` | `strategy_signal_daily_counts` |
| `app/api/strategies.py:1584` | `load_owned_pnl` | `strategy_signals s` — `s.strategy_version` |

`load_control_state` reads `strategy_results_store` / promotions / deployments and is correctly on the result basis; it is untouched.

Both halves had to move. The lookup at `app/api/strategies.py:1729` was `key = (strategy_id, versions[strategy_id])` — the result basis — so rebinding the query alone would have changed nothing visible. **A result-basis filter and a result-basis dict lookup fail closed to the same empty card**, which is why fixing one and not the other is indistinguishable from fixing neither.

## Full-population verification (dev, before the fix)

```
version-set intersection (result vs scan basis):                       0

strategy_signals              total 56,835   result-basis 0   scan-basis 33,655
strategy_signal_daily_counts  total    428   result-basis 0   scan-basis    216
strategy_deployments rows:                                             0
```

Zero, not "few" — the predicate could never match, so this was a permanent empty rather than an intermittent one.

## A/B of the three loaders on dev, both bases

```
BEFORE (result basis): fire_rate keys=0   attribution keys=0   owned_pnl keys=0
AFTER  (scan basis)  : fire_rate keys=10  attribution keys=8   owned_pnl keys=0
```

`owned_pnl` staying at 0 is correct: `strategy_deployments` has no rows, so there is no owned position to report.

Fire rate after the fix, all ten:

```
  s1-time-series-momentum              days=7 fired_days=5 fired=8821 evaluable=16838 share=0.5239
  s10-relative-strength-leader         days=2 fired_days=0 fired=   0 evaluable= 5689 share=0.0000
  s2-cross-sectional-momentum          days=2 fired_days=0 fired=   0 evaluable= 3277 share=0.0000
  s3-mean-reversion-in-trend           days=7 fired_days=5 fired=  25 evaluable=16788 share=0.0015
  s4-volatility-compression-breakout   days=7 fired_days=5 fired= 748 evaluable=27315 share=0.0274
  s5-support-bounce                    days=2 fired_days=1 fired= 244 evaluable= 5589 share=0.0437
  s6-resistance-breakout               days=2 fired_days=1 fired=  25 evaluable= 4015 share=0.0062
  s7-trend-pullback                    days=7 fired_days=5 fired= 167 evaluable=16788 share=0.0099
  s8-range-mean-reversion              days=7 fired_days=5 fired=  24 evaluable=27862 share=0.0009
  s9-squeeze-expansion                 days=7 fired_days=3 fired=   8 evaluable=27922 share=0.0003
```

`share_unavailable_reason` is now `None` on every row rather than `never_scanned` on every row. s2 and s10 report `0.0000` **fired** rather than "never scanned", which is the honest state: both rebalance monthly on the first weekday bar of the month, the scan reached full population after the 2026-08-03 rebalance, and their next decision date is **2026-09-01** — derived here from the live union calendar via `s2_cross_sectional_momentum.rebalance_dates`, and matching #2803's independent statement of the same date.

Reproduce:

```bash
uv run python -c "
import psycopg
from app.config import settings
from app.api.strategies import _current_versions, _current_scan_versions
from app.services.strategy_monitoring import load_fire_rate
with psycopg.connect(settings.database_url) as c:
    for label, vs in (('result', _current_versions()), ('scan', _current_scan_versions())):
        print(label, len(load_fire_rate(c, versions=list(vs.values()))))
"
```

## Why the existing guard could not catch it

`tests/test_2803_scan_card_version_basis.py::test_every_scan_relation_query_is_executed_on_the_scan_basis` **already listed `strategy_signal_daily_counts`** in `_SCAN_RELATIONS`. It matches `cur.execute(NAME, params)` pairs in `inspect.getsource(get_strategy_overview)` — a shape a keyword-argument loader in another module does not have. The guard knew the relation mattered and could not see the call.

Two guards added alongside it, both **deriving** the loader set from the loaders' own SQL rather than from a literal list, so a fourth loader over a scan relation fails the guard without anyone remembering to extend anything:

- `test_every_loader_over_a_scan_relation_is_called_on_the_scan_basis` — the binding.
- `test_the_scan_basis_loaders_are_read_back_at_the_scan_key` — the lookup key, the half that fails identically.

`strategy_signals` joins `_SCAN_RELATIONS`. Substring matching is safe across the three members: `strategy_signal_daily_counts` does not contain `strategy_signals`.

Derived set, asserted by running it: `{load_attribution: strategy_signals, load_fire_rate: strategy_signal_daily_counts, load_owned_pnl: strategy_signals}` — `load_control_state` correctly excluded.

**Revert-probed, all three, each restoring exactly one half of the defect:**

| probe | guard that failed | message |
| --- | --- | --- |
| `load_fire_rate(versions=version_values)` | `..._called_on_the_scan_basis` | `load_fire_rate filters strategy_signal_daily_counts but is called with 'version_values'` |
| `fire_rate_by_strategy.get(key, …)` | `..._read_back_at_the_scan_key` | `fire_rate_by_strategy holds rows keyed by the scan basis but is read at 'key'` |
| `fire_rate_by_strategy.get(key, …)` | `test_scan_health_reads_durable_daily_counts_after_detail_retention` (DB) | `assert 0 == 1 … share_unavailable_reason='never_scanned'` |

The third is the operator-visible symptom reproduced end-to-end against a real DB. That test already seeded `strategy_signal_daily_counts` on the scan basis (#2803) but asserted only `strategy.scan.fired_entries`, so the fire-rate block reading the same table was never covered; this adds the assertions rather than a new file.

## Codex checkpoint 2 — one P2, half right, and the other half was the important half

Codex flagged that `load_owned_pnl`'s version filter is widened with every `strategy_deployments.strategy_version` while the card reads a single key.

It was **wrong** that the extra versions are therefore discardable, and acting on that would have shipped a silent wrong number. `realised_pnl_for_keys` reads the same dict at every `paper_deployment_key` for the capital base, and `load_paper_realised_pnl`'s docstring states the rule: *"Old strategy versions remain part of the shared pot after they are retired; limiting this calculation to the current manifest would make realised gains or losses disappear from the capital base."* Narrowing the filter would not raise — a missing key defaults to `StrategyPnl()`, whose `reconciled_realised_pnl` is `0` with no incomplete reason — so a deployed strategy's realised P&L would have read as a confident zero. The union stays, with that reasoning recorded at the call site.

It was **right** that the single-key card lookup drops a deployment held at a prior scan version. That is pre-existing (identical shape on `main`, `pnl_by_strategy.get(key, …)`), and fixing it means summing `StrategyPnl` across versions with a stated `complete` / `incomplete_reasons` combination rule — a definitional change, not a rider on an identity-basis fix. Filed as **#2807** and named in a comment at the lookup.

That the deployment-key lookup is expected to succeed at all also settles the basis question for this loader: rows are keyed by `strategy_signals.strategy_version`, so a deployment key can only ever match when a deployment carries the version its signals carry — the scan basis. Not testable today at 0 deployment rows.

## Scope

Read-path only. No migration, no parser, no stored value changes — this changes which rows an existing read reaches. No backfill applies (Definition-of-Done clauses 8-11 are ETL/parser/schema clauses; nothing here writes).

## Checks

```
uv run ruff check .          0
uv run ruff format --check . 0
uv run pyright               0
uv run pytest -m "not db"    0
uv run pytest tests/smoke    0
uv run pytest -m db tests/test_api_strategies.py tests/test_strategy_monitoring.py   0
```

Commit `bc8a0a60`.

## Security

No change. Read-only endpoint, same auth dependency, no new user-controlled input reaches SQL — every bound value is derived from the in-process `STRATEGY_MANIFEST`. The change is which server-side constant is bound to an existing parameterised predicate.


Luke-Bradford:
Both findings resolved at `4cb795f1`. They are two views of one weakness — the derivation matched a bare substring anywhere in the SQL text — so one narrowing answers both.

### WARNING — nothing verifies `load_control_state` cannot be swept into the guard

`FIXED 4cb795f1`. Correct concern, and it was implicit rather than asserted. Two changes:

1. `_loaders_over_a_scan_relation` now requires the relation to appear after `FROM`/`JOIN` **and** match a table name **exactly**, and requires the SQL to bind `%(versions)s` at all (a query with no version predicate has no basis to get wrong).
2. `test_a_result_basis_loader_is_not_swept_into_the_scan_guard` pins the exclusion, with a second assertion that `_CONTROL_SQL` still reads `strategy_results_store` — so if `load_control_state` ever stops being a result-basis loader, the guard says to recheck which basis it belongs on rather than silently changing meaning.

The derived set is unchanged and verified by running it: `{load_attribution: strategy_signals, load_fire_rate: strategy_signal_daily_counts, load_owned_pnl: strategy_signals}`.

Worth stating why this was worth asserting rather than reasoning about: the cost of being wrong is asymmetric. A false guard failure on a correct loader teaches the next reader that this file cries wolf — which is exactly how the #2803 guard's real coverage gap would have been dismissed as well.

### NITPICK — substring match could fire on a mere mention

`FIXED 4cb795f1`. Same change, and the `FROM`/`JOIN` context is the stricter check suggested. Revert-probed both directions:

| probe | expected | result |
| --- | --- | --- |
| `strategy_signals` added to `_CONTROL_SQL` **as a comment** | must not trip | passes |
| `FROM strategy_signal_daily_counts` renamed to `..._v2` | must not match a longer name | no match |

### The second probe found a real defect in the narrowing, now also fixed

The rename probe passed for the wrong reason: `load_fire_rate` dropped out of the derived set entirely and every remaining assertion still passed on the survivors. The narrowing had made **coverage loss silent**, which is the same failure mode as the original bug one level up.

Added a floor — `{load_attribution, load_fire_rate, load_owned_pnl} <= set(loaders)` — which keeps automatic pickup of new loaders (the point of deriving) while making a known one disappearing loud. Same "retarget this guard rather than deleting it" pattern the file already uses for its three SQL constants. Re-probed:

```
AssertionError: a known scan-basis loader dropped out of the derived set
(['load_attribution', 'load_owned_pnl']) — retarget this guard rather than
letting it pass on the remainder
```

### Also on this push

`docs/review-prevention-log.md` gains the entry for this class (Definition of Done clause 7 — the defect recurred across #2803 and #2806): *"A guard that matches a CALL SHAPE misses the same defect behind a different call shape."*

Gates re-run green at `4cb795f1`: ruff / ruff format / pyright / `pytest -m "not db"` / `tests/smoke` / `pytest -m db tests/test_api_strategies.py tests/test_strategy_monitoring.py`.

Luke-Bradford:
### NITPICK — the guards match variable NAMES, not provenance (raised twice)

`FIXED 0e7bc4e0`. The narrowing last round made the derivation stricter but left the assertion itself reading the variable's name, which is the half the nitpick was actually about. Now it reads where the value came from.

`_names_derived_from(source, origin)` runs a forward taint over the `ast.Assign` nodes in `get_strategy_overview`, seeded on the origin function and grown to a fixed point. It reaches the real chain — `scan_versions` → `scan_version_values` → `scan_params` / `scan_pnl_versions` / `scan_key` — without naming any of them, so a new intermediate is picked up automatically. Each of the three guards now asserts the bound name is in the scan-derived set **and** not in the result-derived set.

Revert-probed both directions, which is the point of the change:

| probe | old guard | new guard |
| --- | --- | --- |
| every bound name renamed to strip the substring `scan` (`scan_version_values`→`live_version_values`, `scan_params`→`live_params`, `scan_key`→`live_key`) — correct code, misleading names | fails on all three sites | passes |
| `load_fire_rate(conn, versions=version_values)` — the #2806 defect restored under a name containing no `scan` | passes | fails: `load_fire_rate filters strategy_signal_daily_counts but is called with 'version_values', which is not derived from _current_scan_versions()` |

**One thing I built and then removed, because it is the failure mode this file has already been narrowed twice to avoid.** The first cut also asserted the scan-derived and result-derived sets were globally disjoint. It caught probe 2 — but at the wrong place and under the wrong name: taint flows forward, so the broken binding surfaced as `['fire_rate'] mixes both identity bases`, pointing at a downstream card field rather than at the call. Worse, it would false-fail the moment a card field legitimately combines a result row with a scan count, which is a normal thing for this endpoint to do. The mixing check belongs per-name at the three sites that bind a basis, and that is where it now lives; the helper keeps only non-empty floors on both sets so the guard cannot silently degrade to vacuous.

Gates green at `0e7bc4e0`: ruff / ruff format / pyright / `pytest -m "not db"` / `tests/smoke`.

PR #2810, opened 2026-08-21
fix(#2809): measure scan freshness against the frontier a scan would choose

### LINKED ISSUE #2809: fix: the /strategies scan card reads 'stale' on all 10 while the scan sits exactly on the frontier — freshness is measured against a different corpus, with the statistic the scanner forbids
`/strategies/overview` reports `scan.status = "stale"` for all 10 strategies right now, while every strategy's stored `frontier_date` is **equal to the frontier a scan run at this moment would choose**. The verdict is computed against a corpus the scanner does not read, using the statistic the scanner's own module docstring forbids.

## The comparison as it stands

`app/api/strategies.py:1295`

```python
_LATEST_CORPUS_SQL = "SELECT MAX(last_bar) FROM research_price_series"
```

`app/api/strategies.py:1702-1708`

```python
scan_status = ... else "current" if latest_corpus_date is not None and frontier >= latest_corpus_date else "stale"
```

`frontier` is `strategy_scan_watermark.frontier_date`, written by `run_signal_scan`, which reads **`price_daily`** through the quarantine-masked loader (`app/services/price_masked_bars.py::_LAST_BAR_SQL`) and picks the **modal** last bar (`app/services/strategy_signal_scan.py::choose_frontier`).

So the two sides of `>=` are:

| side | table | statistic | population |
| --- | --- | --- | --- |
| `frontier` | `price_daily` | **modal** last bar | validated universe, quarantine-masked |
| `latest_corpus_date` | `research_price_series` | **max** `last_bar` | the whole research archive |

Different table, different statistic, different population.

## Measured on dev, 2026-08-21T03:3x

```
frontier a scan would choose now : 2026-08-19  (modal 5,819 of 6,584 loadable, universe 6,773)
strategy_scan_watermark          : 2026-08-19  on all 10 strategies (written 00:38Z today)
MAX(research_price_series.last_bar): 2026-08-20 — ONE row: series_id 7728, vendor 'cboe', symbol 'VIX'
```

`select last_bar, count(*) from research_price_series group by 1 order by 1 desc limit 3` → `(2026-08-20, 1), (2026-07-08, 1018), (2026-07-07, 3191)`. The single VIX row, refreshed on its own path (`updated_at` 2026-08-21T02:54Z), sets the freshness bar for every strategy card.

The `price_daily` side shows the same shape from the other direction:

```
price_date  rows
2026-08-18  11,014
2026-08-19  11,004
2026-08-20   1,563   <- daily_candle_refresh run 111693 still RUNNING (started 03:00:46Z)
```

`max(price_daily.price_date)` is also 2026-08-20 — held by a **partial, in-flight refresh**. That is precisely the case the modal rule exists to survive:

> **The frontier is the MODAL last bar, never `max(price_date)`.** On the day of measurement 7 instruments carried a bar the other 5,783 did not; keying on the maximum manufactures thousands of terminal refusals out of a refresh still in flight.
> — `app/services/strategy_signal_scan.py` module docstring, item 2

The card re-imports that forbidden statistic through the back door, from a second corpus.

## Why it matters

This is the operator's headline card. Every strategy reads `stale` while the scan is doing exactly what it is specified to do, so the field carries no information: it cannot distinguish "the scan is behind" from "some other series refreshed more recently than the equity corpus". It is the same family as #2803/#2806 — a verdict keyed on a basis the producer never writes — one field over.

## Fix shape

Compare like with like: the freshness bar should be the modal last bar over the **same** population the scan reads, under the **same** tie-break rule, so that `current` means "the scan sits on the frontier a scan run now would choose".

Measured cost of the honest statistic on this box:

| variant | latency | answer |
| --- | --- | --- |
| `load_bar_spans` + `choose_frontier` (full reuse) | 0.60s | 2026-08-19 |
| same as one SQL, unbounded | 0.53s | 2026-08-19 |
| same as one SQL, bounded to `price_date >= today - 10d` | **0.036s** | 2026-08-19 |

The endpoint currently answers in 30-66ms, so the unbounded form is a 10-20x regression and the bounded one is not. The bound needs stating and guarding, not hiding: instruments with no bar in the window cannot set the mode, which is correct for a freshness statistic and wrong for the scan's own eligibility — the two must not share a query, only the tie-break rule.

Refs #2437.

Luke-Bradford:
`/strategies/overview` reported `scan.status = "stale"` for all 10 strategies while every one of them sat **exactly** on the frontier a scan run at that moment would choose. The freshness bar was measured against a different corpus with a statistic the scanner explicitly forbids.

Closes #2809. Follows #2803 and #2806, which were the same family — a verdict keyed on a basis the producer never writes — on the version identity rather than the corpus date.

## The defect

`app/api/strategies.py` compared the stored scan frontier against:

```python
_LATEST_CORPUS_SQL = "SELECT MAX(last_bar) FROM research_price_series"
```

Wrong three ways at once:

| | what the scan does | what the card did |
| --- | --- | --- |
| table | `price_daily`, through the quarantine-masked predicate | `research_price_series`, the backtest archive |
| statistic | **modal** last bar (`choose_frontier`) | **max** |
| population | the validated universe | the whole archive |

The statistic is not a preference. `app/services/strategy_signal_scan.py`'s module docstring, item 2:

> **The frontier is the MODAL last bar, never `max(price_date)`.** On the day of measurement 7 instruments carried a bar the other 5,783 did not; keying on the maximum manufactures thousands of terminal refusals out of a refresh still in flight.

The card re-imported that forbidden max through the back door, from a second corpus.

## Measured on dev, 2026-08-21

```
frontier a scan would choose now  : 2026-08-19  (modal 5,819 of 6,584 loadable, universe 6,773)
strategy_scan_watermark, all 10   : 2026-08-19  (written 00:38Z that day)
MAX(research_price_series.last_bar): 2026-08-20  — ONE row: series_id 7728, vendor 'cboe', symbol 'VIX'
```

`select last_bar, count(*) from research_price_series group by 1 order by 1 desc limit 3` → `(2026-08-20, 1), (2026-07-08, 1018), (2026-07-07, 3191)`. A single VIX row on its own refresh path set the freshness bar for every strategy card.

The `price_daily` side shows the same shape from the other direction:

```
price_date  rows
2026-08-18  11,014
2026-08-19  11,004
2026-08-20   1,563   <- daily_candle_refresh run 111693 still RUNNING
```

`max(price_daily.price_date)` was also 2026-08-20 — held by a **partial, in-flight refresh**, which is precisely the case the modal rule exists to survive.

## The fix

`_corpus_frontier` reads the scan's own population (`load_validated_universe`, masked predicate) and applies the scanner's own tie-break. That tie-break is now `modal_bar_date`, extracted from `choose_frontier` so there is **one rule with two callers** rather than two copies — two copies being how this happened. `current` now means "the stored scan sits on the frontier a new run would pick".

### The cost, and what it forced

The old comment justified the archive read as an optimisation — *"avoids a full scan of the multi-million-row bar corpus on every Strategies page load"* — and that cost is real. Measured here:

| variant | latency | answer |
| --- | --- | --- |
| `load_bar_spans` + `choose_frontier` (full) | 0.60s | 2026-08-19 |
| the same as one SQL, unbounded | 0.53s | 2026-08-19 |
| bounded to `price_date >= today - 10d` | **0.036s** | 2026-08-19 |

The endpoint answers in 30-66ms, so the unbounded form is a 10-20x regression. `_CORPUS_FRESHNESS_WINDOW` is 10 days, fixed **by construction** and stated as such (no published rule exists): it must outlast the longest exchange closure — four sessions (9/11) plus the weekends either side is ~9 calendar days — while staying far short of the ~30 days at which a genuinely dead refresh would still look alive.

### Codex checkpoint 2 found the optimisation could change the answer

[P1] The bound drops every instrument whose last bar predates the window, and dropped instruments do not vote. With most of the universe stale and a minority freshly refreshed, the windowed mode is the minority's date while a scan run now would choose the stale majority's — the card would then call a scan sitting exactly on the frontier `stale`, i.e. this very defect reintroduced by its own fix.

Resolved with a certificate rather than by dropping the optimisation. The bound cannot *corrupt* an instrument's last bar (a `max` over the tail of a series still equals the true max whenever that max is inside the window), so divergence needs the dropped set to outvote the surviving mode. `universe_size - seen` bounds the dropped set, so:

```python
def window_decides_the_mode(*, modal_count, seen, universe_size) -> bool:
    return universe_size - seen <= modal_count
```

When true, no date outside the window can beat the mode even if every dropped instrument shares one, and the windowed answer is *provably* the unbounded one. `<=` and not `<` because an exact tie breaks on the later date, which is the in-window one — the same rule `modal_bar_date` applies. When false, `_corpus_frontier` falls back to `load_bar_spans` + `choose_frontier` and pays the 0.60s.

Live: universe 6,773, seen 5,830, modal 5,819 → decisive, and the windowed answer equals the unbounded one, at 47-66ms.

Codex re-review after the change: *"consistently uses the scanner's validated, quarantine-masked population and modal frontier rule, with a sound fallback when the bounded query is inconclusive. No actionable correctness defects were identified."*

## Behaviour change

`scan.status` flips from `stale` to `current` for all 10 strategies on the current dev corpus. `never_run` / `rotated` are untouched — they are decided before the freshness bar is consulted. The `latest_corpus_date is None` path (nothing in the window and the fallback also finds nothing) still reads `stale`, unchanged.

## Tests

`tests/test_2809_scan_freshness_basis.py`, pure (no Postgres). Every guard revert-probed:

| probe | result |
| --- | --- |
| `_corpus_frontier` reverted to `MAX(...) FROM research_price_series` | fails: *"the freshness bar must be measured over the population the scan is scored on"* |
| `since` bound dropped from the freshness query | fails on the bound assertion |
| quarantine-coverage predicate dropped from the freshness query | fails: *"would count bars the masked loader never returns"* |
| window narrowed to 5 days | fails the closure-length assertion |

Plus the tie-break equality between `modal_bar_date` and `choose_frontier`, the certificate's boundary (tie decisive, one fewer not), and a mirror check that the new SQL carries the same coverage predicate as `_LAST_BAR_SQL` — the "wrong population" leg of the defect, which the fix for the other two legs could otherwise have reintroduced.

## Security

No security surface. Read-only endpoint change; no new input reaches SQL (`instrument_ids` and `since` are parameterised, `since` is server-derived from `datetime.now(tz=UTC)`), no authorisation boundary touched.

## Verification

| clause | evidence |
| --- | --- |
| local gates at `6379842a` | ruff / ruff format / pyright / `pytest -m "not db"` / `tests/smoke` all green |
| Codex ckpt-2 | one P1 found and fixed; re-review clean |
| dev-verify (live endpoint) | pending — recorded in a follow-up comment, since the API serves `~/Dev/eBull` and this branch must be checked out there |

### DIFF
--- a/app/api/strategies.py
+++ b/app/api/strategies.py
@@ -37,7 +37,7 @@ from app.services.cost_model import COST_MODEL_ID
 from app.services.equity_curve import BENCHMARK_RULE_ID, SIZING_RULE_ID
 from app.services.outcome_resolver import RULE_SET_VERSION as OUTCOME_RULE_SET_VERSION
 from app.services.position_builder import RULE_SET_VERSION as POSITION_RULE_SET_VERSION
-from app.services.price_masked_bars import load_recent_last_bar_counts
+from app.services.price_masked_bars import load_bar_spans, load_recent_last_bar_counts
 from app.services.research_price_structure_store import QUARANTINE_RULE_SET_VERSION
 from app.services.runtime_config import (
     RuntimeConfigCorrupt,
@@ -110,7 +110,12 @@ from app.services.strategy_result_ambiguity import (
     composed_holdout_ambiguity_refusals,
     record_sha256,
 )
-from app.services.strategy_signal_scan import SCAN_UNIVERSE, modal_bar_date
+from app.services.strategy_signal_scan import (
+    SCAN_UNIVERSE,
+    choose_frontier,
+    modal_bar_date,
+    window_decides_the_mode,
+)
 from app.services.strategy_wealth import load_strategy_wealth_history
 from app.services.sync_orchestrator.dispatcher import publish_manual_job_request_with_conn
 from app.services.trial_register import TRIAL_REGISTER, TRIAL_REGISTER_VERSION
@@ -1069,13 +1074,22 @@ def _corpus_frontier(conn: psycopg.Connection[Any], *, as_of: date) -> date | No
     The old comment justified the choice as an optimisation — *"avoids a full
     scan of the multi-million-row bar corpus on every Strategies page load"* —
     and the cost was real: the honest statistic unbounded is 0.60s on a 30-66ms
-    endpoint. ``_CORPUS_FRESHNESS_WINDOW`` is what buys it back, at 0.036s for
-    the identical answer.
+    endpoint. The windowed read buys it back at 0.036s, and
+    ``window_decides_the_mode`` is what keeps it from buying a different answer.
     """
     universe = load_validated_universe(conn)
     counts = load_recent_last_bar_counts(conn, universe, since=as_of - _CORPUS_FRESHNESS_WINDOW)
     modal = modal_bar_date(counts)
-    return None if modal is None else modal[0]
+    if modal is not None and window_decides_the_mode(
+        modal_count=modal[1], seen=sum(counts.values()), universe_size=len(universe)
+    ):
+        return modal[0]
+    # The window cannot settle it, so pay for the whole distribution rather than
+    # answer from the part of the corpus that happened to refresh recently.
+    frontier = choose_frontier(
+        {instrument_id: span.last_bar for instrument_id, span in load_bar_spans(conn, universe).items()}
+    )
+    return None if frontier is None else frontier.bar_date
 
 
 def _ambiguity_record_from_result_row(
--- a/app/api/strategies.py
+++ b/app/api/strategies.py
@@ -7,7 +7,7 @@ from collections import Counter, defaultdict
 from collections.abc import Mapping, Sequence
 from datetime import UTC, date, datetime, timedelta
 from decimal import Decimal
-from typing import Literal, cast
+from typing import Any, Literal, cast
 
 import psycopg
 import psycopg.rows
@@ -37,6 +37,7 @@ from app.services.cost_model import COST_MODEL_ID
 from app.services.equity_curve import BENCHMARK_RULE_ID, SIZING_RULE_ID
 from app.services.outcome_resolver import RULE_SET_VERSION as OUTCOME_RULE_SET_VERSION
 from app.services.position_builder import RULE_SET_VERSION as POSITION_RULE_SET_VERSION
+from app.services.price_masked_bars import load_recent_last_bar_counts
 from app.services.research_price_structure_store import QUARANTINE_RULE_SET_VERSION
 from app.services.runtime_config import (
     RuntimeConfigCorrupt,
@@ -44,6 +45,7 @@ from app.services.runtime_config import (
     get_runtime_config,
     update_runtime_config,
 )
+from app.services.strategies.validated_universe import load_validated_universe
 from app.services.strategy_ambiguity_policy import AMBIGUITY_RULE_VERSION
 from app.services.strategy_base_currency import (
     DEPLOYMENT_CURRENCY_UNSUPPORTED,
@@ -108,7 +110,7 @@ from app.services.strategy_result_ambiguity import (
     composed_holdout_ambiguity_refusals,
     record_sha256,
 )
-from app.services.strategy_signal_scan import SCAN_UNIVERSE
+from app.services.strategy_signal_scan import SCAN_UNIVERSE, modal_bar_date
 from app.services.strategy_wealth import load_strategy_wealth_history
 from app.services.sync_orchestrator.dispatcher import publish_manual_job_request_with_conn
 from app.services.trial_register import TRIAL_REGISTER, TRIAL_REGISTER_VERSION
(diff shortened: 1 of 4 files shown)


Luke-Bradford:
## Dev-verify — live endpoint, `~/Dev/eBull` detached at `6379842a`

All 10 strategy cards flip from `stale` to `current`, on the same corpus and the same watermark rows that read `stale` before the change.

```
s1-time-series-momentum              status=current  frontier=2026-08-19  entries=8821  exits=6822
s10-relative-strength-leader         status=current  frontier=2026-08-19  entries=   0  exits=   0
s2-cross-sectional-momentum          status=current  frontier=2026-08-19  entries=   0  exits=   0
s3-mean-reversion-in-trend           status=current  frontier=2026-08-19  entries=  25  exits=9971
s4-volatility-compression-breakout   status=current  frontier=2026-08-19  entries= 748  exits=   0
s5-support-bounce                    status=current  frontier=2026-08-19  entries= 244  exits=   0
s6-resistance-breakout               status=current  frontier=2026-08-19  entries=  25  exits=   0
s7-trend-pullback                    status=current  frontier=2026-08-19  entries= 167  exits=6800
s8-range-mean-reversion              status=current  frontier=2026-08-19  entries=  24  exits=   0
s9-squeeze-expansion                 status=current  frontier=2026-08-19  entries=   8  exits=   0
```

`frontier_date` and the counts are unchanged — the scan wrote those rows at 00:38Z and this PR does not touch the scan. Only the verdict about them moved, which is the whole claim.

### Latency, disclosed

`GET /strategies/overview` warm: **0.099s / 0.103s** against **0.031s / 0.034s** before. The delta matches the measured 47-66ms for `_corpus_frontier` on the decisive path. The undecisive path is not exercised on this corpus (universe 6,773, seen 5,830, modal 5,819 — the dropped set cannot outvote the mode), and would cost ~0.60s when it is.

### Not done this iteration

- **FE screenshot of `/strategies`.** Vite on `:5173` is not running (`curl` returns 000), so the front-end pass was skipped per the loop's own rule. The change is API-only and the page renders the same `scan.status` enum it always has, but the rendered card has not been eyeballed at this commit.
- **`-m db` tier.** Not run; the diff adds no DB-backed test and the fast tier plus smoke are green. The new SQL is exercised against the real dev corpus by the verification above rather than by a seeded fixture, which is the trade the test-tiering rule prefers for one new read query.

PR #2812, opened 2026-08-21
fix(#2811): read a periodic strategy's zero on its decision dates, not every bar

### LINKED ISSUE #2811: S-2 and S-10 have never fired a live signal: the scan has never covered a rebalance date, and an uncovered decision date is invisible
Two of the ten strategies on `/strategies` have fired **zero** live signals — not "0 fired out of N evaluated", but no fired row at any version. Measured on dev 2026-08-21, from `strategy_signal_daily_counts`:

| strategy | versions | scanned days | fired (all versions) |
| --- | --- | --- | --- |
| s1-time-series-momentum | 3 | 15 | 28,044 |
| s5-support-bounce | 4 | 12 | 1,832 |
| **s2-cross-sectional-momentum** | 4 | **13** | **0** |
| **s10-relative-strength-leader** | 2 | 8 | **0** |

## It is not the strategy — the scan has never reached a decision date

Both are periodic. S-2 rebalances on the first weekday bar of a month (`s2_cross_sectional_momentum.rebalance_dates`, as fixed by #2797). Against the real `price_daily` calendar for 2026-07-01..2026-08-20 there is exactly **one** rebalance date: **2026-08-03**.

The dates the live scan has actually written for S-2:

```
2026-07-28, 07-31, 08-04, 08-05, 08-06, 08-10, 08-11, 08-12, 08-13, 08-14, 08-16, 08-17, 08-18
```

`scanned ∩ rebalance = {}` — and note the shape: the scan covers **07-31 and 08-04 but not 08-03**. That is a hole between two covered days, not the edge of a run. So S-2 has never had the opportunity to fire, and its zero says nothing about the strategy.

`#2797` merged today and it is on this path (it changed `rebalance_dates` in the strategy module the scan calls), so the rule itself is now right. But nothing backfills — `write_window_indices`' cold-start rule forbids it, deliberately, to avoid reintroducing look-ahead — so **the next opportunity for either strategy is 2026-09-01**.

## The part that is a defect rather than a wait

An uncovered decision date and a quiet decision date render identically. `docs/review-prevention-log.md` already states this for a neighbouring date:

> The junk instruments were not frontier-eligible, so the scan wrote *no row of any kind* for 2026-08-01 — not a `thin_cross_section` refusal, not a `not_fired`. An absent period and a quiet period render identically, so no count, no alarm and no dashboard could distinguish them. The only detector was asking why a legitimate-looking zero was there.

That is exactly what happened here, one rebalance later, and it cost a full diagnostic pass to notice. A strategy card showing `fired_share_of_evaluable: 0.0000` alongside `share_unavailable_reason: null` reads as a confident measured zero. For a periodic strategy whose decision dates were never scanned, it is not a measurement at all.

## Proposed scope

1. **Why was 2026-08-03 skipped** while 07-31 and 08-04 were written? Frontier refusal (`refused_thin_frontier`) leaves no census row, so `job_runs` for `strategy_signal_scan` around 08-03 is the place to look, not the ledger.
2. **Make an uncovered decision date visible on the card.** For a strategy with a period rule, "this version has not seen a decision date" is a distinct state from "it saw one and did not fire", and only the second is a zero worth reporting. This is the operator-visible half and the reason this is filed as a bug rather than a note.
3. **Confirm 2026-09-01 is covered** when it arrives — the acceptance for both strategies actually firing.

⚠ Not proposing to backfill 08-03. The no-backfill rule is settled and the look-ahead reason still holds.

Refs #2437, #2797.

Luke-Bradford:
## What was wrong

`/strategies` reported `fired_share_of_evaluable: 0.0000` alongside
`share_unavailable_reason: null` for **S-2** and **S-10** — which reads as a confident
measured zero. It was not a measurement at all. Both rebalance monthly; their only
decision date inside live scan coverage was **2026-08-03**, and no `strategy_signal_scan`
run of any kind existed before **2026-08-09**.

`strategy_registry.py:643` rule 3 records a non-decision bar as `not_fired`, and that is
correct and settled (`s2_member`: *"THE PRICE FLOOR IS AN ELIGIBILITY RULE, NOT AN
EVALUABILITY ONE … calling that `not_evaluable` would inflate the refusal counts
criterion 9 reads"*). The ledger is right. The **reader** was summing those bars into the
share's denominator, so the share measured the *calendar* rather than the strategy:

| strategy | old share | denominator | new |
| --- | --- | --- | --- |
| s2-cross-sectional-momentum | `0.0000` | 3,277 non-decision bars | `no_decision_date_scanned` |
| s10-relative-strength-leader | `0.0000` | 5,689 non-decision bars | `no_decision_date_scanned` |

This mechanises #2797's prevention-log lesson — *"when a cadence is coarser than daily,
check the version's expected lifetime against it before concluding a zero is the market
talking"* — which until now was a note to a human.

## ⚠ This corrects the issue's own diagnosis

#2811 asks *"why was 2026-08-03 skipped while 07-31 and 08-04 were written"*, and its
follow-up frames it as a caught-up window omitting one interior bar. **There is no such
omission.** `signal_bar_date` is written **per instrument**, and `write_window_indices`
bounds a cold start to the last bar *strictly before the frontier* **in that instrument's
own sequence**. Every census date before 08-06 is one sparse name:

| symbol | bars ≥ 2026-07-24 | census date it manufactured |
| --- | --- | --- |
| JTKWY | 07-28, 08-07 | 07-28 |
| MARK | 07-30, 07-31, 08-07 | 07-31 |
| SICP | …07-30, **08-04**, 08-07… | 08-04 |
| TRVN | 07-29, **08-05**, 08-07 | 08-05 |

Each `row_count = 1`, all from the single cold start at 08-09 14:43 against frontier
08-07. `GROUP BY signal_bar_date` over a per-instrument relation produces a date axis, and
a date axis reads as a calendar — that is the whole trap, and `row_count` was the tell.

⚠ Deliberately **no** counter-claim of a contiguous "left edge" either: a cold start writes
each instrument's last pre-frontier bar, which can be arbitrarily old, so the written set
is not an interval. The honest statement names the scan *runs*, not a date range.

## Why the calendar is STORED, not recomputed

The obvious read-layer fix (call `load_validated_universe → load_union_calendar →
decision_calendar` per request) was specced first and rejected on two grounds:

1. **It is a reconstruction.** It resolves a historical `strategy_version` against
   *today's* universe, corpus and quarantine state — a bar backfilled earlier in a month
   silently moves which date was "first of the month", with no census row and no version
   changing. That is #2809's shape exactly.
2. **It costs more than the endpoint.** `/strategies/overview` answers in **75–146 ms**;
   the union calendar alone is **0.61 s**, and the cost is intrinsic (`SELECT DISTINCT
   price_date` with no join is 1.02 s). A one-year bound is still 0.56 s and needs a
   lookback constant nothing sources.

So the scan publishes what it used. This is retroactive in the way that matters: a
decision calendar is a function of the corpus, not of coverage, so one run publishes all
73 of S-2's historical dates and the card is immediately right about census rows written
days earlier.

⚠ Publication is **not** gated on the run having work, and that is load-bearing: on the
day this landed every version was already `up_to_date` at frontier 2026-08-19, so tying it
to plan formation would have shipped inert.

## Guard against reproducing the root cause one layer up

A single sparse instrument landing on a rebalance date would set `decision_days = 1` and
re-enable the fake share. A date counts as a decision day only with at least
`min_participants` evaluable entry rows — taken from the strategy's **own** declaration
(`MIN_CROSS_SECTION = 10` for S-2), not invented here: below it the strategy refuses with
`thin_cross_section` and provably cannot fire.

## Evidence

**Full-population A/B** (dev, all 10 current versions): **8 byte-identical** — denominators
*and* `scanned_days` — **2 moved** from `0.0000` to a refusal.

**Full-population checks:** stored calendar == in-process calendar for both cross-sectional
versions (73 == 73, set equality); off-calendar fired signals across the whole census:
**none**. Published versions match existing `strategy_scan_watermark` rows — **no strategy
identity moved**.

**Real endpoint, real dev DB:** `/strategies/overview` → 200 in **85 ms** (inside the
75–146 ms budget), S-2 and S-10 rendering `no_decision_date_scanned`, other eight unchanged.

**Producer:** real `run_signal_scan` executed against the dev corpus; published 73 dates
each for S-2 and S-10 via the `up_to_date` path.

## Codex checkpoint 2 findings, both fixed

- **P1 — calendar/census corpus skew.** A historical backfill can move a month boundary
  under a stationary frontier. Publication now stamps `frontier_date`, and `load_fire_rate`
  warns when the census reaches past it. The skew's direction is fail-safe: a disagreeing
  calendar refuses the share, never reports a wrong one. Kept publish-always because
  publish-with-census reintroduces the inert-on-deploy gap above.
- **P2 — known-empty was unrepresentable.** A rule naming no date stored zero rows, which
  is byte-identical to never having published — and the reader's whole contract is that
  those differ. **My own test asserted the opposite of the truth.** Fixed with a
  publication header table; the reader is driven from the header and left-joins the dates.

## Security

No new surface. One additive read-only table, no user input, no auth path touched. All SQL
parameterised — pyright's `LiteralString` check caught an f-string in a *test* helper,
rewritten as `executemany` with bound params rather than suppressed.

## Notes

- `tests/test_api_strategies.py` has **6 pre-existing failures** on this branch and equally
  on `origin/main` (`resolve_stocks_type_id` → *"'Stocks' resolved to 0 rows"*; the DB
  fixture lacks `etoro_instrument_types`). Verified by revert-probe. Not caused here and
  not fixed here.
- Issue item 3 (confirm 2026-09-01 is covered) is a wait, not work — and is conditional on
  the version surviving, a scan running past that frontier, and the date clearing
  `min_participants`. The card will now say which.

Closes #2811
Refs #2437, #2797, #2809


Luke-Bradford:
## Review resolution — `6f08b447`

**[WARNING] `derive_fire_rate` invariant asserts can 500 the whole endpoint — FIXED `6f08b447`**

Correct, and the repo already had the answer one module over that I did not carry across:
`strategy_signal_scan._commit_strategy` states the identical contract for the writer —
*"One strategy's failure does not stop the others. That is why this returns a result
carrying the error instead of raising."*

Caught per `(strategy_id, strategy_version)` inside `load_fire_rate`. The failing version's
card refuses; the other nine are still measured.

⚠ One correction to the framing, in the direction of less alarm: **the raise is
unreachable from the current fold**, and I have said so at the call site rather than
leaving the guard looking load-bearing. All three invariants hold by construction —
`fired_days` and `decision_days` are both counted over `counted`, `counted` is a subset of
the version's rows, and `_FIRE_RATE_SQL` computes evaluable as `fired + not_fired`. The
guard is for the fold this becomes later, which is exactly when nobody will be re-deriving
those three facts. Kept for that reason, not removed.

**[NITPICK] off-calendar fires only logged, invisible to API consumers — FIXED `6f08b447`**

Taken, and it merged with the WARNING into one fix. `invariant_violated` is now a member of
**both** `ShareUnavailableReason` and `WeeklyRateUnavailableReason`, renders as
"Evidence inconsistent — refused", and reaches `types.ts` + `SHARE_UNAVAILABLE_LABELS`
(the frontend `Record<>` keyed on the union makes the label non-optional at typecheck).

Both enums, not one: the surrounding contract is *"a value is None if and only if its
reason is not"*, so refusing the share while leaving a confident weekly rate beside it
would be the same defect one field over.

**[PREVENTION] fail-isolation — EXTRACTED `docs/review-prevention-log.md`**

Added as *"A `raise` inside a per-item derivation called from an AGGREGATE endpoint turns
one bad row into an empty page (#2811, 2026-08-21)"*. It carries your rule, plus two things
the round itself taught:

- the repo's own writer-side precedent (`_commit_strategy`) existed and did not travel to
  the reader — so the generalised prompt is *"when adding a hard failure to a per-entity
  function, look at how the WRITER of the same entity handles its own failures"*;
- **my first attempt at the test passed for the wrong reason.** It fabricated an
  off-calendar fire to trigger the raise; the partition had already excluded that row, so
  it returned `no_decision_date_scanned` and asserted nothing about isolation. The entry
  records that a test constructing an input the producer cannot emit proves the branch,
  not the isolation — the test now patches the raise in, and is revert-probed (handler
  changed to bare `raise` → fails with `ValueError: simulated producer-invariant
  violation`; restored → 7 pass).

Gates green on `6f08b447`: ruff, ruff format, pyright, 8,845 fast-tier, 155 smoke,
frontend typecheck + 1,525 unit.

PR #2818, opened 2026-08-21
feat(#2817): surface per-regime result cohorts on the strategies overview

### LINKED ISSUE #2817: Surface per-regime result cohorts on /strategies — the backtest writes them, nothing reads them
## Symptom

`docs/proposals/ta/2026-08-14-strategy-set-s5-s10.md` and the operator's own
constraint require walk-forward evidence in **per-year and per-regime blocks,
never one pooled number over the whole span**.

Per-year blocks ship: `RECENT_EVIDENCE_WINDOWS` declares `year-2022`,
`year-2023`, `year-2024` plus three rolling windows, and
`/strategies/overview` renders them as `evidence_windows`.

**Per-regime blocks do not.** `strategy_result_regime_cohorts` is written by
`app/services/backtest_run.py:3598` / `:3631` via
`store_result_regime_cohorts`, one row per `(result_id, regime)` over
`bull_quiet` / `bull_volatile` / `bear_quiet` / `bear_volatile` /
`unclassified`, with per-cohort expectancy, bootstrap CI, profit factor,
worst trade and effective sample size.

Nothing reads it:

```
$ grep -rn "strategy_result_regime_cohorts" app/ migrations/
app/services/strategy_regime_evidence.py:227:  INSERT INTO strategy_result_regime_cohorts (
app/services/strategy_regime_evidence.py:244:  SELECT count(*), ... FROM strategy_result_regime_cohorts WHERE result_id=%s
```

No API surface, no frontend field. `grep -rn regime app/api/strategies.py`
returns only prose in a strategy description. `grep -rln regime frontend/src`
does not hit the strategies page.

## Why it matters now

Backtest run 111610 (`strategy_backtest_run`, started 2026-08-21T02:07:51Z) is
the first run since #2726 that will populate these cohorts. Without a reader,
that compute lands in a table the operator can never see, and the walk-forward
stays pooled-per-window on the card — the exact shape the operator ruled out.

Measured state before the run lands:

```
strategy_result_regime_cohorts : 0 rows
strategy_results_store         : 324 rows (300 namespace='hold_out')
evidence_windows status        : 'missing' on all 6 windows for all 10 strategies
```

## Scope

Read-only surfacing. No migration, no change to the writer, no change to the
cohort definition (settled in #2437 / #2726: the regime is classified on the
entry SIGNAL date, and cohorts explain a parent result rather than being
independently promotable).

1. `app/api/strategies.py` — attach the stored cohorts to each `ResultArm`.
2. `frontend/src/api/types.ts` + `StrategiesPage.tsx` — render the per-regime
   breakdown for the representative arm.

Cohorts hang off `result_id`, and the identity-pin filtering that decides
*which* results are current already lives in `_RESULTS_SQL`. Deriving the
cohort key from the rows that query returned — rather than re-listing the pins
in a second endpoint — is the #2806 "derive, never list" rule applied here.

Part of #2437

Luke-Bradford:
## What changed

`strategy_result_regime_cohorts` had a writer and no reader. `backtest_run.py:3598`
/ `:3631` store one row per `(result_id, regime)` for every hold-out result, with
per-cohort expectancy, block-bootstrap CI, profit factor, worst trade and
effective sample size — and `grep -rn strategy_result_regime_cohorts app/` hit
only the writer module. No API field, no frontend field.

This surfaces them:

- **`app/api/strategies.py`** — `ResultRegimeCohort` model + `regime_cohorts` on
  every `ResultArm`, fetched by `_REGIME_COHORTS_SQL` and ordered bull→bear,
  quiet→volatile, `unclassified` last.
- **`frontend/`** — `StrategyRegimeCohort` type and a `RegimeBreakdown` table
  under "Primary evidence" on the strategies card.

## Why now

The build queue's item 6 requires walk-forward evidence in **per-year and
per-regime blocks, never one pooled number over the whole span**. Per-year blocks
already ship as `evidence_windows`. Per-regime blocks did not reach the operator
at all.

Backtest run 111610 (`strategy_backtest_run`, started 2026-08-21T02:07:51Z) is the
first run since #2726 that populates these cohorts. Without a reader that compute
lands somewhere the operator can never see.

## Design notes

**The cohort query is keyed on the result ids `_RESULTS_SQL` actually returned,
not on a second copy of its ten identity pins.** Restating the pins is the #2806
defect shape — the two predicates drift, the join silently returns nothing, and a
panel reading "no regime evidence" becomes indistinguishable from one measuring a
strategy that never traded. Deriving the key from the producer's output makes the
two sets equal by construction, and
`test_overview_regime_cohorts_follow_the_result_ids_not_a_second_pin_predicate`
pins it by storing cohorts against a result the pin filter EXCLUDES and asserting
they never appear.

**Three states are worded apart, because collapsing any two of them is how a card
starts lying:**

| state | render |
| --- | --- |
| cohorts present | one row per regime that saw a trade; a trailing line names the regimes that saw none (`trade_count >= 1` is a writer invariant, so a short list is a measurement) |
| empty, `trade_count > 0` | "Not measured for this result" — it predates the #2726 writer. Backfilling means re-running the backtest and charging the trial register (#2599/#2616), so these stay empty permanently |
| empty, `trade_count == 0` | "This arm realised no trades" |

**A null `profit_factor` means the cohort has no losing trade**, so the ratio has
no denominator — `RegimeCohort.__post_init__` enforces the iff. Rendered as words;
a dash would read as missing data on the single strongest cohort in the table.

**Trades are shown with names AND dates.** Four trades on one decision date is one
bet, which is why the block bootstrap clusters by date; neither count bounds the
other.

`bootstrap_model_id` is deliberately not exposed — the CI and effective sample
size are null as a group (the writer refuses a partial bootstrap), so their
absence already says "not bootstrapped" without a per-row model string.

## Payload

Measured, not estimated: one serialised cohort is **362 bytes**. Worst case
(10 strategies × 6 windows × 4 arms × 5 regimes) adds **~434 KB** to a response
that is 40,039 bytes today. Accepted rather than split into a second endpoint,
because a separate route would have to re-implement `_RESULTS_SQL`'s identity-pin
filtering to decide which results are current — which is the exact duplication the
#2806 rule forbids. Every field shipped is rendered, so none of it is dead weight.

## Two things fixed in passing

Both are `test_api_strategies.py` fallout from #2809 (merged today), which CI could
not catch: CI runs no pytest, and the pre-push gate is `-m "not db"` while this
module is db-marked. All 8 DB tests in the file failed on `origin/main` before this
branch.

1. **`_seed_universe_anchor`.** `_corpus_frontier` now reads
   `load_validated_universe`, which resolves the §4.0 `Stocks` anchor from
   `etoro_instrument_types` and REFUSES on zero rows. That table is created empty
   by `sql/070` and filled by the nightly eToro sync, so no test DB has it — every
   DB test in the file raised `RuntimeError`. The refusal is correct behaviour for
   an unbootstrapped system; the bootstrap is what was missing. Seeding the anchor
   fixed 6 of the 8.
2. **`test_scan_freshness_reads_the_ingest_census_without_scanning_daily_bars`
   deleted, not repaired.** It seeded one `research_price_series` row and asserted
   `scan.status == "current"` — i.e. it asserted `MAX(last_bar)` over the research
   archive, which is precisely the basis #2809 removed as "wrong table, wrong
   statistic, wrong population" (`_corpus_frontier`'s docstring). The behaviour it
   pinned no longer exists. The replacement contract, including a structural guard
   that the overview never reads that archive again, is
   `tests/test_2809_scan_freshness_basis.py`.

## Verification

| check | result |
| --- | --- |
| `uv run ruff check .` / `ruff format --check .` | pass (1,529 files) |
| `uv run pytest -m "not db"` | exit 0 |
| `uv run pytest tests/smoke` | exit 0 |
| `uv run pytest -m db tests/test_api_strategies.py` | exit 0, 22 collected (was 8 failing on `origin/main`) |
| `pnpm --dir frontend typecheck` | pass |
| `pnpm --dir frontend test:unit` | 1,527 passed / 136 files |
| Codex checkpoint 2 (`codex exec review --base origin/main`) | no actionable defects |

**Dev-verified against the real dev DB** at `30c9b378` by calling
`get_strategy_overview` on `settings.database_url`: 200-equivalent, 0 arms rendered
(every window is `missing` under the current versions), payload 40,039 bytes —
unchanged, as expected while no result carries a split. Measured on the full stored
population:

```
select count(*) total,
       count(*) filter (where r.trade_count > 0 and c.result_id is null) with_trades_no_cohorts
from strategy_results_store r
left join (select distinct result_id from strategy_result_regime_cohorts) c using (result_id);
--  324 | 324
```

All 324 stored results predate #2726, so every arm renders the "not measured for
this result" branch today and fills in as run 111610's results land. The cohort
join itself is exercised against real Postgres and the real writer
(`store_result_regime_cohorts`) in the two new DB tests.

## Security

Read-only. No new mutation path, no new authentication or authorisation surface,
no user-controlled input reaching SQL — the one new query binds a server-derived
list of `bigint` result ids as a parameter.

## Not in scope

No migration, no change to the cohort writer, no change to the cohort definition.
The regime is classified on the entry SIGNAL date and cohorts explain a parent
result rather than being independently promotable — both settled in #2437 / #2726
and untouched here.

Part of #2437
Closes #2817

### DIFF
--- a/app/api/strategies.py
+++ b/app/api/strategies.py
@@ -286,10 +286,10 @@ class ResultRegimeCohort(BaseModel):
     expectancy_ci_high_pct: Decimal | None
     profit_factor: Decimal | None
     worst_trade_pct: Decimal
+    # ⚠ NULL AS A GROUP, never singly: `RegimeCohort.__post_init__` refuses a
+    # partially-populated bootstrap, so a null CI and a null effective sample
+    # mean "this cohort was not bootstrapped", not "one number went missing".
     effective_sample_size: Decimal | None
-    # Ships with the interval it produced, so a null CI reads as "not
-    # bootstrapped" rather than as a missing number under an unnamed model.
-    bootstrap_model_id: str | None
 
 
 class ResultArm(BaseModel):
@@ -1360,8 +1360,7 @@ _REGIME_COHORTS_SQL = """
         expectancy_ci_high_pct,
         profit_factor,
         worst_trade_pct,
-        effective_sample_size,
-        bootstrap_model_id
+        effective_sample_size
     FROM strategy_result_regime_cohorts
     WHERE result_id = ANY(%(result_ids)s::bigint[])
 """
--- a/app/api/strategies.py
+++ b/app/api/strategies.py
@@ -7,7 +7,7 @@ from collections import Counter, defaultdict
 from collections.abc import Mapping, Sequence
 from datetime import UTC, date, datetime, timedelta
 from decimal import Decimal
-from typing import Any, Literal, cast
+from typing import Any, Final, Literal, cast, get_args
 
 import psycopg
 import psycopg.rows
@@ -103,6 +103,7 @@ from app.services.strategy_position_manager import (
     manage_owned_position,
 )
 from app.services.strategy_recent_evidence import RECENT_EVIDENCE_WINDOWS
+from app.services.strategy_regime_evidence import RegimeCohortLabel
 from app.services.strategy_result import TOTAL_RETURN_BASIS
 from app.services.strategy_result_ambiguity import (
     AmbiguityRecord,
@@ -250,6 +251,47 @@ class PriorVersionTrackRecord(BaseModel):
     incomparable_reasons: list[str]
 
 
+class ResultRegimeCohort(BaseModel):
+    """One arm's realised trades, split by the regime that held when it fired (#2437).
+
+    ⚠ Explains a parent arm; NOT independently promotable. Portfolio path
+    statistics — drawdown above all — stay on the arm, because filtering closed
+    trades cannot reconstruct overlapping marked paths
+    (``app/services/strategy_regime_evidence.py`` module docstring).
+
+    Two absences the reader must render differently, both enforced by
+    ``RegimeCohort.__post_init__`` on the writer side:
+
+    - **A regime with no row had no realised trade in this arm.** Every stored
+      cohort carries ``trade_count >= 1``, so the five labels are never all
+      present. "No trades in bear_volatile" is a measurement, not a gap.
+    - **``profit_factor`` is null exactly when ``losing_trade_count == 0``** —
+      the ratio has no denominator. That is the strongest possible cohort, and
+      rendering it as a blank cell says the opposite.
+
+    ``unclassified`` is its own label, not an error: the regime is classified on
+    the entry SIGNAL date, and a date the regime series does not cover cannot be
+    labelled without conditioning on state the strategy did not have.
+    """
+
+    # The producer's own alias, not a copy of its members: a label added there
+    # must not need a matching edit here to be renderable.
+    regime: RegimeCohortLabel
+    trade_count: int
+    instrument_count: int
+    decision_date_count: int
+    losing_trade_count: int
+    expectancy_pct: Decimal
+    expectancy_ci_low_pct: Decimal | None
+    expectancy_ci_high_pct: Decimal | None
+    profit_factor: Decimal | None
+    worst_trade_pct: Decimal
+    effective_sample_size: Decimal | None
+    # Ships with the interval it produced, so a null CI reads as "not
+    # bootstrapped" rather than as a missing number under an unnamed model.
+    bootstrap_model_id: str | None
+
+
 class ResultArm(BaseModel):
     result_version: str
     purpose: Literal["harness_validation", "capital_candidate"]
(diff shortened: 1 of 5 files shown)


Luke-Bradford:
### NITPICK — `frontend/src/pages/StrategiesPage.tsx` CI-range branch not narrowing `expectancy_ci_high_pct`

`FIXED 555e41b6`

The branch now tests both bounds:

```tsx
{cohort.expectancy_ci_low_pct === null || cohort.expectancy_ci_high_pct === null
  ? "not bootstrapped"
  : `${pctPoints(cohort.expectancy_ci_low_pct)} to ${pctPoints(cohort.expectancy_ci_high_pct)}` + …}
```

You were right that the "null as a group" invariant is a writer-side guarantee
(`RegimeCohort.__post_init__` refuses a partial bootstrap) and the reader should
not lean on it. A half-null interval now falls to "not bootstrapped" rather than
printing one bound against a dash and reading as a measured range — the failure
mode that matters is the quiet one, since `pctPoints(null)` renders an em dash
that looks like a legitimate cell.

Gates re-run at `555e41b6`: `ruff check .` / `ruff format --check .` pass,
`pnpm --dir frontend typecheck` pass, `StrategiesPage.test.tsx` 42 passed.

PR #2819, opened 2026-08-21
fix(#2778): a scale-gate refusal costs one arm's control, not the whole run

### LINKED ISSUE #2778: Scale gate's cohort time projection is noise: 3-member sample swings >2x in both directions, and a refusal aborts the whole invocation

Luke-Bradford:
## Problem

`SyntheticControlScaleBudget.reserve` raises `ScaleBudgetExceeded`, and that exception propagated out of the `ArmMeasurement` constructor in `evaluate_arm` / `evaluate_level_arms` — so it aborted the **entire backtest invocation**.

Measured on 2026-08-19 (#2778): one cohort's *estimate* came in 45 seconds over a 1,200s threshold. The refusal discarded a completed 1,000-member cohort and roughly ninety minutes of compute, persisting nothing. The same arm re-projected at 611.3s on the next invocation and actually completed in **528.8s** — the projection that destroyed the run over-stated the work by 2.35x.

The projection is a 3-member serial pilot extrapolated across 8 workers, so it is noise in both directions. But the blast radius is the larger defect and is independent of the threshold, which the ticket says explicitly: *"A refusal that skipped the arm, or that persisted completed work first, would have made the bad estimate survivable."*

Real, current, and on the critical path: backtest run `111610` is the walk-forward evidence run for the S-1..S-10 set, and runs `111447` / `108436` both died mid-flight — `108436` on exactly this gate.

## Change

`_run_cohort_for` returns `(control, refusal)` and absorbs `ScaleBudgetExceeded`. The arm's own measurement is complete before the cohort is attempted, so the arm survives, the run continues, and that arm's rows carry `synthetic_control_not_run`.

**No new vocabulary was invented.** `synthetic_control_not_run` is documented in `app/services/strategy_result.py:289` as *"no cohort exists for this result. The WRITER has not run §9's control"* — which is exactly true of a cohort refused before fan-out.

## Why catching only `ScaleBudgetExceeded` is the safety property

`ArmMeasurement.cohort`'s docstring already states that `None` *"never [means] that a computed control was lost"*. That invariant survives, and the reason is positional: `reserve()` is called at `synthetic_control_run.py:1354`, strictly **before** the `ProcessPoolExecutor` fans out at line 1376. A refused cohort has produced no member outcome, so there is nothing to lose.

Anything raised from inside the fan-out means members **were** computed; reporting that arm as a plain "control not run" would hide a partial cohort, so it stays fatal. `tests/test_2778_scale_refusal_non_fatal.py::test_a_failure_inside_the_fan_out_is_still_fatal` pins this.

## Not silent

The scheduler already refuses `refresh_recent + synthetic_control` outright (`scheduler.py:5601`) on the stated principle that an operator who asked for a control must never silently receive rows saying `synthetic_control_not_run`. A scale refusal is discovered mid-run and so cannot be refused up front, but the same principle binds it. Three surfaces:

- `job_runs.error_msg`, via `tracker.note` — the operator-visible success-path channel.
- `BacktestRunReport.control_refusals` — a derived property keyed by `_arm_label`, mirroring the existing `deflation_refusals` shape ("a refusal an operator can act on"). Derived from `arms` rather than carried as a second field, since a control refusal belongs to exactly one arm that is already in the report.
- `logger.warning` in `_run_cohort_for` and `log_report`.

`_arm_label` is new and exists so the string the gate refuses under and the key an operator looks the refusal up by are the same string by construction, not by two f-strings agreeing. It replaces two hand-written copies.

## Scope left out, and why

**The estimator itself is unchanged.** #2778's first acceptance clause — *"projection error measured across at least one full strategy set, reported as a distribution"* — requires a full-strategy-set run on an idle box. Backtest run `111610` has been running on this host since 02:07 UTC; a projection measured under that contention would measure the contention, and the ticket's own closing line warns against re-fitting constants to a single observation. **Unblock condition: re-measure once `111610` reaches a terminal status, then decide between widening the pilot, spawning the pilot pool, or dropping the per-cohort estimate in favour of the cumulative budget.** The third acceptance clause (a probe for "whatever replaces the current estimate") is unreachable until that choice is made.

Worth noting the threshold half of the ticket has already been fixed on `main`: `SYNTHETIC_CONTROL_MAX_PROJECTED_COHORT_S` is now derived as ⅓ of the 4h invocation budget (4,800s), so the specific 1,245.1s refusal that killed run `108436` can no longer fire. This PR removes the failure *mode*, not just that instance.

## Prevention

`docs/review-prevention-log.md` — extended the #2175 entry ("a monkeypatched entry point must forward the FULL signature"). That entry covers the **signature** axis and `**kwargs`; it does not cover the **return shape**, which rots identically. This change moved a return type and three shadows of the contract had to follow, none of which would fail at import:

- `scripts/verify_2697_metric_axis_ab.py` — whose own why-comment already records this exact rot happening once before, when #2772 added `scale_budget`.
- three `SimpleNamespace(rows_written=N)` stubs in `tests/test_strategy_recent_evidence.py` standing in for `BacktestRunReport`.

Generalised rule added: when a signature **or return shape** moves, grep the name across `scripts/` **and** `tests/` in the same commit.

## Security

No security surface. No auth, no user input, no SQL, no broker path. The change strictly narrows what aborts a read-mostly research job.

## Verification

| check | result |
| --- | --- |
| `uv run ruff check .` / `ruff format --check .` | pass |
| `uv run pytest -m "not db"` (fast tier) | pass — includes the 6 new pure tests, which collect under `not db` |
| `uv run pytest -m db tests/test_backtest_run.py tests/test_synthetic_control_run.py tests/test_strategy_holdout_namespace.py` | exit 0 |
| `uv run pytest tests/smoke -n 0` | exit 0 |
| `codex exec review --base origin/main` (checkpoint 2) | no actionable regression |

Not a corpus/parser/schema change — no migration, no backfill, no stored-data semantics move — so Definition-of-Done clauses 8-12 do not apply. `ScaleBudgetExceeded` was previously unreachable in any stored row; the only behavioural delta is which rows get written at all, and it is strictly more.

Closes #2778

### DIFF
--- a/app/services/backtest_run.py
+++ b/app/services/backtest_run.py
@@ -192,6 +192,7 @@ from app.services.synthetic_control_run import (
     HOLDOUT_CONTROL_REASON,
     CohortCollector,
     CohortResult,
+    ScaleBudgetExceeded,
     SyntheticControlScaleBudget,
     run_cohort,
 )
@@ -532,6 +533,18 @@ class NamespaceMeasurement:
         return self.axis_dates[-1]
 
 
+def _arm_label(strategy_id: str, ambiguity_arm: AmbiguityArm | None, quarantine_arm: QuarantineArm) -> str:
+    """The one spelling of an arm's name, shared by the gate and the report.
+
+    ⚠ ONE FUNCTION AND NOT AN f-STRING AT EACH SITE. This label is what a scale
+    refusal is keyed by in ``BacktestRunReport.control_refusals``, so the string
+    the gate refuses under and the string an operator looks the refusal up by
+    have to be the same string by construction rather than by two format
+    expressions agreeing.
+    """
+    return f"{strategy_id}/{ambiguity_arm or 'shared'}/{quarantine_arm}"
+
+
 @dataclass(frozen=True)
 class ArmMeasurement:
     """Everything one ``(strategy, quarantine arm)`` corpus pass produced."""
@@ -553,7 +566,25 @@ class ArmMeasurement:
     #: invocation asked for one. ⚠ ``None`` is the DEFAULT and means the run did
     #: not compute it — never that a computed control was lost: ``run_cohort``
     #: raises rather than returning a partial one.
+    #:
+    #: ⚠ SINCE #2778 ``None`` HAS A SECOND CAUSE, and the invariant above still
+    #: holds. The scale gate can refuse this cohort, and ``ScaleBudgetExceeded``
+    #: is raised by ``SyntheticControlScaleBudget.reserve`` *before* the member
+    #: pool fans out — so a refused cohort has produced no member outcome and
+    #: there is still nothing lost to report. ``cohort_refusal`` says which of
+    #: the two ``None`` means; anything raised from INSIDE the fan-out is still
+    #: fatal, because that one WOULD be discarding computed work.
     cohort: CohortResult | None = None
+    #: Why ``cohort`` is ``None`` when the invocation asked for one, verbatim
+    #: from the gate. ⚠ ``None`` here means the control was never requested for
+    #: this arm — an operator reading ``synthetic_control_not_run`` on the stored
+    #: row cannot otherwise tell "this run did not ask for a control" from "the
+    #: gate refused to run one", and those want different next moves.
+    cohort_refusal: str | None = None
+
+    @property
+    def label(self) -> str:
+        return _arm_label(self.strategy_id, self.ambiguity_arm, self.quarantine_arm)
 
 
 @dataclass(frozen=True)
@@ -595,6 +626,18 @@ class BacktestRunReport:
     @property
     def rows_written(self) -> int:
         return len(self.rows)
+
+    @property
+    def control_refusals(self) -> Mapping[str, str]:
+        """Why §9's control is absent, per arm the scale gate refused (#2778).
+
+        ⚠ DERIVED FROM ``arms`` RATHER THAN CARRIED AS A FIELD, unlike
+        ``deflation_refusals``. A deflation group is not an arm and has nowhere
+        else to live; a control refusal belongs to exactly one arm, which is
+        already in this report — so a second copy could only ever disagree with
+        the first.
+        """
+        return {arm.label: arm.cohort_refusal for arm in self.arms if arm.cohort_refusal is not None}
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


Luke-Bradford:
**NITPICK** (`_arm_label` `None` branch only covered indirectly) — `FIXED 2a1a5f4c`

Added `test_arm_label_reproduces_the_f_string_it_replaced`, parametrised over `worst_case` / `best_case` / `None`, asserting `_arm_label` equals the literal spelling both replaced call sites used (`evaluate_arm`'s `{ambiguity_arm or 'shared'}` and `evaluate_level_arms`' never-`None` `{ambiguity}`). The `None` case was previously reachable only through `ArmMeasurement.label` in the report test, and was never compared against the old f-string.

Issue #2820, opened 2026-08-21
Backtest run 111610 died at the write phase after 5h11m: the masked arm's metric axis is incoherent and `_expected_refusals` cannot predict `metric_axis_unproven`


Luke-Bradford:
Found while shipping #2778. Not that ticket's cause — the timestamps rule it out — but it is what is actually blocking the walk-forward evidence run (build-queue item 6).

## What happened

Run `111610` (`strategy_backtest_run`, full S-1..S-10 set) started 2026-08-21T02:07:51Z and failed at **07:18:35Z** — 5h11m — at the write phase, persisting nothing:

```
s1-time-series-momentum in_sample/masked would store with refusals
['ambiguity_arms_not_compared', 'harness_validation_only', 'holdout_never_evaluated',
 'metric_axis_unproven', 'promotion_evidence_missing', 'synthetic_control_not_run']
against the expected
['ambiguity_arms_not_compared', 'harness_validation_only', 'holdout_never_evaluated',
 'promotion_evidence_missing', 'synthetic_control_not_run']
— refusing before the first INSERT rather than leaving rows behind whose gate verdict nobody predicted
```

The delta is exactly one code: **`metric_axis_unproven`**.

### 1. The masked arm's axis is incoherent, and the admitted arm's is not

`QUARANTINE_ARM_ORDER` is `('admitted', 'masked')`, so the `admitted` arm was validated first and did **not** raise. The failure is specific to `masked`, not an artefact of which arm happens to be processed first.

`build_result` (`app/services/backtest_run.py:2847`) stamps `metric_axis_dates`, `_start`, `_end` and `_digest` all from the same `outcome.axis_dates`, so the provenance clauses of `metric_axis_is_valid` (`app/services/strategy_result.py:919`) cannot fail by construction. That leaves only the semantic clauses:

* `dates[0] < window_start or dates[-1] > window_end`
* `namespace == "in_sample" and dates[-1] >= HOLDOUT_BOUNDARY`
* the `periods_per_year` / `cagr_pct` reconciliation against `len(dates)`

[...]

PR #2821, opened 2026-08-21
fix(#2820): name the metric-axis clause that refused, instead of a bare bool

Luke-Bradford:
Part 2 of #2820. Deliberately **not** part 1 — the underlying incoherence in the masked arm is untouched here, because diagnosing it costs a five-hour run and this change is what makes that run answer the question.

## Problem

Backtest run `111610` evaluated the full S-1..S-10 set from 2026-08-21T02:07:51Z and died at the write phase at 07:18:35Z — 5h11m — persisting nothing:

```
s1-time-series-momentum in_sample/masked would store with refusals [... 'metric_axis_unproven' ...]
against the expected [...] — refusing before the first INSERT
```

`metric_axis_is_valid` returns a single `bool` over **eight independent clauses**. The refusal-set diff names the code; nothing names the clause. The only way to learn which one closed was to re-run.

## Change

`metric_axis_invalid_reason(identity, metrics) -> str | None` returns the failing clause. `metric_axis_is_valid` becomes a thin `is None` wrapper, so the verdict keeps exactly one implementation and no caller's semantics moved. The pair writer appends the reason to its message, and only when `metric_axis_unproven` is actually the disputed code — every other refusal-set delta reads byte-identically to before.

The three reconciliation clauses carry both numbers (`stored X, axis of N bars gives Y`), because a bare label still costs a re-run to act on.

## The reason is diagnostic, never a verdict

`metric_axis_unproven` stays **one closed refusal code**. Splitting it into eight would change what a frozen `STRUCTURAL_REFUSAL_POLICY_VERSION` declaration means, and those declarations are immutable — the remedy for a reinterpretation is a new `strategy_version`, not a re-freeze. Nothing branches on the string; it is read and printed.

## One deliberate behavioural note

Absence is now tested **before** the rule version, inverting the original `or` chain's short-circuit order.

The verdict is identical either way — both return non-`None` — and only the label moves. It matters because a legacy pre-#2697 row carries `metric_axis_rule_version = NULL`, so the old order would have reported `rule_version_moved` for a row that has no axis at all. Those rows are the bulk of what `strategy_results` holds today (all 24 stored rows, written 2026-08-08), so the misleading label would have been the common one.

## What this already narrows on #2820, before any re-run

`ResultIdentity.__post_init__` (`app/services/strategy_result.py:551-577`) **raises** on every provenance clause: rule version, endpoint agreement, digest, window containment, in-sample hold-out boundary, and the opportunity-set digest.

`build_result` constructs a `ResultIdentity`. So a freshly built result **cannot** be refused for provenance — it would have raised at construction instead. The `s1-time-series-momentum in_sample/masked` refusal is therefore one of the three metrics-reconciliation clauses:

* `periods_per_year_mismatch`
* `total_return_below_negative_100pct`
* `cagr_does_not_reconcile`

That is derived from the constructor's own checks, not measured on the failing arm, and it is why the tests exercise those three and say plainly which returns are unreachable from a constructed identity.

## Security

None. Pure function over an in-memory dataclass; no auth, input, SQL or broker path. It strictly adds detail to an existing refusal and changes no verdict.

## Verification

| check | result |
| --- | --- |
| `uv run ruff check .` / `format --check` | pass |
| `uv run pytest -m "not db"` | pass — 8 new pure tests, collected under `not db` |
| `uv run pytest -m db tests/test_strategy_result.py tests/test_strategy_metric_axis_db.py tests/test_backtest_run.py` | exit 0 |
| `codex exec review --base origin/main` | "preserves the existing boolean verdict… only computes the reason when the metric-axis refusal differs"; no findings |

No stored-data semantics move, no migration, no backfill — Definition-of-Done clauses 8-12 do not apply. No row's verdict changes; the only delta is the text of an exception and of a diagnostic no caller reads.

Refs #2820


Luke-Bradford:
**NITPICK** (`metric_axis_invalid_reason` out of alphabetical order in `__all__`) — `FIXED c6c29073`

Sorted the four `metric_axis_*` entries among themselves. Two clarifications on the premise: the list is **not** globally sorted (`namespace_for_bar` trails the promotion-refusal group), and `metric_axis_promotion_refusals` already preceded `metric_axis_is_valid` on `main` — so the out-of-order pair predates this branch. My inserted line was correctly placed relative to `metric_axis_is_valid`; moving `promotion_refusals` after it is what actually sorts the block, so that is what the commit does rather than re-ordering the whole list.

PR #2822, opened 2026-08-21
fix(#2820): a saturated total return cannot pin a CAGR, so stop demanding one

### LINKED ISSUE #2820: Backtest run 111610 died at the write phase after 5h11m: the masked arm's metric axis is incoherent and `_expected_refusals` cannot predict `metric_axis_unproven`

Luke-Bradford:
Part 1 of #2820, and the actual blocker on build-queue item 6 (walk-forward evidence). Part 2 (PR #2821, merged) is what made this diagnosable in 25 minutes instead of 5 hours.

## The finding

`compute_metrics` derives **both** figures from the same ratio (`app/services/strategy_statistics.py:421,440`):

```python
total_return = (final_equity / starting_equity - 1.0) * 100.0
cagr         = ((final_equity / starting_equity) ** (1.0 / years) - 1.0) * 100.0
```

so they cannot genuinely disagree. The disagreement was manufactured by the **validator**, which reconstructs the ratio from the stored percentage:

```python
final_multiple = 1.0 + metrics.total_return_pct / 100.0
expected_cagr  = -100.0 if final_multiple == 0.0 else ...
```

`(ratio - 1.0) * 100.0` rounds to exactly `-100.0` for **any** ratio below `2**-54`, because doubles in `[0.5, 1.0)` have an ulp of `2**-53`. The percentage physically cannot carry a multiple that small, so the reconstruction yields `0.0` and the rule demanded a `-100%` CAGR from a row whose CAGR was right.

## Measured, backtest run 112188

`s1-time-series-momentum in_sample/masked`, with PR #2821's diagnostic:

```
metric axis refused on: cagr_does_not_reconcile
(stored -63.08857514295101, -100.0% over 59.48528405201917y gives -100.0)
```

Reconstructing forward from the stored CAGR:

| quantity | value |
| --- | --- |
| implied `final_equity / starting_equity` | `1.7882895749018668e-26` |
| `(ratio - 1.0) * 100.0` | `-100.0` — exactly, `== -100.0` is `True` |
| validator's reconstructed multiple | `0.0` |
| validator's expected CAGR | `-100.0` |
| stored CAGR | `-63.08857514295101` |

A sleeve that all but wiped out over six decades. That is a real outcome, not a defect — and it is why the `admitted` arm passed while `masked` failed: masking removes bars and produced the near-zero terminal equity. `QUARANTINE_ARM_ORDER` is `('admitted', 'masked')`, so admitted was validated first and did not raise.

## The fix narrows the gate; it does not skip it

When the stored total return is saturated, the true ratio is known only to lie in `[0, 2**-54]`. So the CAGR is known only to lie in the closed interval those endpoints imply, and the rule now asserts exactly that and no more:

```
-100.0 <= cagr_pct <= (2**-54) ** (1/years) - 1) * 100
```

For run 112188's 59.49 years that is `[-100.0%, -46.6998%]`, and `-63.089%` sits inside it. A positive CAGR, a `-10%` CAGR, or anything below `-100%` is still refused — `test_a_cagr_the_saturated_bound_excludes_is_still_refused` pins all four. The **unsaturated** path is untouched and keeps its `1e-10` equality (`test_an_unsaturated_total_return_still_reconciles_exactly` asserts a `1e-6` drift is still caught).

This is the narrowest change that stops refusing correct rows: it withdraws exactly the one assertion the stored number provably cannot support.

## The constant is derived, not chosen

`SATURATED_TOTAL_RETURN_MAX_MULTIPLE = math.ulp(1.0) / 4` — i.e. `2**-54`, spelled so the platform states it rather than a literal asserting it. `test_the_constant_is_the_actual_saturation_point_on_this_platform` re-derives it by binary search **and** checks `math.nextafter(r, inf)` does not saturate, so it cannot drift into the invented-constant class the instruction set forbids.

## What this does NOT change

No stored row's verdict for any non-wiped-out sleeve; no migration; no backfill; no promotion threshold. `metric_axis_unproven` remains one closed refusal code and `STRUCTURAL_REFUSAL_POLICY_VERSION` does not move — the set of reachable refusal *outcomes* is unchanged, only which inputs land in them. Rows previously refused on this clause were never written (the assertion fires before the first INSERT), so there is nothing stored to re-judge.

## Security

None. Pure arithmetic on an in-memory dataclass; no auth, input, SQL or broker path.

## Verification

| check | result |
| --- | --- |
| `uv run ruff check .` / `format --check` | pass |
| `uv run pytest -m "not db"` | pass — 9 new pure tests |
| `uv run pytest -m db tests/test_strategy_result.py tests/test_strategy_metric_axis_db.py tests/test_backtest_run.py tests/test_strategy_statistics.py` | exit 0 |
| `codex exec review --base origin/main` | "correctly handles IEEE-754 saturation… while retaining mathematically derived CAGR bounds"; no findings |
| dev reproduction | run 112188, s1-scoped, 25 min, failure message quoted above |

⚠ **Not yet verified end-to-end:** that a full run now survives its write phase. That needs a fresh backtest, which is the next step once this lands, and it is the thing that finally populates `evidence_windows` on `/strategies` (all 10 strategies currently read `0/6 missing`). Flagging it rather than implying it is done.

Closes #2820
Refs #2437

### DIFF
--- a/app/services/strategy_result.py
+++ b/app/services/strategy_result.py
@@ -913,21 +913,135 @@ class PromotionCandidate:
 #: no longer carries it. The policy's reachable outcomes changed, so the
 #: version moves: frozen declarations pin this string precisely so a change
 #: like this cannot reinterpret them silently.
-STRUCTURAL_REFUSAL_POLICY_VERSION: Final = "structural-refusal-policy-2026-08-15-v4-metric-axis"
+#: v5 (#2820): the REACHABLE OUTCOMES DID NOT CHANGE — ``metric_axis_unproven``
+#: was already satisfiable — but which rows reach it did: a near-wiped-out
+#: sleeve whose ``total_return_pct`` saturates at -100.0 used to be refused and
+#: now reconciles. That is a false positive corrected rather than a gate
+#: loosened, and the version still moves, because "a row that was refused is
+#: now promotable" is exactly what a frozen declaration pins this string to
+#: notice. Bumping on the narrower reading would have made the change silent.
+#: ⚠ Measured before the bump, 2026-08-21 —
+#: ``select structural_refusal_policy_version, count(*) from
+#: strategy_preregistration_declarations group by 1`` returns
+#: ``v2 -> 3, v3 -> 2``. FIVE declarations exist and NONE is at v4, so every one
+#: of them is already superseded and no trial pays a new cost here. ⚠⚠ Note the
+#: ``-v2`` note above records "returned 0" as at 2026-08-12; that figure is
+#: stale rather than wrong, which is why this one states its query and its date.
+STRUCTURAL_REFUSAL_POLICY_VERSION: Final = "structural-refusal-policy-2026-08-21-v5-cagr-saturation"
 
 
-#: The largest ``final_equity / starting_equity`` whose ``total_return_pct``
-#: still rounds to exactly ``-100.0``.
+#: Relative slack on ``base ** years``. ⚠ DERIVED, THEN MEASURED. The base
+#: carries at most half an ulp of relative error from its own division and
+#: addition, and raising to ``years`` multiplies that by ``years``; ``pow``
+#: itself adds about one ulp. Eight times the product bounds both with margin.
+#: Measured to separate: with it, no honest row is rejected across 11 spans from
+#: 0.25 to 100 years x terminal multiples from 1e-30 to 1e3, while a CAGR drift
+#: of 1e-12 on a well-conditioned row is still refused.
+_POW_RELATIVE_SLACK_ULPS: Final = 8.0
+
+
+#: Absolute slack when inverting ``total_return_pct``/``cagr_pct`` back to a
+#: terminal multiple, in units of ``2**-54``.
 #:
-#: ⚠ DERIVED FROM IEEE-754, NOT CHOSEN. ``total_return_pct`` is
-#: ``(ratio - 1.0) * 100.0``. Doubles in ``[0.5, 1.0)`` have an ulp of
-#: ``2**-53``, so ``ratio - 1.0`` rounds to exactly ``-1.0`` for every ratio up
-#: to half of that — ``2**-54``, spelled here as ``ulp(1.0) / 4`` so the
-#: platform states it rather than a literal asserting it.
-#: ``tests/test_2820_cagr_saturation.py`` re-derives the same value by binary
-#: search and checks the very next double up fails, so this cannot drift
-#: silently.
-SATURATED_TOTAL_RETURN_MAX_MULTIPLE: Final = math.ulp(1.0) / 4
+#: ⚠ DERIVED, THEN MEASURED, AND THE TWO AGREE EXACTLY. The stored value is
+#: ``fl(fl(ratio - 1.0) * 100.0)``. Inverting it crosses two roundings the
+#: neighbour bracket does NOT already cover:
+#:   1. ``ratio - 1.0`` — the result sits in ``[0.5, 1.0)`` in magnitude for any
+#:      near-wipeout, where half an ulp is ``2**-54``; and
+#:   2. dividing the bracket by 100 — a quotient of magnitude ~1.0, so half an
+#:      ulp is ``2**-53``, i.e. two more of the same unit.
+#: Three units total. Measured over 13 spans from 0.25 to 100 years x terminal
+#: multiples from 1e-30 to 1e3, THREE is the smallest multiple that rejects no
+#: honest row — at two, nine are refused — and it is still tight enough to
+#: refuse a multiple of 2e-16 paired with a stored return of -100.0, which
+#: really stores -99.99999999999997. Four admits that pair.
+_MULTIPLE_RECONSTRUCTION_SLACK_UNITS: Final = 3.0
+
+
+def _multiple_reconstruction_slack() -> float:
+    return _MULTIPLE_RECONSTRUCTION_SLACK_UNITS * math.ulp(1.0) / 4.0
+
+
+def _rounding_bracket(value: float) -> tuple[float, float]:
+    """The interval of real numbers that round to this stored double.
+
+    ⚠ ``value + (neighbour - value) / 2`` AND NOT ``(value + neighbour) / 2``.
+    Summing two adjacent doubles near the top of the range overflows to
+    infinity, which would silently turn a bracket into an unbounded one — the
+    direction that admits everything.
+    """
+    return (
+        value + (math.nextafter(value, -math.inf) - value) / 2.0,
+        value + (math.nextafter(value, math.inf) - value) / 2.0,
+    )
+
+
+def _ratio_from_total_return_pct(total_return_pct: float) -> tuple[float, float]:
+    """Terminal multiples consistent with a stored ``total_return_pct``.
+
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
**WARNING** (near-saturated rows bypass the new bound and fall to the old equality check) — `FIXED 1e37a7fe`

Confirmed, then generalised well past the original fix. The review was right that the `final_multiple == 0.0` special case covered only ratios below `2**-54`; measured, true ratios from ~1e-16 to ~1e-8 were still being refused.

Three further holes surfaced across five `codex exec review` rounds, all now closed and all kept as named regression tests:

1. **Comparing in percentage space launders an inconsistent CAGR.** `isclose(rel_tol=1e-10)` near -100 leaves ~1e-8 of absolute slack on the percentage, which spans a decade of multiples: a `-34%` CAGR passed against a `-99.9999999999999%` total return whose true CAGR is `-44.05%`. → `test_the_near_saturated_band_cannot_launder_an_inconsistent_cagr`
2. **A blanket `ulp(1.0)` widening was 4x too wide** at the saturated end, admitting a `2e-16` multiple against a stored `-100.0` (which really stores `-99.99999999999997`). The slack is now three units of `2**-54`, derived term-by-term from the two roundings and measured as the smallest that rejects no honest row. → `test_the_saturated_bracket_is_not_wider_than_saturation_actually_reaches`
3. **NaN collapsed the bracket.** Every NaN comparison is `False`, so `max(0.0, nan)` is `0.0` — a NaN CAGR produced the bracket `(0.0, 0.0)`, which overlaps a saturated total return's and would have promoted a malformed row. `isclose` refused NaN for free; interval arithmetic does not. → `test_a_non_finite_metric_is_refused_before_any_interval_arithmetic`

**The final shape is not the one this PR opened with.** Instead of a special case, the rule brackets the terminal multiples each stored number is consistent with and asks whether the two brackets overlap — symmetric, because either number can saturate depending on the span (the total return over 59 years, the CAGR over a sub-year window like `year-2024`).

⚠ **It is stricter than what it replaces, not looser.** The bracket collapses to near-nothing where the floats are well conditioned, so a well-conditioned row now refuses a 1e-12 CAGR drift that `isclose(rel_tol=1e-10)` admitted. It widens only where a stored number provably cannot distinguish its neighbours.

Measured across 14 spans (0.25 to 100 years) x terminal multiples from 1e-30 to 1e30: **no honest row rejected**, every counterexample above refused, and shallow CAGRs (0%, +5%, -10%, -30%) against a wipeout still refused.

Also in this push, from the same review rounds: `_rounding_bracket` now computes `value + (neighbour - value) / 2` rather than `(value + neighbour) / 2`, which overflowed to infinity near the top of the double range — the direction that silently turns a bracket unbounded.

`STRUCTURAL_REFUSAL_POLICY_VERSION` moves to **v5**. The policy's reachable outcomes did not change, but which rows reach them did, and a frozen declaration pins that string precisely to notice that. Measured before the bump — `select structural_refusal_policy_version, count(*) from strategy_preregistration_declarations group by 1` returns `v2 -> 3, v3 -> 2`: five declarations exist, none at v4, so all are already superseded and none pays a new cost. (The `-v2` comment's "returned 0" note was measured 2026-08-12 and is stale; the new note states its query and its date.)

PR #2824, opened 2026-08-21
feat(#2823): surface criterion 5's purged K-fold split on /strategies

### LINKED ISSUE #2823: Surface the purged walk-forward split on /strategies — the backtest writes folds, nothing reads them
## Symptom

`docs/proposals/ta/2026-08-14-strategy-set-s5-s10.md` and the operator's own
constraint require **walk-forward** evidence. `frontend/src/pages/StrategiesPage.tsx:154`
even carries the constraint as a comment:

>  *  walk-forward evidence is never one pooled number over the whole span.

Two of the three halves ship. Per-year blocks render as `evidence_windows`
(`RECENT_EVIDENCE_WINDOWS`). Per-regime cohorts landed in #2817.

**The split itself — the proof the walk-forward was actually purged and
embargoed — has no reader.** `strategy_result_folds` (`sql/269`) is written by
`app/services/backtest_run.py:3702` via `store_walk_forward_folds`, one row per
`(result_id, fold_index)` carrying the fold's inclusive date block, its MEASURED
panel-axis `embargo_bars`, `bar_count`, and the four-way census
`test_count` / `train_count` / `purged_count` / `embargoed_count`.

A data-access reader exists — `app/services/result_ledger.py:2417
read_walk_forward_folds` — and every one of its callers is a test or a one-off
script:

```
$ grep -rn "read_walk_forward_folds" --exclude-dir=.venv .
app/services/result_ledger.py:2417   (definition)
app/services/result_ledger.py:2495   (__all__)
tests/test_strategy_result_folds.py  (x5)
scripts/audit_2745_in_sample_run.py:340
scripts/verify_2240_result_writers.py:277
scripts/verify_2240_fold_wiring.py:251
```

No `app/api/` caller, no frontend field:

```
$ grep -rn "walk_forward\|walkForward" app/api/          -> (nothing)
$ grep -rniE "walk.?forward|purged_count|embargo" frontend/src
frontend/src/pages/StrategiesPage.tsx:154:  (the comment quoted above, only)
```

`sql/269`'s own header calls the missing consumer *"the phase-6 read"*. It was
anticipated and never built.

## Why it matters now

Backtest run 112352 (`strategy_backtest_run`, started 2026-08-21T09:44:22Z,
params `holdout_purpose: "pinned recent-evidence windows for the S-1..S-10
walk-forward (build-queue item 6)"`) is in flight and writes folds for every
in-sample arm it stores. This is #2817's argument verbatim: without a reader
that compute lands in a table the operator can never see.

Measured on dev, 2026-08-21T10:0xZ, before the run lands:

```
strategy_result_folds        :  64 rows / 16 in_sample results (s1, s2, s3)
                                4 folds each, walk_forward_model_id
                                'c5-purged-walk-forward-v1'
strategy_results_store       : 300 hold_out + 28 in_sample
/strategies/overview         : 10 strategies, 60/60 evidence_windows 'missing',
                               0 result_arms on every card
```

## The join is by identity pins, not by a FK

⚠ Folds hang off **`namespace = 'in_sample'`** results; `_RESULTS_SQL` in
`app/api/strategies.py:1324` selects **`namespace = 'hold_out'`**. So keying a
fold read on the result ids `_RESULTS_SQL` returned — #2817's pattern — returns
nothing, always.

`strategy_result_control_support` does not bridge them either; it holds **0 rows**
on dev, and its `control_result_id` is the control arm, not the in-sample split.

The honest link is the identity pins the overview already enforces. To avoid the
#2806 shape (two copies of the pin list drift, the join silently returns nothing,
and "no walk-forward evidence" becomes indistinguishable from a split that was
never cut), the in-sample select must **bind the same parameter dict**
`_RESULTS_SQL` already binds rather than re-listing the pins by hand.

## Scope

Read-only surfacing. No migration, no writer change, no change to the split
construction, no strategy-version movement (`_module_hash` covers
`strategy_registry.py` only; nothing here touches it).

1. `app/api/strategies.py` — select the current-identity in-sample split and
   attach it to each strategy in the overview.
2. `frontend/src/api/types.ts` + `StrategiesPage.tsx` — render the per-fold
   block: date range, bars, embargo, and the purged/embargoed census.

## Acceptance

- `/strategies/overview` returns a non-null walk-forward split for a strategy
  whose current version has a stored in-sample result with folds.
- A strategy with no such result reports a NAMED reason, never a silent absence
  or a fake zero (#2811's rule, applied to this field).
- The rendered `embargo_bars` is legible: S-1 measures 4,825 panel bars on the
  2024 window, which is correct (`docs/review-prevention-log.md` — the embargo is
  measured on the PANEL axis, and S-1 declares no `max_hold_bars`), and must not
  read as a defect.

Refs #2437

Luke-Bradford:
## What changed

`/strategies` now renders criterion 5's stored split — the per-fold evidence that
the in-sample panel was actually purged and embargoed.

`strategy_result_folds` (`sql/269`) has been written by `store_walk_forward_folds`
since #2240. A data-access reader exists (`result_ledger.py:2417
read_walk_forward_folds`) and every one of its callers is a test or a one-off
script — no `app/api/` route, no frontend field. `sql/269`'s own header calls the
missing consumer *"the phase-6 read"*. #2817 closed the same gap one table over,
on the per-regime cohorts; this is the walk-forward-integrity half of build-queue
item 6.

## Why the join is by identity pins, not by a FK

Folds hang off `namespace = 'in_sample'` results. `_RESULTS_SQL` selects
`namespace = 'hold_out'`. So keying a fold read on the result ids `_RESULTS_SQL`
returned — #2817's pattern, and the right one there — would match nothing, always,
and the panel would read "no strategy has a split" forever.
`strategy_result_control_support` does not bridge them either: 0 rows on dev, and
its `control_result_id` is the control arm rather than the in-sample companion.

Restating the ten pins would be the #2806 shape — two copies drift, the join
silently empties, and "no walk-forward evidence" becomes indistinguishable from a
split that was never cut. They are not restated: `_WALK_FORWARD_SPLIT_SQL` binds
**the same `params` dict** `_RESULTS_SQL` binds, with the same placeholder names,
so a pin added to one is a `ProgrammingError` on the other rather than silent drift.

`LEFT JOIN`, not `INNER` — an in-sample result carrying no folds must still arrive,
or `no_split_stored` collapses into `no_in_sample_result`.

## The grain is one split per strategy — measured, not assumed

A strategy version stores **four** in-sample results, the 2x2 of `ambiguity_arm`
x `quarantine_arm`, each with its own folds. Measured on dev for
`s1-time-series-momentum` at `strategy-registry-v1+2307ee566d7b` (result ids
117-120), fold 0:

| arm | first_date | last_date | bar_count | embargo_bars | train |
| --- | --- | --- | --- | --- | --- |
| best_case / masked | 1962-01-02 | 1999-09-02 | 4,375,006 | 615 | 1,694,103 |
| best_case / admitted | 1962-01-02 | 1999-09-02 | 4,375,006 | 615 | 1,694,643 |
| worst_case / masked | 1962-01-02 | 1999-09-02 | 4,375,006 | 615 | 1,694,103 |
| worst_case / admitted | 1962-01-02 | 1999-09-02 | 4,375,006 | 615 | 1,694,643 |

The **geometry** is identical across all four — `bar_weighted_folds` reads only
the panel axis. The **census** moves with `quarantine_arm` alone; `ambiguity_arm`
is a fill-price assumption and cannot move a population (117 == 119, 118 == 120,
exactly). So one split is rendered, and the arm whose census it is, is NAMED in
the payload rather than left implicit.

## Copy corrections the source required

Every one of these reads backwards by default, and all four are load-bearing for
what the panel says:

- The construction is **purged K-fold over contiguous blocks**, NOT an anchored or
  rolling walk-forward. `app/services/walk_forward.py`: *"ch. 7's purged K-fold
  over contiguous time blocks, not a strictly anchored walk-forward"* — both sides
  of a test block carry training data, which is what leaves room for the embargo.
  The heading says "Purged K-fold".
- `first_date`/`last_date` bound the **test** block, not a training interval. The
  column header is "Test block".
- `purged` = training observations whose label window overlaps the test block.
  `embargoed` = training observations STARTING immediately after it. Separate
  verdicts, never merged.
- **The four counts are one population re-classified per fold**, so a column total
  is always wrong. Verified on dev: every S-1 fold sums to exactly **4,228,628**
  (1,026,750+2,398,068+0+803,810 = 1,148,643+1,024,208+2,542+2,053,235 = …). The
  panel states this and renders no totals.

There is no per-fold performance field and none may be added — §5.3: the split is
a validity gate, not a training loop, and a per-fold Sharpe *"would invite exactly
the 'which fold did best' search criterion 6 exists to bound"*.

## Named absence

`derive_fire_rate`'s contract, one panel over: a silent blank and a measured zero
are indistinguishable to a reader.

- `no_in_sample_result` — no in-sample result under the current pins. The ordinary
  state for a version whose backtest has not been re-run; it does NOT mean the
  split was skipped.
- `no_split_stored` — the run happened, the split never reached storage.
- `invariant_violated` — stored rows contradict the writer's contract. Contained
  to its own card and named there rather than raised: this runs per strategy inside
  an aggregate endpoint, so propagating would blank all ten cards over one bad row.

`folds` is non-empty exactly when `unavailable_reason` is null.

The panel renders on the **blank-card branch too** (`!window || !arm`), because the
split belongs to the in-sample panel and is independent of whether any hold-out arm
survived the pins. Measured 2026-08-21: all ten cards were in exactly that state
(60/60 evidence windows `missing`, 0 result arms), so on those cards this is the
only stored evidence there is.

## Verification

Dev-verified by calling the real endpoint function against the dev DB:

```
s1-time-series-momentum       4 folds  arm=admitted  model=c5-purged-walk-forward-v1  1962-01-02..2024-09-27
    fold 0: 1962-01-02..2003-09-30 bars=8,258,154 emb=1639 test=1,026,750 train=2,398,068 purged=0     embargoed=803,810
    fold 1: 2003-10-01..2011-08-02 bars=8,261,576 emb=4825 test=1,148,643 train=1,024,208 purged=2,542 embargoed=2,053,235
    fold 2: 2011-08-03..2016-09-30 bars=8,258,798 emb=4825 test=1,019,774 train=2,173,419 purged=1,974 embargoed=1,033,461
    fold 3: 2016-10-03..2021-06-28 bars=8,266,544 emb=4825 test=1,033,461 train=3,190,818 purged=4,349 embargoed=0
s10 / s2 / s3 / s4 / s5 / s6 / s7 / s8 / s9   REASON=no_in_sample_result
```

Fold 3's `embargoed = 0` is correct — nothing follows the last block. The large
embargo counts are the measured PANEL-axis embargo (`docs/review-prevention-log.md`:
measure a bound on the axis the window is cut on), and S-1 declares no
`max_hold_bars`, so a wide span is expected rather than a defect.

Backtest run 112352 is in flight and writes folds for every in-sample arm it
stores; the nine `no_in_sample_result` cards fill in as it lands.

Gates: `ruff check` / `ruff format --check` / `pyright` (0 errors) /
`pytest -m "not db"` / `pytest tests/smoke` / `pnpm typecheck` /
`pnpm test:unit` (1529 passed) — all green. Codex checkpoint 1 on the spec
(caught the anchored-vs-K-fold mislabel, the test-block reading and the
cross-fold-sum trap, all of which are fixed here) and checkpoint 2 on the diff
(no findings).

## Scope

Read-only surfacing. No migration, no writer change, no change to the split
construction. No strategy version moves — `_module_hash()` covers
`strategy_registry.py` only, and nothing here touches it.

## Security

No new authz surface. `/strategies/overview` keeps its existing session
requirement; the new statement is parameterised throughout and reads two tables
the endpoint already reads under the same pins. No user-controlled input reaches
the SQL.

Refs #2437
Closes #2823

### DIFF
--- a/app/api/strategies.py
+++ b/app/api/strategies.py
@@ -118,6 +118,10 @@ from app.services.strategy_signal_scan import (
     modal_bar_date,
     window_decides_the_mode,
 )
+from app.services.strategy_walk_forward_evidence import (
+    SplitUnavailableReason,
+    derive_walk_forward_split,
+)
 from app.services.strategy_wealth import load_strategy_wealth_history
 from app.services.sync_orchestrator.dispatcher import publish_manual_job_request_with_conn
 from app.services.trial_register import TRIAL_REGISTER, TRIAL_REGISTER_VERSION
@@ -379,6 +383,77 @@ class EvidenceWindow(BaseModel):
     arms: list[ResultArm]
 
 
+class WalkForwardFold(BaseModel):
+    """One block of criterion 5's split (#2823).
+
+    ⚠ ``first_date`` / ``last_date`` bound the **test** block, not a training
+    interval — ``app/services/walk_forward.py``: *"One contiguous block of the
+    panel axis, held out for testing."*
+
+    ⚠ The four counts are ONE population re-classified by this fold, and
+    ``FoldCensus.total`` asserts exactly that. Summing any of them across folds
+    counts the same observations ``fold_count`` times, so no total is exposed
+    here and none may be rendered.
+
+    ⚠ NO PER-FOLD PERFORMANCE FIELD EXISTS AND NONE MAY BE ADDED. §5.3: the
+    split is a validity GATE, not a training loop, and a per-fold number *"would
+    invite exactly the 'which fold did best' search criterion 6 exists to
+    bound"*.
+
+    Two legitimate zeroes (``sql/269``): ``test_count`` may be 0 on a fold that
+    spans real dates when no observation STARTS inside a thin era, and
+    ``embargo_bars`` 0 means *"nothing to measure on this fold's training
+    side"* — never that the embargo was skipped.
+    """
+
+    fold_index: int
+    first_date: date
+    last_date: date
+    bar_count: int
+    #: MEASURED, on the PANEL-date axis — not instrument bars, not calendar days
+    #: and not a declared holding period. It is the maximum panel-axis label span
+    #: over this fold's own post-purge training set, so it varies per fold and
+    #: needs no constant from any strategy.
+    embargo_bars: int
+    test_count: int
+    train_count: int
+    #: Training observations whose LABEL WINDOW overlaps the test block.
+    purged_count: int
+    #: Training observations that START in the window immediately FOLLOWING the
+    #: test block. A separate verdict from ``purged_count`` and not a second
+    #: symmetric purge — collapsing the two would make §5.3's finding about
+    #: their relative size unreportable.
+    embargoed_count: int
+
+
+class WalkForwardSplit(BaseModel):
+    """The strategy's stored split, or the named reason it has none (#2823).
+
+    Grain is the STRATEGY CARD, not the arm and not the evidence window: the
+    geometry is identical across all four stored arms (measured — see
+    ``strategy_walk_forward_evidence``), and the folds sit inside the in-sample
+    prefix, so they belong to no hold-out window.
+
+    ⚠ The construction is **purged K-fold over contiguous blocks**, NOT an
+    anchored or rolling walk-forward, whatever the table's name suggests. Both
+    sides of a test block carry training data; that is what leaves room for the
+    embargo at all.
+    """
+
+    folds: list[WalkForwardFold]
+    #: The STORED id, never today's constant — a split cut under a superseded
+    #: construction stays readable as that construction.
+    walk_forward_model_id: str | None
+    fold_count: int | None
+    #: Which arm's census the counts are. The geometry is arm-invariant; the
+    #: census is not, because masking changes the universe.
+    quarantine_arm: str | None
+    window_start: date | None
+    window_end: date | None
+    #: Non-null exactly when ``folds`` is empty.
+    unavailable_reason: SplitUnavailableReason | None
+
+
 class StrategyOverview(BaseModel):
     strategy_id: str
     strategy_version: str
@@ -391,6 +466,10 @@ class StrategyOverview(BaseModel):
     exclusion_reason: str | None
     scan: ScanHealth
     evidence_windows: list[EvidenceWindow]
+    #: Criterion 5's split for this version, per strategy rather than per window
+    #: — the folds partition the in-sample prefix and belong to no hold-out
+    #: window (#2823).
+    walk_forward_split: WalkForwardSplit
     prior_versions: list[PriorVersionTrackRecord]
     # ⚠ NOT a prior-version summary, despite the name: `_RESULT_COUNTS_SQL`
     # filters on the CURRENT versions, so this counts rows under the current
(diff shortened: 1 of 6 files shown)


Luke-Bradford:
**[NITPICK] `WalkForwardSplit(**{**split.__dict__, ...})` relies on field names lining up** — `FIXED 2fad0219`.

Agreed, and the splat was weaker here than at the two neighbouring
`StrategyAttributionView(**attribution.__dict__)` call sites: this one also rebuilt
`folds` into a different type on the way through, so a nested conversion was hidden
inside a dict literal.

`_walk_forward_split_view` now copies field by field.

One addition beyond what was asked, because the explicit constructor only closes
half the hole: it catches a RENAME, but a field **added** to the derivation and
never surfaced would still return a valid, quietly incomplete payload with no
request failing. `tests/test_strategy_walk_forward_evidence.py::test_the_response_model_carries_exactly_the_derivation_s_fields`
pins the two field SETS equal so that drift fails at the fast tier instead.

Revert-probed rather than assumed — injecting a field into
`StrategyWalkForwardSplit` fails the guard, and removing it passes.

PR #2826, opened 2026-08-21
feat(#2770): an evidence-bound operator path from candidate to paper

### LINKED ISSUE #2770: Add evidence-bound operator promotion path from research to paper
## Problem

The strategy control plane has strict primitives, but there is no production/API caller that advances a current manifest capital candidate through:

None -> research_candidate -> historical_validated -> forward_observation -> paper_enabled.

The existing lifecycle endpoint only pauses or retires. The Strategies page can allocate an already paper-enabled strategy but cannot get an evidence-qualified candidate there.

Exposing promote_strategy directly would be unsafe: caller-supplied result_ids can name a favourable subset even though each named result individually passes. The operator path must bind the complete declared evidence denominator and must not trust browser-provided evidence IDs or verdicts.

## Scope

- Add explicit, ordered operator actions; no arbitrary to_stage input.
- Recompute current strategy/version and authoritative evidence inside the locked transaction.
- Historical validation must select the complete current pinned hold-out matrix: all six declared recent windows x four ambiguity/quarantine arms, with no missing, duplicate, or extra identity combinations.
- Reuse the control-plane replay for every pinned result and persist a deterministic evidence reference derived from the authoritative bundle.
- Starting forward observation must not weaken or replace historical evidence.
- Paper approval must require a current passing prospective assessment under the effective policy and pin that durable assessment identity.
- Require authenticated operator reason and preserve immutable audit events.
- Harness-validation and unregistered strategies remain unable to advance.
- No action may reach live_enabled; live remains dedicated-gate and fail-closed.
- Surface allowed next action and refusals on the Strategies page and add targeted backend/frontend tests.
- Add concurrency and idempotency tests so duplicate operator submissions cannot fork the chain.

## Acceptance

An authenticated operator can advance an actually qualified current capital candidate in order and then configure bounded demo/paper allocation from the Strategies page. Any incomplete, cherry-picked, stale, mismatched, concurrent, unregistered, harness, or live-reaching request refuses without a partial promotion.

Luke-Bradford:
## What was missing

`promote_strategy` (`app/services/strategy_control_plane.py:422`) is a strict, heavily
guarded primitive — and it had exactly one API caller,
`POST /strategies/{id}/lifecycle` (`app/api/strategies.py`), which only ever targets
`paused` or `retired`. Nothing in `app/` or `scripts/` walked the forward chain
`None → research_candidate → historical_validated → forward_observation →
paper_enabled`. So `/strategies` could allocate an already-`paper_enabled` strategy but
had no way to get one there, and the chain was unreachable outside the test suite.

## Why the primitive could not simply be exposed

It takes caller-supplied `result_ids` and validates each one **individually**. A caller
could pin the three windows where a strategy happened to look good; every pinned row
would pass its own per-row checks, and the stored promotion would be indistinguishable
from an honest one. The denominator, not the row, is what a cherry-pick attacks.

## The denominator is now assembled server-side

`app/services/strategy_operator_promotion.py`. The operator names an **action**; the
evidence is assembled inside the transaction, under the same per-version advisory lock
the primitive takes, from three declarations that already existed:

| component | source | count |
| --- | --- | ---: |
| windows | `strategy_recent_evidence.RECENT_EVIDENCE_WINDOWS` | 6 |
| arms | ambiguity × quarantine, the same cross product `app/api/strategies.py` already uses for `EvidenceWindow.status == "complete"` | 4 |
| comparability | `current_identity_pins()` | 10 predicates |

The third is the one worth naming. **Twenty-four rows with no gaps can still mix results
measured against different corpora, cost models or rule sets** — label-complete and
meaningless. Binding the whole pin dict *is* the cross-row coherence check; there is
deliberately no second one. Its own docstring already said the pins "ARE the result
identity — differing on any of them means the numbers are not comparable".

### `_current_identity_pins` / `_current_versions` move to a service module

To `app/services/strategy_result_identity.py`, aliased at their old names in
`app/api/strategies.py`. A service may not import an API module. This is not churn: the
pin dict exists *because* readers must not drift, so a third reader copying it would
defeat it exactly as a second one would have. The alias keeps the existing call sites and
the tests that monkeypatch `app.api.strategies._current_versions` working unchanged.

## Duplicates are RESOLVED, not refused — and that is measured, not assumed

`result_version` hashes `metric_axis_dates` (`ResultIdentity.version`,
`app/services/strategy_result.py`), and `strategy_results_unique` is on
`(strategy_id, strategy_version, result_version)` — **not** on the window/arm identity.
So a re-run whose corpus lost or gained a single bar writes a second, legitimate row for
the same window and arm. Pinned results are `ON DELETE RESTRICT`, so the older one cannot
be tidied away. "Refuse on duplicate" would make any re-run strategy permanently
unpromotable, and `refresh_recent` re-runs are routine.

`select_latest_rows` takes the highest `result_id`. That is the safe direction — a re-run
that measured **worse** supersedes a better old row, never the reverse — and no caller can
influence it. It is one implementation, shared by the read surface and the transaction,
rather than a SQL `DISTINCT ON` plus a Python copy that could disagree about "latest".

`test_a_re_run_over_a_moved_axis_is_a_second_row_on_one_identity` proves the store really
does accept two rows on one identity, rather than asserting it in a comment.

## The three evidence rules

| action | to_stage | evidence bound | `pinned_result_count` |
| --- | --- | --- | ---: |
| `register_research_candidate` | `research_candidate` | none — not an evidence stage | 0 |
| `validate_historical` | `historical_validated` | complete 24-combination bundle | 24 |
| `start_forward_observation` | `forward_observation` | complete bundle covering everything `historical_validated` covered | 24 |
| `enable_paper` | `paper_enabled` | fresh passing prospective assessment, dated after forward observation began | 0 |

**Weakening is compared on IDENTITIES, not result ids.** Codex checkpoint 2 caught the
first implementation comparing ids, as a P1: a re-run between the two promotions replaces
every id, so every previously pinned id reads as "dropped", and because the store is
append-only that verdict is permanent — one `refresh_recent` would block
`start_forward_observation` forever. It was the same failure the duplicate rule above
exists to avoid, reintroduced two functions later. The rule is: *superseding a row is not
weakening the evidence; ceasing to cover a window or an arm is.* The identity check cannot
fire while both steps demand a complete matrix over the same declared windows; what it
catches is the declared set **shrinking** between the steps, which is a real weakening and
is invisible to every completeness check.

**Paper approval** reuses the overview's own freshness predicate verbatim, including its
five-second future-clock tolerance — one predicate, one behaviour, or the card reads
"fresh" where the transaction reads "stale". It additionally requires
`checked_at >= the forward_observation promotion`: an assessment computed before forward
observation began is not evidence *from* it, and without this all four advances can run
back-to-back on backtest evidence alone, making the stage decorative. Refusals reuse the
existing `prospective_assessment_*` vocabulary rather than minting parallel codes.

**Not built, deliberately:** a minimum forward-observation duration. Duration floors
belong to the live gate (#2599's contract-frozen forward-shadow floor reads `forward_days`
off these windows); a second floor here would put one policy in two places.

## API

`POST /strategies/{strategy_id}/advance`, `require_session`. Body is `{action, reason}` —
**no `strategy_version`, no `to_stage`, no `result_ids`**, the three inputs that would let
a browser choose its own denominator. A test asserts the request model's field set is
exactly `{action, reason}`, so adding one back fails rather than passing review.

`UniqueViolation` is **not** caught. The advisory lock serialises concurrent submissions,
so the second reads the advanced stage and refuses with an invalid-transition 409 before
any INSERT. Catching it would need constraint-name discrimination
(`idx_strategy_promotions_one_initial` and `..._one_successor` mean different things) plus
a savepoint to avoid leaving the transaction aborted — all to handle a race the lock
already excludes. The indexes stay the backstop #2612 designed them to be. Two identical
submissions therefore give one 200 and one 409: duplicate *prevention*, which is what the
ticket asks for, not idempotent replay, which it does not.

`live_enabled` is unreachable — not an `OperatorAction`, and the primitive refuses it
independently.

## Read surface

`StrategyOverview` gains `next_operator_action` and `next_operator_action_refusals`, and
`/strategies` renders the step in the research section. The action is **named alongside
its refusals** rather than nulled by them — an operator who cannot act needs to know which
step is blocked. Codex checkpoint 2 also flagged (P2) that the paper step reported no
refusals at all, so reaching `forward_observation` rendered an enabled button for a
guaranteed 409; it now surfaces the assessment refusals the overview has already computed.
Those are necessarily a subset of what the transaction checks, so an enabled button claims
nothing *known* refuses, not that the request will succeed. The page is advisory by
construction; a stale page gets a 409 and renders it verbatim.

## ⚠ This ships the gate; it cannot yet be walked end to end

Measured on dev, 2026-08-21:

```sql
select strategy_id, evidence_window_id, ambiguity_arm, quarantine_arm, purpose, count(*)
from strategy_results_store group by 1,2,3,4,5;
```

16 groups across s1-s4, **every one with `evidence_window_id IS NULL`** and
`purpose = 'harness_validation'`. No strategy has a matrix to pin, and all ten manifest
entries are `harness_validation`, which `promote_strategy` refuses for evidence stages.
Backtest run **112352** (`refresh_recent: true`) is writing the first matrix now.

That is the sequencing, not a gap: the gate has to exist before evidence can pass through
it. Deciding a strategy is a `capital_candidate` is an evidence decision on its own
ticket, and is explicitly out of scope here. **No end-to-end operator run happened and
this PR does not imply one did.**

## Security model

The endpoint is an operator authorisation: `require_session`, not the router's
`require_session_or_service_token`, because a promotion is recorded with a named
`promoted_by` and a service token has no operator identity to attribute one to — the same
reasoning `update_core_mandate` gives. The request body carries no evidence selector, so
there is no client-trusted input to the promotion decision; version, stage and evidence
are all resolved server-side inside the locked transaction. Nothing here touches the
order path, the kill switch, `enable_auto_trading` or `enable_live_trading`, and
`live_enabled` is unreachable by two independent refusals.

## Verification

| check | result |
| --- | --- |
| `uv run ruff check .` | pass |
| `uv run ruff format --check .` | pass (1538 files) |
| `uv run pytest -m "not db"` | pass |
| `uv run pytest tests/smoke` | pass |
| `uv run pytest tests/test_2770_operator_promotion_path.py` | 57 pass, and they survive `-m "not db"` (verified: the module is not auto-marked, so it is on the push gate) |
| `uv run pytest tests/test_2770_authoritative_evidence_loader.py` | pass |
| `uv run pytest -m db tests/test_strategy_control_plane.py tests/test_api_strategies.py tests/test_result_ledger.py tests/test_strategy_result.py` | 2 failures, **pre-existing on `origin/main`** — see below |
| `pnpm --dir frontend typecheck` | pass |
| `pnpm --dir frontend test:unit` | 136 files, 1536 tests, pass |
| `codex exec review --base origin/main` | round 1: P1 + P2 (both fixed, both with a test); round 2: no findings |

**Pre-existing failures, not from this branch.** `test_the_deployment_currency_refusal_and_its_constraint_agree`
(both parametrisations) fails with `etoro_instrument_types.description = 'Stocks'
resolved to 0 rows`. Reproduced identically in a clean worktree at `origin/main`, so it is
a test-DB seeding gap, not a regression here. Noted rather than ticketed, per the standing
loop instruction not to open audit tickets while building.

**Incidental, noted not ticketed:** the overview's own completeness check
(`app/api/strategies.py`, `arm_keys` built as a `set`) tolerates duplicate identities
silently — eight rows on four identities read as `complete`. This PR's read surface runs
`select_latest_rows` first and so is unaffected; the existing `EvidenceWindow.status`
computation still has the old behaviour.

No schema migration, no parser change, no ETL change, so Definition-of-Done clauses 8-12
do not apply.

Closes #2770

### DIFF
--- a/app/api/strategies.py
+++ b/app/api/strategies.py
@@ -98,6 +98,15 @@ from app.services.strategy_monitoring import (
     pool_owned_pnl_by_strategy,
     realised_pnl_for_keys,
 )
+from app.services.strategy_operator_promotion import (
+    EvidenceRow,
+    OperatorAction,
+    advance_strategy,
+    evidence_refusal_summary,
+    next_operator_action_view,
+    recent_evidence_refusals,
+    select_latest_rows,
+)
 from app.services.strategy_position_manager import (
     StrategyPositionManagerError,
     manage_owned_position,
@@ -112,6 +121,7 @@ from app.services.strategy_result_ambiguity import (
     composed_holdout_ambiguity_refusals,
     record_sha256,
 )
+from app.services.strategy_result_identity import current_identity_pins, current_result_versions
 from app.services.strategy_signal_scan import (
     SCAN_UNIVERSE,
     choose_frontier,
@@ -484,6 +494,14 @@ class StrategyOverview(BaseModel):
     allocation: StrategyAllocationView
     allocation_ready: bool
     allocation_refusals: list[str]
+    #: The one ordered step `POST /{id}/advance` would take from this stage (#2770),
+    #: NAMED ALONGSIDE its refusals rather than nulled by them — an operator who
+    #: cannot act needs to know which step is blocked, not that there is no step.
+    #: `null` means a terminal stage with no forward edge.
+    #: ⚠ ADVISORY. A page cannot be transactionally coupled to a later request, so
+    #: the transaction stays authoritative and a stale page gets a 409 to render.
+    next_operator_action: OperatorAction | None
+    next_operator_action_refusals: list[str]
 
 
 class StrategyAttributionView(BaseModel):
@@ -1139,6 +1157,32 @@ class StrategyLifecycleResponse(BaseModel):
     promotion_id: int
 
 
+class StrategyAdvanceRequest(BaseModel):
+    """⚠ NO `strategy_version`, NO `to_stage`, NO `result_ids` — deliberately (#2770).
+
+    Those are the three inputs that would let a browser choose its own promotion
+    denominator. `promote_strategy` validates caller-supplied `result_ids`
+    INDIVIDUALLY, so a favourable subset passes every per-row check while the
+    promotion rests on a cherry-picked set. The version and evidence are resolved
+    inside the locked transaction instead.
+    """
+
+    action: OperatorAction
+    reason: str = Field(min_length=1, max_length=500)
+
+
+class StrategyAdvanceResponse(BaseModel):
+    strategy_id: str
+    strategy_version: str
+    from_stage: str | None
+    stage: str
+    promotion_id: int
+    evidence_ref: str | None
+    #: 24 for the two result-evidence stages; 0 for `research_candidate` (no evidence)
+    #: and for `paper_enabled`, which pins an assessment rather than results.
+    pinned_result_count: int
+
+
 def _live_gate_view(report: LiveGateReport) -> LiveGateResponse:
     return LiveGateResponse(
         strategy_id=report.strategy_id,
@@ -1162,15 +1206,12 @@ def _live_gate_view(report: LiveGateReport) -> LiveGateResponse:
     )
 
 
-def _current_versions() -> dict[str, str]:
-    """The identities STORED RESULTS carry — the backtest measurement basis.
-
-    ⚠ NOT the identities the live scan writes. See ``_current_scan_versions``.
-    """
-    return {
-        strategy_id: entry.identity(universe=BACKTEST_UNIVERSE, cost_model_id=COST_MODEL_ID).version
-        for strategy_id, entry in STRATEGY_MANIFEST.items()
-    }
+# ⚠ MOVED TO `app.services.strategy_result_identity` (#2770) and aliased here, not
+# reimplemented. `strategy_operator_promotion` binds the same basis to assemble the
+# promotion denominator, and a service may not import an API module. The alias keeps
+# this module's own call sites and the tests that monkeypatch
+# `app.api.strategies._current_versions` working unchanged.
+_current_versions = current_result_versions
 
 
 def _current_scan_versions() -> dict[str, str]:
@@ -1710,27 +1751,11 @@ def _walk_forward_split_view(split: StrategyWalkForwardSplit) -> WalkForwardSpli
     )
 
 
-def _current_identity_pins() -> dict[str, str]:
-    """The pins a stored row must match to sit on today's measurement basis.
-
-    Exactly the equality predicates ``_RESULTS_SQL`` applies, named once so the
-    prior-version reader cannot drift from the current-version one. These ARE the
-    result identity — differing on any of them means the numbers are not
-    comparable, which is why the reader reports the difference instead of either
-    hiding the version or splicing its figures in.
-    """
-    return {
-        "namespace": "hold_out",
-        "corpus_version": corpus_version_for(BACKTEST_UNIVERSE),
-        "cost_model_id": COST_MODEL_ID,
-        "sizing_rule": SIZING_RULE_ID,
-        "benchmark_rule": BENCHMARK_RULE_ID,
-        "return_basis": TOTAL_RETURN_BASIS,
-        "ambiguity_rule_version": AMBIGUITY_RULE_VERSION,
-        "position_rule_set_version": POSITION_RULE_SET_VERSION,
-        "outcome_rule_set_version": OUTCOME_RULE_SET_VERSION,
-        "input_rule_set_version": QUARANTINE_RULE_SET_VERSION,
-    }
+# ⚠ MOVED TO `app.services.strategy_result_identity` (#2770), same reason as
+# `_current_versions` above. The docstring there carries the additional load this
+# dict now bears: it is also the cross-row coherence rule for a promotion
+# denominator, so a second copy would be a correctness defect and not just churn.
+_current_identity_pins = current_identity_pins
 
 
 def build_prior_versions(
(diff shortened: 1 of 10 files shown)


Luke-Bradford:
Both round-1 findings fixed in `69f53579`.

**WARNING** — `PromotionStep` sent a generated `reason`. `FIXED 69f53579`.
The step now takes the operator's own words in a required field, and the button stays
disabled until there is a non-blank rationale. Your framing was the useful part: a
template does not merely fail to record a reason, it fills the audit column with
something that reads as a real justification forever after, which is worse than empty.
Covered by `advances a strategy whose evidence no longer refuses, sending no denominator`
(asserts the button is disabled before a rationale is typed, and that the trimmed operator
text is what reaches the API).

**NITPICK** — `Field(min_length=1, max_length=500)` accepted `" "`. `FIXED 69f53579`.
Now `Annotated[str, StringConstraints(strip_whitespace=True, min_length=1, max_length=500)]`.
Pydantic strips before it measures, so a blank rationale is a 422 on the input that is
actually wrong rather than a 409 out of `promote_strategy._require_text` on a request that
looked well formed. Two new tests: `test_a_whitespace_only_rationale_is_refused_at_the_boundary`
and `test_a_real_rationale_is_stored_stripped`, plus a frontend case
`will not advance on a whitespace-only rationale`.

The frontend trims before sending, so the string the button accepts is exactly the string
the endpoint accepts — the two validations cannot disagree.

Gates re-run on `69f53579`: ruff, ruff format, pyright (0 errors), `pytest -m "not db"`,
`pytest tests/smoke`, `pnpm --dir frontend typecheck`, `pnpm --dir frontend test:unit`
(136 files, 1537 tests) — all green.

Issue #2833, opened 2026-08-22
R5 S-A: cap-weighted beta sleeve (SPY/IVV) — adopt the null model as default sleeve + net benchmark


Luke-Bradford:
Part of #2832. Refs #2437.

## Claim under test

Not an edge. The null model: cap-weighted US beta (SPY or IVV) as the engine's default
sleeve and the **net benchmark line every strategy must beat** on /strategies. SPIVA:
~90% of professional funds fail this over 15 years — which makes it simultaneously the
cheapest thing we can hold and the hardest bar we can set.

## Spike (days, no deflation exposure — no alpha claim is being tested)

1. Instrument-eligibility endpoint for SPY and IVV: confirm real-asset unleveraged
   classification (settlementType real, no overnight fee) + fractional eligibility.
   ⚠ USD listings only — `orderCurrency` is USD-only (VERIFIED-PORTAL); CSPX.L/IUSA.L
   are GBP-quoted and presumed non-orderable.
2. Measure effective half-spread from ~5 trading days of stored quote snapshots on both.
3. All-in one-off entry cost = measured round-trip + documented GBP→USD conversion fee
   for our tier.

## Pre-declared pass bar

All-in one-off cost ≤ 0.60% AND zero overnight financing on the unleveraged long.
Fail: classified CFD-only, or conversion+spread > 0.60%.

## On pass

Adopt as default sleeve + the net benchmark line on /strategies. Carry both caveats
verbatim: (a) for a pure beta sleeve a UK ISA at another broker strictly tax-dominates —
this is the correct ENGINE default, not the correct household allocation; (b) unhedged
USD exposure on a GBP account.

Luke-Bradford:
The attended core/cash product path is now merged in PR #2926 (`608dc879`). The live Portfolio endpoint reads this ticket directly and currently reports SPY.RTH, CSPX.L and IUSA.L at 1/5 observed trading days; cash remains the fallback and execution is blocked. A real attended POST smoke returned 409 before mutation, with orders 1→1 and core trades 0→0. Once this ticket reaches a reviewed selection verdict, the same page can save the mandate and run the guarded demo rebalance; no code bypass or second arm is needed. Refs #2603.

Luke-Bradford:
## Clock/continuity check after the Portfolio reachability merge — 2026-08-24

The first trading day is real and the collector is live. Read-only dev measurements:

[...]

PR #2927, opened 2026-08-24
fix(#2603): expose draft core mandate setup

Luke-Bradford:
## Issue reference

Refs #2603.

## Summary

Makes the Portfolio page actionable during #2833 evidence collection by exposing the backend-supported disabled core-mandate draft. Enablement and demo execution remain gated on a reviewed instrument selection.

## Changes

- Always render core target, reserve, band, minimum, and audit-reason controls.
- Permit only a disabled, instrument-free draft while `can_configure=false`; keep Enable and Rebalance visibly disabled with the gate explained.
- Add a regression test for the collecting-state draft payload.

## Security and audit model

No execution path touched. The existing backend rejects enabled mandates before #2833 selection, and the rebalance control remains disabled. Draft writes still require an authenticated session and an audit reason.

## Testing

- [x] `uv run ruff check .`
- [x] `uv run pyright`
- [x] `uv run pytest -m "not db"` — 9,800 passed, 14 skipped
- [x] `uv run pytest tests/smoke` — 161 passed, 3 skipped
- [x] `pnpm --dir frontend test:unit` — 1,618 passed
- [x] pre-push dark/pill/chart and repository chokepoint gates

### DIFF
--- a/frontend/src/pages/StrategyPortfolioLens.test.tsx
+++ b/frontend/src/pages/StrategyPortfolioLens.test.tsx
@@ -172,6 +172,34 @@ describe("StrategyPortfolioLens", () => {
     expect(screen.getByText(/UK ISA at another broker tax-dominates/i)).toBeInTheDocument();
   });
 
+  it("lets the operator save a disabled draft while evidence is still collecting", async () => {
+    const save = vi.spyOn(strategiesApi, "updateCoreMandate").mockResolvedValue({
+      configured: true,
+      enabled: false,
+      core_instrument_id: null,
+      core_target_pct: "80",
+      liquidity_reserve_pct: "10",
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
WARNING — FIXED 38bd9314 — the collecting-state test now refetches a configured disabled draft after the first save, proves `mandateDirty=false` disables Save, and proves a second click cannot issue another write.

NITPICK — REBUTTED — `sleeve.selected_instrument_id ?? mandate.core_instrument_id ?? null` has two optional inputs. The final `?? null` normalises the case where both are `undefined` to the explicit `number | null` wire shape required by `CoreMandateUpdate`; removing it produces `undefined`, which JSON omits rather than transmitting the backend field explicitly.

PR #2928, opened 2026-08-24
fix(#2525): enforce shared core capital boundary

Luke-Bradford:
## Issue reference

Refs #2525
Refs #2603

## Summary

Binds the core/cash sleeve and alpha strategies to the operator-assigned paper pot instead of whole-account cash or same-instrument holdings. The Portfolio page now fails closed and reports assigned-capital availability honestly when exact live core value is not observable.

## Changes

- add one shared capital-authority read that counts post-epoch alpha reservations, pending core orders, and exact broker position IDs
- reject pre-pot non-terminal lifecycles, malformed ownership, non-USD account snapshots, manual same-instrument holdings, and capital-limit breaches
- serialize alpha allocation, core submission, mandate changes, pool revisions, and position-management reads on the shared allocator lock
- bind core preflight/execution to assigned headroom and exact owned positions while retaining broker cash as an independent ability-to-pay ceiling
- protect pool withdrawals against alpha, pending core, and active recorded core commitments
- expose distinct unavailable/blocker states in the strategy API and Portfolio page instead of advertising unproved rebalance capacity
- document the settled boundary and cover the database, broker parser, API, scheduler, executor, and frontend paths

## Security and audit model

This touches the demo execution path. Core orders remain demo-only and operator-attended; the existing kill switch and execution guard are still required, and durable intent/admission/audit rows are written before any broker submission. The new gate also proves the shared capital lock is held before admission. No order was placed during verification.

## Testing

- [x] Tested locally against Docker PostgreSQL stack
- [x] New service logic exercised with realistic inputs and edge cases (empty results, API errors)
- [x] No raw user input reaches the DB without parameterisation
- [x] `uv run ruff check .`
- [x] `uv run pyright`
- [x] `uv run pytest -m "not db"`
- [x] `uv run pytest tests/smoke`
- [x] focused DB suites for shared capital, withdrawal protection, submission gating, mandate API, paper executor, broker parsing, core executor, and observation job
- [x] `pnpm --dir frontend test:unit` (140 files, 1,618 tests)

Verified at `89e3b895f49433abbdbad370b1cf683df21d7d18`; all CI checks and the latest Claude review are green.

## Checklist

- [x] Branch follows the required naming convention
- [x] Raw API payloads remain available in the broker snapshot used for validation
- [x] No new score/thesis output rows
- [x] No order placement path bypasses execution guard
- [x] CI checks pass

### DIFF
--- a/app/api/strategies.py
+++ b/app/api/strategies.py
@@ -997,6 +997,7 @@ class CoreSleeveBlockerResponse(BaseModel):
         "core_demo_required",
         "core_order_unresolved",
         "core_capital_authority_incomplete",
+        "core_live_snapshot_required",
         "core_paper_pool_unconfigured",
         "core_paper_pool_disabled",
         "core_sandbox_exceeded",
@@ -3526,7 +3527,7 @@ def read_core_sleeve(
     elif capital_authority.core_active_position_ids:
         blockers.append(
             CoreSleeveBlockerResponse(
-                code="core_capital_authority_incomplete",
+                code="core_live_snapshot_required",
                 detail=(
                     "Active core commitment needs an exact broker snapshot; "
                     "the read-only page cannot advertise rebalance headroom."
--- a/app/api/strategies.py
+++ b/app/api/strategies.py
@@ -52,7 +52,7 @@ from app.services.strategy_base_currency import (
     DEPLOYMENT_CURRENCY_UNSUPPORTED,
     SUPPORTED_DEPLOYMENT_CURRENCIES,
 )
-from app.services.strategy_capital_sandbox import sandbox_bound
+from app.services.strategy_capital_sandbox import sandbox_bound, sandbox_headroom
 from app.services.strategy_control_plane import (
     PAPER_ALLOCATOR_ADVISORY_LOCK,
     StrategyControlError,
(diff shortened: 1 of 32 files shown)


Luke-Bradford:
Review resolution at `397f31e718`:

- WARNING FIXED — active exact-owned positions now emit `core_live_snapshot_required`; `core_capital_authority_incomplete` is reserved for a failed/incomplete authority observation. API and frontend unions plus regression coverage were updated.
- NITPICK FIXED — replaced the dead two-value `_effective_capital_bases` return with `_effective_deployment_base`; the shared pool bound continues to come only from the engine-wide authority.

Also made the paper-executor broker fixtures explicitly USD so the cross-currency fail-close is exercised against realistic snapshots. Focused API/executor DB tests, pyright, frontend typecheck, and the full pre-push gate pass.

Luke-Bradford:
Latest review resolution at `89e3b895`:

- WARNING FIXED — withdrawal protection now sums `alpha_committed + core_pending_committed + core_active_recorded_committed`, matching the engine-wide committed-capital arithmetic.
- Added a regression test that attempts to lower the assigned pot beneath recorded active core commitment and requires the named refusal.

Focused DB tests, ruff, pyright, and the complete pre-push hook pass.

PR #2931, opened 2026-08-24
fix(#2833): align Portfolio evidence progress

Luke-Bradford:
## What changed

- align the Portfolio page core-sleeve evidence coverage with the frozen #2833 prospective boundary
- count only observed rows on dates shared by all three declared candidates
- expose overall progress as common-date coverage rather than the minimum of three independent counts
- pin the runtime boundary to the frozen declaration in regression tests

## Why

After the prospective #2833 declaration merged, the verifier correctly excluded pre-declaration observations but the page read path still counted them. That could show operator progress the evidence process was not allowed to use.

## Verification

- `uv run pytest -q tests/test_2603_core_selection.py tests/test_2603_core_mandate_api.py` — 28 passed
- independent Codex review — no actionable findings
- full repository pre-push gate — green
- live dev database read — `evidence_collecting`, 0/5 common dates, all three candidates 0 days

Refs #2833
Refs #2603

### DIFF
--- a/app/services/strategy_core_selection.py
+++ b/app/services/strategy_core_selection.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from datetime import date
+from datetime import UTC, date, datetime
 from typing import Any, Final, Literal
 
 import psycopg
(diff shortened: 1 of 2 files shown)


Luke-Bradford:
REBUTTED The claimed runtime break does not exist. `sql/366_strategy_core_quote_observations.sql` defines `strategy_core_quote_observations.sample_bucket TIMESTAMPTZ NOT NULL`, and the live dev database reports `(sample_bucket, timestamp with time zone)` from `information_schema.columns`. I also executed `load_core_selection(conn)` from this PR against that live database; the exact changed `_COVERAGE_SQL` completed and returned `evidence_collecting`, 0 common days, with SPY.RTH/CSPX.L/IUSA.L each at 0 prospective days.

PREVENTION REBUTTED This is not an unhandled schema-drift class needing a new standing rule: the repository already owns the column in `sql/366_strategy_core_quote_observations.sql`, the full pre-push smoke gate booted against the dev DB, and the requested query-specific real-DB verification has now been performed and recorded above.

Luke-Bradford:
Mandatory rebuttal-only checkpoint 3: Codex independently inspected the schema/query path and found no remaining action. It confirmed migration 366 is the sole table definition, `sample_bucket TIMESTAMPTZ NOT NULL` is already on `main`, startup applies migrations before serving, and `load_core_selection()` directly executes the changed SQL. It also agreed that the recorded successful live-DB execution satisfies the requested verification and that hypothetical manual schema drift is not a PR-specific merge blocker.

All Claude findings are therefore explicitly resolved; CI is green on `e1dfeab1`.

PR #2940, opened 2026-08-25
Clarify provisional core evidence dates

Luke-Bradford:
## Summary

- rename the Portfolio evidence stat from an implied completed trading-day count to `Common dates seen`
- state that the visible count remains provisional until the sealed verifier opens
- make the server-authored blocker explain that #2833 checks complete populations only after the fifth date closes

## Why

The first prospective common date became visible at 14:23 UTC on 2026-08-25 as soon as each frozen candidate had an observed bucket. That is the correct progress count, but the existing `Trading days` hint and `has 1 of 5 required trading days` blocker overstated it: the date is not final evidence until its session has completed, the following UTC boundary has passed, and the sealed verifier has checked the full internal population. Actions already fail closed; this change makes the operator copy equally precise.

## Verification

- `uv run pytest -q tests/test_2603_core_selection.py tests/test_2603_core_mandate_api.py` — 28 passed
- `npm test -- --run src/pages/StrategyPortfolioLens.test.tsx` — 28 passed
- mandatory pre-push gate — green (ruff, format, pyright, invariant lints, full non-DB tier, smoke and frontend checks)
- live source state at discovery: common date 1/5, candidate intervals internally continuous, sealed verifier unopened

## Safety

Copy and test change only. No evidence row, declaration, verifier, mandate, capital boundary, kill switch or broker state is changed.

Refs #2603. Refs #2833.

### DIFF
--- a/app/api/strategies.py
+++ b/app/api/strategies.py
@@ -3620,8 +3620,9 @@ def read_core_sleeve(
                 CoreSleeveBlockerResponse(
                     code="core_evidence_collecting",
                     detail=(
-                        f"#2833 has {selection.observed_trading_days} of "
-                        f"{selection.required_trading_days} required trading days; cash remains the fallback."
+                        f"#2833 has observations on {selection.observed_trading_days} of "
+                        f"{selection.required_trading_days} required common dates. The sealed verifier checks "
+                        "the complete populations after the fifth date closes; cash remains the fallback."
                     ),
                 )
             )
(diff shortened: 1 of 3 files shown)


Luke-Bradford:
Resolved the review nit in `26375cad`: the per-candidate table header now says `Dates seen`, matching the provisional `Common dates seen` stat without implying completed evidence. The separate `Evidence-selected` hint remains intentionally tied to the post-verdict instrument state. The focused 28-test frontend file and the full mandatory pre-push gate are green on the follow-up commit.

